PR #14, opened 2018-09-08
Many things

sclevine:
Add support for `.append` and `.override`, also:

- Revert symlinks
- Remove all dependency on packs
- Switch to go modules
- Move images into lifecycle
- Move all compilation and buildpack downloading into Docker
- Introduce github.com/pkg/errors


env.go
```diff
@@ -43,14 +44,16 @@ func suffix(s string, suffix byte) string {
...
-	return eachEnvFile(envDir, func(k, v string) error {
-		return p.Setenv(k, v)
-	})
-}
-
 func (p *Env) AddEnvDir(envDir string) error {
 	return eachEnvFile(envDir, func(k, v string) error {
+		if strings.HasSuffix(k, ".append") {
+			name := strings.TrimSuffix(k, ".append")
+			return p.Setenv(name, p.Getenv(name)+string(v))
```
jkutner: Will `Setenv` expand any `$ENV_VARS` in `v`? That is, can you put env vars in the value of an env var? I don't have a preference here, but wonder if we should be explicit about that behavior in the spec.

jkutner:
@sclevine what does this mean for the packs repo? Will it be exclusively v2 images?

env.go
sclevine: `Setenv` is `os.Setenv`, which does not to expand. The spec should definitely be explicit about this.

sclevine: Clarified: https://github.com/buildpack/spec/commit/032f1be2e39235b5b471b05b8004e4b6b12d6763

sclevine:
That seems reasonable to me. Maybe we could rename it to better indicate that the purpose of the repo is to provide builder images for the existing Heroku/CF buildpacks?

jkutner:
@sclevine 👍

PR #19, opened 2018-09-14
Accept metadata json as cli arg

dgodd:
This enables users to extract the previous image's metadata seperately and then pass it to the docker image to run


dgodd:
We chose to pass metadata as a file instead of an arg

PR #22, opened 2018-10-03
Add script for quick development

dgodd:
Running "images/bin/devbuild" will compile the lifecycle executables
locally and then generate "packs/samples:dev" from "packs/samples" with
the new lifecycle binaries. This enables the abililty to quickly modify
the lifecycle binaries and then test them out.

### DIFF
--- a/images/bin/build-dev
+++ b/images/bin/build-dev
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+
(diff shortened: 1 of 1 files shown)


sclevine:
Could we call it `images/bin/build-dev`?

dgodd:
@sclevine File name changed

PR #23, opened 2018-10-11
Fix runimage/sha in export & publish case

dgodd:
It was storing the image sha, should be storing the sha for the top layer of the image

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -26,7 +26,11 @@ type Exporter struct {
 }
 
 func (e *Exporter) Export(launchDir string, runImage, origImage v1.Image) (v1.Image, error) {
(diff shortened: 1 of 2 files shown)


sclevine:
Can we call this `runImage` instead of `runimage`?

dgodd:
@sclevine Commit above changes the name

sclevine:
Still has an issue with import order consistency. Internal imports should be at the bottom of the list in the exporter test (or all imports in all files should use the two group style). Approving, but on the condition this is fixed before or shortly after merging.

dgodd:
Duplicate of https://github.com/buildpack/lifecycle/pull/28

PR #26, opened 2018-10-17
Uses hardcoded defaults for most lifecycle inputs

ekcasey:
* PACK_USER_ID and PACK_GROUP_ID defaults are still provided with stack env vars

[buildpack/roadmap#14]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,18 +13,17 @@ ARG yj_url=https://github.com/sclevine/yj/releases/download/v2.0/yj-linux
 ARG pack_uid=1000
 ARG pack_gid=1000
 
-LABEL io.buildpacks.stack.id="io.buildpacks.stacks.bionic"
 
 RUN apt-get update && \
   apt-get install -y wget xz-utils ca-certificates && \
   rm -rf /var/lib/apt/lists/*
-
 
 RUN \
   groupadd pack --gid ${pack_gid} && \
   useradd --uid ${pack_uid} --gid ${pack_gid} -m -s /bin/bash pack
 
 ENV PACK_USER_ID=${pack_uid}
+ENV PACK_GROUP_ID=${pack_gid}
 ENV PACK_USER_GID=${pack_gid}
 
 COPY --from=builder /go/bin /lifecycle
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -27,7 +27,7 @@ func init() {
 	cmd.FlagLaunchDir(&launchDir)
 	cmd.FlagGroupPath(&groupPath)
 	cmd.FlagUseDaemon(&useDaemon)
-	cmd.FlagUseHelpers(&useHelpers)
+	cmd.FlagUseCredHelpers(&useHelpers)
 	cmd.FlagMetadataPath(&metadataPath)
 }
 
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -34,7 +34,7 @@ func init() {
 func main() {
 	flag.Parse()
 	repoName = flag.Arg(0)
-	if flag.NArg() > 1 || repoName == "" || launchDir == "" {
+	if flag.NArg() > 1 || repoName == "" {
 		cmd.Exit(cmd.FailCode(cmd.CodeInvalidArgs, "parse arguments"))
 	}
 	cmd.Exit(analyzer())
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -10,15 +10,15 @@ import (
 )
 
 const (
-	DefaultLaunchDir            = "/workspace"
-	DefaultCacheDir             = "/cache"
-	DefaultBuildpacksDir        = "/buildpacks"
-	DefaultPlatformDir          = "/platform"
-	DefaultOrderPath            = "/buildpacks/order.toml"
-	DefaultGroupPath            = `./group.toml`
-	DefaultPlanPath             = "./plan.toml"
-	DefaultUseDaemon            = false
-	DefaultUseCredentialHelpers = false
+	DefaultLaunchDir      = "/workspace"
+	DefaultCacheDir       = "/cache"
+	DefaultBuildpacksDir  = "/buildpacks"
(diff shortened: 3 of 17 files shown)


cmd/analyzer/main.go
```diff
@@ -34,7 +34,7 @@ func init() {
 func main() {
 	flag.Parse()
 	repoName = flag.Arg(0)
-	if flag.NArg() > 1 || repoName == "" || launchDir == "" {
+	if flag.NArg() > 1 || repoName == "" {
```
sclevine: Empty launch dir isn't valid here.

cmd/cmd.go
```diff
@@ -10,56 +10,62 @@ import (
...
-	EnvUseHelpers = "PACK_USE_HELPERS"
+	DefaultLaunchDir            = "/workspace"
+	DefaultCacheDir             = "/cache"
+	DefaultBuildpacksDir        = "/buildpacks"
+	DefaultPlatformDir          = "/platform"
+	DefaultOrderPath            = "/buildpacks/order.toml"
+	DefaultGroupPath            = `./group.toml`
+	DefaultPlanPath             = "./plan.toml"
+	DefaultUseDaemon            = false
+	DefaultUseCredentialHelpers = false
```
sclevine: How about `DefaultUseCredHelpers` to match the convention elsewhere?

cmd/analyzer/main.go
ekcasey: Yes that was the thought. We were very inconsistent about which values we were validating weren't empty. I decided to go with none. An invalid dir should be caught at a later I think.

cmd/launcher/main.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/buildpack/lifecycle/cmd"
 )
 
+var (
+	launchDir string
+)
+
+func init() {
+	cmd.FlagLaunchDir(&launchDir)
```
sclevine: This flag is never parsed. I'm not sure if the launcher should have flags, given how it's used as an `ENTRYPOINT`. Maybe leave this unconfigurable for now?

cmd/cmd.go
ekcasey: makes sense, I just changed it

images/run/Dockerfile
```diff
@@ -9,6 +10,8 @@ HEALTHCHECK --interval=30s --timeout=1s --start-period=60s --retries=1 \
 
 USER pack
 
+LABEL io.buildpacks.stack.id="${stack_id}"
+ENV PACK_STACK_ID="${stack_id}"
```
sclevine: Should not be exposed for launch.

Dockerfile
```diff
@@ -13,18 +13,17 @@ ARG yj_url=https://github.com/sclevine/yj/releases/download/v2.0/yj-linux
...
 RUN apt-get update && \
   apt-get install -y wget xz-utils ca-certificates && \
   rm -rf /var/lib/apt/lists/*
 
-
 RUN \
   groupadd pack --gid ${pack_gid} && \
   useradd --uid ${pack_uid} --gid ${pack_gid} -m -s /bin/bash pack
 
 ENV PACK_USER_ID=${pack_uid}
```
sclevine: These three shouldn't be exposed during launch.

cmd/cmd.go
```diff
@@ -10,56 +10,62 @@ import (
...
 func FlagMetadataPath(metadata *string) {
 	flag.StringVar(metadata, "metadata", "", "path to json containing image metadata for previous image")
 }
 
 func FlagUseDaemon(use *bool) {
-	flag.BoolVar(use, "daemon", boolEnv(EnvUseDaemon), "export to docker daemon")
+	flag.BoolVar(use, "daemon", DefaultUseDaemon, "export to docker daemon")
 }
 
 func FlagUseHelpers(use *bool) {
```
sclevine: Maybe `FlagUseCredHelpers` to match convention and the default?

cmd/analyzer/main.go
sclevine: 👍

cmd/cmd.go
sclevine: 👍

cmd/launcher/main.go
ekcasey: Hmm, looks like the launcher.go is using the launch dir value. Wouldn't things like this
```
	if err := os.Chdir(filepath.Join(l.DefaultLaunchDir, "app")); err != nil {
		return errors.Wrap(err, "change to app directory")
	}
```
break if it was configurable in builder but had a different default in launch. Maybe we should be added `-launch=<launch-dir>` with the correct value from build to the `ENTRYPOINT`

images/run/Dockerfile
ekcasey: Makes sense for `PACK_STACK_ID` but the label is useful because the platform could use it for validation. For example, when `pack build` is given a `--run-image` flag, we need to check the stack.

cmd/cmd.go
ekcasey: fixed it

cmd/launcher/main.go
nebhale: Maybe write it into `/workspace/config` during the build?

ekcasey: @sclevine Working on this I am realizing that `exporter` needs another input. Right now during export we are overloading the `<launch-dir>` concept. We need to know both the directory where the assets from which to generate layers live (`<layer>.toml`s, <layer-dir>s, etc.). AND we need to know at what path we want these layers to live in the exported image.

For example, when we call invoke `lifecycle.Exporter#export` from the `pack` cli, the assets actually live in a temp dir on the host even though we want `<launch-dir>` to be `/workspace` in the final image.

After getting some feedback from @dgodd and @jchesterpivotal, I propose we call the former concept `launchAssetDir` and add another flag to the reference implementation.

We needs to clear up these concepts so `<launch-dir>` can be truly configurable and so `exporter` can set that right value for `PACK_LAUNCH_DIR` on the exported image

PR #27, opened 2018-10-18
Exporter can accept UID & GID

dgodd:
Setting UID & GID will set the values on image layer directories and
files.

This allows files to be temporarily stored on file systems where the
user does not have privileges to set uid & gid.

[https://github.com/buildpack/pack/issues/50]

Signed-off-by: Dave Goddard <dave@goddard.id.au>

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -11,11 +11,10 @@ import (
 	"strings"
 
 	"github.com/BurntSushi/toml"
+	"github.com/buildpack/lifecycle/img"
 	"github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/pkg/errors"
-
-	"github.com/buildpack/lifecycle/img"
 )
 
 type Exporter struct {
@@ -23,6 +22,7 @@ type Exporter struct {
 	TmpDir     string
 	In         []byte
 	Out, Err   io.Writer
+	UID, GID   int
 }
 
 func (e *Exporter) Export(launchDir string, runImage, origImage v1.Image) (v1.Image, error) {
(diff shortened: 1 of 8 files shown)


exporter.go
```diff
@@ -11,18 +11,18 @@ import (
 	"strings"
 
 	"github.com/BurntSushi/toml"
+	"github.com/buildpack/lifecycle/img"
 	"github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/pkg/errors"
-
-	"github.com/buildpack/lifecycle/img"
```
sclevine: Seems like a mistake?

exporter_test.go
```diff
@@ -221,6 +270,34 @@ func getImageFile(image v1.Image, layerDigest, path string) (string, error) {
 	return getLayerFile(layer, path)
 }
 
+func getImageFileUidGid(image v1.Image, layerDigest, path string) (int, int, error) {
```
sclevine: How about:
```
getImageFileOwner(image v1.Image, layerDigest, path string) (uid int, gid int, err error)
```

exporter.go
dgodd: vim just merged all the imports together and sorted them. It seemed a good whitespace change so I left it in

exporter_test.go
dgodd: Test helper function name changed in new commit

exporter.go
sclevine: The import pattern we've used so far is:
```
<stdlib>

<external>

<internal>
```

dgodd: I would rather let go tooling do it's thing.

dgodd: I have disabled go tooling to edit the file and splt this back out

PR #28, opened 2018-10-19
Metadata records run image top layer

ekcasey:
* Stores top layer SHA instead of image SHA
* Allows rebasing without access to original run image


ekcasey:
I'm also open (and maybe partial) to storing all the run layers in the metadata.

ekcasey:
@sclevine please review changes

PR #35, opened 2018-11-02
Add support for custom app directory with cmd arg

jkutner:
This allows for a custom named AppDir, or an AppDir that does not reside under the LaunchDir. The default remains `workspace/app`, but this also cleans up some hard coded references to `"app"`.

I've also introduced a `PACK_APP_DIR` env var to support the launcher, which does not take flags. I'm not sure if this should be a flag. If not, we should maybe add this env var to the spec.

The primary use case is backwards compatibility with pre-v3 platforms, and the potential of a compatibility layer with older buildpacks (i.e. a "shim").

### Questions

* Should the cmd.DefaultAppDir change if the LaunchDir is set to a custom value?

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -26,7 +26,7 @@ type Exporter struct {
 	UID, GID   int
 }
 
-func (e *Exporter) Export(launchDirSrc, launchDirDst string, appDirSrc string, appDirDst string, runImage, origImage v1.Image) (v1.Image, error) {
+func (e *Exporter) Export(launchDirSrc, launchDirDst, appDirSrc, appDirDst string, runImage, origImage v1.Image) (v1.Image, error) {
 	metadata := AppImageMetadata{}
 
 	if err := addRunImageMetadata(runImage, &metadata); err != nil {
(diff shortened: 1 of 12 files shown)


exporter.go
```diff
@@ -26,7 +26,7 @@ type Exporter struct {
 	UID, GID   int
 }
 
-func (e *Exporter) Export(launchDirSrc, launchDirDst string, runImage, origImage v1.Image) (v1.Image, error) {
+func (e *Exporter) Export(launchDirSrc, launchDirDst string, appDirSrc string, appDirDst string, runImage, origImage v1.Image) (v1.Image, error) {
```
Zanadar: This function signature is maybe getting a bit crowded? Can we clean it up with something like:

```suggestion
func (e *Exporter) Export(launchDirSrc, launchDirDst, appDirSrc, appDirDst string, runImage, origImage v1.Image) (v1.Image, error) {
```

Zanadar:
This seems like a great idea!

jkutner:
Feedback addressed

PR #41, opened 2018-11-24
Support arbitrary directory names for buildpacks

jkutner:
This change adds support for arbitrary naming of the `/buildpacks/<buildpack>` directories. There is no longer an assumption that this directory name will match (or even derive from) the Buildpack ID. 

In this way, the `/` (and potentially other characters) can be substituted without needing to establish a contract between the lifecycle and it's consumer.

I've also spiced up the test strings used for the IDs.


map_test.go
```diff
@@ -175,13 +176,17 @@ func testMap(t *testing.T, when spec.G, it spec.S) {
...
-func mkBuildpackTOML(t *testing.T, dir, id, version string) {
-	mkfile(t, fmt.Sprintf(buildpackTOML, id, version),
-		filepath.Join(dir, id, version, "buildpack.toml"),
+func mkBuildpackTOML(t *testing.T, dir, id, name, version string) {
+	mkfile(t, fmt.Sprintf(buildpackTOML, id, name, version),
+		filepath.Join(dir, escapeId(id), version, "buildpack.toml"),
 	)
 }
+
+func escapeId(id string) string {
```
sclevine: Should be `escapeID` as per https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

Also, the spec restricts the ID to `.`, `-`, and `/`, so a single underscore should be okay. (Less relevant in the tests,  just FYI.)

jkutner:
@sclevine i've added some new commits that escape the ID during build and export (I was hoping I could get away without that for now but can't). In any case, it still doesn't matter if pack and lifecycle use the same substitution pattern (because detect reads whatever dir is there, and pack doesn't care what name the lifecycle gives to the layer/cache dirs).

ekcasey:
@jkutner @sclevine Would it make sense to add a `buildpacks` key to `order.toml` that specifies a map of buildpack IDs to buildpacks objects that include the `name` and `dir` for the buildpack? That way, if the `order.toml` references a buildpack more than once, we don't need to redundantly specify the other values. 

Also, I could imagine a builder image having a buildpack that is not referenced in the `groups` section the builder's `order.toml` (it can only be invoked explicitly with `pack build --buildpack`). If I am running `pack build --buildpack id@version`, how will we know which directory to find the buildpack in?

sclevine:
@ekcasey `name` and `dir` are not valid keys in order.toml -- those struct fields are populated by combining order.toml with the buildpack directory, which has a defined layout. This is also how a buildpack not specified in order.toml is located.

See: https://github.com/buildpack/spec/blob/master/platform.md#order.toml-(toml)

ekcasey:
@sclevine I thought, from the comments, this PR was suggesting allow arbitrary names instead of naming the dirs based on the ID. I'll take a closer look. I must be missing something here.

jkutner:
Yes, as @ekcasey said, this (in efffect) allows arbitrary directory names. The `map[string]*Buildpack` is created by traversing the `/buildpacks` directory, looking for `buildpack.toml` files and reading the `id`. This works because both `buildpack.toml` and `id` are required by the spec.

sclevine:
Just to clarify: this allows arbitrary directory names, but it doesn't require that those arbitrary directory names be specified in `order.toml`.

jkutner:
@sclevine correct

PR #44, opened 2018-12-07
Set entrypoint to /lifecycle/launcher during export

jkutner:
This change set `ENTRYPOINT` to a fixed value of `/lifecycle/launcher`. This will ensure that CMDs like `/bin/bash` are started with the app environment.

It also adds a `Cmd` func for use by pack CLI.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -179,6 +179,11 @@ func (e *Exporter) ExportImage(launchDirDst, appDirDst string, runImage, origIma
 		return nil, errors.Wrap(err, "set app dir env var")
 	}
 
+	repoImage, err = img.Entrypoint(repoImage, []string{"/lifecycle/launcher"})
+	if err != nil {
+		return nil, errors.Wrap(err, "set entrypoint")
+	}
+
 	return repoImage, nil
 }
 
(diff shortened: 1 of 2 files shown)


sclevine:
**For the entrypoint change:**
Currently, the platform spec specifies that the run image must have a copy of the launcher and must have the entrypoint set to the launcher. https://github.com/buildpack/roadmap/issues/21 will adjust this so that the exporter is responsible for copying the launcher to the image and setting the entrypoint (such that the run image shouldn't have either).

**For the cmd change:**
I don't think we should couple this the pack CLI. Otherwise, apps would have different ways of specifying a default start command across different platforms. Could we introduce something like a Procfile instead? What if the lifecycle read a `process.toml` (or `app.toml`?) from the app root that had the same format as the process list in `launch.toml`? Then we could merge that on top of the buildpack-provided list before we pick the CMD.

Also, just FYI, there are massive changes to the exporter/analyzer for the great layer merger here: https://github.com/buildpack/lifecycle/pull/43

jkutner:
@sclevine i can rework this onto https://github.com/buildpack/lifecycle/pull/43 once it's merged. This isn't urgent.

I figured we'd need to discuss a `process.toml` or `app.toml` kind of thing eventually. We probably need something like it.

sclevine:
Closing as this was implemented a while ago.

PR #47, opened 2018-12-12
launcher succeeds when buildpack dir does not exist

ekcasey:
* fixes bug in env setting logic
* succeed when metadata has buildpacks that did not contibute layers


sclevine:
Not entirely sure I understand how this happens, but looks good.

ekcasey:
@sclevine The example @nebhale provided was a case where a buildpack contribruted a process to the app `metadata.toml` (by writing a `launch.toml`) but did not contribute a launch layer. I imagine the same thing could happen in any case that a buildpack did not provide a launch layer (e.g. it provides only build layers for subsequent buildpacks).

PR #53, opened 2018-12-14
Write uid/gid/pemissions for all parent directories in layer archives

matthewmcnew:
This prevents exported layers from having the wrong owner in an app image
https://github.com/buildpack/pack/issues/80

Signed-off-by: Matthew McNew mmcnew@pivotal.io


ekcasey:
Closing, see https://github.com/buildpack/lifecycle/pull/58 instead for uid/gid fix

PR #58, opened 2018-12-18
Remove layers-src and app-src exporter params

ekcasey:
Fixes https://github.com/buildpack/pack/issues/80


ekcasey:
Note: This zeros out the access and mod times of some directories. Perhaps we should do this for all files in exported layers?

sclevine:
We should zero out all times to support de-duplication. go-containerregistry might already be doing this, not sure.

ekcasey:
@sclevine go-containerregistry (at least the version we are using) does not do this, but I agree that we should

PR #64, opened 2019-01-10
Analyzer bug 61

ekcasey:
Doesn't write metadata written for `build = true` layers
Rewrites metadata for launch layers from then previous image label, even when the cached layer contents are valid


cache.go
```diff
@@ -0,0 +1,78 @@
...
+		layersDir:   layersDir,
+		buildpackID: buildpackID,
+		layers:      layers,
+	}, nil
+}
+
+type cacheType int
+
+const (
+	staleNoMetadata cacheType = iota
```
sclevine: I'd suggest prefixing these. Not sure we want a global constant called `valid`. Maybe:
```
layerValid = iota
layerStaleNoMetadata
layerStaleWrongSHA
layerNotForLaunch
```

testdata/analyzer/cached-layers/buildpack.node/node_modules.toml
```diff
@@ -1,2 +1,3 @@
+launch = true
 [metadata]
   stalekey = "staleval"
```
sclevine: Missing newline

cache.go
ekcasey: I like the suggestion but how do you feel about prefixing with `cache` instead?

cache.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+type cache struct {
+	layersDir   string
+	buildpackID string
+	layers      map[string]*LayerMetadata
+}
+
+func readCache(layersDir, buildpackID string) (cache, error) {
+	layers := make(map[string]*LayerMetadata)
```
sclevine: Can use `layers := map[string]*LayerMetadata` if no pre-allocation is needed.

cache.go
```diff
@@ -0,0 +1,78 @@
...
+	tomls, err := filepath.Glob(filepath.Join(layersDir, buildpackID, "*.toml"))
+	if err != nil {
+		return cache{}, err
+	}
+	for _, toml := range tomls {
+		name := strings.TrimRight(filepath.Base(toml), ".toml")
+		layers[name], _ = readTOML(toml)
```
sclevine: Seems like we should check this error?

cache.go
```diff
@@ -0,0 +1,78 @@
...
+
+const (
+	staleNoMetadata cacheType = iota
+	staleWrongSHA
+	nonLaunch // we can't determine whether the cache is stale for launch=false layers
+	valid
+)
+
+func (c *cache) classifyLayer(layerName string, metadata *BuildpackMetadata) cacheType {
+	cachedLayer, ok := c.layers[layerName]
```
sclevine: `ok` is defined but not used

analyzer_test.go
```diff
@@ -176,9 +190,37 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+						} else {
+							expected := `build = false
+launch = true
+cache = false
+
+[metadata]
+  akey = "avalue"
+  bkey = "bvalue"
+`
+							actual := string(txt)
```
sclevine: Could inline, not reused

analyzer.go
```diff
@@ -49,179 +46,61 @@ func (a *Analyzer) analyze(metadata AppImageMetadata) error {
...
-	}
-	if err := os.Remove(abd.layerPath(name) + ".toml"); err != nil && !os.IsNotExist(err) {
-		return err
-	}
-	return nil
-}
-
-func appImageMetadata(groupBP string, metadata AppImageMetadata) (*BuildpackMetadata, bool) {
+func bpMetadata(buildpackID string, metadata AppImageMetadata) *BuildpackMetadata {
 	for _, buildpackMetaData := range metadata.Buildpacks {
```
sclevine: Metadata is one word (`MetaData` -> `Metadata`)

cache.go
sclevine: Definitely like `cacheX` better!

PR #65, opened 2019-01-11
Added Cacher mechanism that exports cached layers to an image

jkutner:
This change creates a `Cacher` struct with two implementations:

* `NoopCacher`: for the existing behavior, which works with Docker volumes
* `LocalImageCacher`: for putting the cache layers in an image

The `LocalImageCacher` can be activated with two options to `/lifecycle/exporter`: 

* `-daemon`
* `-cache-image <base-image-to-use-for-cache>`

An optional flag `-cache-tag` changes the default tag from `cache`. The same image name is use for run and cache.


ekcasey:
@jkutner what do you think about implementing `cacher` as a separate binary? I like the idea of providing a reference caching implementation but I'm not as keen on adding flags to `exporter` that are tied to a specific caching mechanism.

jkutner:
@ekcasey when you say a `cacher` binary you mean under `cmd/` to produce another `/lifecycle/*` binary, correct? I'm cool with that _but_ I wonder how hard it is to decouple the cacher from the export. Some caching logic needs to be woven into the export logic. For example, see [this issue i ran into with volumes](https://github.com/buildpack/lifecycle/issues/63).

We'd also have to harden the interface between the export and cache though, or accept a performance hit (like reading toml files and creating tarballs twice). 

Right now, the export passes metadata to the cacher, and it asks the cache (via `RequiresTar`) if it should create the tarball for a cache-only layer.

Do we foresee N caching mechanisms? Or will the Image and Volume mechanisms be the only ones?

jkutner:
this is superseded by #83

PR #69, opened 2019-01-17
exporter treates missing layer.toml files like empty layer.toml files

ekcasey:
* throws an error when metadata file fails to parse

[buildpack/lifecycle#63]


layers.go
```diff
@@ -12,54 +12,67 @@ import (
...
+	for _, fi := range fis {
+		if fi.IsDir() {
+			bpDir.layers[fi.Name()] = bpDir.newBPLayer(fi.Name())
+		}
+	}
+
+	tomls, err := filepath.Glob(filepath.Join(path, "*.toml"))
+	for _, toml := range tomls {
+		name := strings.TrimRight(filepath.Base(toml), ".toml")
+		if name == "launch" {
```
matthewmcnew: It would probably make sense to remove this before it gets to the exporter and analyzer. (buildpack/lifecycle#70)

And/or explicitly test this as desired exporter behavior.

analyzer.go
```diff
@@ -85,7 +85,8 @@ func (a *Analyzer) analyze(metadata AppImageMetadata) error {
 		for layer, data := range metadataLayers {
 			if !data.Build {
 				a.Out.Printf("writing metadata for uncached layer '%s/%s'", buildpackID, layer)
-				if err := cache.newBPLayer(layer).writeMetadata(metadataLayers); err != nil {
+				layer := cache.newBPLayer(layer)
```
sclevine: Not sure if shadowing `layer` improves this.

layers.go
sclevine: +1

layers.go
```diff
@@ -12,54 +12,67 @@ import (
...
-		name := strings.TrimRight(filepath.Base(toml), ".toml")
-		layers = append(layers, bpLayer{
-			layer{
-				path:       filepath.Join(layersDir, buildpackID, name),
-				identifier: fmt.Sprintf("%s/%s", buildpackID, name),
-			},
-		})
-	}
-	return bpLayersDir{
+	layers := map[string]bpLayer{}
```
sclevine: If `layers` is extracted, use it later instead of `bpDir.layers`. Otherwise, I suggest inlining it.

layers.go
ekcasey: It's interesting that `exporter` doesn't need the `launch.toml` files. That might imply that `builder` should clean them up

PR #72, opened 2019-01-18
extract buildpack id/path escaping into utils

jkutner:
This fixed a problem where the escaped buildpack ID used in a map did not match the ID in the image metadata, which caused the analyzer to invalidate the cache. 

We moved the escape/unescape logic intil `util.go`. The new util funcs are invoked in detector and analyzer. We also fixed a spot where the escaped buildpack ID was used as a key in a map (the real ID should always be used unless it's dealing with the filesystem). Finally, we fixed a spot where analyzer was not unescaping the buildpack dir into an ID.

I think we can avoid problems like this in the future by encapsulating all Buildpack logic related to the filesystem and metadata into a standard `type Buildpack struct` that can be used throughout the codebase. Currently there are numerous buildpack structs, each with their own logic for reading the fs and creating maps of metadata.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -20,10 +20,10 @@ func WriteTOML(path string, data interface{}) error {
 	return toml.NewEncoder(f).Encode(data)
 }
 
-func BuildpackIdToDir(id string) string {
+func buildpackIDToDir(id string) string {
 	return strings.Replace(id, "/", "_", -1)
 }
 
-func BuildpackDirToId(dirname string) string {
+func buildpackDirToID(dirname string) string {
 	return strings.Replace(dirname, "_", "/", -1)
 }
--- a/utils.go
+++ b/utils.go
@@ -3,6 +3,7 @@ package lifecycle
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/BurntSushi/toml"
 )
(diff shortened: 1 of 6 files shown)


utils.go
```diff
@@ -18,3 +19,11 @@ func WriteTOML(path string, data interface{}) error {
 	defer f.Close()
 	return toml.NewEncoder(f).Encode(data)
 }
+
+func BuildpackIdToDir(id string) string {
```
sclevine: Could we call this `buildpackIDToDir`?
I don't think it needs to be exported, and see [here](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

utils.go
```diff
@@ -18,3 +19,11 @@ func WriteTOML(path string, data interface{}) error {
...
+	return strings.Replace(id, "/", "_", -1)
+}
+
+func BuildpackDirToId(dirname string) string {
```
sclevine: Suggest: `buildpackDirToID`

utils.go
jkutner: fixed in d0a7f80

utils.go
jkutner: fixed in d0a7f80

PR #83, opened 2019-02-11
Cacher/Restorer

ekcasey:
I am opening this PR for feedback, I don't think we want to merge it until the `pack` work to use `cacher/restorer` and it can be fully integration tested


ekcasey:
There are two things I am interested in changing:

1. Instead of `cacher <cache-image-name>` I think the interface should be `cacher -image=<cache-image-name>` so that the volume cacher can use the same binary with the interface`cacher -volume=<cache-volume-mount-path>`

2. Instead of using the same metadata label we use for app images I think we should use a different labels `io.buildpacks.cache.metadata`

thoughts on these interface changes?

utils.go
```diff
@@ -20,10 +20,6 @@ func WriteTOML(path string, data interface{}) error {
 	return toml.NewEncoder(f).Encode(data)
 }
 
-func buildpackIDToDir(id string) string {
+func escape(id string) string {
```
jkutner: The new name of this func feels a little ambiguous. I get that we need it for more than dir names now, but maybe we should still capture the intent, like `escapeBuildpackID`? Or maybe I've just been Java-ing too long.

jkutner:
The recommendation for `cacher --image` and `cache --volume` seems sane. But I don't think I have a use case for volume caching at all (even locally, I'd prefer to use images). So I don't have a preference. I do want to explore local versus remote image caches some more (but I'm not even sure I have a use case for the latter).

jkutner:
Because `cacher` and `restorer` are so symmetric, I wonder if we should combine them to have `cacher store -image=myimage` and cacher restart -image=myimage` commands (but leave the `cacher.go` and `restorer.go` as-is under the hood). I think the two binaries are fine for now, but it might be something good to capture for future work.

djoyahoy:
@ekcasey can you please review our changes?

image/local.go
```diff
@@ -207,6 +220,19 @@ func (l *local) TopLayer() (string, error) {
 	return topLayer, nil
 }
 
+func (l *local) GetLayer(sha string) (io.ReadCloser, error) {
+	l.prevDownload()
+	layerID, ok := l.prevMap[sha]
+	if !ok {
+		return nil, fmt.Errorf("image '%s' does not contain layer with diff ID '%s'", l.RepoName, sha)
+	}
+	rc, err := os.Open(filepath.Join(l.prevDir, layerID))
```
sclevine:
```
return os.Open(filepath.Join(l.prevDir, layerID))
```

image/local.go
```diff
@@ -59,11 +59,24 @@ func (f *Factory) NewLocal(repoName string, pull bool) (Image, error) {
...
+func (f *Factory) NewEmptyLocal(repoName string) Image {
+	inspect := dockertypes.ImageInspect{}
+	inspect.Config = &container.Config{
+		Labels: map[string]string{},
+	}
+	return &local{
+		RepoName:   repoName,
+		Docker:     f.Docker,
+		Inspect:    inspect,
+		layerPaths: []string{},
```
sclevine: `nil` slice is more commonly used instead of empty slice (unlike for maps, which must be pre-allocated). This line could be removed.

image/remote.go
```diff
@@ -190,6 +191,10 @@ func (r *remote) TopLayer() (string, error) {
 	return hex.String(), nil
 }
 
+func (r *remote) GetLayer(string) (io.ReadCloser, error) {
+	panic("not implemented")
```
sclevine: Considering returning an error here instead of panicking. (Interfaces are evaluated at runtime, so may not always be a programmer error.)

metadata.go
```diff
@@ -51,3 +63,58 @@ func (m *AppImageMetadata) metadataForBuildpack(id string) BuildpackMetadata {
...
+func getAppMetadata(image image.Image, log *log.Logger) (AppImageMetadata, error) {
+	metadata := AppImageMetadata{}
+	contents, err := getMetadata(image, MetadataLabel, log)
+	if err != nil {
+		return metadata, err
+	}
+
+	if err := json.Unmarshal([]byte(contents), &metadata); err != nil {
+		log.Printf("WARNING: image '%s' has incompatible '%s' label\n", image.Name(), MetadataLabel)
+		return metadata, nil
```
sclevine: Seems like we shouldn't be returning a metadata with undefined contents + no error here. Should this be an empty `AppImageMetadata` struct?

metadata.go
```diff
@@ -51,3 +63,58 @@ func (m *AppImageMetadata) metadataForBuildpack(id string) BuildpackMetadata {
...
+		log.Printf("WARNING: image '%s' not found or requires authentication to access\n", image.Name())
+		return "", nil
+	}
+	contents, err := image.Label(metadataLabel)
+	if err != nil {
+		return "", err
+	}
+	if contents == "" {
+		log.Printf("WARNING: image '%s' does not have '%s' label", image.Name(), metadataLabel)
+		return "metadata", nil
```
sclevine: Should this be `return "", nil`? Returning the string `"metadata"` seems confusing.

restorer.go
```diff
@@ -0,0 +1,100 @@
...
+			if layer.Launch {
+				if err := bpLayer.writeSha(layer.SHA); err != nil {
+					return err
+				}
+			}
+			rc, err := cacheImage.GetLayer(layer.SHA)
+			if err != nil {
+				return err
+			}
+			defer rc.Close()
```
sclevine: This will close all of the layers at the end of the whole method, not at the end of each iteration. This will leak file descriptors. I suggest keeping the defer, but extracting a method here.

restorer.go
```diff
@@ -0,0 +1,100 @@
...
+	fis, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+	for _, fi := range fis {
+		path := filepath.Join(path, fi.Name())
+		if err := os.Chown(path, uid, gid); err != nil {
+			return err
+		}
+		if fi.IsDir() {
```
sclevine: Assuming `recursiveChown` also chowns the provided `path`, it seems like we're chowning the top-level directories twice.

testhelpers/testhelpers.go
```diff
@@ -326,3 +326,18 @@ func RecursiveCopy(t *testing.T, src, dst string) {
 	err = os.Chmod(dst, 0775)
 	AssertNil(t, err)
 }
+
+func CreateSingleFileTar(path, txt string) (io.Reader, error) {
+	var buf bytes.Buffer
+	tw := tar.NewWriter(&buf)
+	if err := tw.WriteHeader(&tar.Header{Name: path, Size: int64(len(txt)), Mode: 0666}); err != nil {
```
sclevine: Suggest `0644` here, as tar headers aren't transformed by the OS umask.

cmd/restorer/main.go
```diff
@@ -0,0 +1,65 @@
+package main
+
+import (
+	"flag"
```
sclevine: Fix import path grouping.

image/remote.go
djoyahoy: We'd like to leave the panics because we do think that although interfaces are evaluated at runtime, using the remote image in this way is strictly programmer error. That begs another discussion on the Image interface (it could be a more limited interface), and that would fail at compile time. But, that seems like a separate refactor.

sclevine: My thinking is:
1. This is not the pattern that I've observed elsewhere (e.g., in ggcr: https://github.com/google/go-containerregistry/blob/52aa2729ea7597c8d5f7d6a977fa5cab9fc3a77a/pkg/v1/stream/layer.go#L78)
2. It seems conceivable that a combination of user choices (e.g., remote + get layer) would rely on this method to inform the user that the combination is invalid. A panic is much less friendly in that case.
3. `lifecycle/image` is importable from outside of the lifecycle module, and it's generally bad practice in Go to panic across package boundaries. An external library could recover from this case (e.g., by propagating an error that the user selection was invalid).

(I'm not going to block merging the PR on consensus here, just providing feedback. 😄 )

jkutner:
My comment about "a use case for volume caching" is bunk. I'll approve pending resolution of stephen's requests

Issue #85, opened 2019-03-01
Buildpack versions are not populated in app image metadata


ameyer-pivotal:
Repro steps:

1. `pack build testapp -p acceptance/testdata/node_app`
1. `docker inspect testapp`

Notice that the buildpack version is empty in the `io.buildpacks.lifecycle.metadata`:
```
\"buildpacks\":[{\"key\":\"io.buildpacks.samples.nodejs\",\"version\":\"\", ...
```

djoyahoy:
Fixed in PR #86

PR #86, opened 2019-03-01
Ensure buildpack versions are written to app image metadata

ameyer-pivotal:
Resolves issue #85

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -197,7 +197,7 @@ func mergeTOML(l *log.Logger, out io.Writer, in ...io.Reader) {
 
 type BuildpackOrder []BuildpackGroup
 
-func (bo BuildpackOrder) Detect(c *DetectConfig) (plan []byte, group *BuildpackGroup) {
+func (bo BuildpackOrder) Detect(c *DetectConfig) ([]byte, *BuildpackGroup) {
 	for i := range bo {
 		if plan, group, ok := bo[i].Detect(c); ok {
 			return plan, group
--- a/detector.go
+++ b/detector.go
@@ -197,10 +197,10 @@ func mergeTOML(l *log.Logger, out io.Writer, in ...io.Reader) {
 
 type BuildpackOrder []BuildpackGroup
 
-func (bo BuildpackOrder) Detect(c *DetectConfig) ([]byte, *BuildpackGroup) {
+func (bo BuildpackOrder) Detect(c *DetectConfig) (plan []byte, group *BuildpackGroup) {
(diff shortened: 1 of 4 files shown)


detector.go
```diff
@@ -197,7 +197,7 @@ func mergeTOML(l *log.Logger, out io.Writer, in ...io.Reader) {
 
 type BuildpackOrder []BuildpackGroup
 
-func (bo BuildpackOrder) Detect(c *DetectConfig) (plan []byte, group *BuildpackGroup) {
+func (bo BuildpackOrder) Detect(c *DetectConfig) ([]byte, *BuildpackGroup) {
```
sclevine: Keep named parameters for better documentation.

map_test.go
```diff
@@ -85,9 +85,9 @@ func testMap(t *testing.T, when spec.G, it spec.S) {
 
 		it("should return an ordering of buildpacks", func() {
 			m := lifecycle.BuildpackMap{
-				"buildpack1@version1.1": {Name: "buildpack1-1.1"},
-				"buildpack1@version1.2": {Name: "buildpack1-1.2"},
-				"buildpack2@latest":     {Name: "buildpack2"},
+				"buildpack1@version1.1": {Name: "buildpack1-1.1", Version: "1.1"},
```
sclevine: What does this change accomplish?

detector.go
joaopapereira: We removed them because they were showing as obfuscated in the debugged

map_test.go
joaopapereira: Ensure the version is being passed to the output.

sclevine: The returned objects are just copies of the original objects (see `lookup`), so the name seemed sufficient. No strong opinion, but we might want to make the same changes in the `#ReadGroup` test.

detector.go
sclevine: I think replacing `plan` and `group` on 202 with `p` and `g` would solve this and keep the godoc more clear, but approving either way.

ameyer-pivotal: Ah yeah, I agree with @sclevine thinking about the gdocs. Optionally we can just call `return` but that never felt very readable.

map_test.go
ameyer-pivotal: I guess I see your point. This change was simply one we chose to keep as we were writing tests to triage the bug. I'm OK either way.

sclevine: Would prefer to remove them then, to keep the test clear and consistent.

PR #87, opened 2019-03-04
Rename buildpack key to id for app image metadata label

djoyahoy:


ameyer-pivotal:
Will this cause issues for apps already created which use the old key name?

ashwin-venkatesh:
@sclevine Should I merge this? 🤷‍♂️

djoyahoy:
This is not an important change, so I'll close it. If it's important it will resurface again.

PR #91, opened 2019-03-05
Lifecycle only gets remote or local image, without concern for pulling

ashwin-venkatesh:
- Lifecycle farms out responsibility of pulling to pack
- We want to ensure that pack does not return an error in the
  situation where we can pull from remote but the image only exists
locally

[buildpack/pack#90]

### DIFF
--- a/image/local.go
+++ b/image/local.go
@@ -43,15 +43,23 @@ type local struct {
 }
 
 func (f *Factory) NewLocal(repoName string, pull bool) (Image, error) {
+	var pullFailureError error
+	var inspect types.ImageInspect
+	var err error
 	if pull {
-		if err := f.pullImage(f.Out, f.Docker, repoName); err != nil {
-			return nil, fmt.Errorf("failed to pull image '%s' : %s", repoName, err)
+		err = f.pullImage(f.Out, f.Docker, repoName)
+		if err != nil {
+			pullFailureError = fmt.Errorf("failed to pull image '%s' : %s", repoName, err)
 		}
-	}
-
-	inspect, _, err := f.Docker.ImageInspectWithRaw(context.Background(), repoName)
-	if err != nil && !dockerclient.IsErrNotFound(err) {
-		return nil, errors.Wrap(err, "analyze read previous image config")
+		inspect, _ = f.inspectLocalImage(repoName)
+		if inspect.Config == nil && pullFailureError != nil {
+			return nil, pullFailureError
+		}
+	} else {
+		inspect, err = f.inspectLocalImage(repoName)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return &local{
@@ -61,6 +69,14 @@ func (f *Factory) NewLocal(repoName string, pull bool) (Image, error) {
 		layerPaths: make([]string, len(inspect.RootFS.Layers)),
 		prevOnce:   &sync.Once{},
 	}, nil
+}
+
+func (f *Factory) inspectLocalImage(repoName string) (types.ImageInspect, error) {
+	inspect, _, err := f.Docker.ImageInspectWithRaw(context.Background(), repoName)
+	if err != nil && !dockerclient.IsErrNotFound(err) {
+		return types.ImageInspect{}, errors.Wrap(err, "analyze read previous image config")
+	}
+	return inspect, nil
 }
 
 func (f *Factory) NewEmptyLocal(repoName string) Image {
--- a/image/local.go
+++ b/image/local.go
@@ -4,7 +4,6 @@ import (
 	"archive/tar"
 	"context"
 	"crypto/sha256"
-	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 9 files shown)


ekcasey:
@ashwin-venkatesh @joaopapereira 

I wonder if it makes sense to take the pulling logic out of `NewLocal` and move it into `pack`. The lifecycle binaries never actually pull images. 

I'm thinking about the pull logs and this upcoming bug https://github.com/buildpack/roadmap/issues/49, both of which would be less complicated if the pull logic lived in pack

Additionally I'm worried that pack might print confusing logs if it can't determine whether the image was pulled or not.

image/local.go
```diff
@@ -43,15 +43,23 @@ type local struct {
...
+	var inspect types.ImageInspect
+	var err error
 	if pull {
-		if err := f.pullImage(f.Out, f.Docker, repoName); err != nil {
-			return nil, fmt.Errorf("failed to pull image '%s' : %s", repoName, err)
+		err = f.pullImage(f.Out, f.Docker, repoName)
+		if err != nil {
+			pullFailureError = fmt.Errorf("failed to pull image '%s' : %s", repoName, err)
+		}
+		inspect, _ = f.inspectLocalImage(repoName)
```
sclevine: Returning an error usually implies that the other arguments have undefined contents. Check the error here, instead of checking `inspect.Config == nil`. Otherwise you miss cases where `inspect.Config != nil` and `err != nil`.

image/local.go
```diff
@@ -43,15 +43,23 @@ type local struct {
 }
 
 func (f *Factory) NewLocal(repoName string, pull bool) (Image, error) {
+	var pullFailureError error
+	var inspect types.ImageInspect
+	var err error
```
sclevine: I don't think `pullFailureError` or `err` need to be declared here.

image/local_test.go
```diff
@@ -55,6 +55,105 @@ func testLocal(t *testing.T, when spec.G, it spec.S) {
...
+						labels := make(map[string]string)
+						labels["some.label"] = "some.value"
+
+						h.CreateImageOnLocal(t, dockerCli, repoName, fmt.Sprintf(`
+							FROM scratch
+							LABEL repo_name_for_randomisation=%s
+							ENV MY_VAR=my_val
+							`, repoName), labels)
+
+						localImage, e := factory.NewLocal(repoName, true)
```
sclevine: suggest `e` -> `err`

image/local_test.go
```diff
@@ -55,6 +55,105 @@ func testLocal(t *testing.T, when spec.G, it spec.S) {
...
+					localLabels := make(map[string]string)
+					localLabels["local.label"] = "local.value"
+
+					h.CreateImageOnRemote(t, dockerCli, repoName, fmt.Sprintf(`
+							FROM scratch
+							LABEL repo_name_for_randomisation=%s
+							ENV MY_VAR=my_val
+							`, repoName), remoteLabels)
+
+					localImage, e := factory.NewLocal(repoName, true)
```
sclevine: suggest `e` -> `err`

image/local_test.go
```diff
@@ -73,7 +172,7 @@ func testLocal(t *testing.T, when spec.G, it spec.S) {
 					FROM scratch
 					LABEL repo_name_for_randomisation=%s
 					LABEL mykey=myvalue other=data
-				`, repoName))
+				`, repoName), make(map[string]string))
```
sclevine: I'd suggest `nil` instead of `make(map[string]string)` for these tests.

image/local_test.go
```diff
@@ -55,6 +55,105 @@ func testLocal(t *testing.T, when spec.G, it spec.S) {
...
+			})
+
+			when("image is not available locally", func() {
+				it("returns an error", func() {
+					repoName = "localhost:" + localTestRegistry.Port + "/pack-image-test-" + h.RandString(10)
+					h.CreateImageOnRemote(t, dockerCli, repoName, fmt.Sprintf(`
+							FROM scratch
+							LABEL repo_name_for_randomisation=%s
+							ENV MY_VAR=my_val
+							`, repoName), map[string]string{})
```
sclevine: Suggest `nil` instead of empty map in this case.

image/local.go
```diff
@@ -430,56 +420,3 @@ func (l *local) prevDownload() error {
 	})
 	return outerErr
 }
-
-func (f *Factory) pullImage(output io.Writer, dockerCli *dockerclient.Client, ref string) error {
```
ameyer-pivotal: Moved into pack

image/local.go
```diff
@@ -117,8 +107,8 @@ func (l *local) Found() (bool, error) {
 }
 
 func (l *local) Digest() (string, error) {
-	if l.Inspect.Config == nil {
-		return "", fmt.Errorf("failed to get digest, image '%s' does not exist", l.RepoName)
+	if found, _ := l.Found(); !found {
+		return "", fmt.Errorf("failed to get digest, image '%s' does n1ot exist", l.RepoName)
```
ameyer-pivotal: Typo, will correct

PR #94, opened 2019-03-14
Add Delete to local image

djoyahoy:
- Make image/local and image/remote structs public and
  return them from the image factory instead of the
  image interface
- Delete the original cache image after the cache step

### DIFF
--- a/image/local.go
+++ b/image/local.go
@@ -53,7 +53,7 @@ func (f *Factory) NewLocal(repoName string) (*Local, error) {
 	}, nil
 }
 
-func (f *Factory) NewEmptyLocal(repoName string) Image {
+func (f *Factory) NewEmptyLocal(repoName string) *Local {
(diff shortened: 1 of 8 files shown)


image/local.go
```diff
@@ -38,13 +38,13 @@ type local struct {
 	easyAddLayers    []string
 }
 
-func (f *Factory) NewLocal(repoName string) (Image, error) {
+func (f *Factory) NewLocal(repoName string) (*Local, error) {
```
sclevine: Looks like we forgot to change `NewEmptyLocal` as well? Also, should we remove the interface from this package?

I'd like to chat about the direction here before we merge.

djoyahoy:
@sclevine Reverted changes that exposed private local and remote images. Added delete to the image interface and made remote return a not implemented error for now.

PR #96, opened 2019-03-25
Use version from toml file as BuildpackMap key

jkutner:
This facilitates the implementation of https://github.com/buildpack/pack/issues/113

### DIFF
--- a/map.go
+++ b/map.go
@@ -8,6 +8,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const buildpackVersionLatest = "latest"
+
 type BuildpackMap map[string]*Buildpack
 
 type buildpackTOML struct {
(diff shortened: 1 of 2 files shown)


sclevine:
I think this will break the `latest` symlinks.

jkutner:
@sclevine what do you think about treating `latest` as a special value (either in lifecycle or the spec). That is, a buildpack would not be allowed to define it's version as `"latest"`. That would allow me to work around this.

sclevine:
That restriction seems like a great idea to me. I don't think it makes the current proposed changes work with the `latest` symlink though.

jkutner:
@sclevine correct, i would still have to adjust this PR, but I wanted to float that idea before doing so

jkutner:
@sclevine @ekcasey added a test, but didn't use a symlink (i created a real dir with real files). I think that's probably a good idea?

PR #98, opened 2019-03-27
removes "wall of shas"

ekcasey:
* temporarily set output destination for standard logger to discard before writing a remote image
* surpresses ggcr output

[buildpack/roadmap#45]

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"io/ioutil"
 	"log"
 	"os"
 
@@ -34,6 +35,9 @@ func init() {
 }
 
 func main() {
+	// suppress output from libraries, lifecycle will not use standard logger
+	log.SetOutput(ioutil.Discard)
+
 	flag.Parse()
 	repoName = flag.Arg(0)
 	if flag.NArg() > 1 || repoName == "" {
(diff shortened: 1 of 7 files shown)


image/remote.go
```diff
@@ -258,6 +261,8 @@ func (r *remote) Save() (string, error) {
 		return "", err
 	}
 
+	log.SetOutput(ioutil.Discard)
```
sclevine: Mixed feelings about this.

Upside:
- Probably no other way to do this without an upstream PR
- `log.SetOutput` is re-entrant, so a global data race isn't possible

Downside:
- This temporarily disables logging for any functions running in parallel that use the global logger (and the global logger is designed to be used in parallel). This could introduce a very surprising bug: some unrelated logs could occasionally disappear.

Would it make more sense to move this logic into a `main` package? If naively moving it up the stack would disable too many logs, could we fix this by inverting control?

image/remote.go
```diff
@@ -258,6 +261,8 @@ func (r *remote) Save() (string, error) {
...
+	defer log.SetOutput(os.Stdout)
```
sclevine: The default value for the global logger is `os.Stderr` (and it doesn't appear to be possible to determine the current global `io.Writer` in the `log` package). Might be better to use `os.Stderr` here, but this also makes me feel like pushing this into the `main` package might be better.

PR #99, opened 2019-03-27
Fixes issues that occurs in some local image rebuilds

djoyahoy:
* when new image is built from different base, don't assume you can reuse layers without fetching the layer tar

### DIFF
--- a/image/local.go
+++ b/image/local.go
@@ -89,13 +89,25 @@ func (l *local) Env(key string) (string, error) {
 
 func (l *local) Rename(name string) {
 	l.easyAddLayers = nil
-	if inspect, _, err := l.Docker.ImageInspectWithRaw(context.TODO(), name); err == nil {
-		if len(inspect.RootFS.Layers) > len(l.Inspect.RootFS.Layers) {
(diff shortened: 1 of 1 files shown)


image/local.go
```diff
@@ -270,16 +282,21 @@ func (l *local) Save() (string, error) {
...
 		}
-		done <- err
+		defer res.Body.Close()
+		io.Copy(ioutil.Discard, res.Body)
+
+		done <- nil
 	}()
 
 	tw := tar.NewWriter(pw)
+	defer tw.Close()
```
sclevine: This is also closed further down (and there appear to be a number of double-closes in this method).

djoyahoy: We need to close the pipe later on in the function to terminate the go routine that is writing the tar. We wanted to defer here for the cases in which we return early due to error. As far as I know, Close is idempotent.

sclevine: This is [not always true](https://golang.org/pkg/io/#Closer), but seems okay in the case of `tar.Writer`. Okay with this for now, but I think we should consider refactoring some of this later.

PR #103, opened 2019-03-28
Chown symlinks not their targets

danajp:
`os.Chown` follows the symlink then tries to `chown` the target file. This returns `os.PathError` on broken symlinks (symlinks pointing at a file that doesn't exist).


sclevine:
Looks great. Thanks!

PR #106, opened 2019-03-29
Adds correct number of history entries when saving local image

ekcasey:
* enables pushing local images to registries that enforce shema 1 compat
* enables remote rebasing of images that were originally built locally

[buildpack/lifecycle#104]

### DIFF
--- a/image/local.go
+++ b/image/local.go
@@ -298,20 +298,13 @@ func (l *local) Save() (string, error) {
 	tw := tar.NewWriter(pw)
 	defer tw.Close()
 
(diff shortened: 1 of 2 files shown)


ekcasey:
We originally tried to preserve the history for reused layers, however the docker client doesn't provide an easy & performant way for us to do this. The results of `ImageHistory` may contain more items than the number of layers. The only way to get the history we want from the manifest seems to be a `ImageSave`, but this is a slow operation. Maybe at some point the client will support `ManifestInspect`.

PR #111, opened 2019-04-08
SetupCredHelpers will not overwrite cred helpers from docker config

djoyahoy:
* Helpers will only be added if they do not already exist in the docker config
* The credHelpers section is re-used from the docker config rather than re-written

[buildpack/lifecycle#71]

### DIFF
--- a/cred_helpers.go
+++ b/cred_helpers.go
@@ -2,15 +2,14 @@ package lifecycle
 
 import (
 	"encoding/json"
-	"fmt"
+	"errors"
 	"github.com/google/go-containerregistry/pkg/name"
 	"os"
 	"path/filepath"
 	"regexp"
 )
 
-func SetupCredHelpers(home string, refs ...string) error {
-	dockerPath := filepath.Join(home, ".docker")
+func SetupCredHelpers(dockerPath string, refs ...string) error {
 	configPath := filepath.Join(dockerPath, "config.json")
 
 	config := map[string]interface{}{}
@@ -55,17 +54,15 @@ func SetupCredHelpers(home string, refs ...string) error {
 		return nil
 	}
 
+	ch, ok := config["credHelpers"].(map[string]interface{})
+	if !ok {
+		return errors.New("failed to parse docker config 'credHelpers'")
+	}
+
 	for k, v := range credHelpers {
-		ch, ok := config["credHelpers"].(map[string]interface{})
-		if !ok {
-			return fmt.Errorf("failed to parse docker config 'credHelpers'")
+		if _, ok := ch[k]; !ok {
+			ch[k] = v
 		}
-
-		if _, ok := ch[k]; ok {
-			continue
-		}
-
-		ch[k] = v
 	}
 
 	if err := os.MkdirAll(dockerPath, 0777); err != nil {
--- a/cred_helpers.go
+++ b/cred_helpers.go
@@ -2,18 +2,18 @@ package lifecycle
 
 import (
 	"encoding/json"
+	"fmt"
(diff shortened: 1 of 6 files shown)


cred_helpers.go
```diff
@@ -42,19 +48,35 @@ func SetupCredHelpers(refs ...string) error {
...
-	if !added {
+
+	if len(credHelpers) == 0 {
 		return nil
 	}
+
+	for k, v := range credHelpers {
+		ch, ok := config["credHelpers"].(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("failed to parse docker config 'credHelpers'")
```
sclevine: Use `errors.New`

cred_helpers_test.go
```diff
@@ -0,0 +1,119 @@
+package lifecycle_test
```
sclevine: Bravo for adding tests for this stuff!

cred_helpers_test.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+
+	when("a docker config exists", func() {
+		var (
+			dockerDir string
+		)
+
+		it.Before(func() {
+			dockerDir = filepath.Join(tmpDir, ".docker")
+			h.AssertNil(t, os.MkdirAll(dockerDir, 0755))
```
sclevine: `0777`?

cred_helpers.go
```diff
@@ -42,19 +48,35 @@ func SetupCredHelpers(refs ...string) error {
...
 		}
 	}
-	if !added {
+
+	if len(credHelpers) == 0 {
 		return nil
 	}
+
+	for k, v := range credHelpers {
+		ch, ok := config["credHelpers"].(map[string]interface{})
```
sclevine: Extract this from the loop -- no need to re-read the value every time.

cred_helpers.go
```diff
@@ -42,19 +48,35 @@ func SetupCredHelpers(refs ...string) error {
...
+		if !ok {
+			return fmt.Errorf("failed to parse docker config 'credHelpers'")
+		}
+
+		if _, ok := ch[k]; ok {
+			continue
+		}
+
+		ch[k] = v
```
sclevine: Suggestion: swap `ok` -> `!ok` and move this line into the loop above in place of `continue`.

cred_helpers_test.go
```diff
@@ -0,0 +1,119 @@
...
+			config := parseCredHelpers(t, tmpDir)
+			h.AssertEq(t, config["fake.gcr.io"], "gcr")
+		})
+	})
+}
+
+func copyFile(t *testing.T, src, dest string) {
+	buf, err := ioutil.ReadFile(src)
+	h.AssertNil(t, err)
+	h.AssertNil(t, ioutil.WriteFile(dest, buf, 0644))
```
sclevine: `0666`

cred_helpers.go
```diff
@@ -2,18 +2,18 @@ package lifecycle
...
+	"fmt"
 	"github.com/google/go-containerregistry/pkg/name"
 	"os"
 	"path/filepath"
 	"regexp"
 )
 
-func SetupCredHelpers(refs ...string) error {
-	dockerPath := filepath.Join(os.Getenv("HOME"), ".docker")
+func SetupCredHelpers(home string, refs ...string) error {
```
sclevine: Instead of coupling this to the idea of `$HOME`, maybe we inject the full path to `.docker` instead?

djoyahoy: That sounds good to me.

PR #112, opened 2019-04-08
Fix detect to log 'Trying group X of Y...' [buildpack/lifecycle#105]

djoyahoy:

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -198,7 +198,7 @@ type BuildpackOrder []BuildpackGroup
 
 func (bo BuildpackOrder) Detect(c *DetectConfig) (plan []byte, group *BuildpackGroup) {
 	for i := range bo {
-		c.Out.Printf("Trying group %d of %d...", i + 1, len(bo))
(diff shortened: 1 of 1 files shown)


detector.go
```diff
@@ -199,6 +198,7 @@ type BuildpackOrder []BuildpackGroup
 
 func (bo BuildpackOrder) Detect(c *DetectConfig) (plan []byte, group *BuildpackGroup) {
 	for i := range bo {
+		c.Out.Printf("Trying group %d of %d...", i + 1, len(bo))
```
sclevine: Now we've lost the buildpack count, which was the original purpose of the line.
If we feel we need more info here, how about:
```
Trying group #3 out of 5 with 8 buildpacks...
```
```go
c.Out.Printf("Trying group %d out of %d with %d buildpacks...", i+1, len(bo), len(bo[i].Buildpacks))
```

djoyahoy: Ah, ok. I thought the intent was to simply log which group we were currently trying.

PR #113, opened 2019-04-08
Add CreatedAt on Image interface

ashwin-venkatesh:
This gives us access to the CreatedAt time of an image

### DIFF
--- a/image/local.go
+++ b/image/local.go
@@ -134,6 +134,16 @@ func (l *local) Digest() (string, error) {
 	return parts[1], nil
 }
 
+func (l *local) CreatedAt() (time.Time, error) {
+	createdAtTime := l.Inspect.Created
+	createdTime, err := time.Parse(time.RFC3339Nano, createdAtTime)
+
+	if err != nil {
+		return time.Time{}, err
+	}
+	return createdTime, nil
+}
+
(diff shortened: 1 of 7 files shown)


image/local.go
```diff
@@ -134,6 +134,16 @@ func (l *local) Digest() (string, error) {
 	return parts[1], nil
 }
 
+func (l *local) CreatedAt() (time.Time, error) {
+	createdAtTime := l.Inspect.Created
+	createdTime, err := time.Parse(time.RFC3339Nano, createdAtTime)
+
+	if err != nil {
+		return time.Now(), err
```
sclevine: Return empty time here instead of now. No need to make an extra syscall for undefined values.

image/remote.go
```diff
@@ -107,6 +107,14 @@ func (r *remote) Digest() (string, error) {
 	return hash.String(), nil
 }
 
+func (r *remote) CreatedAt() (time.Time, error) {
+	configFile, err := r.Image.ConfigFile()
+	if err != nil {
+		return time.Now(), fmt.Errorf("failed to get createdAt time for image '%s': %s", r.RepoName, err)
```
sclevine: Return empty time instead of now.

PR #120, opened 2019-04-12
Add volume cache support

ameyer-pivotal:
- Add cache interface
- Make image cache adhere to interface
- Add new volume cache implementation


cache/image_cache.go
```diff
@@ -0,0 +1,90 @@
...
+func (c *ImageCache) Name() string {
+	return c.origImage.Name()
+}
+
+func (c *ImageCache) SetMetadata(metadata lifecycle.CacheMetadata) error {
+	data, err := json.Marshal(metadata)
+	if err != nil {
+		return errors.Wrap(err, "marshall metadata")
+	}
+	return c.newImage.SetLabel(lifecycle.CacheMetadataLabel, string(data))
```
ekcasey: If the CacheMetadataLabel name only used in the cache package and is only relevant to one of the caching implementations the label name constant should probably move into the cache package

cache/image_cache.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (c *ImageCache) RetrieveMetadata() (lifecycle.CacheMetadata, bool, error) {
+	contents, err := lifecycle.GetMetadata(c.origImage, lifecycle.CacheMetadataLabel, c.logger)
+	if err != nil {
+		return lifecycle.CacheMetadata{}, false, err
+	}
+
+	metadata := lifecycle.CacheMetadata{}
+	if err := json.Unmarshal([]byte(contents), &metadata); err != nil {
+		c.logger.Printf("WARNING: image '%s' has incompatible '%s' label\n", c.origImage.Name(), lifecycle.CacheMetadataLabel)
```
ekcasey: I think this warning gets printed inappropriately when the origImage doesn't exist on a normal first build (in addition to a warning about the image not existing). That problem predates this PR but maybe we can fix it while we are moving this code around

cache/image_cache.go
```diff
@@ -0,0 +1,90 @@
...
+func (c *ImageCache) SetMetadata(metadata lifecycle.CacheMetadata) error {
+	data, err := json.Marshal(metadata)
+	if err != nil {
+		return errors.Wrap(err, "marshall metadata")
+	}
+	return c.newImage.SetLabel(lifecycle.CacheMetadataLabel, string(data))
+}
+
+func (c *ImageCache) RetrieveMetadata() (lifecycle.CacheMetadata, bool, error) {
+	contents, err := lifecycle.GetMetadata(c.origImage, lifecycle.CacheMetadataLabel, c.logger)
```
ekcasey: I feel a little weird about the cache package importing the lifecycle package. Is it worth it for this utility method?

cache/image_cache_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+	"github.com/buildpack/lifecycle"
+	"github.com/buildpack/lifecycle/image/fakes"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	"github.com/buildpack/lifecycle/cache"
+
+	h "github.com/buildpack/lifecycle/testhelpers"
```
ekcasey: fix the imports

cache/volume_cache.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := copyFile(tarPath, filepath.Join(c.stagingDir, sha+".tar")); err != nil {
+		return errors.Wrapf(err, "caching layer '%s' (%s)", identifier, sha)
+	}
+	return nil
+}
+
+func (c *VolumeCache) ReuseLayer(identifier string, sha string) error {
+	c.logger.Printf("reusing layer '%s' with diffID '%s'\n", identifier, sha)
+	if err := copyFile(filepath.Join(c.committedDir, sha+".tar"), filepath.Join(c.stagingDir, sha+".tar")); err != nil {
+		return errors.Wrapf(err, "reusing layer '%s' (%s)", identifier, sha)
```
ekcasey: maybe we should give a more helpful error message in the specific case where the copy fails because the reused layer doesn't exist in the cache. "layer missing from cache" or something like that may be more helpful than the error from os.Open

cache/volume_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	"testing"
+	"time"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	"github.com/buildpack/lifecycle"
+	"github.com/buildpack/lifecycle/cache"
+
+	h "github.com/buildpack/lifecycle/testhelpers"
```
ekcasey: consolidate the last two import groups

cmd/cacher/main.go
```diff
@@ -65,19 +69,30 @@ func cache() error {
...
 
-	origCacheImage, err := factory.NewLocal(cacheImageTag)
-	if err != nil {
-		return err
+		origCacheImage, err := factory.NewLocal(cacheImageTag)
+		if err != nil {
+			return err
+		}
+
+		cacherCache = cache.NewImageCache(cacher.Out, factory, origCacheImage)
```
ekcasey: cacherCache feels a little awkward, maybe cacheStore?

cache/image_cache.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (c *ImageCache) SetMetadata(metadata lifecycle.CacheMetadata) error {
+	data, err := json.Marshal(metadata)
+	if err != nil {
+		return errors.Wrap(err, "marshall metadata")
+	}
+	return c.newImage.SetLabel(lifecycle.CacheMetadataLabel, string(data))
+}
+
+func (c *ImageCache) RetrieveMetadata() (lifecycle.CacheMetadata, bool, error) {
```
sclevine: Is this `bool` really useful? Is `false` + empty struct ever any different from `true` + empty struct? If the distinction is necessary, name the parameters so it's clear what the `bool` is for.

cache/volume_cache.go
ameyer-pivotal: @ekcasey Wouldn't the error from os.Open say that the file is missing? I suppose we could be explicit and check for os.IsNotExist here

cache/volume_cache_test.go
ameyer-pivotal: @sclevine How did you hack into Emily's account...? ;P

cmd/cacher/main.go
ameyer-pivotal: Agreed. Will think of something different.

cache/image_cache.go
ameyer-pivotal: @sclevine Yeah we considered this. Ultimately we landed here because there didn't seem to be an idiomatic way to check for this value on the caller side, and using pointers also felt weird. We felt a bool was clearest. Will opt to name it if you're fine with that (unless you have a better idea).

cache/image_cache.go
ameyer-pivotal: @ekcasey I'm actually surprised we're not running into cyclic dependency issues here...

cache/image_cache.go
```diff
@@ -0,0 +1,90 @@
...
+	newImage  *image.LoggingImage
+}
+
+func NewImageCache(logger *log.Logger, factory ImageFactory, origImage image.Image) *ImageCache {
+	newImage := factory.NewEmptyLocal(origImage.Name())
+	return &ImageCache{
+		logger:    logger,
+		factory:   factory,
+		origImage: origImage,
+		newImage:  image.NewLoggingImage(logger, newImage),
```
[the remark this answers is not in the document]
ameyer-pivotal: I was thinking of something similar. I'll probably just remove it and move the logging into the cache implementation

cache/image_cache.go
ameyer-pivotal: I'll create a `metadata` package as I had mentioned before

cache/image_cache.go
ameyer-pivotal: @sclevine Ended up removing the bool. You're right that it wasn't super necessary

cache/image_cache.go
```diff
@@ -0,0 +1,90 @@
+package cache
+
+import (
+	"encoding/json"
+	"io"
+	"log"
+
+	"github.com/buildpack/lifecycle"
+	"github.com/buildpack/lifecycle/image"
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
ameyer-pivotal: Done.

cache/image_cache_test.go
ameyer-pivotal: Done.

cache/volume_cache.go
ameyer-pivotal: Wondering if this is enough an edge case to not worry about. Happy to handle it otherwise. If we did, would we need it anywhere else in this file (like AddLayer)?

cache/volume_cache_test.go
ameyer-pivotal: Imports fixed.

cache/image_cache.go
ameyer-pivotal: Done.

cache/image_cache.go
ameyer-pivotal: Done.

cache/image_cache.go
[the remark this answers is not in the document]
ameyer-pivotal: Done.

cache/image_cache.go
ameyer-pivotal: Done.

cmd/cacher/main.go
ameyer-pivotal: Done.

cache/image_cache.go
ameyer-pivotal: Done.

cache/image_cache.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (c *ImageCache) RetrieveMetadata() (Metadata, error) {
+	contents, err := metadata.GetRawMetadata(c.origImage, MetadataLabel)
+	if err != nil {
+		return Metadata{}, errors.Wrap(err, "retrieving metadata")
+	}
+
+	meta := Metadata{}
+	_ = json.Unmarshal([]byte(contents), &meta)
```
sclevine: Don't ignore this error:
```
return meta, json.Unmarshal([]byte(contents), &meta)
```

cache/image_cache_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+
```
sclevine: Remove newline.

cache/volume_cache.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return Metadata{}, nil
+		}
+		return Metadata{}, errors.Wrapf(err, "opening metadata file '%s'", metadataPath)
+	}
+	defer file.Close()
+
+	metadata := Metadata{}
+	_ = json.NewDecoder(file).Decode(&metadata)
```
sclevine: Same as above, don't ignore the error.

cache/volume_cache.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := os.Rename(c.committedDir, c.backupDir); err != nil {
+		return errors.Wrap(err, "backing up cache")
+	}
+	defer os.RemoveAll(c.backupDir)
+
+	if err := os.Rename(c.stagingDir, c.committedDir); err != nil {
+		if err := os.Rename(c.backupDir, c.committedDir); err != nil {
+			return errors.Wrap(err, "rolling back cache")
+		}
+		return nil
```
sclevine: Are we ignoring the error intentionally when the rename fails? If so, what condition are we trying to ignore?

exporter.go
```diff
@@ -21,45 +22,42 @@ type Exporter struct {
 	UID, GID     int
 }
 
-func (e *Exporter) Export(layersDir, appDir string, runImage, origImage image.Image, launcher string, stack StackMetadata) error {
+func (e *Exporter) Export(layersDir, appDir string, runImage, origImage image.Image, launcher string, stack metadata.StackMetadata) error {
 	var err error
 
-	metadata := AppImageMetadata{}
+	meta := metadata.AppImageMetadata{}
```
sclevine: In case it's helpful, I've seen `md` used as a common short name for metadata elsewhere.

cache/image_cache.go
ameyer-pivotal: @sclevine This is intentional for the cache behavior. In this case, if the metadata is malformed, we're returning an empty metadata object.

cache/volume_cache.go
ameyer-pivotal: Same as previous comment

cache/volume_cache.go
ameyer-pivotal: Yes, if committing the cache fails, we attempt a rollback (no error). If the rollback fails, we error.

ameyer-pivotal:
@sclevine Potentially. The `metadata` package came about in moving some of the cache implementations into their own `cache` package. Without `metadata` being its own package, we have a cyclic dependency (lifecycle -> cache -> lifecycle) vs (lifecycle -> cache -> metadata).

I do think it's a good move though to be more modular rather than having a monopackage. Eventually, we'd like to move some things to `internal`, and it makes sense to have it broken out (I'm assuming you can't have part of a package in `internal` though, so correct me if that's incorrect).

cache/image_cache_test.go
ameyer-pivotal: Done.

cache/image_cache.go
sclevine: if `json.Unmarshal` returns an error, you have no guarantee that the values returned are safe to use. I suggest checking for an error and returning empty struct (with no error) if you'd like to accept malformed metadata.

metadata/metadata.go
```diff
@@ -74,56 +68,28 @@ func (m *AppImageMetadata) metadataForBuildpack(id string) BuildpackMetadata {
...
 	} else if !found {
-		log.Printf("WARNING: image '%s' not found or requires authentication to access\n", image.Name())
 		return "", nil
 	}
 	contents, err := image.Label(metadataLabel)
 	if err != nil {
-		return "", err
+		return "", errors.Wrapf(err, "retrieving label '%s' for image '%s'", metadataLabel, image.Name())
 	}
 	if contents == "" {
```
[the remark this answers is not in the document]
ameyer-pivotal: Oops. I didn't catch that when I was removing the warnings... Thanks!

cache/volume_cache.go
sclevine: 👍

metadata/metadata.go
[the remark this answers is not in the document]
ameyer-pivotal: Done.

cache/image_cache.go
ameyer-pivotal: Gotcha. I didn't realize the same thing goes for `meta` since it's not technically returned. But I see your point and think it's clearer to do what you mentioned.

sclevine: Approving conditional on this change 👍 

For some more context: `json.Unmarshal` acts directly on the fields in `meta` (without copying it first or rolling back on failure), and doesn't reset anything when it encounters an error (it actually panics and runs up the stack).

ameyer-pivotal: Great point @sclevine. Thanks for the context. Done.

PR #124, opened 2019-04-17
Fake image tracks saved layers in a temp dir

djoyahoy:
- Copy all layers on save to a temp dir to mimic the
  actual behavior of an image
- Add cleanup method to clean up the temp dir


djoyahoy:
This will have to be ported to the upcoming `imgutil` repo but we'd like to get this in to reduce merge conflicts for some pack changes.

PR #125, opened 2019-04-17
Use external imgutil library

ameyer-pivotal:
- Also removes `image.Factory`


testmock/image.go
```diff
@@ -5,7 +5,7 @@
 package testmock
```
ameyer-pivotal: It's a little weird that we still have this since the interface is no longer in this repo. @djoyahoy Is going to eliminate it by moving to the fake image instead of mock image. It may or may not be part of this PR.

ameyer-pivotal: I suppose we could also duplicate the interface here on the consumer side (it will still be needed in imgutil as well though). Thoughts?

analyzer_test.go
```diff
@@ -473,7 +476,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 
 		when("the image cannot found", func() {
 			it.Before(func() {
-				image.EXPECT().Found().Return(false, nil)
+				image.Delete()
```
sclevine: Might want to check the error here.

analyzer_test.go
```diff
@@ -493,21 +496,9 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
-				err := analyzer.Analyze(image)
-				h.AssertError(t, err, "some-error")
-			})
-		})
-
 		when("the image does not have the required label", func() {
 			it.Before(func() {
-				image.EXPECT().Found().Return(true, nil)
-				image.EXPECT().Label("io.buildpacks.lifecycle.metadata").Return("", nil)
+				image.SetLabel("io.buildpacks.lifecycle.metadata", "")
```
sclevine: missing error check

analyzer_test.go
```diff
@@ -530,8 +521,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 
 		when("the image label has incompatible metadata", func() {
 			it.Before(func() {
-				image.EXPECT().Found().Return(true, nil)
-				image.EXPECT().Label("io.buildpacks.lifecycle.metadata").Return(`{["bad", "metadata"]}`, nil)
+				image.SetLabel("io.buildpacks.lifecycle.metadata", `{["bad", "metadata"]}`)
```
sclevine: missing error check

cache/image_cache.go
```diff
@@ -4,29 +4,24 @@ import (
...
-	origImage image.Image
-	newImage  image.Image
+	initializer func(string) imgutil.Image
+	origImage   imgutil.Image
+	newImage    imgutil.Image
 }
 
-func NewImageCache(factory ImageFactory, origImage image.Image) *ImageCache {
-	newImage := factory.NewEmptyLocal(origImage.Name())
+func NewImageCache(origImage imgutil.Image, initializer func(string) imgutil.Image) *ImageCache {
```
ekcasey: I think we should just inject the new empty image instead of injecting this function. It seems like it exists to enable generating a new image after Save happens, but the behavior there feel unintuitive to me. If I add x, and y layers to my cache object, then save, then add z layer to the same cache object I would expect the cache to have x, y, and z layers. Right now I think we would only end up with z layer in that final cache.

cache/image_cache.go
```diff
@@ -78,7 +73,7 @@ func (c *ImageCache) Commit() error {
 	}
 
 	c.origImage = c.newImage
-	c.newImage = c.factory.NewEmptyLocal(c.origImage.Name())
+	c.newImage = c.initializer(c.origImage.Name())
```
ekcasey: I think we should keep the original origImage and cache image after saving

ekcasey: Volume cache would need to change to behave analogously if we made the change I am suggesting

cache/image_cache.go
djoyahoy: @ekcasey I think that our cache doesn't work that way be design. Our cache is totally destructive unless you mark things you'd like to keep. In your example, I'd have to "reuse" x and y after my initial commit if wished to continue seeing x any y. I feel like the cache fell into this design due to the underlying design of the image library. To get different behavior, I think we might have to make a larger set of changes. 

I'm wondering if there's a solution where, for now, caches in lifecycle can only be used once per invocation of cacher or restorer.

PR #127, opened 2019-04-24
Exporter defaults to `stack.toml` run image

### LINKED ISSUE #119: exporter chooses run-image from stack.toml when -runImage not provided
Right now `-runImage` is a required flag on the `exporter` We should make this flag optional and exporter should choose an appropriate mirror from `stack.toml` when the flag is not provided. This will allow us to easily take advantage of run image mirrors in the knative template and other cases where the lifecycle run outside of pack.

joaopapereira:
Remove the required `--image` command line argument and default
to `stack.toml` configuration.
When the argument is passed it overrides the `stack.toml` value

Fixes #119

### DIFF
--- a/cmd/exporter/main.go
+++ b/cmd/exporter/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
@@ -51,8 +50,7 @@ func main() {
 
 	flag.Parse()
 	if flag.NArg() > 1 || flag.Arg(0) == "" {
-		args := map[string]interface{}{"narg": flag.NArg(), "layersDir": layersDir}
-		cmd.Exit(cmd.FailCode(cmd.CodeInvalidArgs, "parse arguments", fmt.Sprintf("%+v", args)))
+		cmd.Exit(cmd.FailCode(cmd.CodeInvalidArgs, "parse arguments"))
 	}
 	repoName = flag.Arg(0)
 	cmd.Exit(export())
@@ -91,8 +89,9 @@ func export() error {
 
 	if runImageRef == "" {
 		if stack.RunImage.Image == "" {
-			return cmd.FailCode(cmd.CodeInvalidArgs, "--run-image is required when stack.toml is not present in builder")
+			return cmd.FailCode(cmd.CodeInvalidArgs, "-image is required when there is no stack metadata available")
 		}
+
 		runImageRef = stack.RunImage.Image
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/exporter/main.go
```diff
@@ -87,17 +81,30 @@ func export() error {
...
-
 	var stack metadata.StackMetadata
 	_, err = toml.DecodeFile(stackPath, &stack)
 	if err != nil {
 		outLog.Printf("no stack.toml found at path '%s', stack metadata will not be exported\n", stackPath)
 	}
 
+	if runImageRef == "" {
+		if stack.RunImage.Image == "" {
+			return cmd.FailCode(cmd.CodeInvalidArgs, "--run-image is required when stack.toml is not present in builder")
```
ekcasey: The lifecycle flag is flag is `-image` rather than `--run-image`. Given that the `stack.toml` can have an arbitrary name if the path is supplied with a flag or an environment variable, perhaps we should say something more generic like "-image is required when there is no stack metadata available". Also, I'm not sure we should mention builders here, that feels like a platform specific concept.

cmd/exporter/main.go
```diff
@@ -87,17 +81,30 @@ func export() error {
...
+		}
+		runImageRef = stack.RunImage.Image
```
ekcasey: We want to select the best mirror (the one whose registry matches  `repoName`) instead of always using `stack.RunImage.Image`. `stack.RunImage.Image` can be the default if none of the mirrors have a matching registry.

joaopapereira:
@ekcasey I added the functionality of choosing the Run Image by registry inside the image package. But only after doing so I realize that it was split to a separated repo.

This feels like a functionality that should be in imageUtil, do you think I should create it in there and when it is merge have update this PR?

ekcasey:
@joaopapereira given that the only thing in imgutil right now is our image abstraction, I think this functionality fits better in the lifecycle repo

image/image.go
```diff
@@ -0,0 +1,32 @@
...
+		return "", errors.New("no images provided to search")
+	}
+
+	for _, i := range images {
+		reg, err := ParseRegistry(i)
+		if err != nil {
+			continue
+		}
+		if registry == reg {
+			return i, nil
```
ekcasey: This is a minor thing but I was confused the first time I glanced over this code b/c the variable name `i` led me to assume it was referring to the array index.

cmd/exporter/main.go
```diff
@@ -2,14 +2,14 @@ package main
...
 	"flag"
-	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
 
 	"github.com/BurntSushi/toml"
 	"github.com/buildpack/imgutil"
+	"github.com/buildpack/lifecycle/image"
```
ekcasey: we should group this with the other imports from lifecycle

joaopapereira:
@ekcasey done 😺

PR #134, opened 2019-05-13
Show group detection conditions first

jchesterpivotal:
Currently, when outputting from group detection, the name of the buildpack is displayed ahead of the condition. For example:

```
  Buildpack-1: pass
  Buildpack-with-longer-name: pass
  BP-short: fail
```

As the example shows, conditions are not vertically aligned. This makes it harder to visually scan for 'pass', 'fail' etc. It also makes it harder to write scripts that automatically parse lifecycle output.

This commit puts the conditions ahead of the buildpack name:

```
  pass: Buildpack-1
  pass: Buildpack-with-longer-name
  fail: BP-short
```

(This PR replaces the pig's breakfast that I and git-duet created in #133)

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -116,17 +116,17 @@ func (bg *BuildpackGroup) Detect(c *DetectConfig) (plan []byte, group *Buildpack
 		optional := bg.Buildpacks[i].Optional
 		switch code {
 		case CodeDetectPass:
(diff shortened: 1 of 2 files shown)


jchesterpivotal:
@jkutner @ekcasey FYI.

jchesterpivotal:
Are we awaiting a second approval?

PR #140, opened 2019-06-21
Allow for multiple tags

### LINKED ISSUE #138: Distinguish previous image tag from publish tag(s)
### Motivations:

This issue accomplishes three goals.

(1) Avoids unexpected behavior when a new image has been published to a given tag after `analyzer` has run, but before `exporter` has completed. Currently a buildpack may indicate (by leaving a `layer.toml` file with no corresponding directory) that it wishes to reuse a layer. However, if that layer changes due to this race condition, the layer that is reused might be different than the layer validated by the buildpack

(2) Provides an easy way to publish the same image to multiple tags (e.g. an immutable `rc-17` tag and a mutable `:latest` tag). This is important for auditability and rollbacks when using registries like artifactory that garbage collect untagged images.

(3) Explicitly disentangles two concepts which the current lifecycle API blurs together. The previous (analyzed) image and the desired export tag.

### `analyzer`

**When** I run `analyzer some/image:tag -analyzed /path/to/analyzed.toml`
**Then** `analyzer`, in additional to the current behavior, writes a `analyzed.toml` file to the given path with the following information
```
repository: some/image
digest: <image-digest>
[metadata]
...
```

### `exporter`
**Given** I have a file at path `/path/to/analyzed.toml` with the following contents
```
repository: some/image
digest: <image-digest>
[metadata]
...
```

**When** I run `exporter some/image:tag some/other:tag -analyzed /path/to/analyzed.toml`
**Then** the `exporter` will fetch reused layers from `some/image@<image-digest>`
**And** the `exporter` will export to all tags provided

### Flag details
`-analyzed` will be an optional flag on both `analyzer` and `exporter` and will default to `./analyzed.toml`

### Backwards compatibility
If exporter cannot find the `analyzed.toml` file it will fails (this means that a previous version on the `analyzer` binary cannot be used with the newest version of the `exporter` binary, which seems like an acceptable limitation).

jromero:
- Lock analyzed image and metadata for exporting
- Add support for additional image names (tags)

Resolves #138


exporter.go
```diff
@@ -114,33 +114,56 @@ func (e *Exporter) Export(layersDir, appDir string, runImage, origImage imgutil.
...
+	if digestErr != nil {
+		return errors.Wrap(digestErr, "getting digest")
 	}
-	e.Out.Printf("\n*** Image: %s@%s\n", runImage.Name(), sha)
+	e.Out.Printf("\n*** Digest: %s\n", digest)
 
-	return nil
+	e.Out.Println("*** Images:")
+	for _, n := range append([]string{workingImage.Name()}, additionalNames...) {
+		e.Out.Printf("      %s - %s\n", n, getSaveStatus(err, n))
```
ekcasey: I think we should avoid reinterpreting this error again later. Let's extract what we need if the first error handling block.

metadata/metadata.go
```diff
@@ -10,47 +10,60 @@ import (
...
+	Stack      StackMetadata       `json:"stack" toml:"stack"`
 }
 
 type AppMetadata struct {
-	SHA string `json:"sha"`
+	SHA string `json:"sha" toml:"sha"`
+}
+
+type AnalyzedMetadata struct {
+	Repository string           `toml:"repository"`
```
ekcasey: If we call this repository it should be just the repository instead of the entire tag reference. We should either break out the pieces of the reference or rename this.

metadata/metadata.go
```diff
@@ -10,47 +10,60 @@ import (
...
-	Version string                   `json:"version"`
-	Layers  map[string]LayerMetadata `json:"layers"`
+	ID      string                   `json:"key" toml:"key"`
+	Version string                   `json:"version" toml:"version"`
+	Layers  map[string]LayerMetadata `json:"layers" toml:"layers"`
 }
 
 type LayerMetadata struct {
-	SHA    string      `json:"sha" toml:"-"`
+	SHA    string      `json:"sha" toml:"sha"`
```
ekcasey: We don't want to write the layer sha to the layer metadata files in the layers dir. This might cause issues.

metadata/metadata.go
```diff
@@ -10,47 +10,60 @@ import (
...
 	Data   interface{} `json:"data" toml:"metadata"`
 	Build  bool        `json:"build" toml:"build"`
 	Launch bool        `json:"launch" toml:"launch"`
 	Cache  bool        `json:"cache" toml:"cache"`
 }
 
 type RunImageMetadata struct {
-	TopLayer string `json:"topLayer"`
```
ekcasey: hyphenate key for toml

hone:
Pretty excited for this!

ekcasey:
Due to degraded Travis availability we are going to merge w/o waiting for green Travis builds.

PR #151, opened 2019-07-15
Add -version flag to all binaries

### LINKED ISSUE #143: -version flag
### Motivation

As a user I may want to determine the provenance of a given lifecycle binary. Lifecycle binaries should be able to report their version and the git commit they were built from, like most other CLI tools.

### Release Candidates
**Given** a have built a lifecycle binary using `VERSION=0.4.0 make build`
**When** I run `<lifecycle-binary> -version`
**Then** The output should report the given version and git sha
Example:
```
$ lifecycle/out/exporter -version
0.4.0+f49a44e
```

### Local Development
**Given** a have built a lifecycle binary locally using `make build`
**When** I run `<lifecycle-binary> -version`
**Then** The output should report the version as dev
Example:
```
$ lifecycle/out/exporter -version
0.0.0+f49a44e
```

### Edge case
**Given** a lifecycle binary
**When** a user passes other flags or positional arguments in addition to the `-version` flag
**Then** all other arguments should be ignored and the lifecycle will print the version

jromero:
- Added acceptance test suite

Resolves #143 

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -1,0 +1,87 @@
+// +build acceptance
+
+package acceptance
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	h "github.com/buildpack/lifecycle/testhelpers"
+)
+
+func TestAcceptance(t *testing.T) {
+	spec.Run(t, "acceptance", testAcceptance, spec.Parallel(), spec.Report(report.Terminal{}))
+}
+
+func testAcceptance(t *testing.T, when spec.G, it spec.S) {
+
+	when("All", func() {
+		when("version flag is set", func() {
+			for _, data := range [][]string{
+				{"analyzer: only -version is present", "analyzer -version"},
+				{"analyzer: other params are set", "analyzer -daemon -version some/image"},
+
+				{"builder: only -version is present", "builder -version"},
+				{"builder: other params are set", "builder -app=/some/dir -version some/image"},
+
+				{"cacher: only -version is present", "cacher -version"},
+				{"cacher: other params are set", "cacher -path=/some/dir -version"},
+
+				{"detector: only -version is present", "detector -version"},
+				{"detector: other params are set", "detector -app=/some/dir -version"},
+
+				{"exporter: only -version is present", "exporter -version"},
+				{"exporter: other params are set", "exporter -analyzed=/some/file -version some/image"},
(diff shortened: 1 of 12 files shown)


lifecycle.go
```diff
@@ -23,3 +23,5 @@ var POSIXLaunchEnv = map[string][]string{
 	"bin": {"PATH"},
 	"lib": {"LD_LIBRARY_PATH"},
 }
+
+var VersionString string
```
sclevine: Suggest:
```
var BuildVersion = "0.0.0"
```

acceptance/acceptance_test.go
```diff
@@ -0,0 +1,88 @@
...
+	h "github.com/buildpack/lifecycle/testhelpers"
+)
+
+func TestAcceptance(t *testing.T) {
+	spec.Run(t, "acceptance", testAcceptance, spec.Sequential(), spec.Report(report.Terminal{}))
+}
+
+func testAcceptance(t *testing.T, when spec.G, it spec.S) {
+
+	var lifecycleCmd = func(name string, args ...string) *exec.Cmd {
```
sclevine: Separate into real function with `t` as first arg.

Also, for all testing helpers, call `t.Helper()`.

acceptance/acceptance_test.go
```diff
@@ -0,0 +1,87 @@
...
+	h "github.com/buildpack/lifecycle/testhelpers"
+)
+
+func TestAcceptance(t *testing.T) {
+	spec.Run(t, "acceptance", testAcceptance, spec.Sequential(), spec.Report(report.Terminal{}))
+}
+
+func testAcceptance(t *testing.T, when spec.G, it spec.S) {
+
+	when("All", func() {
```
sclevine: Why use two levels of nesting here?

acceptance/acceptance_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	h "github.com/buildpack/lifecycle/testhelpers"
+)
+
+func TestAcceptance(t *testing.T) {
+	spec.Run(t, "acceptance", testAcceptance, spec.Sequential(), spec.Report(report.Terminal{}))
```
sclevine: Drop `spec.Sequential()`, it's the default is only needed adjust the behavior in sub-Gs.

acceptance/acceptance_test.go
ameyer-pivotal: There will be later sections specific to each binary, but since this was common to all, we chose to group them here.

PR #159, opened 2019-07-26
Add launcher metadata to build metadata label

### LINKED ISSUE #146: Add launcher version to build metadata
As a user of Cloud Native Buildpacks, I may want to audit the contents of my built image. Given that the lifecycle adds the `launcher` to every generated dependency, a full audit will require the launcher version and metadata specifying the exact source code from which the launcher was built.

**Given** I am exporting an image using lifecycle version `<version>` built from a git commit with sha `<some-git-commit-sha>`
**When** `exporter` runs
**Then** `exporter` should add the following metadata to the `io.buildpacks.build.metadata` label

`{"launcher": {"version": "<version>", "source": {"git": {"repository": "github.com/buildpack/lifecycle", "commit": "<some-git-commit-sha>"}}}, ...}`

** This story assumes that the exporter launcher will be from the same lifecycle release as the other lifecycle binaries

jromero:
Resolves #146

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -52,7 +52,7 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
 							t.Error(err)
 						}
 
-						h.AssertStringContains(t, string(output), "some-version")
+						h.AssertStringContains(t, string(output), "some-version+asdf123")
 					})
 				})
 			}
@@ -64,7 +64,9 @@ func lifecycleCmd(t *testing.T, name string, args ...string) *exec.Cmd {
 	cmdArgs := append(
 		[]string{
 			"run",
-			"-ldflags", "-X github.com/buildpack/lifecycle/cmd.buildVersion=some-version",
+			"-ldflags",
+			"-X github.com/buildpack/lifecycle/cmd.Version=some-version " +
+				"-X github.com/buildpack/lifecycle/cmd.SCMCommit=asdf123",
 			"./cmd/" + name + "/main.go",
 		}, args...,
 	)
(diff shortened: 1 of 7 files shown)


acceptance/acceptance_test.go
```diff
@@ -64,7 +64,10 @@ func lifecycleCmd(t *testing.T, name string, args ...string) *exec.Cmd {
 	cmdArgs := append(
 		[]string{
 			"run",
-			"-ldflags", "-X github.com/buildpack/lifecycle/cmd.buildVersion=some-version",
+			"-ldflags",
+			"-X github.com/buildpack/lifecycle.Version=some-version " +
+				"-X github.com/buildpack/lifecycle.SCMCommit=asdf123 " +
+				"-X github.com/buildpack/lifecycle.SCMRepository=github.com/buildpack/lifecycle",
```
ekcasey: minor nitpick: if we aren't asserting on anything that uses SCMRepository it probably doesn't make sense to set it for the tests

cmd/cmd.go
```diff
@@ -134,11 +134,26 @@ const (
...
+	// SCMCommit is the commit information provided by SCM. It is injected at compile time.
+	SCMCommit = ""
+	// SCMRepository is the source repository. It is injected at compile time.
+	SCMRepository = ""
+	OutLogger     = log.New(os.Stdout, "", 0)
+	ErrLogger     = log.New(os.Stderr, "", 0)
 )
 
+// BuildVersion is a display format of the version and build metadata in compliance with semver.
+func BuildVersion() string {
```
ekcasey: does this function need to be exported?

PR #163, opened 2019-08-09
Load user-provided env vars by default

sclevine:
Implements: https://github.com/buildpack/rfcs/pull/14


sclevine:
Fixed the failure. I can't get it to fail without https://github.com/buildpack/lifecycle/pull/163/commits/9326554a4042c43c18c13807f32fd61df087a7ea on my mac, but 🤷‍♂

PR #165, opened 2019-08-09
Handle version resolution when latest symlink is present

### LINKED ISSUE #164: "too many buildpacks with matching ID" when latest symlink is present
**Reproduction**
* Builder created by `pack` master (0.4.0) and `lifecycle` master (0.4.0).
* Run on (pack 0.3.0) `pack build some/name -p . --buildpack ...`
* NOTICE: Detection error with "too many buildpacks with matching ID '...'"

When `pack` 0.4.0 creates a builder it also creates a symlink of "latest" for backwards compatibility causing version resolution in `lifecycle` to find multiple versions which are exactly the same but treats them as a conflict.

jromero:
Resolves #164 

### DIFF
--- a/compat/compat.go
+++ b/compat/compat.go
@@ -84,7 +84,7 @@ func resolveVersion(bpRef buildpackRefConfig, buildpacksDir string) (string, err
 		return "", err
 	}
 
-	var matchVersions []string
+	matchVersions := map[string]interface{}{}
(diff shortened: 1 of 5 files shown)


jromero:
No longer necessary, we are changing the logic on the platform `pack` to not produce a `latest` symlink making this issue moot.

PR #168, opened 2019-08-19
rename providers to provides

pyama86:
Hi guys!
According to the specs, the correct spelling should be `provided`, not `providers`.
https://github.com/buildpack/spec/blob/master/buildpack.md#build-plan-toml

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -45,7 +45,7 @@ type BuildPlan struct {
 }
 
 type BuildPlanEntry struct {
-	Providers []Buildpack `toml:"providers"`
+	Providers []Buildpack `toml:"provides"`
 	Requires  []Require   `toml:"requires"`
 }
 


sclevine:
The struct names are definitely confusing, but the fields names are not wrong. `BuildPlan` and `BuildPlanEntry` structs are used for the output "Build Plan" from executions of `/bin/detect`, before the entries are fed into executions of `/bin/build`. The matching fields from the spec are actually here: https://github.com/buildpack/lifecycle/blob/master/detector.go#L341

I'm hoping to clean up the struct names before the next release.

pyama86:
Thank you for your response.
I was mistaken and I thought that the build plan would be linked from `/bin/detect` to `/bin/build`.
If the column name of toml was wrong and it would not parse to well, but i understand your message.
thanks.

PR #173, opened 2019-08-29
Truncate SHAs when displaying them

jkutner:
This adds a util function that will truncate SHAs and remove the `sha256:` prefix. This is similar to what commands like `docker ps` and `docker images` do, where they truncate the sha to 12 characters.

We might also add `--no-trunc` option as with the [docker commands](https://docs.docker.com/engine/reference/commandline/ps/).

Alternatively, we could remove the SHAs from the logs entirely, and only show them when a verbose flag is passed.

End result looks like:

```
[exporter] Reusing layers from image with id 32c0a8586f36
[exporter] Exporting layer 'app' with SHA 75b3f03ac61d
[exporter] Reusing layer 'config' with SHA 227f48eccef2
[exporter] Exporting layer 'launcher' with SHA 942ac40d4b66
[exporter] Reusing layer 'heroku/java:jre' with SHA 3cec1444bdec
[exporter] *** Images:
[exporter]       index.docker.io/library/myimage:latest - succeeded
[exporter]
[exporter] *** Image ID: ecc176797be5
```

### DIFF
--- a/cacher.go
+++ b/cacher.go
@@ -80,10 +80,10 @@ func (c *Cacher) addOrReuseLayer(cache Cache, layer bpLayer, previousSHA string)
 	}
 
 	if sha == previousSHA {
-		c.Out.Printf("Reusing layer '%s' with SHA %s\n", layer.Identifier(), sha)
+		c.Out.Printf("Reusing layer '%s' with SHA %s\n", layer.Identifier(), DisplaySha(sha))
 		return sha, cache.ReuseLayer(previousSHA)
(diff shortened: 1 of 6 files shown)


sclevine:
Unfortunately, the short SHAs are only useful with local, uncompressed layer/image digests. Docker registries require a fully-qualified SHA in all cases. Maybe we could hide all but the final image digest (in the registry case) or final image ID (in the daemon case) behind a verbose flag?

PR #179, opened 2019-09-08
Add logging levels

jkutner:
@sclevine @ekcasey @ameyer-pivotal @jromero before i go any further, can you confirm that this is the along the lines of what we discussed?


detector.go
```diff
@@ -160,7 +161,7 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
 }
 
 func (c *DetectConfig) runTrial(i int, trial detectTrial) (depMap, detectTrial, error) {
-	c.Out.Printf("Resolving plan... (try #%d)", i)
+	c.Logger.Infof("Resolving plan... (try #%d)", i)
```
sclevine: `Debugf`?

internal/logging/logging.go
```diff
@@ -0,0 +1,134 @@
...
+		lw.Level = log.InfoLevel
+	} else if f == DebugLevel {
+		lw.Level = log.DebugLevel
+	} else {
+		lw.Level = log.ErrorLevel
+	}
+}
+
+// NewLogWithWriters creates a logger to be used with pack CLI.
+func NewLogWithWriters() *logWithWriters {
```
sclevine: Returning an unexported struct can make life difficult for the consumer. They are unable to type the object they receive. Instead, return the struct through the interface if there's a standard interface for it (`log.Logger`). 

It's often an anti-pattern to define an interface and unexported struct in the same package. But not in this case, where the purpose is to a define a standard logging interface, and provide one possible implementation of it.

sclevine:
Looking great to me 😄

internal/logging/logging.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	}
+
+	if h.wantTime {
+		ts := h.timer().Format(timeFmt)
+		_, _ = fmt.Fprint(h.writer, appendMissingLineFeed(fmt.Sprintf("%s %s%-25s", ts, formatLevel(e.Level), e.Message)))
+		return nil
+	}
+
+	_, _ = fmt.Fprint(h.writer, appendMissingLineFeed(fmt.Sprintf("%s%-25s", formatLevel(e.Level), e.Message)))
```
jromero: What's the purpose of the 25 character padding? We ended up removing it in `pack` because it made testing/asserting slightly more difficult.

internal/logging/logging.go
```diff
@@ -0,0 +1,134 @@
+// Package logging implements the logger for the pack CLI.
```
jromero: Is the intent that `pack` would move to start using this implementation by having `lifecycle` as a dependency so that there is no duplication (and maintenance overhead)?

internal/logging/logging.go
```diff
@@ -0,0 +1,134 @@
...
+func (lw *logWithWriters) WantTime(f bool) {
+	lw.handler.wantTime = f
+}
+
+func (lw *logWithWriters) WantLevel(f string) {
+	if f == InfoLevel {
+		lw.Level = log.InfoLevel
+	} else if f == DebugLevel {
+		lw.Level = log.DebugLevel
+	} else {
```
jromero: I would think that there would be a case for someone to want to see `Warnings` as well as `Errors` but not `Info` or `Debug`. Is there a reason why that option isn't available?

internal/logging/logging.go
jkutner: yea, i was wondering about that, but I was cribbing from `pack`. i'll fix here, but wonder about pack too

internal/logging/logging.go
jkutner: idk. maybe we split the logging stuff out as its only lib?

internal/logging/logging.go
jkutner: hmm, actually, I do need to invoke `WantLevel` outside this package. Should we just export `logWithWriters`?

internal/logging/logging.go
jromero: @ekcasey what are your thoughts here?

internal/logging/logging.go
sclevine: Could all external input be accomplished via the interface or via the constructor?

sclevine: E.g., you could use functional options in the constructor: 
https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

jkutner: Yea, I can do that, but then I'm really starting to diverge from what we're doing in pack. Is that ok?

jkutner:
@sclevine @ekcasey Can you help figure out what I need to do to move this forward? 

- Should I break from the pattern pack uses for `NewLogWithWriters`?
- Do we need to split out a logging lib, or wait?
- Do we need to add log levels to all binaries, or can we merge just these first? (I don't actually think `/lifecycle/builder` will work the same way because we might want to preserve stdout/stderr from the buildpack, so I'd like to wait).

internal/logging/logging.go
ameyer-pivotal: I considered the same thing way back when we began talking about a similar logger in lifecycle. I think a separate repo might have some overhead with maintenance, but not sure if I like pack having to continue to depend on lifecycle...

internal/logging/logging.go
ameyer-pivotal: I think it's a good pattern and we'll probably want to move pack towards it anyway.

jkutner: The problem I'm up against is that `NewLogWithWriters()` is called before we know the logging level. I can remove `cmd.Logger` and have each `main.go` create it's own instance once it gets the flag. Or is there another way?

jromero: > Yea, I can do that, but then I'm really starting to diverge from what we're doing in pack. Is that ok?

... and here start the problems... We've made minor changes to pack logger (some more in review) since this PR opened. We really need to determine what the best way of minimizing maintenance.

jkutner: I've added `WantLevel(string)` to the `Logger` interface.

internal/logging/logging.go
jromero: I think we could consider this issue resolved. It seems like we are aligned on what the future would look like but not part of this issue.

jkutner:
This is what it looks like in practice:


```
$ pack build myimage --builder heroku/buildpacks:18 --no-pull
Selected run image heroku/pack:18
Executing lifecycle version 0.0.0
Using build cache volume pack-cache-67f7abf58243.build
===> DETECTING
[detector] Success! (1)
===> RESTORING
[restorer] Restoring cached layer 'heroku/java:jdk'
[restorer] Restoring cached layer 'heroku/java:maven_m2'
===> ANALYZING
[analyzer] Analyzing image '3f066b64904aecb1541ad7468cf13efe913e068af7a1373a72bd4c352e926782'
[analyzer] Using cached layer 'heroku/java:jdk'
[analyzer] Using cached layer 'heroku/java:maven_m2'
[analyzer] Writing metadata for uncached layer 'heroku/java:jre'
===> BUILDING
[builder]
...
[builder]
===> EXPORTING
[exporter] Reusing layers from image with id '3f066b64904aecb1541ad7468cf13efe913e068af7a1373a72bd4c352e926782'
[exporter] *** Images:
[exporter]       index.docker.io/library/myimage:latest - succeeded
[exporter]
[exporter] *** Image ID: e6c614dd4f6d9da837391533c647fc562fbeffe2c29f8ae5800fd84f804dcec6
===> CACHING
[cacher] Reusing layer 'heroku/java:jdk'
[cacher] Caching layer 'heroku/java:maven_m2'
Successfully built image myimage
```

cacher.go
```diff
@@ -80,10 +80,12 @@ func (c *Cacher) addOrReuseLayer(cache Cache, layer bpLayer, previousSHA string)
 	}
 
 	if sha == previousSHA {
-		c.Out.Printf("Reusing layer '%s' with SHA %s\n", layer.Identifier(), sha)
+		c.Logger.Infof("Reusing layer '%s'", layer.Identifier())
+		c.Logger.Debugf("  - with SHA %s\n", sha)
```
jromero: This seems a little odd to me to be honest. Log messages should be able to stand on their own. The fact that there is no new line character in the `Infof` call might not work as you'd expect based on some of the opinions we've made in certain logger implementations. I believe we add a new line character per log message if not present when writing out.

IMO, a better approach would be two log messages.

```go
c.Logger.Infof("Reusing layer '%s'\n'", layer.Identifier())
c.Logger.Debugf("Layer '%s' SHA: %s\n", layer.Identifier(), sha)
```

cacher.go
```diff
@@ -80,10 +80,12 @@ func (c *Cacher) addOrReuseLayer(cache Cache, layer bpLayer, previousSHA string)
...
 		return sha, cache.ReuseLayer(previousSHA)
 	}
 
-	c.Out.Printf("Caching layer '%s' with SHA %s\n", layer.Identifier(), sha)
+	c.Logger.Infof("Caching layer '%s'\n", layer.Identifier())
+	c.Logger.Debugf("  - with SHA %s\n", sha)
```
jromero: Ditto

detector.go
```diff
@@ -101,36 +102,36 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
...
-			c.Out.Printf("pass: %s", bp)
+			c.Logger.Debugf("pass: %s", bp)
 			results = append(results, detectResult{bp, run})
 		case CodeDetectFail:
 			if bp.Optional {
-				c.Out.Printf("skip: %s", bp)
+				c.Logger.Debugf("skip: %s", bp)
 			} else {
-				c.Out.Printf("fail: %s", bp)
+				c.Logger.Debugf("fail: %s", bp)
```
jromero: Example of where no new line characters are present yet they should be treated as such when writing out.

PR #183, opened 2019-09-24
Make logs more user friendly

jkutner:
Supersedes #173

Here's an example of the info level logging:

```
$ pack build myimage --builder heroku/buildpacks:18 --no-pull
===> DETECTING
[detector] heroku/java 0.14
===> RESTORING
[restorer] Restoring cached layer 'heroku/java:jdk'
[restorer] Restoring cached layer 'heroku/java:maven_m2'
===> ANALYZING
[analyzer] Using cached layer 'heroku/java:jdk'
[analyzer] Using cached layer 'heroku/java:maven_m2'
[analyzer] Writing metadata for uncached layer 'heroku/java:jre'
===> BUILDING
[builder]
...
[builder]
===> EXPORTING
[exporter] Adding layer 'app'
[exporter] Reusing layer 'config'
[exporter] Adding layer 'launcher'
[exporter] Reusing layer 'heroku/java:jre'
[exporter] *** Images:
[exporter]       index.docker.io/library/myimage:latest (a94409e2e599)
===> CACHING
[cacher] Reusing layer 'heroku/java:jdk'
[cacher] Caching layer 'heroku/java:maven_m2'
Successfully built image myimage
```

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -21,7 +21,7 @@ const (
 	CodeDetectFail = 100
 )
 
-var ErrFail = errors.New("No buildpack groups passed detection.\nPlease check that you are running against the correct path.")
+var ErrFail = errors.New("no buildpacks participating")
 
 type Buildpack struct {
 	ID       string `toml:"id" json:"id"`
--- a/detector.go
+++ b/detector.go
@@ -19,9 +19,11 @@ import (
 const (
 	CodeDetectPass = 0
 	CodeDetectFail = 100
+
+	DetectFailureMessage = "None of the provided buildpack groups passed detection for this repository.\nPlease check that you are running against the correct path."
 )
 
-var ErrFail = errors.New("detection failed")
+var ErrFail = errors.New("no buildpacks match this repository")
 
 type Buildpack struct {
 	ID       string `toml:"id" json:"id"`
@@ -146,6 +148,7 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
 		return c.runTrial(i, trial)
 	})
 	if err != nil {
+		c.Logger.Info(DetectFailureMessage)
 		return nil, nil, err
 	}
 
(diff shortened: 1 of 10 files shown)


jkutner:
Rebased on top of #184 while keeping the detect output from #184

detector.go
```diff
@@ -21,7 +21,7 @@ const (
 	CodeDetectFail = 100
 )
 
-var ErrFail = errors.New("detection failed")
+var ErrFail = errors.New("No buildpack groups passed detection.\nPlease check that you are running against the correct path.")
```
jromero: This seems like it's intended for UI/UX. It doesn't follow the standard go conventions (ex. start with logger case). The fact that it has a new line character is a little strange too. 

I'm curious if a better approach would be to write out these messages as part of the `cmd/detector` by checking if `err == ErrorFail` somewhere around here: https://github.com/buildpack/lifecycle/blob/9d0ea0f8f5b934f8dec2fbb09e2f618e1c44c2b7/cmd/detector/main.go#L82-L84

exporter.go
```diff
@@ -217,19 +215,18 @@ func (e *Exporter) saveImage(image imgutil.Image, additionalNames []string) erro
...
-		e.Logger.Infof("\n*** Digest: %s\n", v.Digest.DigestStr())
-	default:
-		e.Logger.Infof("\n*** Reference: %s\n", v.String())
+	refType, ref := getReference(id)
+	e.Logger.Info("*** Images:")
+	for _, n := range append([]string{image.Name()}, additionalNames...) {
+		if ok, message := getSaveStatus(saveErr, n); !ok {
+			e.Logger.Infof("      %s - %s\n", n, message)
+		} else {
+			e.Logger.Infof("      %s (%s)\n", n, TruncateSha(ref))
```
jromero: It seems a little strange that we'll be displaying the same value over and over for each image tag. Not sure what a better solution would be. Just calling it out. :D

jkutner: I have mixed feelings and struggled with this. In some sense, it's good that each line stands on its own. And most people won't be building with more than one tag i imagine. I think like Ben's detect output, this is not an end state. We'll keep improving.

detector.go
jkutner: 👍

PR #185, opened 2019-09-30
Add `rebase` to lifecycle

ashwin-venkatesh:
- rename metdata -> metadata
- move RunImageMirror to metadata package

Signed-off-by: Ashwin Venkatesh <avenkatesh@pivotal.io>


metadata/layers.go
```diff
@@ -65,6 +66,19 @@ type StackRunImageMetadata struct {
 	Mirrors []string `toml:"mirrors" json:"mirrors,omitempty"`
 }
 
+func (sm *StackMetadata) BestRunImageMirror(registry string) (string, error) {
```
jromero: Minor: Personally I think it made more sense to be a standalone function instead of it being on the `StackMetadata` but it's only aesthetics.

Makefile
```diff
@@ -32,6 +32,7 @@ build:
 	$(GOENV) $(GOBUILD) -o ./out/lifecycle/exporter -a ./cmd/exporter
 	$(GOENV) $(GOBUILD) -o ./out/lifecycle/cacher -a ./cmd/cacher
 	$(GOENV) $(GOBUILD) -o ./out/lifecycle/launcher -a ./cmd/launcher
+	$(GOENV) $(GOBUILD) -o ./out/lifecycle/rebaser -a ./cmd/rebaser
```
jromero: We probably want to document this new artifact in our README.md

ashwin-venkatesh: ohh..good point. On it.

metadata/layers.go
ashwin-venkatesh: The decision to put it on the StackMetadata was on the back of the idea that it has all the RunImage + Mirrors information so it felt like the right receiver. But as you mentioned, it is an aesthetics thing :)

cmd/rebaser/main.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	newBaseImage, err := newImage(runImageRef)
+	if err != nil || !newBaseImage.Found() {
+		return cmd.FailErr(err, "access run image")
+	}
+
+	if err := rebaser.Rebase(appImage, newBaseImage, imageNames[1:]); err != nil {
+		if _, isSaveError := err.(*imgutil.SaveError); isSaveError {
```
sclevine: Could use `_, ok` idiom instead of `_, isSaveError`.

rebaser.go
```diff
@@ -0,0 +1,53 @@
...
+		return errors.Wrap(err, "get run image id or digest")
+	}
+	origMetadata.RunImage.Reference = identifier.String()
+
+	data, err := json.Marshal(origMetadata)
+	if err != nil {
+		return errors.Wrap(err, "marshall metadata")
+	}
+
+	if err = workingImage.SetLabel(metadata.LayerMetadataLabel, string(data)); err != nil {
```
sclevine: Using `if err =` here instead of `if err :=` suggests that you're trying to preserve the error in the outer context.

cmd/rebaser/main.go
ashwin-venkatesh: Given we are casting the error, I like the suggested change.

rebaser.go
ashwin-venkatesh: Good catch. I prefer the consistency of `:=` across the file as well.

PR #186, opened 2019-10-03
Rebase fails if new base image has a different stack

ashwin-venkatesh:
Ensure rebase does not take place if the new base image is based on a different stack

### DIFF
--- a/rebaser.go
+++ b/rebaser.go
@@ -2,6 +2,7 @@ package lifecycle
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/buildpack/imgutil"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


rebaser_test.go
```diff
@@ -104,5 +104,16 @@ func testRebaser(t *testing.T, when spec.G, it spec.S) {
...
+		when("app image and run image are based on different stacks", func() {
+			it("returns an error and prevents the rebase from taking place", func() {
+				h.AssertNil(t, fakeWorkingImage.SetLabel(metadata.StackMetadataLabel, "io.buildpacks.stacks.bionic"))
+
+				h.AssertNil(t, fakeNewBaseImage.SetLabel(metadata.StackMetadataLabel, "io.buildpacks.stacks.cflinuxfs3"))
+
+				err := rebaser.Rebase(fakeWorkingImage, fakeNewBaseImage, additionalNames)
+				h.AssertError(t, err, "incompatible stack")
+			})
+
```
ekcasey: nit: extra space

rebaser_test.go
```diff
@@ -104,5 +104,16 @@ func testRebaser(t *testing.T, when spec.G, it spec.S) {
 				h.AssertEq(t, md.Buildpacks[0].ID, "buildpack.id")
 			})
 		})
```
ekcasey: nit: space needed below this line

rebaser.go
```diff
@@ -24,6 +24,20 @@ func (r *Rebaser) Rebase(
...
+		return errors.Wrap(err, "get working image stack")
+	}
+
+	newBaseImageStack, err := metadata.GetStackMetadata(newBaseImage)
+	if err != nil {
+		return errors.Wrap(err, "get new base image stack")
+	}
+
+	if workingImageStack.Id != newBaseImageStack.Id {
+		return errors.New("incompatible stack")
```
jromero: We should probably output a little bit more information. Something like `incompatible stack: stack 'stack.new' is not compatible with 'stack.original'`

ashwin-venkatesh: Done

rebaser.go
```diff
@@ -24,6 +24,20 @@ func (r *Rebaser) Rebase(
...
 	}
 
+	workingImageStack, err := metadata.GetStackMetadata(workingImage)
+	if err != nil {
+		return errors.Wrap(err, "get working image stack")
+	}
+
+	newBaseImageStack, err := metadata.GetStackMetadata(newBaseImage)
+	if err != nil {
+		return errors.Wrap(err, "get new base image stack")
```
ekcasey: @ashwin-venkatesh I don't think this handles the empty label case. Unhappy path tests around empty labels should drive that out.

PR #195, opened 2019-10-14
Clean up usage of assertNil in tests

lukasberger:
exporter_test.go referred to non-existent 'assertNil'
analyzer_test.go re-implemented a function from testhelpers


jromero:
Thanks for the contribution!

It's nice to see the improvements.

PR #198, opened 2019-10-15
Clean up logging

ekcasey:
- Use heroku/color for colors
- Use apex/logger for base logger
- Fix compiling for Windows

Signed-off-by: Micah Young <myoung@pivotal.io>


detector_test.go
```diff
@@ -42,14 +41,13 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
 
-		outLog = &bytes.Buffer{}
 		config = &lifecycle.DetectConfig{
 			FullEnv:       append(os.Environ(), "ENV_TYPE=full"),
 			ClearEnv:      append(os.Environ(), "ENV_TYPE=clear"),
 			AppDir:        appDir,
 			PlatformDir:   platformDir,
 			BuildpacksDir: buildpacksDir,
-			Logger:        mocks.NewMockLogger(io.MultiWriter(outLog, it.Out())),
+			Logger:        &log.Logger{Handler: logHandler},
```
sclevine: Did you intend to drop the test output (`it.Out()`)?

detector_test.go
```diff
@@ -25,9 +24,9 @@ func TestDetector(t *testing.T) {
 func testDetector(t *testing.T, when spec.G, it spec.S) {
 	var (
 		config      *lifecycle.DetectConfig
-		outLog      *bytes.Buffer
 		platformDir string
 		tmpDir      string
+		logHandler  = &memory.Handler{}
```
sclevine: Things in this context are executed many times and not shared across all `it`s. Was this intentional?

cmd/flags.go
```diff
@@ -26,7 +22,7 @@ const (
 	DefaultPlanPath      = "./plan.toml"
 	DefaultProcessType   = "web"
 	DefaultLauncherPath  = "/cnb/lifecycle/launcher"
-	DefaultLogLevel      = logging.InfoLevel
+	DefaultLogLevel      = "info"
```
jkutner: why not reference a constant here?

cmd/flags.go
```diff
@@ -132,41 +129,18 @@ func FlagVersion(version *bool) {
 }
 
 func FlagLogLevel(level *string) {
-	flag.StringVar(level, "log-level", DefaultLogLevel, "set the logging level")
+	flag.StringVar(level, "log-level", envOrDefault(EnvLogLevel, DefaultLogLevel), "logging level")
```
jkutner: does "set the" or similar get added somewhere else in the help/description. "logging level" on it's own doesn't add any value of "log-level".

cmd/logs.go
```diff
@@ -0,0 +1,63 @@
...
+	Logger = &log.Logger{
+		Handler: &handler{
+			writer: os.Stdout,
+		},
+	}
+	warnStyle  = color.New(color.FgYellow, color.Bold).SprintfFunc()
+	errorStyle = color.New(color.FgRed, color.Bold).SprintfFunc()
+)
+
+func SetLogLevel(level string) {
```
jkutner: why not `func (l *Logger) SetLogLevel(level string)`?

detector.go
```diff
@@ -76,7 +74,7 @@ type DetectConfig struct {
 	AppDir        string
 	PlatformDir   string
 	BuildpacksDir string
-	Logger        logging.Logger
+	Logger        Logger
```
jkutner: does using unscoped `Logger` for both the instance and the type create a risk of collision?

detector_test.go
ekcasey: I don't believe we want to share the same `logHandler` across all `it`s

cmd/flags.go
ekcasey: We changed it to be consistent with the other flag help texts which set things but don't begin with "set the". I'm not sure "set the logging level" or "logging level" add anything over "log-level", but I am happy to change it back or just remove the help text entirely.

detector_test.go
micahyoung: Fixed

detector.go
ekcasey: Given that the instance is a field I'm not worried about collision. Defining the interface type in the package that consumes it, rather than the package that provides the implementation is idiomatic.

cmd/logs.go
ekcasey: We didn't define our own Logger type. We are using an apex `*log.Logger` directly. Given that we have a package level Logger variable, a package level function to set the log level makes sense to me. Happy to discuss.

detector_test.go
ekcasey: We haven't been writing the log output to `it.Out()` in the other tests files. We should probably make them all consistent one way or another.

cmd/flags.go
micahyoung: apex/log unfortunately does not expose these constants but we do end up using apex for doing all the validating and parsing. If we were to add our own constants for the each possible value, we would only ever reference `Info`. Knowing those constraints, do you have any strong preferences?

cmd/flags.go
jkutner: /shrug idk, given how few people use lifecycle directly it probably doesn't matter rn

PR #201, opened 2019-10-16
Add missing directory when building

ameyer-pivotal:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,14 +24,17 @@ build: build-linux
 
 build-macos:
 	@echo "> Building for macos..."
+	mkdir -p ./out/lifecycle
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,14 +24,17 @@ build: build-linux
 
 build-macos:
 	@echo "> Building for macos..."
+	mkdir -p ./out/lifecycle
```
ekcasey: I think we should build into `./out/lifecycle-<os>` dirs, to avoid surprising results or `*.exe` files in our lifecycle archives. We may also want packaging targets for different OSes but given we don't have a use case yet we can hold on that.

ekcasey:
I'll approve given the issue I mentioned already existed

Makefile
ameyer-pivotal: Deferring to separate PR, as that issue existed previously

PR #203, opened 2019-10-22
Fix permissions when untarring files with umask

lukasberger:
Umask can change the permissions of the file being written leading to the created file having different permissions than it should according to the tar archive.

For example, with umask 027, two tests fail:

```
Suite: tar
Total: 9 | Focused: 0 | Pending: 0
x.x......
Passed: 7 | Failed: 2 | Skipped: 0

--- FAIL: TestTar (0.00s)
    --- FAIL: TestTar/tar (0.00s)
        --- FAIL: TestTar/tar/#ReadTarArchive/extracts_a_tar_file (0.00s)
            tar_test.go:77:   os.FileMode(
                -       s"-rw-r-----",
                +       s"-rw-r--r--",
                  )

        --- FAIL: TestTar/tar/#ReadTarArchive/doesn't_alter_permissions_of_existing_folders (0.00s)
            tar_test.go:92:   os.FileMode(
                -       s"drwxr-----",
                +       s"drwxr--r--",
                  )

FAIL
FAIL    command-line-arguments  0.073s
```

### DIFF
--- a/archive/tar.go
+++ b/archive/tar.go
@@ -147,6 +147,10 @@ func Untar(r io.Reader, dest string) error {
 			if err := writeFile(tr, path, hdr.FileInfo().Mode()); err != nil {
 				return err
 			}
+			// Update permissions in case umask was applied.
+			if err := os.Chmod(path, hdr.FileInfo().Mode()); err != nil {
+				return err
+			}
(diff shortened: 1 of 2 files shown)


sclevine:
@lukasberger Great find! Merging now

PR #204, opened 2019-10-22
Combined exporter cacher

lukasberger:
Combines the exporter and cacher phases into one exporter phase to layer tarballs.

* Move logic from cacher.go to exporter.go.
* Move tests from cacher_test.go to exporter_test.go.
* Rename cache image flag from '-image' to '-cache-image' and '-path' to '-cache-dir'.
* Remove unused 'In' field of the Exporter struct.
* Remove cacher implementation, tests, and command.

Part of the implementation of https://github.com/buildpack/rfcs/pull/21.


exporter.go
```diff
@@ -40,26 +53,26 @@ func (e *Exporter) Export(
 	var err error
 
 	meta := metadata.LayersMetadata{}
-
 	meta.RunImage.TopLayer, err = workingImage.TopLayer()
 	if err != nil {
 		return errors.Wrap(err, "get run image top layer SHA")
 	}
 
+	// TODO why are we saving stack in addition to runImageRef when the two can be different?
```
lukasberger: Question to the reviewers. Since metadata.Stack only contains the run image which may or may not have been used, why is it saved in addition to the actual run image in metadata.RunImage?

PR #205, opened 2019-10-24
Combine exporter and cacher phases

lukasberger:
Combines the exporter and cacher phases to reuse layer tarballs.

* Move logic from cacher.go to exporter.go.
* Move tests from cacher_test.go to exporter_test.go.
* Rename cache image flag from '-image' to '-cache-image'.
* Rename cache directory flag from '-path' to '-cache-dir'.
* Remove unused 'In' field of the Exporter struct.
* Remove cacher implementation, tests, and command.
* Upgrade Platform API version to 0.2 as this PR contains breaking changes.

Part of the implementation of buildpack/rfcs#21.

Signed-off-by: Lukas Berger <bergerl@google.com>


exporter.go
```diff
@@ -40,26 +53,26 @@ func (e *Exporter) Export(
 	var err error
 
 	meta := metadata.LayersMetadata{}
-
 	meta.RunImage.TopLayer, err = workingImage.TopLayer()
 	if err != nil {
 		return errors.Wrap(err, "get run image top layer SHA")
 	}
 
+	// TODO why are we saving stack in addition to runImageRef when the two can be different?
```
lukasberger: Question to the reviewers. Since metadata.Stack only contains the run image which may or may not have been used, why is it saved in addition to the actual run image in metadata.RunImage?

cmd/flags.go
```diff
@@ -61,11 +61,11 @@ func FlagBuildpacksDir(dir *string) {
 }
 
 func FlagCacheDir(dir *string) {
-	flag.StringVar(dir, "path", os.Getenv(EnvCacheDir), "path to cache directory")
+	flag.StringVar(dir, "cache-dir", os.Getenv(EnvCacheDir), "path to cache directory")
```
jkutner: could we accept `-path` or `-cache-dir` to prevent a hard break?

jkutner:
This will mean a breaking change for `pack`, which is probably fine, but we could consider avoiding it or allowing a grace period (maybe until we can combine it with other breaking changes). We could do the following methinks: 

* Leave the `cacher` in place
* Allow `-path` on `restorer` but ignore it (or use it when `-cache-dir` isn't provided)
* Because `pack` won't pass `-cache-image` to `exporter` it won't create the cache

I think we're probably still at an ok place to make breaking changes, but if this were to be the only breaking in change in a release I would want us to consider ^

cmd/flags.go
sclevine: `cacher` is going away entirely, so this is going to be a hard break for everyone anyways.

Unrelated comment: In the RFC we discussed this being `-cache-volume`, and it seems to have been changed to `-cache-dir` here. (Perhaps this is because it doesn't need to be a volume?) It might be worth pointing out that for the rest of the flags, we don't append `-dir` or `-path` the argument takes a path. Would `-cache` be more appropriate?

exporter.go
sclevine: It looks like `metadata.Stack.RunImage` is the set of (intentionally mutable) run image tags (including mirrors) that the rebaser should consider when selecting a run image based on the target registry during the rebase. 

`metadata.RunImageMetadata` is information about the specific run image that was used during the export.

exporter.go
```diff
@@ -151,11 +167,75 @@ func (e *Exporter) Export(
...
+			return errors.Wrapf(err, "reading layers for buildpack '%s'", bp.ID)
+		}
+
+		bpMD := metadata.BuildpackLayersMetadata{
+			ID:      bp.ID,
+			Version: bp.Version,
+			Layers:  map[string]metadata.BuildpackLayerMetadata{},
+		}
+		for _, layer := range bpDir.findLayers(cached) {
+			layer := layer
```
sclevine: Is this just future-proofing against additional closures that may be introduced and executed in parallel?

lukasberger: One of the checks that ran during make reported a potential(?) error when using '&layer' below when the variable wasn't explicitly reassigned. But parallelizing the layer export something I want to explore as well.

cmd/flags.go
lukasberger: I chose cache-dir to match the existing CNB_CACHE_DIR environment variable. If we want to go with -cache-volume instead, we should probably update the env var to match. While it doesn't necessarily need to be a volume, I think it will likely always be as the cache would otherwise effectively get discarded after each build. That being said if we're okay with changing the env var, I don't have a strong preference for either -cache-dir or -cache-volume.

I slightly prefer longer '-cache-dir' as opposed to just '-cache' as it's matches other other cache flag, '-cache-image' as well as the env var above. But if you'd like to maintain flag consistency, I'm not opposed to caching it.

exporter.go
lukasberger: Does that mean that if I export my application with a custom run image through the -image flag and then run rebaser on the image, it will default to rebasing the app image on top of the stack run image? That seems like it could lead to unexpected behavior.

If we save only one of metadata.RunImage or metadata.Stack, rebaser can raise an error when the '-image'  flag isn't provided and the app is based on a custom run image.

cmd/flags.go
jkutner: @sclevine `-path` is on `restorer`, which is not going away?

Instead of immediately `rm -rf`ing `cacher` in the next release, we could leave the binary in place as a no-op and have it emit a warning that it has been deprecated and will be removed in a future release. I think we're at a point where we should consider this, especially if it's the only breaking change in a given release.

ekcasey: @jkutner If a platform is unaware of the platform API change, having `cacher` no-op here might create unexpected behavior.

If we wanted to prevent a hard break, these are the options as I see them:
1. We have `exporter` skip the caching step when it is provided neither a dir nor an image. This will be slow but won't implode. It has some weird edge cases given that `restorer` will still restore the last successful cache.
2. We could give `exporter` default caching behavior that matches `pack`'s current usage (i.e cache to `/cache`). This works for `pack` but would do the wrong thing and or break for a platform that is currently using `cacher` with `-image` for example

I don't particularly like either of those options.

Given that an increase in the Platform API to `0.2` indicates a breaking change, I think we could just trust platforms to know if they aren't compatible with this lifecycle rather than trying to prevent a breaking change (`pack` will already bail out in this case).

exporter.go
ekcasey: @lukasberger I'm not sure I understand what you are proposing. How would the `rebaser` know if the image was based on a custom run image w/o knowing both the specific run-image that was used and the default mutable run-image and mirror tag references?

If a user of the lifecycle want to change the default stack for rebase it can provide a `stack.toml` file to the `exporter` to set the `metadata.Stack.RunImage`. If this stack metadata is left empty, `rebaser` requires an `-image` flag. Right now both the specific base image to use and the metadata to set for rebase are independently configurable, which I believe gives platforms the most flexibility.

cmd/flags.go
jromero: I was thinking the same thing as @jkutner. I don't think it would be difficult to provide an intermediate shim. This would prevent unexpectedly breaking from one builder to the next. This would affect all _platform_ implementations like `pack` as well as tekton, and knative definitions. Allowing for some grace period would probably be beneficial by preventing unnecessary support (via slack, issues, or otherwise).

ekcasey: In that case, I think we would need to leave an actual implementation of `cacher` , not a no-op binary, b/c those platforms won't be providing the necessary cache flags to `exporter`.

jkutner: Yea i prefer leaving the `cacher` option, which is what i already suggested in https://github.com/buildpack/lifecycle/pull/205#issuecomment-546950819 (not sure why switched :P). Also still open to making a breaking change, but it feels like we don't *have* to.

lukasberger: > @sclevine -path is on restorer, which is not going away?
The flag name affects restorer as well, since it also uses FlagCacheDir.

If we leave the cacher binary as is, what happens in the case where the platform sets flags through env vars? Platforms aware of this change can remove the cacher step, but for those that keep it would both exporter and cacher write to the same cache image twice?

Part of the RFC is switching the order of analyzer and restorer (in progress as a separate PR), which significantly alters their behavior and may necessitate a breaking change anyway. (At least I'm not sure I see a way to implement the changes in a backwards-compatible way.)

exporter.go
lukasberger: The `rebaser` would know that the app image was based on a custom run image (specified through `-image` on `exporter`) if there was no stack specified in the metadata. It would raise an error when attempting to rebase such app image without also providing the new image to rebase to through `-image`.

What is the intended use case for exporting with image X as the base and setting image Y as the rebase target? Why not use Y from the start if the app image can reasonably be expected to end up with Y as the base at some point anyway?

cmd/exporter/main.go
```diff
@@ -231,13 +235,51 @@ func export() error {
 		if _, isSaveError := err.(*imgutil.SaveError); isSaveError {
 			return cmd.FailErrCode(err, cmd.CodeFailedSave, "export")
 		}
-
 		return cmd.FailErr(err, "export")
 	}
 
+	// Failing to export cache should not be an error if the app image export was successful.
+	if cacheErr := exportCache(exporter); cacheErr != nil {
```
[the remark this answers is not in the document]
lukasberger: The latter is what happens if cacheImageTag is not set. This purpose of this check is to not prevent a build from succeeding if there are errors while saving the cache but the application image was pushed correctly.

exporter.go
lukasberger: Resolving conversation, discussed offline and working as intended.

cmd/flags.go
lukasberger: Discussed offline, we will resolve this by updating the platform api version and releasing an updating version of pack.

cmd/exporter/main.go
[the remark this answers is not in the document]
lukasberger: Please re-open if you think we should fail the build instead.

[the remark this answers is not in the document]
ekcasey: The current behavior defaults to a volume cache but there is no default value for the -`cache-dir` flag.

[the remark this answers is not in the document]
ekcasey: Thinking about this more carefully. We need a way for users to skip caching all together. I think caching should be skipped when neither flag is passed, but we should print a warning.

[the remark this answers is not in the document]
lukasberger: Added warning if neither cache image nor cache directory are set.

cmd/flags.go
jkutner: Sounds good, thanks @lukasberger. I'd like to keep track of all the changes so that we can write a migration guide for 0.6.0 (or whatever version this lands in). I think the updates to `pack` will help us catalog those changes.

lukasberger: Please see https://github.com/buildpack/pack/pull/384 for the proposed pack changes.

jromero: Sorry for the nitpick, specially if the conversation was had and I just couldn't find references to it. I'm in favor for the name `cache-dir` as oppose to `cache-volume` since technically, depending on the system, it might not actually be a volume, it's really just a directory.

My nitpick is around consistency, we have other directories that don't have the `dir` suffix. For example, `layers` , `platform`, `launch-cache`. Same for `image` suffix. For consistencies sake and to prevent yet another breaking change to rename them in the future what do we think about renaming them all at once?

Proposal:

- `layers` -> `layers-dir`
- `platform` -> `platform-dir`
- `launch-cache` -> `launch-cache-dir`
- `image` -> `run-image`

IMO, they are more understandable with the slight addition of a suffix.

ekcasey:
@lukasberger I approved this and #207. This one could use a tiny bit of cleanup once it can be rebased on #207 (see comments).

I think we need `pack` to work with platform API 0.2 before we can merge either of these to master or we will lose the ability to test `master` of lifecycle with `pack`. You mentioned you were working on those `pack` changes in parallel with these lifecycle changes. Do you plan to submit that PR or should we find somebody else to do it?

lukasberger:
@ekcasey The pack changes were merged in https://github.com/buildpack/pack/pull/384. Depending on how everyone else feels about the flag name cleanup that @jromero proposed being part of this PR, I can make the changes in pack as well. Regarding the cleanup, did you mean after #207 is merged, or do you want me to rebase this PR on top of the other branch?

cmd/flags.go
ekcasey: @jromero modifying all of these flag names would be a big change require spec changes and probably deserves it's own RFC. I think we should discuss that idea separately from this PR.

ekcasey:
@lukasberger Now that I realize the changes are already merge in `pack` I am inclined to merge both of these at once (merging just one of them will put us in a breaking state).

After they are merged you (or another contributor) can open a PR refactoring `cmd/exporter/main.go` to use `cache.NewImageCacheFromName` from the other PR.

cmd/exporter/main.go
```diff
@@ -230,13 +234,55 @@ func export() error {
...
+		cmd.Logger.Warnf("Failed to export cache: %v\n", cacheErr)
+	}
 	return nil
 }
 
+func exportCache(exporter *lifecycle.Exporter) error {
+	var cacheStore lifecycle.Cache
+	switch {
+	case cacheImageTag != "":
+		origCacheImage, err := remote.NewImage(
```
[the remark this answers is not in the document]
ekcasey: we can defer this refactor until after we have merged all the breaking API changes

cmd/exporter/main.go
```diff
@@ -230,13 +234,55 @@ func export() error {
...
+			emptyImage,
+		)
+	case cacheDir != "":
+		var err error
+		cacheStore, err = cache.NewVolumeCache(cacheDir)
+		if err != nil {
+			return cmd.FailErr(err, "create volume cache")
+		}
+	default:
+		exporter.Logger.Warn("Not exporting cache: no cache flag specified.")
```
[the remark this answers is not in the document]
ekcasey: we can defer this refactor until after we have merged all the breaking API changes

PR #207, opened 2019-11-01
Switch analyzer and restorer phases

lukasberger:
* Analyzer restores metadata for cache and app layers in /layers:
  * Restores metadata and sha for all cache=true layers.
  * Restores metadata and sha for all launch=true layers.
  * Does not restore metadata for cache=false and build=true layers.
* Restorer attempts to restore all cache layers in /layers.
  * Restores data for all cache=true layers that are in the cache.
  * Removes metadata for all cache=true layers that are not in cache.
  * Leaves all cache=false metadata as they are.
* Remove obsolete analyzer testdata.
* Remove unused classifiyCache.

Part 2 of the implementation of buildpack/rfcs#21.


analyzer.go
```diff
@@ -19,92 +20,110 @@ type Analyzer struct {
...
+		// The restorer step will restore the layer data for cache=true layers if possible or delete the layer.
+		appLayers := appMeta.MetadataForBuildpack(buildpack.ID).Layers
+		for name, layer := range appLayers {
+			identifier := fmt.Sprintf("%s:%s", buildpack.ID, name)
+			if !layer.Launch {
+				a.Logger.Infof("Not restoring metadata for %q, marked as launch=false", identifier)
+				continue
+			}
+			if layer.Build && !layer.Cache {
+				a.Logger.Infof("Not restoring metadata for %q, marked as build=true, cache=false", identifier)
```
ekcasey: Maybe this should also be debug level output?

analyzer.go
```diff
@@ -19,92 +20,110 @@ type Analyzer struct {
...
+			return nil, errors.Wrap(err, "reading buildpack layer directory")
+		}
+
+		// Restore metadata for launch=true layers.
+		// The restorer step will restore the layer data for cache=true layers if possible or delete the layer.
+		appLayers := appMeta.MetadataForBuildpack(buildpack.ID).Layers
+		for name, layer := range appLayers {
+			identifier := fmt.Sprintf("%s:%s", buildpack.ID, name)
+			if !layer.Launch {
+				a.Logger.Infof("Not restoring metadata for %q, marked as launch=false", identifier)
```
ekcasey: This should probably be debug level output for symmetry with line 79

cmd/analyzer/main.go
```diff
@@ -101,25 +106,57 @@ func analyzer() error {
...
+		)
+		if err != nil {
+			return cmd.FailErr(err, "creating new cache image")
+		}
+
+		cacheStore = cache.NewImageCache(
+			origCacheImage,
+			emptyImage,
+		)
+	} else {
```
ekcasey: Similar to the exporter/cacher PR, I think we need to handle the case where there is no cache. When we check args in the `main` func we should warn if both `cacheDir`and `cacheImage` are empty and provide a noop cache to the analyzer.

analyzer_test.go
```diff
@@ -86,515 +101,382 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
 
-				when("there cached launch layers that are missing from metadata", func() {
-					it("removes the layer dir and metadata", func() {
-						// copy to layerDir
-						h.RecursiveCopy(t, filepath.Join("testdata", "analyzer", "cached-layers"), layerDir)
-
-						if _, err := analyzer.Analyze(image); err != nil {
-							t.Fatalf("Error: %s\n", err)
-						}
+				it("restores escaped buildpack layer metadata", func() {
```
ekcasey: This looks like a copy of the previous test

restorer_test.go
```diff
@@ -189,89 +212,208 @@ func testRestorer(t *testing.T, when spec.G, it spec.S) {
...
+					h.AssertEq(t, string(got), escapedLayerSHA)
+				})
+				it("restores data", func() {
+					got := h.MustReadFile(t, filepath.Join(layersDir, "escaped_buildpack_id", "escaped-bp-layer", "file-from-escaped-bp"))
+					want := "echo text from escaped bp layer\n"
+					h.AssertEq(t, string(got), want)
+				})
+			})
+
+			when("there is a cache=true an unknown buildpack", func() {
```
ekcasey: This one seems a bit strange to me. I don't think we ever want only metadata for a `cache=true`, `launch=false`, `build=false` layer. It would be hard to get into this situation but in general I think the restorer should never leave the metadata and sha for a cached layer, if it can't restore the layer.

restorer_test.go
```diff
@@ -189,89 +212,208 @@ func testRestorer(t *testing.T, when spec.G, it spec.S) {
...
+			})
+
+			when("there is a cache=true layer in cache but not in group", func() {
+				it.Before(func() {
+					meta := "cache=true"
+					h.AssertNil(t, writeLayer(layersDir, "nogroup.buildpack.id", "some-layer", meta, noGroupLayerSHA))
+					h.AssertNil(t, restorer.Restore(testCache))
+				})
+
+				it("keeps metadata and sha file", func() {
```
ekcasey: Same as above

restorer_test.go
ekcasey: Given that it would be nearly impossible to get into this situation, I'm not sure we need to change the implementation, but writing a test around it makes it seem like intentional behavior.

analyzer.go
lukasberger: Good point, changed to debug.

analyzer.go
lukasberger: Changed to debug.

cmd/analyzer/main.go
lukasberger: Added a warning and modified analyzer to use empty cache metadata.

analyzer_test.go
lukasberger: Good catch, removed.

restorer_test.go
lukasberger: The issue here is that the buildpack is not part of the group which makes it difficult to find it. I agree it should be removed. Is there a way to find layers that don't match the group? Should we add it or is it excessive?

Removed the test for now.

restorer_test.go
lukasberger: Removed the 'keeps metadata' part but kept the 'does not restore data', because I think that's still relevant.

restorer_test.go
ekcasey: We could look for unrecognized buildpacks in the layers directory, but given that the analyzer uses the group to create all those directories in the first place, I think it might be excessive.

cache/common.go
```diff
@@ -2,6 +2,44 @@ package cache
...
+
+	"github.com/buildpack/lifecycle"
+	"github.com/buildpack/lifecycle/auth"
+	"github.com/buildpack/lifecycle/cmd"
 )
 
 var errCacheCommitted = errors.New("cache cannot be modified after commit")
+
+// MaybeCache returns a Cache object if one can be created from given arguments or nil.
+func MaybeCache(image, dir string) (lifecycle.Cache, error) {
```
ekcasey: I understand the utility of this function but it feels weird to me for it to live here, from a package interface perspective. Given that the logic here is more about the user interface of the cli (which flag or env var has precedence) than the implementation of the cache, I would suggest moving it to the `cmd` package.

cache/common.go
```diff
@@ -2,6 +2,44 @@ package cache
 
 import (
 	"errors"
+	"fmt"
+
+	"github.com/buildpack/imgutil/remote"
+
+	"github.com/buildpack/lifecycle"
+	"github.com/buildpack/lifecycle/auth"
+	"github.com/buildpack/lifecycle/cmd"
```
ekcasey: I know `exporter.go` currently makes this mistake, but I would prefer if nothing outside of `/cmd` imported the `cmd` package. Everything in `cmd` should be about the user interface we are providing with the CLIs. The `cmd` package probably shouldn't be required by packages that implement the core functionality.

cache/common.go
ekcasey: I could see putting just the contents of the `if` block in a function in the `cache` package, to reduce biolerplate construction logic required for a remote `ImageCache`.

cmd/analyzer/main.go
```diff
@@ -101,25 +106,32 @@ func analyzer() error {
...
 		}
 	}
 
-	md, err := analyzer.Analyze(img)
+	cacheStore, err := cache.MaybeCache(cacheImageTag, cacheDir)
+	if err != nil {
+		return cmd.FailErr(err, "set up cache")
+	}
+	if cacheStore == nil {
+		analyzer.Logger.Warn("Not restoring cached layer metadata, no cache flag specified.")
```
ekcasey: Why not log this warning in the `main` function so it is grouped with the other flag validations?

cmd/restorer/main.go
```diff
@@ -73,38 +70,13 @@ func restore() error {
...
-		cacheStore, err = cache.NewVolumeCache(cacheDir)
-		if err != nil {
-			return cmd.FailErr(err, "create volume cache")
-		}
+	cacheStore, err := cache.MaybeCache(cacheImageTag, cacheDir)
+	if err != nil {
+		return cmd.FailErr(err, "set up cache")
+	}
+	if cacheStore == nil {
+		restorer.Logger.Warn("Not restoring cached layer data, no cache flag specified.")
```
ekcasey: Is it even possible to hit this warning given line 53? It seems like we probably want to replace the error on line 54 with this warning and execute `restorer.Restore` regardless, so it can clean up metadata (in the extreme edge case where that is necessary).

ekcasey:
@lukasberger It occurs to me that there is a race condition here when an image is used as a cache. The image digest at the given tag can change between the execution of `analyzer` and `restorer`. 

Once upon a time there was a similar race condition between `analyzer` and `exporter`. The "previous image" at a given tag may have changed during the time interval after it was analyzed and before we attempt to reuse layers. We solved that race condition by putting the digest reference to the previous image in the `analyzed.toml` file and using the immutable image at the digest reference from which to consume layers during export.

We may want to store the digest reference of a cache image during the analyze phase so that the `restorer` can use the exact same image. Otherwise, there could be some wacky edge cases where the metadata doesn't accurately describe the restored layer contents.

lukasberger:
> @lukasberger It occurs to me that there is a race condition here when an image is used as a cache. The image digest at the given tag can change between the execution of `analyzer` and `restorer`.
> 
> Once upon a time there was a similar race condition between `analyzer` and `exporter`. The "previous image" at a given tag may have changed during the time interval after it was analyzed and before we attempt to reuse layers. We solved that race condition by putting the digest reference to the previous image in the `analyzed.toml` file and using the immutable image at the digest reference from which to consume layers during export.
> 
> We may want to store the digest reference of a cache image during the analyze phase so that the `restorer` can use the exact same image. Otherwise, there could be some wacky edge cases where the metadata doesn't accurately describe the restored layer contents.

Since the restorer verifies the metadata sha restored by analyzer matches the layer sha in the cache image and deletes the layer metadata if the two don't match, is this really an issue? Presumably if the layer sha still matches, we don't care if the cache image changed since the layer data is the same.

ekcasey:
@lukasberger good point. The previous case was different b/c there was no sha to compare. I retract my concern.

cmd/restorer/main.go
lukasberger: I completely missed that check. Changed to a warning and removed these lines.

cmd/analyzer/main.go
lukasberger: Good point, moved.

cache/common.go
lukasberger: Inlined MaybeCache in both restorer/main and analyzer/main and added a new NewImageCacheFromName function to cache/image_cache. Does that work?

PR #210, opened 2019-11-04
Slices

elbandito:
Draft PR for slices:

https://github.com/buildpack/spec/blob/master/buildpack.md#launchtoml-toml

**Order of operations:**

1) `[[slices]]` are read from `launch.toml` and saved to `metadata.toml` in the same order
2) Exporter examines the globs that are defined in the `paths` of each slice.  All the matches from the `paths` globs are used to create a new layer (tar file).
3) The file matches are removed from the `appDir` to remove duplications in the slice and app slice layer
4) The slice layer (slice1, slice2, sliceN, app dir) are added to the working image


exporter_test.go
```diff
@@ -655,6 +655,23 @@ type = "Apache-2.0"
 				nonExistingOriginalImage.Cleanup()
 			})
 
+			it("create a slice layer on the Run image", func() {
+				h.AssertNil(t, ioutil.WriteFile(filepath.Join(appDir, "static", "resources", "assets", "config.txt"), []byte("app-config"), os.ModePerm))
+				h.AssertNil(t, ioutil.WriteFile(filepath.Join(appDir, "static", "resources", "assets", "image.jpeg"), []byte("app-image"), os.ModePerm))
+
+				h.AssertNil(t, exporter.Export(layersDir, appDir, fakeAppImage, runImageRef, metadata.LayersMetadata{}, additionalNames, launcherConfig, stack))
+
+				sliceLayerPath := fakeAppImage.AppLayerPath()
```
elbandito: Need to think this through.  With the introduction of additional Slice layers, the image layer paths are now getting staggered by the number of slice layers that get saved to the image.  For this particular test, a single `slice` layer was introduced, causing the `AppLayerPath` (hardcoded to index 0) to now point to the `slice` layer and the `ConfigLayerPath` (hardcoded to index 1) now points to the app layer.   Will probably need to make changes to the `imageUtil` lib.

nebhale:
Some more follow ups.

1.  The names of the slice layers should be simpler/shorter.  I'd like to see something like `slice-0`, `slice-1`, etc. as the names are not important as part of exporting, they're only used to log out.
2. I think the `config` layers should be moved to the very last layer of the image, after all the slices and the app layer.  @ekcasey any opinions on this?

elbandito:
@nebhale Currently, layers are being **added** in the following order:
1) Slices
2) App
3) Config
4) Launcher
5) Buildpacks

Are you suggesting the **add** order be?

1) Launcher
2) Buildpacks
3) App
4) Slices
5) Config

ekcasey:
I think that reordering makes sense. I don't have a sense whether slices or app (i.e. the everything else slice) are more likely to change. I guess it would depend how specific buildpacks choose to use the slice functionality

nebhale:
@elbandito I concur that the latter order is correct.  My instinct is that if you're slicing the application layer, you're likely not leaving anything on the application layer (you care explicitly about layering and ordering of _all_ files) so I don't think it matters in practice which of app or slice layers comes first.  My preference is 

1. Launcher
1. ~Buildpacks~ Layers
1. Slices
1. App
1. Config

for gremlins-in-my-head reasons, but beyond that, I have no opinion.

elbandito:
@ekcasey and @nebhale I rearranged the layers.  Let me know if you see anything else.

archive/tar.go
```diff
@@ -11,6 +11,82 @@ import (
...
+	tw := tar.NewWriter(w)
+
+	fileSet := map[string]struct{}{}
+	for _, file := range files {
+		if AddFileToArchive(tw, file, uid, gid, fileSet) != nil {
+			return "", nil, err
+		}
+	}
+	_ = tw.Close()
+	defer f.Close()
```
ekcasey: This defer `f.Close()` should move directly below the error check after `os.Create`, so the file still gets closed even if the function exits with an error on line 25.

builder.go
```diff
@@ -202,6 +211,7 @@ func isBuild(path string) bool {
 }
 
 type processMap map[string]Process
+type sliceList []Slice
```
ekcasey: I don't think we need this type. `[]Slice` is explicit already at it is symmetrical with `[]BOMEntry`.

exporter.go
```diff
@@ -48,21 +49,45 @@ func (e *Exporter) Export(
...
 	meta.Stack = stack
 
-	meta.App.SHA, err = e.addLayer(workingImage, &layer{path: appDir, identifier: "app"}, origMetadata.App.SHA)
-	if err != nil {
-		return errors.Wrap(err, "exporting app layer")
+	var buildMetadata BuildMetadata
+	sliceLayers := map[string]identifiableLayer{}
+
+	metadataTomlPath := filepath.Join(layersDir, "config", "metadata.toml")
+	_, err = toml.DecodeFile(metadataTomlPath, &buildMetadata)
```
ekcasey: It looks like we are decoding the metadata file twice (also on line 170)

elbandito: good catch!

exporter.go
```diff
@@ -169,3 +207,82 @@ func (e *Exporter) addLayer(image imgutil.Image, layer identifiableLayer, previo
...
+		found := false
+		for _, previous := range previousLayer {
+			if sha == previous.SHA {
+				found = true
+				break
+			}
+		}
+		if found {
+			e.Logger.Infof("Reusing layer '%s'\n", layer.Identifier())
+			e.Logger.Infof("Layer '%s' SHA: %s\n", layer.Identifier(), sha)
```
nebhale: Given the current output:

```
===> EXPORTING
[exporter] Adding layer 'launcher'
[exporter] Adding layer 'io.pivotal.openjdk:class-counter'
[exporter] Adding layer 'io.pivotal.openjdk:java-security-properties'
[exporter] Adding layer 'io.pivotal.openjdk:jvmkill'
[exporter] Adding layer 'io.pivotal.openjdk:link-local-dns'
[exporter] Adding layer 'io.pivotal.openjdk:memory-calculator'
[exporter] Adding layer 'io.pivotal.openjdk:openjdk-jre'
[exporter] Adding layer 'io.pivotal.openjdk:security-provider-configurer'
[exporter] Adding layer 'io.pivotal.jvmapplication:executable-jar'
[exporter] Adding layer 'io.pivotal.springboot:spring-boot'
[exporter] Adding layer 'io.pivotal.clientcertificatemapper:client-certificate-mapper'
[exporter] Adding layer 'io.pivotal.containersecurityprovider:container-security-provider'
[exporter] Adding layer 'io.pivotal.springautoreconfiguration:auto-reconfiguration'
[exporter] Adding layer 'slice-2'
[exporter] Layer 'slice-2' SHA: sha256:a30d09749a148759df26eaca697dee994db3c37adde6e68406f57fe54e76e573
[exporter] Adding layer 'slice-3'
[exporter] Layer 'slice-3' SHA: sha256:a90f94780489ffb2a74d973a9bea7a8b0c1c3df0b1703c9408dfc65e2649aaf8
[exporter] Adding layer 'slice-4'
[exporter] Layer 'slice-4' SHA: sha256:55a0304911d6eb9f4338f2fa9a9cfe9291c83125af151b7e8155b7b10615e8c7
[exporter] Adding layer 'slice-5'
[exporter] Layer 'slice-5' SHA: sha256:aa0914e173b579980f29a25bc151fc5494cdc0e0c4676109f5340bf7e8f830a3
[exporter] Adding layer 'slice-1'
[exporter] Layer 'slice-1' SHA: sha256:48e99c4ff85a50bb2ec6bb1feceadbe5628d3227ebce111bf3ed2f5702b2f115
[exporter] Adding layer 'app'
[exporter] Adding layer 'config'
[exporter] *** Images (7d8a851c3dd5):
[exporter]       index.docker.io/applications/jar-source:latest
```

and the [existing code](https://github.com/buildpack/lifecycle/blob/master/exporter.go#L96), this like should be `Debugf`.

exporter.go
```diff
@@ -169,3 +207,82 @@ func (e *Exporter) addLayer(image imgutil.Image, layer identifiableLayer, previo
...
+			err = image.ReuseLayer(sha)
+		} else {
+			e.Logger.Infof("Adding layer '%s'\n", layer.Identifier())
+			e.Logger.Infof("Layer '%s' SHA: %s\n", layer.Identifier(), sha)
```
nebhale: Given the current output:

```
===> EXPORTING
[exporter] Reusing layer 'launcher'
[exporter] Reusing layer 'io.pivotal.openjdk:class-counter'
[exporter] Reusing layer 'io.pivotal.openjdk:java-security-properties'
[exporter] Reusing layer 'io.pivotal.openjdk:jvmkill'
[exporter] Reusing layer 'io.pivotal.openjdk:link-local-dns'
[exporter] Reusing layer 'io.pivotal.openjdk:memory-calculator'
[exporter] Reusing layer 'io.pivotal.openjdk:openjdk-jre'
[exporter] Reusing layer 'io.pivotal.openjdk:security-provider-configurer'
[exporter] Reusing layer 'io.pivotal.jvmapplication:executable-jar'
[exporter] Reusing layer 'io.pivotal.springboot:spring-boot'
[exporter] Reusing layer 'io.pivotal.clientcertificatemapper:client-certificate-mapper'
[exporter] Reusing layer 'io.pivotal.containersecurityprovider:container-security-provider'
[exporter] Reusing layer 'io.pivotal.springautoreconfiguration:auto-reconfiguration'
[exporter] Reusing layer 'slice-3'
[exporter] Layer 'slice-3' SHA: sha256:a90f94780489ffb2a74d973a9bea7a8b0c1c3df0b1703c9408dfc65e2649aaf8
[exporter] Reusing layer 'slice-4'
[exporter] Layer 'slice-4' SHA: sha256:55a0304911d6eb9f4338f2fa9a9cfe9291c83125af151b7e8155b7b10615e8c7
[exporter] Reusing layer 'slice-5'
[exporter] Layer 'slice-5' SHA: sha256:aa0914e173b579980f29a25bc151fc5494cdc0e0c4676109f5340bf7e8f830a3
[exporter] Reusing layer 'slice-1'
[exporter] Layer 'slice-1' SHA: sha256:48e99c4ff85a50bb2ec6bb1feceadbe5628d3227ebce111bf3ed2f5702b2f115
[exporter] Reusing layer 'slice-2'
[exporter] Layer 'slice-2' SHA: sha256:a30d09749a148759df26eaca697dee994db3c37adde6e68406f57fe54e76e573
[exporter] Adding layer 'app'
[exporter] Reusing layer 'config'
[exporter] *** Images (b1b33820c01a):
[exporter]       index.docker.io/applications/jar-source:latest
```

and the [existing code](https://github.com/buildpack/lifecycle/blob/master/exporter.go#L96), this like should be `Debugf`.

nebhale: Looks good in latest commit.

exporter.go
nebhale: Looks good in latest commit.

elbandito:
@nebhale I'm fine with adding a warning to the user if they are adding an empty slice layer

elbandito:
Example output
```
===> EXPORTING
[exporter] Adding layer 'launcher'
[exporter] Reusing layer 'io.buildpacks.samples.java-maven:jdk'
[exporter] Warning: Adding empty layer 'slice-1'
[exporter] Warning: Adding empty layer 'slice-2'
[exporter] Reusing layer 'slice-3'
[exporter] Warning: Adding empty layer 'slice-4'
[exporter] Reusing layer 'slice-5'
[exporter] Adding layer 'app'
[exporter] Reusing layer 'config'
[exporter] *** Images (9c6d47968665):
[exporter]       index.docker.io/longoria/refactorslices:latest
```

sclevine:
@nebhale @elbandito, after chatting with @ekcasey and @simonjjones, how would you feel about something like this?

In the case that slices are reused:
```
Reusing 0.5 MB / 123.4 MB (3/19 slices) as app layers
```
In the case that slices are not reused, but there are slices:
```
Reusing 0.0 MB / 123.4 MB (0/19 slices) as app layers
```
In the case where there are no slices specified by the buildpack:
```
Adding layer 'app'
```
(which is the current output in this case)

exporter_test.go
```diff
@@ -102,6 +103,93 @@ func testExporter(t *testing.T, when spec.G, it spec.S) {
...
+				sliceSHA          string
+			)
+
+			it.Before(func() {
+				var err error
+
+				appDir, err = filepath.Abs(filepath.Join("testdata", "exporter", "previous-slice-image-exists", "layers", "app"))
+				h.AssertNil(t, err)
+
+				sliceConfigFilePath := filepath.Join(appDir, "static", "assets", "config.txt")
```
ekcasey: Any reason not to make these test fixtures, like the other files in the `appDir`?

exporter_test.go
```diff
@@ -102,6 +103,93 @@ func testExporter(t *testing.T, when spec.G, it spec.S) {
...
+			var (
+				fakeOriginalImage *fakes.Image
+				fakeImageMetadata lifecycle.LayersMetadata
+				sliceSHA          string
+			)
+
+			it.Before(func() {
+				var err error
+
+				appDir, err = filepath.Abs(filepath.Join("testdata", "exporter", "previous-slice-image-exists", "layers", "app"))
```
ekcasey: Now that the `exporter` modifies the app dir (deletes files) we should work in a temp dir. Given that `app` is a subdir of the `layers`, and we recursively copy to a temp dir on line 126, we can just use that temp dir as `appDir` instead of the dir in the source tree.

exporter_test.go
```diff
@@ -102,6 +103,93 @@ func testExporter(t *testing.T, when spec.G, it spec.S) {
...
+					local.IDIdentifier{ImageID: "some-original-run-image-digest"},
+				)
+				fakeAppImage.AddPreviousLayer("sha256:"+localReusableLayerSha, "")
+				fakeAppImage.AddPreviousLayer("sha256:"+launcherSHA, "")
+				fakeAppImage.AddPreviousLayer("sha256:orig-launch-layer-no-local-dir-sha", "")
+				fakeAppImage.AddPreviousLayer("sha256:"+sliceSHA, "")
+
+				h.AssertNil(t, fakeOriginalImage.SetLabel("io.buildpacks.lifecycle.metadata",
+					fmt.Sprintf(`{
+                  "app": [{
```
ekcasey: This json formatting looks off

exporter_test.go
```diff
@@ -640,6 +734,44 @@ type = "Apache-2.0"
...
 
+		when("previous slice image doesn't exist", func() {
+			var (
+				nonExistingOriginalImage *fakes.Image
+			)
+
+			it.Before(func() {
+				h.RecursiveCopy(t, filepath.Join("testdata", "exporter", "previous-slice-image-not-exist", "layers"), layersDir)
+				var err error
+				appDir, err = filepath.Abs(filepath.Join("testdata", "exporter", "previous-slice-image-not-exist", "layers", "app"))
```
ekcasey: Same as above, I think we should use the appDir from the temporary location.

exporter_test.go
```diff
@@ -640,6 +734,44 @@ type = "Apache-2.0"
...
+				nonExistingOriginalImage.Cleanup()
+			})
+
+			it("create a slice layer on the Run image", func() {
+				h.AssertNil(t, ioutil.WriteFile(filepath.Join(appDir, "static", "assets", "config.txt"), []byte("app-config"), os.ModePerm))
+				h.AssertNil(t, ioutil.WriteFile(filepath.Join(appDir, "static", "assets", "image.jpeg"), []byte("app-image"), os.ModePerm))
+
+				h.AssertNil(t, exporter.Export(layersDir, appDir, fakeAppImage, runImageRef, lifecycle.LayersMetadata{}, additionalNames, launcherConfig, stack))
+
+				sliceLayerPath, _ := fakeAppImage.FindLayerWithPath(filepath.Join(appDir, "static", "assets", "config.txt"))
```
ekcasey: missing error check

exporter_test.go
```diff
@@ -640,6 +734,44 @@ type = "Apache-2.0"
...
+
+				assertTarFileExists(t, sliceLayerPath, filepath.Join(appDir, "static", "assets", "config.txt"), true)
+				assertTarFileExists(t, sliceLayerPath, filepath.Join(appDir, "static", "assets", "image.jpeg"), true)
+
+				appLayerPath, _ := fakeAppImage.FindLayerWithPath(filepath.Join(appDir, "static", "assets", "file.keep"))
```
ekcasey: missing error check

exporter_test.go
```diff
@@ -640,6 +734,44 @@ type = "Apache-2.0"
...
+
+				assertTarFileExists(t, appLayerPath, filepath.Join(appDir, "static", "assets", "config.txt"), false)
```
ekcasey: We should assert that the remaining (non-sliced) app files are in the appLayer (final slice)

exporter_test.go
```diff
@@ -492,7 +585,8 @@ type = "Apache-2.0"
       "command": "/worker/command",
       "args": ["worker", "command", "args"]
     }
-  ]
+  ],
+  "slices": null
```
ekcasey: I am not sure we should add slices to this label. This information would be more useful paired with the sha in the array of slice layers in `io.buildpacks.lifecycle.metadata`. (adding to that label doesn't have to be a requirement right now)

exporter.go
```diff
@@ -169,3 +210,85 @@ func (e *Exporter) addLayer(image imgutil.Image, layer identifiableLayer, previo
...
+		var err error
+
+		found := false
+		for _, previous := range previousLayer {
+			if slice.Sha == previous.SHA {
+				found = true
+				break
+			}
+		}
+		if found && !slice.Empty {
```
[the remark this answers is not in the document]
elbandito: @ekcasey good question.  I was consitently seeing the error below whenever I tried to reuse an empty layer.  It wasn't clear to me what was causing this (I was planning to circle back with on it), so for the short term, I added this condition.

```
[exporter] ERROR: failed to export: exporting slice layers: reusing layer (sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef): link /launch-cache/committed/sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef.tar /launch-cache/staging/sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef.tar: file exists
ERROR: failed with status code: 1
```

exporter_test.go
elbandito: @ekcasey now that I'm modifying the temp `appDir` (inside the copied temp dir) there maybe opportunities to consolidate the fixture data.

exporter.go
```diff
@@ -113,6 +139,25 @@ func (e *Exporter) Export(
...
 	}
 
+	// slices
+	meta.App, err = e.addSliceLayers(workingImage, sliceLayers, origMetadata.App)
+	if err != nil {
+		return errors.Wrap(err, "exporting slice layers")
+	}
+
+	// app
+	appSHA, err := e.addLayer(workingImage, &layer{path: appDir, identifier: "app"}, "")
```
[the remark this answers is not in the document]
elbandito: @ekcasey I agree with your logic.  It wasn't clear to me whether or not I should preserve to concept of an `app` layer.  Would we need to clarify _moving away_ from an app layer concept in lieu of multiple slice layers in any of the external documentation?

exporter.go
```diff
@@ -26,6 +27,12 @@ type LauncherConfig struct {
 	Metadata LauncherMetadata
 }
 
+type SliceLayer struct {
+	Sha   string
```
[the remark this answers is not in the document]
elbandito: @ekcasey I'll give it some thought.  In the meantime, I'm going to make some changes to steer away from using the existing `identifiableLayer`.

exporter.go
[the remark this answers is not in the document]
ekcasey: @elbandito It looks like there is a problem in `VolumeCache`. This issue doesn't appear specific to empty layers, but rather the cache errors out inappropriately when it is told to reuse two identical layers during a single build.

exporter.go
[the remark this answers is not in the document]
ekcasey: @elbandito I can't think of any external facing documentation that references the app layer. I am okay with referring to an "app layer" in the logs in the specific case where there are no slices (@sclevine's suggested log output)

exporter_test.go
ekcasey: That makes sense to me. We could use the same app fixture in both the slices/no-slices cases. Either all of the above files are added to a single layer or they are carved up into slice layers, depending on the values in `metadata.toml`.

nebhale:
> how would you feel about something like this?

@sclevine What prompts the addition of sizes there?  None of the other re-use logging indicates how much data was re-used.  I'd prefer consistency across the messages, but broadly I'm on board with the `3/19`-style logging.

elbandito:
@nebhale, regarding the output, I don't have a very strong opinion, but I do share @nebhale thoughts around consistency.  Specifically, the addition of the size indicator - what's the motivation behind this?

ekcasey:
After a discussion in WG it seems like the consensus is that layer sizes are useful for all layers, not just slices. However, to avoid growing the scope of this issue or making slices inconsistent, we can introduce that change in a different PR.

Therefore I would propose:

```
Reusing 3/19 app layer(s)
Adding 17/19 app layer(s)
```

In the case where nothing is reused we can omit the `Reusing` line:
```
Adding 19/19 app layer(s)
```
In the case where there are no slices:
```
Adding 1/1 app layer(s)
```

exporter.go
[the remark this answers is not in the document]
ekcasey: @elbandito I think this PR should fix that issue https://github.com/buildpack/lifecycle/pull/214

testdata/exporter/app-slices/layers/app/metadata.toml
```diff
@@ -0,0 +1,11 @@
+[[processes]]
```
ekcasey: This `metadata.toml` file in the `app` dir is a bit confusing, the one we actually respect lives in the `config` dir. I don't think this file was introduces in this PR but merely copied to a new location.

ekcasey:
@elbandito While this PR looks good on it's own, it seems like it will break `pack inspect-image` b/c of the change to the metadata schema in the `io.buildpacks.lifecycle.metadata`

```
pack inspect-image app
ERROR: inspecting local image 'app': unmarshalling label io.buildpacks.lifecycle.metadata: json: cannot unmarshal array into Go struct field LayersMetadata.app of type lifecycle.LayerMetadata
```

We should update master of `pack` to deserialize correctly for both images built with lifecycle v0.5.0 and  the upcoming lifecycle v0.6.0, before we merge.

testdata/exporter/app-slices/layers/app/metadata.toml
elbandito: @ekcasey I copied most of the fixture structure from an existing fixture.  If I don't need this, I'm more than happy to remove it.

Zanadar:
Accepting this was a bit tricky but i figured it out:
I modified the sample ruby bundler buildpack to slice `vendor` in a few places: https://github.com/buildpacks/samples/blob/slices-test/buildpacks/ruby-bundler/bin/build#L66

```toml

[[slices]]
paths = ["vendor/ruby/2.5.0/bin/*", "vendor/ruby/2.5.0/cache/*", "vendor/ruby/2.5.0/gems/*", "2", "3"]
```

I modified the app to have these vendor folders: https://github.com/buildpacks/samples/tree/slices-test/apps/ruby-bundler/vendor/ruby/2.5.0

And I ran the build with a custom builder created from a master built version of the lifecycle:
```
± zm |slices-test ✓| → pack inspect-builder cnbs/sample-builder:bionic-test
Inspecting builder: cnbs/sample-builder:bionic-test

REMOTE:
(not present)

LOCAL:

Created By:
  Name: Pack CLI
  Version: v0.6.0 (git sha: 109b629d388cec0ed3836b9fed6717727a9187c1)

Stack:
  ID: io.buildpacks.samples.stacks.bionic

Lifecycle:
  Version: 0.0.0
  Buildpack API: 0.2
  Platform API: 0.2

Run Images:
  cnbs/sample-stack-run:bionic

Buildpacks:
  ID                                         VERSION
  io.buildpacks.samples.java-maven           0.0.1
  io.buildpacks.samples.kotlin-gradle        0.0.1
  io.buildpacks.samples.ruby-bundler         0.0.1

Detection Order:
  Group #1:
    io.buildpacks.samples.java-maven@0.0.1
  Group #2:
    io.buildpacks.samples.kotlin-gradle@0.0.1
  Group #3:
    io.buildpacks.samples.ruby-bundler@0.0.1
```

Build:
```
± zm |slices-test ✓| → pack build sample-app:slices --builder cnbs/sample-builder:bionic-test --buildpack ./buildpacks/ruby-bundler --path ./apps/ruby-bundler/
bionic: Pulling from cnbs/sample-stack-run
Digest: sha256:2ca156537bad95e4870e5e8a74b10e7b46b722add5ebb6eda9939e8ce8c7143b
Status: Image is up to date for cnbs/sample-stack-run:bionic
===> DETECTING
[detector] io.buildpacks.samples.ruby-bundler2 0.0.2

===> ANALYZING
[analyzer] Restoring metadata for "io.buildpacks.samples.ruby-bundler2:bundler" from app image
[analyzer] Restoring metadata for "io.buildpacks.samples.ruby-bundler2:ruby" from app image
===> RESTORING
[restorer] Restoring data for "io.buildpacks.samples.ruby-bundler2:bundler" from cache
===> BUILDING
[builder] ---> Ruby Buildpack
[builder] ---> Downloading and extracting Ruby 2.5.0
[builder] ---> Installing bundler
[builder] Successfully installed bundler-2.1.4
[builder] 1 gem installed
[builder] ---> Reusing gems
===> EXPORTING
[exporter] Reusing layer 'launcher'
[exporter] Reusing layer 'io.buildpacks.samples.ruby-bundler2:bundler'
[exporter] Reusing layer 'io.buildpacks.samples.ruby-bundler2:ruby'
[exporter] Reusing 2/2 app layer(s)
[exporter] Reusing layer 'config'
[exporter] *** Images (c277a939811f):
[exporter]       index.docker.io/library/sample-app:slices
[exporter] Reusing cache layer 'io.buildpacks.samples.ruby-bundler2:bundler'
Successfully built image sample-app:slices
```

[asciicast](https://asciinema.org/a/294139)

Issue #123, opened 2019-04-15
Consider making $HOME = /app


josegonzalez:
In v2 buildpacks, the `$HOME` directory was typically `/app`. This was useful for cases where - for whatever reason - the entire path to a file had to be specified during the running of an application. A good example is the `Procfile` entry for some `R` buildpacks, where the fullpath had to be used because of `fakechroot` usage.

Switching from `/workspace` back to `/app` may increase compatibility with older apps by decreasing the number of changes necessary to migrate from v2 to v3 buildpacks.

jkutner:
In the `heroku/pack:18` image used by the `heroku/buildpacks` builder image, there is a symlink from `/app` to `/workspace`, and `/workspace` is the HOME. We build our run-image with:

```Dockerfile
RUN ln -s /workspace /app

RUN useradd -u ${pack_uid} -g ${pack_gid} -ms /bin/bash -d /workspace heroku
```

(note: the Heroku builds never happened in `/app` so we don't need it in the build image)

This works pretty well with all the Heroku buildpacks that use the [cnb-shim](https://github.com/heroku/cnb-shim).

I think this approach is good: accommodate it in the stack image. I would like to move away from `/app` because we often hear from users that they aren't building an "app" (it's a "service" or something).

jkutner:
that said, i think setting `HOME` as `/workspace` (or whatever is configured in place of `/workspace`) makes sense. But IIRC, there is some reason we can't introduce that to the spec.

sclevine:
Where are you setting `HOME` to workspace in the Heroku images?

Setting `HOME` to `/workspace` during build seems risky to me. Tools often use dot-files in `HOME` to cache downloads (not necessary for launch) or store credentials (bad for security if they end up in the final image).

jkutner:
@sclevine I'm only suggestion we set HOME to `/workspace` for the run-image

nebhale:
I agree with @sclevine on this that given how many runtimes expose the root of the application (think exploded JAR/WAR files) that overlapping `$HOME` with the application directory is problematic.  For reference the only CVE ever opened against the CF Java Buildpack was due to this very behavior ☹️

ekcasey:
Given the concerns @sclevine @nebhale have outlined and the lack of activity on this issue, I am going to go ahead and close it.

Feel free to reopen if you would like to restart the discussion.

Issue #77, opened 2019-01-27
As a CNB user, the default layers directory should be `/layers` and the default app directory should be `/workspace`


sclevine:
**Acceptance Criteria**
**When** I build an OCI image using the lifecycle:
- Via `pack build` or
- Via manual orchestration of the lifecycle binaries
**And** I manually inspect the image contents
**Then** I observe that the layers directory is `/layers`
**And** I observe that the app directory is `/workspace`

**Notes**
- Remove any logic that allows `app` to be a subdirectory of `/layers`
- It's okay if re-built images are unable to reuse layers due to the updated lifecycle binaries in the builder, but it would be preferable if the re-builds did not fail. Let's discuss if this is not possible

mgibson1121:
This story passes acceptance.

PR #223, opened 2020-01-03
Prevent duplicate app dir paths

### LINKED ISSUE #222: Handling "." in path when image is built with local implementations.
When I tried to build a docker image using exporter, I found that the output image contains duplicate files. And this is because "." exists in the file path. 

```
failed to register layer: Error processing tar file(duplicates of file paths not supported): time="2020-01-02T07:31:51Z" level=info msg="SUSE:secrets :: enabled"
```

I'm using lifecycle v0.5.0 with the tekton task here https://github.com/tektoncd/catalog/blob/master/buildpacks/buildpacks-v3.yaml

And If I set -app= to a path which ends with "/"(e.g. /workspace/source/ or /workspace/source/.), the image would have duplicated dirs.

jromero:
### Changes

* Make app dir absolute
    * In regards to discussions with @ekcasey: it's unexpected to think that `-app ./some/path/` and `-app /my-app/some/path/` would place the application source in different directories within the final image
* Clean src path to tar to prevent duplicate entries eg. `some/dir/` vs `some/dir` vs `some/dir/.`
* Update tool dependency `golangci` - I did this in an attempt to solve issues with [scopelint](https://github.com/kyoh86/scopelint/issues/4).
* Update `Makefile` to support incoming GOFLAGS - IntelliJ/Goland's built-in terminal provides this flag that then collides with `-mod=vendor` and we get an error stating the it can only be specified once.

Resolves #222 

### Sidenote

~~A concern that come up with the current implementation is that the app dir isn't _configurable_ and is based directly off of the `-app` argument. For example, in theory `-app=/home/jromero/my-app` would yield with that directory location in the final image as `/home/jromero/my-app/`. My initial thought would be that there would be a standard app path location. Seems like this might have existed in PR #35 but has since been removed.~~ 

- After discussing this with @ekcasey, it appears that this is intended for now due to the fact that it's intended to run in a container in which case paths could be modified to suit the need. Some concerns over specific CI systems still exist by chose to defer until such case arises.

Related to:
 - https://github.com/buildpacks/lifecycle/issues/123
 - https://github.com/buildpacks/lifecycle/issues/77
 - https://github.com/buildpacks/lifecycle/pull/35


sclevine:
Also may be a good idea to clean the layer directory, like we do for builder:
https://github.com/buildpacks/lifecycle/blob/5a1ef31fd654cd53abfb5d1e01dc11b4f7537602/builder.go#L58-L69

jromero:
RE: @ekcasey 
Although I agree with the sentiment as a whole I find it cumbersome to think about maintaining two robust implementations of archiving functionality. In practice, we've resolved these issues already in pack and have a different solution for how this occurs. I wonder if a better approach wouldn't be to instead move (or centralize) archiving functionality so that pack and lifecycle share the same solution.

ekcasey:
@jromero One could make a case for moving some of this logic to `imgutil` I suppose, given that the package in question provides utilities for creating layers. I'm fairly ambivalent on that question. With less duplication we won't have to fix bugs in two places. However, the logic is not completely identical due to the differing contexts (e.g lifecycle adds parent dirs) and moving it to `imgutil` adds a bit more friction to the process of making changes.

Whether we extract this package to a shared location or not, I do believe we should fix the core issue in the `archive` package rather than, or in addition to, working around it.

Zanadar:
@ekcasey Would you be okay with that extraction happen as enhancement to actually fixing this bug for the `0.6.0` release?

exporter_test.go
```diff
@@ -816,6 +816,51 @@ type = "Apache-2.0"
 				assertLogEntry(t, logHandler, "Adding 1/1 app layer(s)")
 			})
 
+			when("app dir has ending path separator", func() {
```
ekcasey: Now that we have coverage in the archive package tests, I think we should remove this test and "app dir has dot element". It seems like a lower level concern and these higher level exporter tests are more trouble to maintain.

archive/tar_test.go
```diff
@@ -268,7 +277,6 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
 
 func tarContains(t *testing.T, m string, r func()) {
 	t.Helper()
-	t.Log(m)
```
ekcasey: don't like the log statements? I think they can make long tests easier to read and debug

archive/tar_test.go
```diff
@@ -96,82 +103,73 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
...
+			filepath.Join("testdata", "dir-to-tar") + string(filepath.Separator) + ".",
+		} {
+			src := src
+			it(fmt.Sprintf("writes a tar with the src filesystem contents (%s)", src), func() {
+				uid := 1234
+				gid := 4567
+
+				file, err := os.Create(filepath.Join(tmpDir, "tar_test-go.tar"))
+				h.AssertNil(t, err)
+				defer file.Close()
```
ekcasey: We seem to be closing the same file multiple times (see line 121). It succeeds only b/c we aren't checking for errors here, which feels a bit messy.

archive/tar_test.go
```diff
@@ -182,10 +180,14 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
 				absoluteFilePath, err := filepath.Abs(filepath.Join("testdata", "dir-to-tar"))
 				h.AssertNil(t, err)
 
-				h.AssertNil(t, archive.WriteTarArchive(file, absoluteFilePath, uid, gid))
+				file, err := os.Create(filepath.Join(tmpDir, "tar_test-go.tar"))
+				h.AssertNil(t, err)
+				defer file.Close()
```
ekcasey: closing the same file multiple times again

archive/tar_test.go
jromero: These log messages in particular didn't add any value in my opinion since when the assertion fails it specifies exactly where the assertion resides. Since we added 3 additional iterations it simply added noise to the logs.

archive/tar_test.go
jromero: This is intentional. The defer is there for reassurance given that we might not actually reach the inlined `file.Close()` if something where to go wrong within `archive.WriteTarArchive`.

archive/tar_test.go
jromero: See prior comment. We can continue the conversation there.

archive/tar_test.go
ekcasey: If this was the implementation rather than a test I would suggest moving these few lines into a function and allowing the defer to close the file in both cases.

But given that this is test I am happy to be less nitpicky about it.

natalieparellano:
Accepting - on `5b9f575` (without the fix) invoking the `exporter` with `-app /workspace/.` produces the following error:

```
[exporter] ERROR: failed to export: failed to write image to the following tags: [index.docker.io/library/fake-app-new:latest: save image 'index.docker.io/library/fake-app-new:latest': Error: No such image: 471f8ab89748b6e40135612e8ef4c89715df0e661b7c9eed8f93cfefa4293248]
```

On master, we were able to successfully build the app image with the same invocation.

Note, our acceptance steps used `pack` as the platform. We did the following:
* `make build` the lifecycle binaries
* `cp out/lifecycle/exporter /tmp/fake-app/real-exporter`
* wrote a shell script saved to `out/lifecycle/exporter` to run `real-exporter` from `/workspace` in the build container
* `make package && pack create-builder test-builder --builder-config ./builder.toml`
* `pack build test-app --path /tmp/fake-app --builder test-builder --buildpack /tmp/fake-buildpack` <- observe the build fail or succeed

PR #225, opened 2020-01-09
verify jq is installed before running unit tests

hone:
`jq` is used by the test buildpacks. This PR fails out early if `jq` isn't detected on the `$PATH`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ SCM_REPO?=
 SCM_COMMIT=$$(git rev-parse --short HEAD)
 ARCHIVE_NAME=lifecycle-v$(LIFECYCLE_VERSION)+linux.x86-64
 
+.PHONY: verify-jq
+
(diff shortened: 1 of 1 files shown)


jromero:
Where is `jq` being used? It seems to be only used inside buildpack scripts. Are those buildpacks not being executed in a container?

Makefile
```diff
@@ -71,9 +71,14 @@ format: install-goimports
 	@echo "> Formating code..."
 	test -z $$(goimports -l -w -local github.com/buildpacks/lifecycle $$(find . -type f -name '*.go' -not -path "*/vendor/*"))
 
+verify-jq:
+	ifeq (, $(shell which jq))
```
ekcasey: I played around with this and I think Travis is failing because of the indentation here. `ifeq` should have the same indentation as the target.

hone: yeah, I figured that out eventually. pushed changes!

PR #230, opened 2020-01-21
Implements RFC0013 for app source project metadata

iainsproat:
Implements https://github.com/buildpacks/spec/pull/68

- Adds ProjectMetadataPath flag to Export command
- ~~If flag is not provided,~~ sets default path to "./project-metadata.toml"
- If project metadata path is not present or cannot be unmarshalled, it logs for info and
continues
- Provides types for ProjectMetadata ~~, and Git source code~~
- Amends Export method parameters to include ProjectMetadata
- Tests provided for Export method for context when Project Metadata is
provided

[#170698084]

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -10,41 +10,43 @@ import (
 )
 
 const (
-	DefaultLayersDir     = "/layers"
-	DefaultAppDir        = "/workspace"
-	DefaultBuildpacksDir = "/cnb/buildpacks"
-	DefaultPlatformDir   = "/platform"
-	DefaultOrderPath     = "/cnb/order.toml"
-	DefaultGroupPath     = "./group.toml"
-	DefaultStackPath     = "/cnb/stack.toml"
-	DefaultAnalyzedPath  = "./analyzed.toml"
-	DefaultPlanPath      = "./plan.toml"
-	DefaultProcessType   = "web"
-	DefaultLauncherPath  = "/cnb/lifecycle/launcher"
-	DefaultLogLevel      = "info"
+	DefaultLayersDir           = "/layers"
+	DefaultAppDir              = "/workspace"
+	DefaultBuildpacksDir       = "/cnb/buildpacks"
+	DefaultPlatformDir         = "/platform"
+	DefaultOrderPath           = "/cnb/order.toml"
+	DefaultGroupPath           = "./group.toml"
+	DefaultStackPath           = "/cnb/stack.toml"
+	DefaultAnalyzedPath        = "./analyzed.toml"
+	DefaultPlanPath            = "./plan.toml"
+	DefaultProcessType         = "web"
+	DefaultLauncherPath        = "/cnb/lifecycle/launcher"
+	DefaultLogLevel            = "info"
+	DefaultProjectMetadataPath = "./project-metadata.toml"
 
-	EnvLayersDir         = "CNB_LAYERS_DIR"
-	EnvAppDir            = "CNB_APP_DIR"
-	EnvBuildpacksDir     = "CNB_BUILDPACKS_DIR"
-	EnvPlatformDir       = "CNB_PLATFORM_DIR"
-	EnvAnalyzedPath      = "CNB_ANALYZED_PATH"
-	EnvOrderPath         = "CNB_ORDER_PATH"
-	EnvGroupPath         = "CNB_GROUP_PATH"
-	EnvStackPath         = "CNB_STACK_PATH"
-	EnvPlanPath          = "CNB_PLAN_PATH"
-	EnvUseDaemon         = "CNB_USE_DAEMON"       // defaults to false
-	EnvUseHelpers        = "CNB_USE_CRED_HELPERS" // defaults to false
-	EnvRunImage          = "CNB_RUN_IMAGE"
-	EnvCacheImage        = "CNB_CACHE_IMAGE"
-	EnvCacheDir          = "CNB_CACHE_DIR"
-	EnvLaunchCacheDir    = "CNB_LAUNCH_CACHE_DIR"
-	EnvUID               = "CNB_USER_ID"
-	EnvGID               = "CNB_GROUP_ID"
-	EnvRegistryAuth      = "CNB_REGISTRY_AUTH"
-	EnvSkipLayers        = "CNB_ANALYZE_SKIP_LAYERS" // defaults to false
-	EnvProcessType       = "CNB_PROCESS_TYPE"
-	EnvProcessTypeLegacy = "PACK_PROCESS_TYPE" // deprecated
-	EnvLogLevel          = "CNB_LOG_LEVEL"
+	EnvLayersDir           = "CNB_LAYERS_DIR"
+	EnvAppDir              = "CNB_APP_DIR"
+	EnvBuildpacksDir       = "CNB_BUILDPACKS_DIR"
+	EnvPlatformDir         = "CNB_PLATFORM_DIR"
+	EnvAnalyzedPath        = "CNB_ANALYZED_PATH"
+	EnvOrderPath           = "CNB_ORDER_PATH"
+	EnvGroupPath           = "CNB_GROUP_PATH"
+	EnvStackPath           = "CNB_STACK_PATH"
+	EnvPlanPath            = "CNB_PLAN_PATH"
+	EnvUseDaemon           = "CNB_USE_DAEMON"       // defaults to false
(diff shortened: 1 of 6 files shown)


cmd/flags.go
```diff
@@ -11,41 +11,43 @@ import (
...
+	DefaultPlatformDir         = "/platform"
+	DefaultOrderPath           = "/cnb/order.toml"
+	DefaultGroupPath           = "./group.toml"
+	DefaultStackPath           = "/cnb/stack.toml"
+	DefaultAnalyzedPath        = "./analyzed.toml"
+	DefaultPlanPath            = "./plan.toml"
+	DefaultProcessType         = "web"
+	DefaultLauncherPath        = "/cnb/lifecycle/launcher"
+	DefaultLogLevel            = "info"
+	DefaultProjectMetadataPath = "./project-metadata.toml"
```
ekcasey: Given that this metadata is supplied to the lifecycle by the platform, I would advocate that it should live in `/platform/project-metadata.toml` by default. Organizing data supplied by the platform is the intended purpose of the platform directory. This would be analogous to the default location of env vars supplied by the platform (`/platform/env`).

Given that the platform dir is currently configurable, I don't think this file path needs to be separately configurable.

When you see other files have defaults like `./group.toml`, those are files written by one lifecycle phase for consumption by another. These files end up in the `/layers` dir. Currently the platform does not modify the layers dir and I think we should stick with that convention.

exporter_test.go
```diff
@@ -602,40 +603,71 @@ type = "Apache-2.0"
...
+					projectJSON, err := fakeAppImage.Label("io.buildpacks.project")
+					h.AssertNil(t, err)
+
+					var projectMD lifecycle.ProjectMetadata
+					if err := json.Unmarshal([]byte(projectJSON), &projectMD); err != nil {
+						t.Fatalf("badly formatted metadata: %s", err)
+					}
+
+					t.Log("adds project metadata to label")
+					h.AssertEq(t, projectMD.Source.Type, "git")
```
ekcasey: Just a suggestion: since you unmarshaled into a struct of the same type you passed in, you could simply assert that the two structs are equal for expediency.

ekcasey:
@iainsproat all commits must by signed off so that this PR passes the DCO check

cmd/flags.go
iainsproat: Thanks @ekcasey 
We've removed the flag for the project metadata file path, and instead expecting it at `project-metadata.toml` in the Platform directory.  Hopefully we've interpreted your comments correctly, do let us know if any other changes are required.

exporter_test.go
iainsproat: Good point, we've amended this 👍

project_metadata.go
```diff
@@ -0,0 +1,16 @@
...
+
+type ProjectMetadata struct {
+	Source ProjectSource `toml:"source" json:"source"`
+}
+
+type ProjectSource struct {
+	Type     string      `toml:"type" json:"type"`
+	Version  interface{} `toml:"version" json:"version"`
+	Metadata interface{} `toml:"metadata" json:"metadata"`
+}
```
Zanadar: Think we might need some `omitemptys` here if this is going on and off disk?

ekcasey: Instead of omitting metadata or version when they are not supplied I would prefer if the type of `Version` and `Metadata` changed to  `map[string]interface{}` and we ensured they were never nil. If, for example there was no metadata provided, I would prefer to see `{..., "metadata": {}}` in my label to `{..., "metadata": null}`.

This suggestion better matches the RFC in which all example values for `version` and `metadata` keys are maps.

iainsproat: Thanks for the comments @zmackie and @ekcasey 
I've amended `Version` and `Metadata` to be of type `map[string]interface{}` as it matches the intent of the RFC.

ekcasey:
@iainsproat I apologize for pushing you to move `project-metadata.toml` to the platform dir. After discussion during WG it seems that is not the best place given that this file is platform->lifecycle communication rather than platform->buildpack.

The layers dir `./project-metadata.toml` (where you originally had it) is still a slightly weird place to put this file, but given that we don't have a directory designed for these types of files yet, that is probably the best solution for now. Sorry again for the whiplash.

If you move the default locaiton back to `./project-metadata.toml`, I will accept and merge.

iainsproat:
@ekcasey - hopefully I've understood correctly; I've pushed a change.
The project metadata file defaults to `./project-metadata.toml`.
This is not-configurable by a flag, and is always expected in that location.

cmd/exporter/main.go
```diff
@@ -63,6 +65,7 @@ func init() {
 	cmd.FlagLogLevel(&logLevel)
 	cmd.FlagCacheImage(&cacheImageTag)
 	cmd.FlagCacheDir(&cacheDir)
+	cmd.FlagPlatformDir(&platformDir)
```
ekcasey: Apologies, I know these changes are necessary b/c I originally pushed you in the wrong direction, but if we remove this unnescessary `-platform` flag on exporter and put the `-project-metadata` flag back we will be good to merge

iainsproat: Good spot; this should now be resolved.  Thanks @ekcasey

zmackie:
@iainsproat Looks like this needs a rebase

carlo-colombo:
I squashed and rebased the commit on top of master. Let us know if everything is fine. Thanks!

carlo-colombo:
One final question about the location of the `project-metadata.toml` file. The location on the final image is `/layers/project-metadata.toml`, is this what you expect?

As the working dir of the builder is defined as `"WorkingDir": "/layers",` in the builder image (at least `cloudfoundry/cnb:bionic` `./project-metadata.toml` is `/layers/project-metadata.toml`.

cmd/lifecycle/exporter.go
```diff
@@ -109,6 +110,13 @@ func (e *exportCmd) Exec() error {
 		cmd.Logger.Infof("no stack metadata found at path '%s', stack metadata will not be exported\n", e.stackPath)
 	}
 
+	projectMetadataPath := path.Join(".", lifecycle.ProjectMetadataFileName)
+	var projectMD lifecycle.ProjectMetadata
+	_, err = toml.DecodeFile(projectMetadataPath, &projectMD)
+	if err != nil {
```
ekcasey: We should use `os.IsNotExist` to distinguish between missing project metadata and another type of error and fail in the latter case.

cmd/lifecycle/exporter.go
```diff
@@ -109,6 +110,13 @@ func (e *exportCmd) Exec() error {
 		cmd.Logger.Infof("no stack metadata found at path '%s', stack metadata will not be exported\n", e.stackPath)
 	}
 
+	projectMetadataPath := path.Join(".", lifecycle.ProjectMetadataFileName)
```
ekcasey: We should put the `-project-metadata` flag back to make this configurable

project_metadata.go
```diff
@@ -0,0 +1,16 @@
+package lifecycle
+
+const (
+	ProjectMetadataLabel    = "io.buildpacks.project"
+	ProjectMetadataFileName = "project-metadata.toml"
```
ekcasey: The default value for this file should live in the `cmd` package with the other defaults

ekcasey:
@carlo-colombo The default location looks right but we are missing the flag to make it configurable

matthewmcnew:
Will this constitute a bump to the platform api?

carlo-colombo:
@matthewmcnew I am not sure about the api policies for this repository, but the change introduced by the PR is retro compatible and not breaking as it just log a message if the file is not found in the expected location.

@ekcasey I introduced back the flag for passing the `project-metadata.toml` path

matthewmcnew:
A bump in the platform api would be useful to know when the `project-metadata` flag is available.

cmd/flags.go
```diff
@@ -133,6 +134,10 @@ func FlagLogLevel(level *string) {
 	flagSet.StringVar(level, "log-level", envOrDefault(EnvLogLevel, DefaultLogLevel), "logging level")
 }
 
+func FlagProjectMetadataPath(projectMetadataPath *string) {
+	flagSet.StringVar(projectMetadataPath, "project-metadata", envOrDefault(EnvProjectMetadataPath, "./project-metadata.toml"), "path to project-metadata.toml")
```
ekcasey: the default value should be grouped with the other defaults near the top of the file

carlo-colombo: Sorry I miss the pattern here. I add a constant `DefaultProjectMetadataPath`

PR #232, opened 2020-01-21
Move lifecycle build phases into single binary

### LINKED ISSUE #191: symlink lifecycle binaries to a single binary
Copying the six binaries of the lifecycle adds 72Mb to every builder image.

We could build a single binary instead and create symlinks for each phase of the lifecycle. It would work like busybox.

This would also make the build much faster.

wdyt?

ekcasey:
* excludes launcher
* Closes #191

Signed-off-by: Ashwin Venkatesh <avenkatesh@pivotal.io>


ekcasey:
@buildpacks/owners  do we feel like this requires an RFC?

Makefile
```diff
@@ -25,26 +26,44 @@ all: test build package
...
+	ln -sf detector $(BUILD_DIR)/lifecycle/exporter
+	ln -sf detector $(BUILD_DIR)/lifecycle/rebaser
 
 build-windows:
 	@echo "> Building for windows..."
-	mkdir -p ./out/lifecycle
-	GOOS=windows $(GOENV) $(GOBUILD) -o ./out/lifecycle -a ./cmd/...
+	mkdir -p $(BUILD_DIR)/lifecycle
+	GOOS=linux $(GOENV) $(GOBUILD) -o $(BUILD_DIR)/lifecycle -a ./cmd/launcher
+	GOOS=linux $(GOENV) $(GOBUILD) -o $(BUILD_DIR)/lifecycle/detector -a ./cmd/build
```
jromero: Couple things:

- `GOOS` should be windows.
- output file should end in .exe

Are we using this? Should we remove windows build until we verify that it works? AFAIK that support doesn't exist yet.

acceptance/acceptance_test.go
```diff
@@ -15,40 +14,111 @@ import (
 	h "github.com/buildpacks/lifecycle/testhelpers"
 )
 
+var dir string
+
 func TestAcceptance(t *testing.T) {
+	if os.Getenv("ACCEPTANCE") == "" {
```
jromero: Why not just use the `+build acceptance` annotation?

jromero: How do you then execute this via IDE do you need to configure a run configuration with the env var? That seems more painful IMO.

We haven't had any issues with build constraints in `pack`. Could you elaborate?

acceptance/acceptance_test.go
```diff
@@ -57,31 +127,11 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
...
-	return cmd
-}
-
-func parseCommand(command string) (binary string, args []string) {
-	parts := strings.Split(command, " ")
-	return parts[0], parts[1:]
+func buildBinaries(t *testing.T) {
+	cmd := exec.Command("make", "build-"+runtime.GOOS)
+	cmd.Dir = ".."
+	cmd.Env = append(os.Environ(), "BUILD_DIR="+dir, "LIFECYCLE_VERSION=some-version", "SCM_COMMIT=asdf123")
```
jromero: A little pedantic but could we simply pass in `dir` here instead of having a "global" variable.

acceptance/acceptance_test.go
ekcasey: It's about the same mount of work either way. In Goland for example, you either need a Run Configuration for the acceptance tests with `ACCEPTANCE` set or you need to add a custom build tag to your Project Settings.

I would say that the advantages of the env var are:
* Goland will compile the file for you by default regardless (and show compilation errors)
* If you try to run it without the appropriate setup it explicitly prints out why it is skipping

acceptance/acceptance_test.go
ekcasey: `dir` is also used by the tests themselves https://github.com/buildpacks/lifecycle/blob/one-binary/acceptance/acceptance_test.go#L117

acceptance/acceptance_test.go
jromero: In my experience adding the build constraint is an auto-fix action with the IDE. Is that not the case for you?

ekcasey:
@jromero yes, let's hold on approval of buildpacks/rfcs#45

ekcasey:
Implements bulidpacks/rfcs#45

cmd/lifecycle/detector.go
```diff
@@ -1,60 +1,40 @@
...
-func main() {
-	// suppress output from libraries, lifecycle will not use standard logger
-	log.SetOutput(ioutil.Discard)
-
-	flag.Parse()
-
-	if printVersion {
-		cmd.ExitWithVersion()
-	}
+func (d *detectCmd) Flags() {
```
sclevine: Maybe `InitFlags` or even just `Init`?

cmd/lifecycle/detector.go
```diff
@@ -1,60 +1,40 @@
...
+	cmd.FlagOrderPath(&d.orderPath)
+	cmd.FlagGroupPath(&d.groupPath)
+	cmd.FlagPlanPath(&d.planPath)
+}
 
-	if err := cmd.SetLogLevel(logLevel); err != nil {
-		cmd.Exit(err)
+func (d *detectCmd) Args(nargs int, args []string) error {
+	if nargs != 0 {
+		return cmd.FailErrCode(errors.New("received unexpected Args"), cmd.CodeInvalidArgs, "parse arguments")
```
sclevine: Args -> arguments

Makefile
```diff
@@ -94,13 +114,17 @@ unit: verify-jq format lint install-yj
...
+	$(GOTEST) -v -count=1 -tags=acceptance ./acceptance/...
 
 clean:
 	@echo "> Cleaning workspace..."
 	rm -rf $(BUILD_DIR)
 
 package: descriptor
 	@echo "> Packaging lifecycle..."
-	tar czf $(BUILD_DIR)/$(ARCHIVE_NAME).tgz -C out lifecycle.toml lifecycle
+	tar czf ./out/$(ARCHIVE_NAME).tgz -C out lifecycle.toml lifecycle
```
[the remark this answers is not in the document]
ekcasey: nope

cmd/flags.go
```diff
@@ -47,88 +47,94 @@ const (
 	EnvLogLevel          = "CNB_LOG_LEVEL"
 )
 
+var flagSet *flag.FlagSet
+
+func init() {
+	flagSet = flag.NewFlagSet("lifecycle", flag.ExitOnError)
```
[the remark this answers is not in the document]
ekcasey: TIL

PR #233, opened 2020-01-22
Implementation of Platform API verification

### LINKED ISSUE #224: Support platform API verification
Implement https://github.com/buildpacks/spec/pull/67 (https://github.com/buildpacks/rfcs/pull/34)

### Acceptance Criteria

The acceptance criteria apply to all phase related binaries:

- analyzer
- builder
- detector
- exporter
- restorer

#### AC 1: Env var is not set
Given environment variable `CNB_PLATFORM_API` is not set
When I run `detector` with valid inputs
Then completes successfully
And exits with exit code `0`

#### AC 2: Env var is set to supported version
Given environment variable `CNB_PLATFORM_API` is set to `0.2`
And `detector` is compiled to support `0.2`
When I run `detector` with valid inputs
Then completes successfully
And exits with exit code `0`

#### AC 3: Env var is set to unsupported version
Given environment variable `CNB_PLATFORM_API` is set to `0.3`
And `detector` is compiled to support `0.2`
When I run `detector` with valid inputs
Then outputs error `The Lifecycle's Platform API version is 0.2 which is incompatible with Platform API version 0.3.`
And exits with exit code `11`

jromero:
Implements verification according to https://github.com/buildpacks/spec/pull/67.

Caveats:
- I did not create a _compatible_ test for every binary due to the level of scaffolding necessary.
- There isn't any mention of `rebaser` in the RFC or spec so I left it out. Should it be included?

Notes: 
- @ekcasey Can I get extra 👀on the assertions regarding _compatiblity_ [here](https://github.com/buildpacks/lifecycle/pull/233/files#diff-4da20a655492cafa733c079fc42c1985)?
    At this time it's not all too relevant given we are pre-stable but would like to document it accurately.
- Stole some cool stuff from https://github.com/buildpacks/lifecycle/pull/232. Nonetheless, we're going to have some merge conflicts.

Resolves #224


api/version.go
```diff
@@ -0,0 +1,130 @@
+package api
```
jromero: Lifted this from https://github.com/buildpacks/pack/blob/master/internal/api/version.go. If this is the right way to go we could remove pack's version and simply rely on this one.

acceptance/acceptance_test.go
```diff
@@ -15,13 +17,83 @@ import (
...
+			tmpWorkingDir := func(sourceDir string) string {
+				d, err := ioutil.TempDir(tmpDir, "")
+				h.AssertNil(t, err)
+				h.RecursiveCopy(t, sourceDir, d)
+				return d
+			}
+
+			when("is not set", func() {
+				it("should complete successfully", func() {
+					cmd := lifecycleCmd("analyzer", "some/image")
```
ekcasey: If there was ever a "some/image" added to dockerhub this would write a bunch of layers to the workstation filesystem. I think we should point at an image that will never exist if that is the intention `example.com/com/image`

ekcasey: Also, we should point the layers dir and to a temp directory or even `/dev/null`. This doesn't pollute the workstation because, right now, when the image doesn't exist nothing gets written. But, if the behavior changes or people add more acceptance tests following this pattern it could get messy.

acceptance/acceptance_test.go
```diff
@@ -15,13 +17,83 @@ import (
...
+	it.After(func() {
+		os.RemoveAll(tmpDir)
+	})
 
 	when("All", func() {
+		when("CNB_PLATFORM_API", func() {
+			tmpWorkingDir := func(sourceDir string) string {
+				d, err := ioutil.TempDir(tmpDir, "")
+				h.AssertNil(t, err)
+				h.RecursiveCopy(t, sourceDir, d)
```
ekcasey: If the only purpose of this is to provide a `group.toml` file, why not just pass `-group=<path/to/test/fixture>` instead of moving this to the default path

api/version.go
```diff
@@ -0,0 +1,132 @@
...
+	if fromLifecycle.major != 0 {
+		return fromLifecycle.major == fromPlatform.major && fromLifecycle.minor >= fromPlatform.minor
+	}
+
+	return false
+}
+
+// IsBuildpackAPICompatible determines if the lifecycle's Buildpack API version is compatible with the buildpack's
+// Buildpack API version based on the spec.
+func IsBuildpackAPICompatible(fromLifecycle, fromBuildpack *Version) bool {
```
ekcasey: This is duplicate of `IsPlatformAPICompatible` with a couple different variable names. We should just have a single `IsAPICompatible`

cmd/analyzer/main.go
```diff
@@ -36,6 +36,10 @@ var (
 )
 
 func init() {
+	if err := cmd.VerifyCompatibility(); err != nil {
```
ekcasey: I think `analzyer -version` should still work even if the API isn't compatible. It would help folks figure out why the API isn't compatible.

ekcasey: same for the rest of the phases

cmd/compatibility.go
```diff
@@ -0,0 +1,32 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/buildpacks/lifecycle/api"
+)
+
+func VerifyCompatibility() error {
```
ekcasey: I don't feel strongly but this could probably move into `version.go` where PlatformAPI is defined

cmd/compatibility.go
```diff
@@ -0,0 +1,32 @@
...
+	platformsAPI := os.Getenv("CNB_PLATFORM_API")
```
ekcasey:
```suggestion
	platformAPI := os.Getenv("CNB_PLATFORM_API")
```

acceptance/acceptance_test.go
jromero: The working directory is a temp directory. That's the purpose and function of `tmpWorkingDir`.

cmd/analyzer/main.go
ekcasey: given the difficulty entailed in implementing this suggestion, we can punt on it for now

ekcasey:
@jromero I think `rebaser` should be included. It should be part of the platform API descpite the fact that the platform API is incompletely specified.

zmackie:
Accepted with:
```bash
± zm |master U:1 ✗| → docker run -it -e CNB_PLATFORM_API=0.1 cnb/sample:accept-lifecycle-0.0.6
cnb@7a46687c39a6:/layers$ /lifecycle/detector
ERROR: failed to : the Lifecycle's Platform API version is 0.2 which is incompatible with Platform API version 0.1
cnb@7a46687c39a6:/layers$ exit
  |2.6.3| NY-Floater-15565 in ~/workspace/samples
± zm |master U:1 ✗| → docker run -it -e CNB_PLATFORM_API=0.2 cnb/sample:accept-lifecycle-0.0.6
cnb@5d6c4e2002ea:/layers$ /lifecycle/detector
ERROR: No buildpack groups passed detection.
ERROR: Please check that you are running against the correct path.
ERROR: failed to detect: no buildpacks participating
cnb@5d6c4e2002ea:/layers$ exit
  |2.6.3| NY-Floater-15565 in ~/workspace/samples
± zm |master U:1 ✗| → docker run -it -e CNB_PLATFORM_API=0.3 cnb/sample:accept-lifecycle-0.0.6
cnb@e252b0b054a7:/layers$ /lifecycle/detector
ERROR: failed to : the Lifecycle's Platform API version is 0.2 which is incompatible with Platform API version 0.3
cnb@e252b0b054a7:/layers$ exit
  |2.6.3| NY-Floater-15565 in ~/workspace/samples
± zm |master U:1 ✗| → docker run -it -e CNB_PLATFORM_API= cnb/sample:accept-lifecycle-0.0.6
cnb@90d6bce17f99:/layers$ /lifecycle/detector
ERROR: No buildpack groups passed detection.
ERROR: Please check that you are running against the correct path.
ERROR: failed to detect: no buildpacks participating
cnb@90d6bce17f99:/layers$ exit
  |2.6.3| NY-Floater-15565 in ~/workspace/samples
± zm |master U:1 ✗| → docker run -it -e CNB_PLATFORM_API='' cnb/sample:accept-lifecycle-0.0.6
cnb@97c404171261:/layers$ /lifecycle/detector
ERROR: No buildpack groups passed detection.
ERROR: Please check that you are running against the correct path.
ERROR: failed to detect: no buildpacks participating
cnb@97c404171261:/layers$ exit
```

PR #242, opened 2020-01-30
Only calculate layer shas once

ekcasey:
* improves exporter performance
* removes unused testmocks

Cannot merge until https://github.com/buildpacks/imgutil/pull/22 merges


cache/image_cache.go
```diff
@@ -72,22 +72,22 @@ func (c *ImageCache) RetrieveMetadata() (lifecycle.CacheMetadata, error) {
 	return meta, nil
 }
 
-func (c *ImageCache) AddLayerFile(sha string, tarPath string) error {
+func (c *ImageCache) AddLayerFile(tarPath string, diffID string) error {
```
jromero: This seems to be a **breaking change** in the API. It's particularly scary because of how ellusive it could be given that the method signature is the same as far as types are concerned.

ekcasey: It seems unlikely that github.com/buildpacks/lifecycle/cache has external consumers https://godoc.org/github.com/buildpacks/lifecycle/cache?importers

PR #243, opened 2020-02-03
Lifecycle/creator

ekcasey:
Requires buildpacks/rfcs#46 to merge


builder.go
```diff
@@ -78,6 +81,11 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 	var bom []BOMEntry
 	var slices []Slice
 
+	currentUser := os.Getuid()
```
jromero: Is this a separate feature or somehow necessary for this to work?

ekcasey: `pack` is going to run `creator` as `root`, when exporting to the daemon. Therefore we need to fix permissions and explicitly run buildpacks as an unprivileged user.

user_unix.go
```diff
@@ -0,0 +1,68 @@
...
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Credential: &syscall.Credential{
+			Uid: uint32(uid),
+			Gid: uint32(gid),
+		},
+	}
+	return cmd, nil
+}
+
+func ensureOwner(path string, uid, gid int) error {
```
sclevine: This method behaves unexpected if you aren't root (by failing to ensure ownership and not returning an error).

user_unix.go
```diff
@@ -0,0 +1,68 @@
...
+	"os/user"
+	"path/filepath"
+	"strconv"
+	"syscall"
+)
+
+var currentUser = os.Getuid()
+
+// asUser sets HOME and non-root user credentials on command
+func asUser(cmd *exec.Cmd, uid, gid int) (*exec.Cmd, error) {
```
sclevine: We should probably check that `uid, gid int` actually matches `os.Getuid` and `os.Getgid` when non-root.

env/launch_test.go
```diff
@@ -0,0 +1,49 @@
+package env
```
sclevine: Use `env_test` to match other testing patterns

env/env_test.go
```diff
@@ -1,19 +1,16 @@
-package lifecycle_test
+package env
```
sclevine: `env_test`

env/env.go
```diff
@@ -8,31 +8,39 @@ import (
...
-	Environ   func() []string
-	Map       map[string][]string
+	RootDirMap map[string][]string
+	vars       map[string]string
+}
+
+func varsFromEnviron(environ []string, removeKey func(string) bool) map[string]string {
+	vars := make(map[string]string)
+	for _, kv := range environ {
+		parts := strings.Split(kv, "=")
```
sclevine: Use `strings.SplitN` to guarantee that the string is only split on the first `=`

cmd/lifecycle/analyzer.go
```diff
@@ -73,20 +78,15 @@ func (a *analyzeCmd) Exec() error {
...
-	}
-
-	cacheStore, err := initCache(a.cacheImageTag, a.cacheDir)
-	if err != nil {
-		return err
+		return cmd.FailErr(err, "get previous image")
 	}
 
-	md, err := analyzer.Analyze(img, cacheStore)
+	analyzedMd, err := analyzer.Analyze(img, cacheStore)
```
sclevine: `Md` -> `MD`

detector.go
```diff
@@ -235,12 +204,24 @@ func (bp *buildpackTOML) Detect(c *DetectConfig) detectRun {
 		return detectRun{Code: -1, Err: err}
 	}
 	defer os.RemoveAll(planDir)
+	if err := ensureOwner(planDir, c.UID, c.GID); err != nil {
```
sclevine: Let's chat about these changes to `Builder` and `Detector`. Having these sections of code run as root while writing files and modifying the environment seems error prone and worries me from a security perspective. I'm especially worried about untrusted input from app/buildpack artifacts that could accidentally be executed as root.

cmd/lifecycle/builder.go
```diff
@@ -36,6 +37,11 @@ func (b *buildCmd) Args(nargs int, args []string) error {
 	return nil
 }
 
+func (b *buildCmd) DropPrivileges() error {
+	// builder should never be run with privileges
```
sclevine: Maybe return an error if UID != 0?

cmd/lifecycle/creator.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return cmd.FailErrCode(err, cmd.CodeFailedBuild, "build")
+	}
+
+	if err := lifecycle.WriteTOML(lifecycle.MetadataFilePath(c.layersDir), md); err != nil {
+		return cmd.FailErr(err, "write metadata")
+	}
+
+	cmd.Logger.Info("---> EXPORTING")
+	return export(exportArgs{
```
sclevine: Maybe a shared/embedded struct with a subset of the fields could clean this up?

env/bulid.go
```diff
@@ -0,0 +1,46 @@
...
+
+var BuildEnvWhitelist = []string{
+	"CNB_STACK_ID",
+	"HOSTNAME",
+	"HOME",
+}
+
+func NewBuildEnv(environ []string) *Env {
+	return &Env{
+		RootDirMap: POSIXBuildEnv,
```
sclevine: Do these fields need to be exported?

ekcasey: If we don't put the tests in the `env` package allow us to test all of the behavior. I know there are some work arounds but maybe we can come back to that later.

PR #246, opened 2020-02-06
Faster Untar

dgageot:
+ Avoid changing file permissions two times
 + Reuse the copy buffer
 + Don’t check if a folder exists if we just created it

On my Mac, with a 1.2G tarball (go-build cache folder), this takes consistently around 4.5s rather than consistently around 6s.
It’s harder to test with a cold fs cache, though.

Signed-off-by: David Gageot <david@gageot.net>

### DIFF
--- a/archive/tar.go
+++ b/archive/tar.go
@@ -223,8 +223,8 @@ type PathMode struct {
 
 func Untar(r io.Reader, dest string) error {
 	// Avoid umask from changing the file permissions in the tar file.
-	old := unix.Umask(0)
-	defer unix.Umask(old)
+	umask := unix.Umask(0)
+	defer unix.Umask(umask)
 
 	buf := make([]byte, 32*32*1024)
 	dirsFound := make(map[string]bool)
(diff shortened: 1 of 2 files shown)


dgageot:
Obviously, I broke something ;-)

archive/tar.go
```diff
@@ -247,18 +256,20 @@ func Untar(r io.Reader, dest string) error {
...
+
 		case tar.TypeReg, tar.TypeRegA:
-			_, err := os.Stat(filepath.Dir(path))
-			if os.IsNotExist(err) {
-				if err := os.MkdirAll(filepath.Dir(path), os.ModePerm); err != nil {
-					return err
+			dirPath := filepath.Dir(path)
+			if !dirsFound[dirPath] {
+				if _, err := os.Stat(dirPath); os.IsNotExist(err) {
+					if err := os.MkdirAll(dirPath, os.ModePerm); err != nil {
```
ekcasey: I believe this PR has changed the behavior that governs the directory permissions in this particular case. If there is no header for a directory in the archive, the umask should be applied. Because we have unset the umask, directories that fall into this case will all be created with perm `0777`.

dgageot: Thanks for spotting that! I'll fix that and most probably add a test case

PR #252, opened 2020-02-12
Create issue templates

zmackie:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,39 @@
...
+
+##### Expected
+<!--- What did you expect to happen -->
+
+
+---
+
+### Environment
+
+##### pack report
```
ekcasey: Lifecycle bugs might show up in platforms other than `pack`. Instead of `pack report` and `docker info`  we could put something more generic like `platform` and then provide `pack` and `pack report` as an example of how this could be filled out.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,39 @@
...
+3.
+
+##### Expected
+<!--- What did you expect to happen -->
+
+
+---
+
+### Environment
+
```
ekcasey: we should add lifecycle version to this section

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,39 @@
...
+### Reproduction
+
+##### Steps
+<!--- What steps should be taken to reproduce the issue? -->
+
+1.
+2.
+3.
+
+##### Expected
```
ekcasey: We probably want to include current behavior in addition to expected behavior

Issue #244, opened 2020-02-04
Implement store.toml


ekcasey:
Implement [store.toml](https://github.com/buildpacks/spec/blob/master/buildpack.md#storetoml-toml) as outlined in the spec.

### `exporter`
For each buildpack `<bp.id>`, that writes a `<layer-dir>/<bp.id>/store.toml` file during the build phase, `exporter` should add the JSON equivalent of the `store.toml` contents to a key named `store` in the corresponding buildpack map in the `io.buildpacks.lifecycle.metadata` label

### `analyzer`
`analyzer` should read the previous image label and restore the contents of each buildpack's store to `<layer-dir>/<bp.id>/store.toml`.

`analyzer` should restore `store.toml` even when the `-skip-layers` flag is passed.

### Examples
**Example File:**
`/layers/bp.one/store.toml`
```toml
[metadata]
  some-key = "some-val"
```

**Example Label:**
`io.buildpack.lifecycle.metadata`
```json
{
  "buildpacks": [
    {
      "key": "bp.one",
      "version": "bp.one.version",
      "layers": {...},
      "store": {"metadata": {"some-key": "some-val"}}
   },
  ...
}
```

PR #256, opened 2020-02-14
store.toml

ekcasey:
Resolves: #244

### DIFF
--- a/analyzer_test.go
+++ b/analyzer_test.go
@@ -174,6 +174,19 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				h.AssertEq(t, md.Metadata, appImageMetadata)
 			})
 
+			it("restores each store metadata", func() {
+				_, err := analyzer.Analyze(image, testCache)
+				h.AssertNil(t, err)
+				for _, data := range []struct{ name, want string }{
+					// store.toml files.
+					{"metadata.buildpack/store.toml", "[metadata]\n  [metadata.metadata-buildpack-store-data]\n    store-key = \"store-val\""},
+					{"no.cache.buildpack/store.toml", "[metadata]\n  [metadata.no-cache-buildpack-store-data]\n    store-key = \"store-val\""},
+				} {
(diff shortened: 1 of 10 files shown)


analyzer_test.go
```diff
@@ -347,7 +363,48 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+						h.AssertStringContains(t, string(got), data.want)
+					}
+				})
+
+				when("analyzer is running as root", func() {
+					it.Before(func() {
+						if os.Getuid() != 0 {
+							t.Skip("Skipped when not running as root")
+						}
+					})
```
zmackie: Do tests usually get run as root? I worry that this could sneak past with just a warning. Maybe jsut being paranoid....

testdata/analyzer/app_metadata.json
```diff
@@ -1,59 +1,73 @@
...
+          },
+          "launch": true,
+          "sha": "some-layer-sha"
+        }
+      },
+      "store": {
+        "metadata": {
+          "no-cache-buildpack-store-data": {
+            "store-key": "store-val"
+          }
```
zmackie: is the access path for this data always going to be `store.metadata.<BUILDPACK_STORE_DATA`? I wonder if we need the second level of `metadata`?

analyzer_test.go
ekcasey: These do run as root in concourse, which is essentially the only time this particular test gets run. We should probably find a better way to test this behavior in the future.

ekcasey:
@zmackie maybe @hone can jump in here. I know he has a couple specific examples from real buildpacks, which would be better than my made up examples.

hone:
@zmackie the intent of `store.toml` is to have a persistent key/value store. In the heroku ruby buildpack, we have a default Ruby version if none is specified. We pin the version that they use upon first deployment. This way, if we update the default Ruby version it won't break their application. If the cache gets clear, we still want to ensure these values are saved. /cc @schneems

PR #258, opened 2020-02-19
introduces ExporterOptions

ekcasey:
replace all 10,000 parameters with an options struct


hone:
Every time I pair with @ekcasey on a feature, we get distracted by refactoring lifecycle 🤔

ekcasey:
included in https://github.com/buildpacks/lifecycle/pull/260

PR #261, opened 2020-02-25
Build linux artifact in a Github Action

simonjjones:
Signed-off-by: Simon Jones <simonjones@vmware.com>

This PR adds a Github actions for new PRs and pushes to master which triggers unit testing, building and packaging of the lifecycle artifact for Linux.

I do have an outstanding concern that this defaults to version `0.0.0`, however this is not intended to immediately replace the current build process, so versioned artifacts will continue to be created and uploaded to our GCS bucket with the correct version number.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,40 @@
+name: build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  test-and-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up go
+        uses: actions/setup-go@v2-beta
+        with:
+          go-version: '1.13'
+      - name: Set up go env
+        run: |
+          echo "::set-env name=GOPATH::$(go env GOPATH)"
+          echo "::add-path::$(go env GOPATH)/bin"
+        shell: bash
+      - name: Install jq
+        run: |
+          mkdir -p deps/bin
+          curl -s -L -o deps/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+          chmod +x deps/bin/jq
+          echo "::add-path::${PWD}/deps/bin"
+      - name: Test
+        run: make test
+      - name: Build
+        run: |
+          make build-linux
+          make package-linux
+      - uses: actions/upload-artifact@v1
+        with:
+          name: lifecycle
+          path: out/lifecycle-v0.0.0+linux.x86-64.tgz
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,8 @@ on:
 jobs:
   test-and-build:
     runs-on: ubuntu-latest
+    env:
+      LIFECYCLE_VERSION: 0.7.0
     steps:
       - uses: actions/checkout@v2
       - name: Set up go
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          go-version: '1.13'
+      - name: Set up go env
+        run: |
+          echo "::set-env name=GOPATH::$(go env GOPATH)"
+          echo "::add-path::$(go env GOPATH)/bin"
+        shell: bash
+      - name: Install jq
+        run: |
+          mkdir -p deps/bin
```
jromero: I believe jq can be installed via apt-get in Ubuntu

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+          echo "::add-path::${PWD}/deps/bin"
+      - name: Test
+        run: make test
+      - name: Build
+        run: |
+          make build-linux
+          make package-linux
+      - uses: actions/upload-artifact@v1
+        with:
+          name: lifecycle
```
jromero: Should this at least be named `lifecycle-linux-x86-64` to future proof against Windows?

simonjjones: Perhaps, I feel like by the time we arrive at that world, this will be versioned anyway and the format will have to change regardless, so did the minimal to get things moving.

.github/workflows/build.yml
simonjjones: We had chosen to hard code our version in our existing CI, I wasn't sure whether or not that was important, so carried over the practice. Happy to explore aptitude installation if we think that's a preference.

simonjjones:
Any thoughts on removing the travis integration as part of this PR, separately or not at all?

jromero:
> Any thoughts on removing the travis integration as part of this PR, separately or not at all?

FWIW, I'm in favor of removing travis integration.

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+          path: out/lifecycle-v0.0.0+linux.x86-64.tgz
```
ekcasey: Can we change from hardcoding `0.0.0`, to hardcoding the next version `0.7.0`, in the action and in the Makefile? I think that would be better until we set up automation to bump the version after releases.

.github/workflows/build.yml
ekcasey: I agree with @jromero on this one

.github/workflows/build.yml
ekcasey: Does `path` support globs here (i.e. `out/lifecycle-v*+linux.x86-64.tgz`)?

simonjjones: > Does path support globs here (i.e. out/lifecycle-v*+linux.x86-64.tgz)?

No, however https://github.com/actions/upload-artifact/issues/11 has recent-ish activity, so perhaps soon?

We could use variables rather than hardcoding, which would be a step towards automation of the bumpings

.github/workflows/build.yml
```diff
@@ -0,0 +1,42 @@
...
+        shell: bash
+      - name: Install jq
+        run: |
+          mkdir -p deps/bin
+          curl -s -L -o deps/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+          chmod +x deps/bin/jq
+          echo "::add-path::${PWD}/deps/bin"
+      - name: Test
+        run: make test
+      - name: Build
```
ekcasey: We should also `build-windows` here to ensure it continues to compile even though we don't need to upload the not-yet-usable artifact.

PR #267, opened 2020-03-06
Fix scopes on auth when app/run images are on different registries

nanzhong:
This bumps the github.com/google/go-containerregistry dependency so that a fix for supporting app and run images on from different registries is pulled in.

In particular, https://github.com/google/go-containerregistry/pull/604 is the fix. Previously, when the app and run image are on different registries, the app image registry would be asked for the pull scope of the run image which will fail since it's not present.

I'm not sure if it's ideal to bump to the lastest commit or to the specific one that fixes the issue. Happy to change to whatever is more desired.


nanzhong:
@ekcasey Done.

PR #268, opened 2020-03-11
uses musl libc setres{g,u}id

ekcasey:
* libc pushes uid/gid change to all threads
* pull in imgutil w/ bugfixes that ensure that docker client only needs a single socket connection


Makefile
```diff
@@ -1,21 +1,21 @@
 GOCMD?=go
 GOFLAGS?=-mod=vendor
-GOENV=GOARCH=amd64 CGO_ENABLED=0
+GOENV=GOARCH=amd64 CGO_ENABLED=1
```
sclevine: CGO should remain disabled for launcher.

priv/docker.go
```diff
@@ -1,7 +1,8 @@
-package docker
+package priv
```
sclevine: Maybe use `internal` if this package is supposed to be private?

priv/docker.go
```diff
@@ -46,11 +46,15 @@ func (*unclosableConn) Close() error {
 
 // connectSockOpt overrides the client dialer to always return the same socket connection
 func connectSockOpt(host *url.URL) (client.Opt, error) {
-	socketConn, err := net.Dial("unix", host.Path)
-	if err != nil {
-		return nil, err
+	conns := make(chan *unclosableConn, 5)
+	for i := 0; i < 5; i++ {
```
sclevine: `len(conns)` (or even better, `runtime.NumCPU`?)

priv/docker.go
ekcasey: `priv` is for "privileges" not "private". Better naming suggestions welcome. I don't want to introduce `internal` w/o taking the time to be more thoughtful about which existing packages should be in it.

priv/docker.go
ekcasey: outdated: moving in a different direction (preventing the need for multiple conns)

PR #270, opened 2020-03-13
launcher sets PATH in env before looking up executable

ekcasey:
This seemed like the quickest/simplest fix for the issue but there are other ways we could approach it.


env/bulid.go
```diff
@@ -1,5 +1,7 @@
 package env
```
dgageot: Not related but the file should be `build.go`, not `bulid.go`

ekcasey: thx

Issue #271, opened 2020-03-13
Analyze fails with kubernetes mounted volumes


matthewmcnew:
When executing the lifecycle inside a kubernetes pod we specify the pod's `securityContext.fsgroup` to mount volumes as the CNB_USER. This will be result in the volume mounts for `/layers` and `/cache` owned by root with the group specified `fsGroup`. 

The [EnsureOwner](https://github.com/buildpacks/lifecycle/blob/master/priv/user_linux.go#L45) lifecycle check will still attempt to chown the volume mounts on these volumes because it checks both the uid and gid. This results in an error on analyze:

```
failed to chown volumes: chown /layers: operation not permitted
```

ekcasey:
I think we can assume that, if the `gid`, is correctly, lifecycle doesn't need to fix the permissions. This shouldn't break the `pack` case when we need to chown the volumes, but should fix the permissions error you are seeing in k8s.

matthewmcnew:
After testing #272 on minikube, it seems that not all storage classes will accurately respect the `securityContext.fsgroup`. Is it possible to just check that those directories are writeable?

PR #273, opened 2020-03-13
support world writable pvcs on k8

ekcasey:
If user does not own volume, check thoroughly for write permissions before attempting to chown.

context: https://github.com/buildpacks/lifecycle/issues/271#issuecomment-598926266


ashwin-venkatesh:
Verified that this fix works against minikube, a gke and pks cluster

PR #289, opened 2020-05-01
Only lookup user when necessary

### LINKED ISSUE #286: Only lookup user when necessary
### Problem

If we do https://github.com/buildpacks/lifecycle/issues/285 the implementation will be complicated by [this check](https://github.com/buildpacks/lifecycle/blob/4e6f6de75ef4a8c183ac5f4bfa5a0f645852de0c/priv/user_linux.go#L107-L110) and the use of `user` to set `HOME`, `USER`, and `DOCKER_CONFIG`.

We will not know when creating the "lifecycle image" what the `CNB_USER_ID` will be in the stack that users provide. Therefore, to use the image during a build, platforms like `pack` would need to modify the image on the fly, which is slow and cumbersome. It's not clear how platforms like Tekton would get around this.

### Possible solution

On spiking this out I think the referenced code can be removed for the `analyze`, `restore`, and `export` phases (or at least, `pack` acceptance tests are happy with it gone). We'd just need to set the `CNB_USER_ID` in the environment.

### Questions

* Do we want to do this? Is there another way around this problem?
* Could we make the change for other lifecycle phases besides `analyze`, `restore`, and `export`?

natalieparellano:
Fixes #286 

### DIFF
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -63,7 +63,7 @@ func (r *rebaseCmd) Privileges() error {
 			return cmd.FailErr(err, "initialize docker client")
 		}
 	}
-	if err := priv.RunAs(r.uid, r.gid); err != nil {
+	if err := priv.RunAs(r.uid, r.gid, true); err != nil {
 		cmd.FailErr(err, fmt.Sprintf("exec as user %d:%d", r.uid, r.gid))
 	}
 	return nil
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -63,7 +63,7 @@ func (r *rebaseCmd) Privileges() error {
 			return cmd.FailErr(err, "initialize docker client")
 		}
 	}
-	if err := priv.RunAs(r.uid, r.gid, false); err != nil {
+	if err := priv.RunAs(r.uid, r.gid); err != nil {
 		cmd.FailErr(err, fmt.Sprintf("exec as user %d:%d", r.uid, r.gid))
 	}
 	return nil
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -63,7 +63,7 @@ func (r *rebaseCmd) Privileges() error {
 			return cmd.FailErr(err, "initialize docker client")
 		}
 	}
-	if err := priv.RunAs(r.uid, r.gid, true); err != nil {
+	if err := priv.RunAs(r.uid, r.gid, false); err != nil {
 		cmd.FailErr(err, fmt.Sprintf("exec as user %d:%d", r.uid, r.gid))
 	}
 	return nil
--- a/priv/user_linux.go
+++ b/priv/user_linux.go
@@ -101,8 +101,7 @@ func recursiveEnsureOwner(path string, uid, gid int) error {
 }
 
 // RunAs sets the user ID and group ID of the calling process.
-// If withUserLookup is true, the user ID must be valid in the image where the process is running.
-func RunAs(uid, gid int, withUserLookup bool) error {
+func RunAs(uid, gid int) error {
 	if uid == os.Getuid() && gid == os.Getgid() {
 		return nil
 	}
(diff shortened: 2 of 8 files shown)


cmd/lifecycle/rebaser.go
```diff
@@ -63,7 +63,7 @@ func (r *rebaseCmd) Privileges() error {
 			return cmd.FailErr(err, "initialize docker client")
 		}
 	}
-	if err := priv.RunAs(r.uid, r.gid); err != nil {
+	if err := priv.RunAs(r.uid, r.gid, true); err != nil {
```
natalieparellano, line 66: I wasn't sure if rebaser needs a user lookup, so left it as it is for now.

priv/user_linux.go
```diff
@@ -100,11 +100,20 @@ func recursiveEnsureOwner(path string, uid, gid int) error {
...
 	}
-	user, err := user.LookupId(strconv.Itoa(uid))
+
+	var (
+		foundUser *user.User
+		err       error
+	)
+	if withUserLookup {
+		foundUser, err = user.LookupId(strconv.Itoa(uid))
+	}
```
natalieparellano, line 116: This code could be made simpler if moved below line 134, but I am not sure if there is value in doing the user check before dropping privileges.

natalieparellano:
I tested a modified lifecycle with these changes against a branch of pack that runs analyze, restore, and export in a minimal image where a user with the given ID has not been created. It appears to be working...

priv/user_linux.go
```diff
@@ -151,3 +150,20 @@ func setresuid(ruid, euid, suid int) error {
...
+	if err := os.Setenv("USER", user.Name); err != nil {
+		return err
+	}
+	if _, ok := os.LookupEnv("DOCKER_CONFIG"); ok {
+		return nil
+	}
+	// ggcr sets default docker config during init, fix for user
+	if err := os.Setenv("DOCKER_CONFIG", filepath.Join(user.HomeDir, ".docker")); err != nil {
+		return err
+	}
```
ekcasey, line 167: I think we should remove this code, `DOCKER_CONFIG` is only relevant to analyzer/exporter and this function only runs with `creator`, we wouldn't want analyzer/exporter to lookup docker config differently as part of creator vs. as a stand alone phase. We originally put this in to be extra cautious, but I think it was the wrong decision for following reasons:

1. This code only runs when the lifecycle is executes as root
2. `pack` doesn't use the docker config to pass creds to the lifecycle
3. other platforms that do use the docker config don't run these phases as root
4. other platforms that do use the docker config (`tekton`, `kpack`) explicitly set `HOME` already, in order to provide docker config generated from secrets, this logic would probably break them rather than help them if they did run as root

cmd/lifecycle/rebaser.go
ekcasey: It doesn't need a user lookup. Having a real user and preparing the env is only important when the phase executes buildpacks. `rebaser` is more like `analyzer/exporter` than like `creator`

priv/user_linux.go
ekcasey: `/etc/passwd` should be world readable so doing the user check after dropping privs shouldn't cause any problems

priv/user_linux.go
```diff
@@ -120,20 +118,16 @@ func RunAs(uid, gid int) error {
...
-		return err
-	}
-	if _, ok := os.LookupEnv("DOCKER_CONFIG"); ok {
+	if !withUserLookup {
 		return nil
 	}
-	// ggcr sets default docker config during init, fix for user
-	if err := os.Setenv("DOCKER_CONFIG", filepath.Join(user.HomeDir, ".docker")); err != nil {
+
+	foundUser, err := user.LookupId(strconv.Itoa(uid))
```
ekcasey, line 125: Can we move this into `setEnvironmentForUser`, export that function, and call it directly from `createCmd.Privileges`? That way we could take the bool out of the `RunAs` signature. I think that usage would read a little more obviously/nicely in the lifecycle commands. The lookup is a impl. detail, the env changes are the behavior.

natalieparellano: Sounds good, I will make the change

priv/user_linux.go
```diff
@@ -151,3 +135,17 @@ func setresuid(ruid, euid, suid int) error {
...
+
+func SetEnvironmentForUser(uid int) error {
+	user, err := user.LookupId(strconv.Itoa(uid))
+	if err != nil {
+		return err
+	}
+	if err := os.Setenv("HOME", user.HomeDir); err != nil {
+		return err
+	}
+	if err := os.Setenv("USER", user.Name); err != nil {
```
sclevine, line 147: Could just `return` this line directly.

PR #294, opened 2020-05-15
Build and push "lifecycle image" in GitHub actions

### LINKED ISSUE #285: Ship "lifecycle image" with new versions of lifecycle
When we ship new versions of the lifecycle, we should also ship a "lifecycle image" that contains the minimum configuration necessary to run the `analyze`, `restore`, and `export` phases of the lifecycle.

Requirements:
* It should be based on `gcr.io/distroless/static`
* It should contain the lifecycle binaries inside `/cnb/lifecycle`
* It should contain a `/tmp` directory
* Its working directory should be `/layers`

When platforms are working with this image, they'll need to do the following (I believe):
* Provide a `stack.toml` - the default is `/cnb/stack.toml`
* Set `CNB_USER_ID` and `CNB_GROUP_ID` in the environment - **note** currently, these ids must be valid in the image because of [this check](https://github.com/buildpacks/lifecycle/blob/4e6f6de75ef4a8c183ac5f4bfa5a0f645852de0c/priv/user_linux.go#L107-L110), we should determine if this can be removed

Context:
* This would be in support of https://github.com/buildpacks/pack/issues/528
* [Slack conversation](https://buildpacks.slack.com/archives/CSFE49VGR/p1587752732046400) with further context
* [Spike](https://github.com/buildpacks/pack/issues/528#issuecomment-618045406) where various requirements were determined

Other:
* We probably want to use GitHub actions for this

natalieparellano:
Fixes #285 

Some evidence that this works: https://github.com/buildpacks/lifecycle/issues/285#issuecomment-629411552


lifecycle-image/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+VOLUME /tmp/lifecycle
+
+
+# Stage 2
+FROM gcr.io/distroless/static
+
+WORKDIR /cnb/lifecycle
+
+COPY --from=download-and-extract-lifecycle /tmp/lifecycle .
```
jromero, line 25:
```suggestion
COPY --from=download-and-extract-lifecycle /tmp/lifecycle /cnb/lifecycle
```

Would this not work instead of using the `WORKDIR` instruction? [Reasons](https://stackoverflow.com/a/41638134) might not be strongly justified but thought it might be clearer anyway.

lifecycle-image/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+WORKDIR /tmp
+
+RUN apt-get update && apt-get install -y wget
+RUN wget $LIFECYCLE_URI && tar xzvf $LIFECYCLE_TGZ
+
+RUN mkdir -p /cnb/lifecycle && mkdir /layers
+
+
+VOLUME /tmp/lifecycle
```
jromero, line 17: Not sure I understand the purpose of the `VOLUME`. You should be able to just copy from any location.

Ref: https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

.github/workflows/lifecycle-image.yml
```diff
@@ -0,0 +1,24 @@
...
+      - published
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Get release version
+        id: get_release_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\/v/}
```
jromero, line 15: Minor: might be safer to use the tag name from the `release` object.

Something like `${{ github.event.release.tag_name }}` or `${{ github.event.tag_name }}`, don't know exactly how it's reflected.

Ref: https://developer.github.com/webhooks/event-payloads/#release

lifecycle-image/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Stage 1
```
jromero, line 1: Wonder if it's worth scoping this file to the pipeline by putting it under `.github/workflows/` similar to [pack](https://github.com/buildpacks/pack/tree/master/.github/workflows/delivery).

.github/workflows/lifecycle-image.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Publish to Registry
+        uses: elgohr/Publish-Docker-Github-Action@master
+        with:
+          name: buildpacksio/lifecycle-image
```
jromero, line 19:
```suggestion
          name: buildpacksio/lifecycle:${{ steps.get_release_version.outputs.VERSION }}
```

1. I don't think there's a need for the name to include "image"
2. We'd probably want versioned tags along with `latest`.

matthewmcnew:
Should this include metadata about the included lifecycle descriptor?

.github/workflows/lifecycle-image.yml
ekcasey: +1

lifecycle-image/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+FROM ubuntu:bionic AS download-and-extract-lifecycle
+
+ARG LIFECYCLE_VERSION
+ARG LIFECYCLE_GITHUB_ORG
```
ekcasey, line 5: Can we give this a default? Or better yet just hardcode `buildpacks` in the `LIFECYCLE_URI` default. I imagine folks will typically override the entire `LIFECYCLE_URI` if they want to change the artifact location.

lifecycle-image/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ARG LIFECYCLE_TGZ=lifecycle-v${LIFECYCLE_VERSION}+linux.x86-64.tgz
+ARG LIFECYCLE_URI="https://github.com/${LIFECYCLE_GITHUB_ORG}/lifecycle/releases/download/v${LIFECYCLE_VERSION}/${LIFECYCLE_TGZ}"
+
+WORKDIR /tmp
+
+RUN apt-get update && apt-get install -y wget
+RUN wget $LIFECYCLE_URI && tar xzvf $LIFECYCLE_TGZ
```
ekcasey, line 12:
```suggestion
RUN apt-get update && apt-get install -y wget -O lifecycle.tgz
RUN wget $LIFECYCLE_URI && tar xzvf lifecycle.tgz
```

If we name the output document explicitly we won't have to correctly derive the name of the file at `LIFECYCLE_URI`. This will make it easier use a build arg to change the lifecycle uri correctly.

ekcasey:
> Should this include metadata about the included lifecycle descriptor?
That is a good idea, we should add a label. Unfortunately `io.buildpacks.lifecycle.metadata` is used for layers already (I would call it `io.buildpacks.layers.metadata` if we had a do-over).

we could:
1. call it `io.buildpacks.lifecycle.metadata`, accept the potential for confusion and assume we will change the other one eventually
2. `io.buildpacks.builder.metadata` already contains a `lifecycle` key that we use for this data on a builder. We could reuse that label, and include only a subset (if you stretch your mind this is a builder without buildpacks?)
3. Add smaller targeted labels `io.buildpacks.lifecycle.version`, `io.buildpacks.platform.api` and `io.buildpacks.buildpack.api` and add those to builders as well. In this case, I would like to make the api label values object to future proof (in case we need to indicate support for multiple apis simultaneously)

ekcasey:
I tried using these images with the tekton template, export failed because there was no `stack.toml` file:
```
no stack metadata found at path '/cnb/stack.toml', stack metadata will not be exported
ERROR: failed to parse arguments: -run-image is required when there is no stack metadata available
```
We won't be able bake a `stack.toml` into this image b/c it can be used with different stacks. However, I wonder if we should point `CNB_STACK_PATH` to a path that is traditionally on a volume in these lifecycle images, like `/layers/stack.toml`, so that platforms have an easy place to copy in this file.

jromero:
> > Should this include metadata about the included lifecycle descriptor?
> 
> That is a good idea, we should add a label. Unfortunately `io.buildpacks.lifecycle.metadata` is used for layers already (I would call it `io.buildpacks.layers.metadata` if we had a do-over).
> 
> we could:
> 
> 1. call it `io.buildpacks.lifecycle.metadata`, accept the potential for confusion and assume we will change the other one eventually
> 2. `io.buildpacks.builder.metadata` already contains a `lifecycle` key that we use for this data on a builder. We could reuse that label, and include only a subset (if you stretch your mind this is a builder without buildpacks?)
> 3. Add smaller targeted labels (e.g. `io.buildpacks.lifecycle.version`, `io.buildpacks.platform.api` and `io.buildpacks.buildpack.api`) and add those to builders as well. In this case, I would like to make the api label values objects (to future proof in case we need to indicate support for multiple apis simultaneously)

I vote for **3**. There really isn't even a reason to namespace these labels but I can see why we'd like to keep a convention.

matthewmcnew:
It seems #2 would open us up for using other Builders as well as a replacement for this image?

lifecycle-image/Dockerfile
natalieparellano: Yeah, this is an outgrowth of using my fork for testing... I will make the change.

natalieparellano:
> I tried using these images with the tekton template, export failed because there was no `stack.toml` file:
> 
> ```
> no stack metadata found at path '/cnb/stack.toml', stack metadata will not be exported
> ERROR: failed to parse arguments: -run-image is required when there is no stack metadata available
> ```
> 
> We won't be able bake a `stack.toml` into this image b/c it can be used with different stacks. However, I wonder if we should point `CNB_STACK_PATH` to a path that is traditionally on a volume in these lifecycle images, like `/layers/stack.toml`, so that platforms have an easy place to copy in this file.

In pack we are (draft PR in review) currently bind mounting the stack.toml to its current location in /cnb/stack.toml. But I believe there were performance concerns about bind mounting directly from the host.

.github/workflows/lifecycle-image.yml
natalieparellano: I'll remove `-image`. We're already getting the versioned tag + latest here: https://github.com/buildpacks/lifecycle/blob/5fa1ed064fc4f032dc32c1586e03939a7342bdfc/.github/workflows/lifecycle-image.yml#L24

.github/workflows/lifecycle-image.yml
natalieparellano: Giving it a try... we still need to trim off the "v" in the release tag.

natalieparellano:
The latest changes should incorporate most of the feedback except labels and a Makefile target for dev.

Will update with labels once we have consensus on the naming. 

Do you think it would be possible to add the Makefile target for dev as a separate unit of work? Just trying to be mindful of our timing as `pack` will enter CCB tomorrow and we're holding the "lifecycle image+trust-builder" stuff in `pack` on getting this image published. (This is probably obvious but worth calling out that we'll need to do an out-of-band docker build and push for lifecycle release 0.7.5 as that's already out.)

.github/workflows/lifecycle-image/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+FROM ubuntu:bionic AS download-and-extract-lifecycle
+
+ARG LIFECYCLE_URI
+
+WORKDIR /tmp
+
+RUN apt-get update && apt-get install -y wget
+RUN wget $LIFECYCLE_URI -O lifecycle.tgz && tar xzvf lifecycle.tgz
+
+RUN mkdir -p /cnb/lifecycle && mkdir /layers
```
jromero, line 11: I believe this should be part of the second stage.

natalieparellano: We can't run these commands in distroless/static

natalieparellano:
Re: labels - on chatting with @ekcasey let's go with option #2 (since option #1 has the potential for confusion and option #3 invites other changes).

.github/workflows/lifecycle-image/Dockerfile
natalieparellano: I tried chmod'ing `/layers` and then copying it over but it's still only writeable as root...

ekcasey: lets just leave `/layers` how it is for now, it likely won't be a problem

.github/workflows/lifecycle-image.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: |
+          - echo ::set-output name=LABEL_METADATA::`"{\"lifecycle\":{\"version\":\"${{ steps.get_metadata.outputs.LIFECYCLE_VERSION }}\",\"api\":{\"buildpack\":\"${{ steps.get_metadata.outputs.BUILDPACK_API }}\",\"platform\":\"${{ steps.get_metadata.outputs.PLATFORM_API }}\"}}}"`
+      - name: Publish to Registry
+        uses: elgohr/Publish-Docker-Github-Action@master
+        with:
+          name: buildpacksio/lifecycle
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          dockerfile: .github/workflows/lifecycle-image/Dockerfile
+          buildargs: LIFECYCLE_URI=https://github.com/buildpacks/lifecycle/releases/download/v${{ steps.get_metadata.outputs.LIFECYCLE_VERSION }}/lifecycle-v${{ steps.get_metadata.outputs.LIFECYCLE_VERSION }}+linux.x86-64.tgz, LABEL_METADATA=${{ steps.get_metadata.outputs.LABEL_METADATA }}
```
natalieparellano, line 29: It's annoying to have so many args, but it appears we need to format the label directly in the Dockerfile to avoid (somehow) unescaping the surrounding double quotes and terminating the label at the first comma ¯\\_(ツ)_/¯

.github/workflows/lifecycle-image.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Get metadata
+        id: get_metadata
+        run: |
+          - echo ::set-output name=LIFECYCLE_VERSION::`echo ${{ github.event.release.tag_name }} | cut -d "v" -f2`
+          - echo ::set-output name=BUILDPACK_API::`cat Makefile | grep BUILDPACK_API?= | cut -d "=" -f2`
```
natalieparellano, line 17: Potentially instead of reading the Makefile we could have a separate file(s) that just contains the metadata values.

.github/workflows/lifecycle-image.yml
```diff
@@ -0,0 +1,30 @@
...
+          - echo ::set-output name=PLATFORM_API::`cat Makefile | grep PLATFORM_API?= | cut -d "=" -f2`
```
ekcasey, line 18: We should really make this data easier to get, but there aren't many better options right now. The alternative would be to extract the descriptor from the archive, but I am okay with this so long as we recognize it's a bit brittle and should be improved (probably with the help of better formatted lifecycle releases) when we can get around to it.

jromero:
Image looks good:

image

Issue #295, opened 2020-05-18
Confusing error message when creator is given an invalid image reference


scottfrederick:
### Summary

When the `creator` is given an image name that it can't parse, a somewhat confusing error message is returned to the builder client. 

##### Steps

This issue can't be re-created with `pack` because `pack` doesn't currently use the `creator` and it parses the image reference itself before passing to the lifecycle anyway. 

It can be re-created with a trivial Spring Boot application using either the Maven plugin with [`mvn spring-boot:build-image`](https://docs.spring.io/spring-boot/docs/2.3.0.BUILD-SNAPSHOT/maven-plugin/reference/html/#build-image) or the Gradle plugin with [`gradle bootBuildImage`](https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/gradle-plugin/reference/html/#build-image). Providing an image name that contains upper-case characters will cause an image reference parsing error. 

* Create a project using [Spring Initializr](https://start.spring.io/), selecting Maven, Spring Boot 2.3.0, and the name `DemoProject`
* Build an image from the project with `mvn spring-boot:build-image -Dspring-boot.build-image.imageName=docker.io/library/DemoProject:1.0` 

##### Current behavior
The example above results in an error like this when the Spring Boot Maven plugin invokes `creator`: 

```
[INFO]  > Running creator
[INFO]     [creator]     ERROR: failed to all tags must have the same registry as the exported image: could not parse reference: docker.io/library/DemoProject:1.0
```

The `failed to all tags must...` part has strange phrasing, and the `all tags must have the same registry as the exported image` part is somewhat confusing in the context of trying to create an image. Beyond that, it is not obvious why the reference could not be parsed. 

##### Expected behavior

An error message that more concisely indicates that the image reference can't be parsed would be an improvement. An indication of what part of the image reference isn't acceptable would be better.

ekcasey:
@scottfrederick I agree we should fix that error message. After some experimentation I think `ggcr` is failing to parse the image reference because it doesn't like the capital letters in the tag. We could lowercase the user inputs before using them but, fwiw, `docker tag` etc. also chokes on capital letters (albeit with a better error message).

PR #296, opened 2020-05-19
Better error message when an image reference fails to parse

ekcasey:
* Fixes: #295

Doesn't validate that all destination tags have a matching registry when
exporting to the daemon (this is unnecessary in the daemon case)

### DIFF
--- a/cmd/lifecycle/creator.go
+++ b/cmd/lifecycle/creator.go
@@ -77,8 +77,8 @@ func (c *createCmd) Args(nargs int, args []string) error {
 		c.previousImage = c.imageName
 	}
 
-	if err := image.EnsureSingleRegistry(append(c.additionalTags, c.imageName)...); err != nil {
-		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "all tags must have the same registry as the exported image")
+	if err := image.ValidateDestinationTags(c.useDaemon, append(c.additionalTags, c.imageName)...); err != nil {
+		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "failed to validate image tag(s)")
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


image/image_test.go
```diff
@@ -14,19 +14,34 @@ func TestImage(t *testing.T) {
...
 		when("a single registry is provided", func() {
 			it("does not return an error", func() {
-				err := image.EnsureSingleRegistry("gcr.io/some/repo", "gcr.io/other-repo:latest", "gcr.io/final-repo")
+				err := image.ValidateDestinationTags(false, "gcr.io/some/repo", "gcr.io/other-repo:latest", "gcr.io/final-repo")
 				h.AssertNil(t, err)
 			})
 		})
+
+		when("the tag reference is invalid", func() {
+			it("error", func() {
```
natalieparellano, line 41:
```suggestion
			it("errors", func() {
```

cmd/lifecycle/creator.go
```diff
@@ -77,8 +77,8 @@ func (c *createCmd) Args(nargs int, args []string) error {
 		c.previousImage = c.imageName
 	}
 
-	if err := image.EnsureSingleRegistry(append(c.additionalTags, c.imageName)...); err != nil {
-		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "all tags must have the same registry as the exported image")
+	if err := image.ValidateDestinationTags(c.useDaemon, append(c.additionalTags, c.imageName)...); err != nil {
+		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "failed to validate image tag(s)")
```
natalieparellano, line 81: Looking at the issue, is the strange phrasing mentioned: `ERROR: failed to all tags must...` going to become `ERROR: failed to failed to validate image tag(s)`? Should the message just start with `validate...`?

ekcasey: good catch

natalieparellano:
Looks good to me!

PR #300, opened 2020-05-21
creator prints colored phase markers

### LINKED ISSUE #299: Single-container eye-catcher are pretty ugly
The eye catchers in the single-container execution style are a pretty big step down from what it used to look like

image

This should have some color and otherwise look a little bit nicer than it currently does.

ekcasey:
* Resolves #299
* style changed to match pack phases

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -14,16 +14,32 @@ const (
 	warnLevelText  = "Warning: "
 )
 
+func init() {
+	// TODO: uncomment when buildpacks/pack#493 (lifecycle containers with a tty) is implemented
+	//color.Disable(!terminal.IsTerminal(int(os.Stdout.Fd())))
+}
+
 // Default logger
 var (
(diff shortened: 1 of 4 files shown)


cmd/logs.go
```diff
@@ -14,17 +14,33 @@ const (
 	warnLevelText  = "Warning: "
 )
 
+func init() {
+	// TODO: uncomment when pack runs lifecycle with a tty
```
jromero, line 18: Is there an issue associated with this work? It would be nice to have the issue link here.

nebhale: https://github.com/buildpacks/pack/issues/493 which appears no closer than ever to being worked.  And as predicted, the longer we wait to implement it, the more people are just forcing TTY-driven behaviors to enabled, causing a sub-par experience for systems that aren't TTY (like `kpack` and log warehousing).

PR #301, opened 2020-05-26
Adds acceptance tests for launcher

ekcasey:


acceptance/launcher_test.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func removeLaunchImage(t *testing.T) {
+	cmd := exec.Command("docker", "rmi", launchImage)
+	if output, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("failed to run %v\n OUTPUT: %s\n ERROR: %s\n", cmd.Args, string(output), err)
+	}
+}
+
+func buildLauncher(t *testing.T) {
```
natalieparellano, line 197: Is there a reason why we're not making use of `buildBinaries` from acceptance_test.go?

acceptance/launcher_test.go
```diff
@@ -0,0 +1,215 @@
...
+			expected := "Executing other-process process-type"
+			if !strings.Contains(string(output), expected) {
+				t.Fatalf("failed to execute other-process:\n\t got: %s\n\t want: %s", output, expected)
+			}
+		})
+	})
+
+	when("CMD provided", func() {
+		it("sources profiles and launches command with shell", func() {
+			cmd := exec.Command("docker", "run", "--rm", launchImage, "echo my-cmd")
```
natalieparellano, line 74: I wonder if we can change `my-cmd` to something else - this name makes me think that `my-cmd` is a fake command that we have a fixture for...

acceptance/launcher_test.go
```diff
@@ -0,0 +1,215 @@
...
+				"-eCNB_APP_DIR=/workspace",
+				"-eSOME_USER_VAR=some-user-val",
+				"-eOTHER_VAR=other-user-val",
+				launchImage,
+				"echo $SOME_USER_VAR $CNB_APP_DIR $OTHER_VAR")
+			output, err := cmd.CombinedOutput()
+			if err != nil {
+				t.Fatalf("failed to run %v\n OUTPUT: %s\n ERROR: %s\n", cmd.Args, output, err)
+			}
+			// bp appends other-bp-val with delimeter '**'
```
natalieparellano, line 108: Perhaps it's worth adding a separate test for this behavior? Or would that belong somewhere else?

acceptance/testdata/launcher/container/layers/config/metadata.toml
```diff
@@ -0,0 +1,13 @@
+[[buildpacks]]
+  id = "some/buildpack"
+  version = "0.0.1"
+
+[[processes]]
+  type = "web"
+  command = "echo \"Executing web process-type\""
+  direct = false
```
natalieparellano, line 8: It doesn't seem that we are testing any `direct = true` cases, is that true? Would that be something worth adding?

acceptance/launcher_test.go
```diff
@@ -0,0 +1,215 @@
...
+				t.Fatalf("failed to run %v\n OUTPUT: %s\n ERROR: %s\n", cmd.Args, output, err)
+			}
+			expected := "Executing web process-type"
+			if !strings.Contains(string(output), expected) {
+				t.Fatalf("failed to execute web:\n\t got: %s\n\t want: %s", output, expected)
+			}
+		})
+	})
+
+	when("no args and CNB_PROCESS_TYPE", func() {
```
natalieparellano, line 44: It doesn't look like we have any tests around setting `CNB_LAYERS_DIR` and `CNB_APP_DIR`. I suppose that would require either a separate container fixture, or some manipulation with `docker` before running the tests.

acceptance/launcher_test.go
```diff
@@ -0,0 +1,215 @@
...
+				t.Fatalf("failed to run %v\n OUTPUT: %s\n ERROR: %s\n", cmd.Args, output, err)
+			}
+			expected := "Executing other-process process-type"
+			if !strings.Contains(string(output), expected) {
+				t.Fatalf("failed to execute other-process:\n\t got: %s\n\t want: %s", output, expected)
+			}
+		})
+	})
+
+	when("process provided in CMD", func() {
```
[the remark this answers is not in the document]
natalieparellano: If I'm following correctly `other-process` is actually a process type whereas the following `when` is a command. Maybe the phrasing could be clearer.

acceptance/launcher_test.go
ekcasey: We only need the `launcher` for this test, and building just the launcher is considerably faster (it doesn't need to be compiled in a container).

acceptance/launcher_test.go
ekcasey: I was trying to strike a balance between small targeted tests and speed (not starting too many containers). Happy to break it out though if it makes this test confusing.

acceptance/testdata/launcher/container/layers/config/metadata.toml
ekcasey: We are executing `direct = true` process in some of the tests where we override the entire command, but yes I'll change one of these buildpack-provided processes.

acceptance/launcher_test.go
ekcasey: Yeah, it would require either a different fixture, or multiple dirs we can switch between in the same fixture. I'll add a test.

acceptance/launcher_test.go
```diff
@@ -0,0 +1,215 @@
...
+			}
+			expected := "Executing web process-type"
+			if !strings.Contains(string(output), expected) {
+				t.Fatalf("failed to execute web:\n\t got: %s\n\t want: %s", output, expected)
+			}
+		})
+	})
+
+	when("no args and CNB_PROCESS_TYPE", func() {
+		it("launches given default process", func() {
```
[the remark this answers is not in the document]
ekcasey: In the past we have referred to `CNB_PROCESS_TYPE` as changing the "default process", especially when the env var is being set by `exporter`. I can make this wording more clear.

acceptance/launcher_test.go
```diff
@@ -0,0 +1,215 @@
...
+			if !strings.Contains(string(output), expected) {
+				t.Fatalf("failed to execute provided CMD:\n\t got: %s\n\t want: %s", output, expected)
+			}
+
+			if strings.Contains(string(output), "CNB_APP_DIR") {
+				t.Fatalf("env contained white listed env far CNB_APP_DIR:\n\t got: %s\n", output)
+			}
+		})
+
+		it("adds buildpack bin dirs to the path before looking up command", func() {
```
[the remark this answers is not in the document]
ekcasey: The `launcher` would fail while looking up and executing `bp-executable` if it hadn't been added to the path

acceptance/launcher_test.go
natalieparellano: That makes sense to me. I think we can leave it where it is for now...

acceptance/launcher_test.go
```diff
@@ -27,36 +27,38 @@ func TestLauncher(t *testing.T) {
...
-			cmd := exec.Command("docker", "run", "--rm", launchImage)
-			output, err := cmd.CombinedOutput()
-			if err != nil {
-				t.Fatalf("failed to run %v\n OUTPUT: %s\n ERROR: %s\n", cmd.Args, output, err)
-			}
-			expected := "Executing web process-type"
-			if !strings.Contains(string(output), expected) {
-				t.Fatalf("failed to execute web:\n\t got: %s\n\t want: %s", output, expected)
-			}
+	when("there is no CMD provided", func() {
```
natalieparellano, line 30: I'm wondering if we can make this even clearer. If I'm understanding correctly, `CMD` is the argument(s) provided to the launcher. For a minute I was wondering if it was the `Cmd` from the OCI container config...

ekcasey: I did mean it in the OCI sense. When launcher is the `entrypoint`, the arguments provided to launcher and the container `cmd` are one and the same. But I am happy to reword it from the perspective of `launcher` rather than the user who runs the container, if it's clearer.

natalieparellano: Ahhh you are exactly right... I think I confused myself.

PR #305, opened 2020-05-29
Prepare for 0.8.0 release.

ekcasey:
* Adds VERSION file
* bump upload-artifact github action to v2
* Uses wildcard for artifact upload, now that it's supported

Geting the next version from the VERSION file should make the version number easier to find and increment.
This should makes local builds easier, users no longer need to set LIFECYCLE_VERSION explicitly for every build.


.github/workflows/build.yml
```diff
@@ -36,7 +34,7 @@ jobs:
         run: |
           make build
           make package
-      - uses: actions/upload-artifact@v1
+      - uses: actions/upload-artifact@v2
         with:
           name: lifecycle-linux-x86-64
-          path: out/lifecycle-v${{ env.LIFECYCLE_VERSION }}+linux.x86-64.tgz
+          path: out/lifecycle-v*+linux.x86-64.tgz
```
natalieparellano, line 40: Nice

PR #308, opened 2020-06-01
Do not delete cache image when exporting cache if it is the same as before

lukasberger:
With the reproducibility improvements, it often happens that the old and
new cache images are the same. The original image should not be
deleted in this case.

This is an issue with remote images, where the image is deleted by
digest and both the old and new images share the same digest.

Signed-off-by: Lukas Berger <bergerl@google.com>

### DIFF
--- a/cache/image_cache.go
+++ b/cache/image_cache.go
@@ -119,7 +119,7 @@ func (c *ImageCache) DeleteOrigImage() error {
 	if err != nil {
 		return errors.Wrap(err, "getting identifier for new image")
 	}
-	if origIdentifier == newIdentifier {
+	if origIdentifier.String() == newIdentifier.String() {
 		return nil
 	}
 	return c.origImage.Delete()
--- a/cache/image_cache.go
+++ b/cache/image_cache.go
@@ -102,10 +102,25 @@ func (c *ImageCache) Commit() error {
 
 	if c.origImage.Found() {
 		// Deleting the original image is for cleanup only and should not fail the commit.
-		if err := c.origImage.Delete(); err != nil {
+		if err := c.DeleteOrigImage(); err != nil {
 			fmt.Printf("Unable to delete previous cache image: %v", err)
 		}
 	}
 	c.origImage = c.newImage
 	return nil
+}
+
+func (c *ImageCache) DeleteOrigImage() error {
+	origIdentifier, err := c.origImage.Identifier()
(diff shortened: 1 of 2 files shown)


cache/image_cache.go
```diff
@@ -102,10 +102,25 @@ func (c *ImageCache) Commit() error {
...
+func (c *ImageCache) DeleteOrigImage() error {
+	origIdentifier, err := c.origImage.Identifier()
+	if err != nil {
+		return errors.Wrap(err, "getting identifier for original image")
+	}
+	newIdentifier, err := c.newImage.Identifier()
+	if err != nil {
+		return errors.Wrap(err, "getting identifier for new image")
+	}
+	if origIdentifier == newIdentifier {
```
ekcasey, line 122: I think it would be safer to compare the string values here in case the identifier is a pointer.

lukasberger: Is the worry here that the signature of `Identifier()` might change to return a pointer and cause this line to always evaluate to false and the original image to be deleted? The second test added in the PR should catch that. Converting to a string has drawbacks like discarding the identifier type. However, if you feel strongly this should be a string comparison, I'm happy to change it.

ekcasey: b/c `Identifier()` returns an interface, a given implementation of the `imgutil.Image` interface can already validly return a pointer. For example if I changed your test setup from
```
fakes.NewImage("fake-image", "", local.IDIdentifier{ImageID: "fakeOrigImage"})
```
to
```
fakes.NewImage("fake-image", "", &local.IDIdentifier{ImageID: "fakeOrigImage"})
```
everything compiles but the tests fail.

I am not worried about losing the `Identifier` type information in the comparison here. We already implicitly assume that the string representation carries all necessary information when add the string value to a label or the `analyzed.toml` file, for example, and assume that string still uniquely identifies an image.

lukasberger: That's a very good point, changed to string comparison. I also had to update some tests because they returned nil for the identifiers which cause a panic. We could be more defensive and check for nil the `DeleteOrigImage` function, but if feels like a return value `(nil, nil)` from `Identifier()` should be invalid and was an artifact of the test setup. Thoughts?

ekcasey: > feels like a return value (nil, nil) from Identifier() should be invalid and was an artifact of the test setup

I agree :)

PR #310, opened 2020-06-09
Add ability to export/launch Windows-based images

ameyer-pivotal:
In general, changes relate to the following:

- Replacing `tar.NewWriter` calls with a factory call which can get a layer writer appropriate for the target image OS
- Fixing OS-related path and delimiter issues
- Implements full launcher support for Windows
- Makefile/CI-related changes to support Windows testing and artifact upload

Signed-off-by: Andrew Meyer <meyeran@vmware.com>
Signed-off-by: Anthony Emengo <aemengo@vmware.com>


ameyer-pivotal:
@aemengo 
@micahyoung

cache/caching_image_test.go
```diff
@@ -69,7 +70,7 @@ func testCachingImage(t *testing.T, when spec.G, it spec.S) {
 				from, err := os.Open(layerPath)
 				h.AssertNil(t, err)
 				defer from.Close()
-				to, err := os.Create(filepath.Join(tmpDir, "committed", layerSHA+".tar"))
+				to, err := os.Create(filepath.Join(tmpDir, "committed", strings.TrimPrefix(layerSHA, "sha256:")+".tar"))
```
ameyer-pivotal, line 73: `sha256:` in a file name causes problems on windows

cache/volume_cache.go
```diff
@@ -180,6 +183,10 @@ func (c *VolumeCache) Commit() error {
 	return nil
 }
 
+func diffIDPath(basePath, diffID string) string {
+	return filepath.Join(basePath, strings.TrimPrefix(diffID, "sha256:")+".tar")
```
ameyer-pivotal, line 187: `sha256:` in a file name causes problems on windows

Makefile
```diff
@@ -14,32 +18,35 @@ BUILDPACK_API?=0.2
...
+BUILD_CTR?=lifecycle-ctr
+DOCKER_CMD?=make test
 
 all: test build package
 
 build: build-linux build-windows
 
+build-linux: build-linux-lifecycle build-linux-symlinks build-linux-launcher
+
+ifeq ($(OS),Windows_NT)
```
micahyoung, line 35: Nearly all the make targets can run on Windows/Linux/MacOS. `build-windows` is the only one required for CI that has to switch currently

Makefile
```diff
@@ -151,23 +166,36 @@ package-linux: GOOS:=linux
 package-linux: GOOS_DIR:=$(BUILD_DIR)/$(GOOS)
 package-linux: ARCHIVE_NAME=lifecycle-v$(LIFECYCLE_VERSION)+$(GOOS).x86-64
 package-linux:
-	@echo "> Writing descriptor file for $(GOOS)..."
-	mkdir -p $(GOOS_DIR)
-	echo "$${LIFECYCLE_DESCRIPTOR}" > $(GOOS_DIR)/lifecycle.toml
-
 	@echo "> Packaging lifecycle for $(GOOS)..."
-	tar czf $(BUILD_DIR)/$(ARCHIVE_NAME).tgz -C $(GOOS_DIR) lifecycle.toml lifecycle
+	$(GOCMD) run tools/packager/main.go -os $(GOOS) -launcherExePath $(GOOS_DIR)/lifecycle/launcher -lifecycleExePath $(GOOS_DIR)/lifecycle/lifecycle -lifecycleVersion $(LIFECYCLE_VERSION) -platformAPI $(PLATFORM_API) -buildpackAPI $(BUILDPACK_API) -outputPackagePath $(BUILD_DIR)/$(ARCHIVE_NAME).tgz
```
micahyoung, line 172: Using the `packager` allows consistent tarball artifacts regardless of OS and identical target syntax. Issues were:
* `tar.exe` is not equivalent to POSIX tar (i.e. can't do symlinks)
* echoing `LIFECYCLE_DESCRIPTOR` to a file was messy on Windows and felt better outside `Makefile`.

micahyoung: I'm not attached to `go run` over `go install` like our other tools. Happy to change if preferred.

Makefile
```diff
@@ -151,23 +166,36 @@ package-linux: GOOS:=linux
...
+	docker build -f tools/Dockerfile.windows --tag $(SOURCE_COMPILATION_IMAGE) --build-arg image_tag=$(WINDOWS_COMPILATION_IMAGE) --cache-from=$(SOURCE_COMPILATION_IMAGE) --isolation=process --quiet .
+
+docker-run-windows: docker-build-source-image-windows
+docker-run-windows:
+	@echo "> Running '$(DOCKER_CMD)' in docker windows..."
+	@docker volume rm -f lifecycle-out
+	docker run -v lifecycle-out:c:/lifecycle/out -e LIFECYCLE_VERSION -e PLATFORM_API -e BUILDPACK_API -v gopathcache:c:/gopath -v '\\.\pipe\docker_engine:\\.\pipe\docker_engine' --isolation=process --rm $(SOURCE_COMPILATION_IMAGE) $(DOCKER_CMD)
+	docker run -v lifecycle-out:c:/lifecycle/out --rm $(SOURCE_COMPILATION_IMAGE) tar -cf- out | tar -xf-
+	@docker volume rm -f lifecycle-out
+
```
micahyoung, line 192: This target is meant to be an easy and consistent wrapper for any Windows command (i.e. `make unit|acceptance|test`, `make build-windows package-windows`, `go run ...`)

Our hope is to make building, running, and troubleshooting on Windows extremely consistent for any CNB contributor and CI by ensuring everyone's using an identical Dockerfile-based environment.

micahyoung: An alternative implementation we tried was `docker build` w/o source, `docker create`, `docker cp`, `docker start` but symlinks were not copied over properly (ended up as empty files)

Makefile
```diff
@@ -151,23 +168,36 @@ package-linux: GOOS:=linux
...
+docker-build-source-image-linux:
+	docker build --tag $(SOURCE_COMPILATION_IMAGE) -f tools/Dockerfile.linux --build-arg image_tag=$(LINUX_COMPILATION_IMAGE) --cache-from=$(SOURCE_COMPILATION_IMAGE) --quiet .
+
+docker-run-linux: docker-build-source-image-linux
+docker-run-linux:
+	@echo "> Running '$(DOCKER_CMD)' in docker linux..."
+	@docker volume rm -f lifecycle-out
+	docker run -v lifecycle-out:/lifecycle -e LIFECYCLE_VERSION -v gopathcache:/go -v /var/run/docker.sock:/var/run/docker.sock --rm $(SOURCE_COMPILATION_IMAGE) $(DOCKER_CMD)
+	docker run -v lifecycle-out:/lifecycle --rm $(SOURCE_COMPILATION_IMAGE) tar -cf- out | tar -xf-
+	@docker volume rm -f lifecycle-out
```
micahyoung, line 202: This is not currently used but is functionally equivalent to `docker-run-windows`. It could be incorporated into in `build-linux-lifecycle` to avoid the uncached `apk add` call and would reduce setup setups in the GHA  `build.yml`

.github/workflows/build.yml
```diff
@@ -32,9 +32,30 @@ jobs:
...
+          make docker-run-windows DOCKER_CMD="make test"
+        shell: cmd
+      - name: Build
+        run: |
+          make docker-run-windows DOCKER_CMD="make build-windows package-windows"
+        shell: cmd
+      - uses: actions/upload-artifact@v2
+        with:
+          name: lifecycle-windows-x86-64
+          path: out/lifecycle-v*+windows.x86-64.tgz
```
micahyoung, line 61: The Windows jobs ends up simpler than Linux since dependencies are in the `tools/Dockerfile.windows` and it's run through the `make` target wrapper.

The tradeoff is that Windows `build` is slower than Linux (~4 mins vs ~7 mins resp) but I feel the consistency/simplicity is worth it. More thoughts on the `Makefile#docker-run-windows` target below

Makefile
```diff
@@ -8,38 +15,40 @@ LDFLAGS+=-X 'github.com/buildpacks/lifecycle/cmd.SCMCommit=$(SCM_COMMIT)'
 LDFLAGS+=-X 'github.com/buildpacks/lifecycle/cmd.PlatformAPI=$(PLATFORM_API)'
 GOBUILD=go build $(GOFLAGS) -ldflags "$(LDFLAGS)"
 GOTEST=$(GOCMD) test $(GOFLAGS)
-LIFECYCLE_VERSION?=$(shell cat VERSION)
 PLATFORM_API?=0.3
 BUILDPACK_API?=0.2
 SCM_REPO?=github.com/buildpacks/lifecycle
 PARSED_COMMIT:=$(shell git rev-parse --short HEAD)
 SCM_COMMIT?=$(PARSED_COMMIT)
+PWD?=${CURDIR}
```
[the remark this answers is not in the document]
ameyer-pivotal: Will do

Makefile
```diff
@@ -151,23 +168,36 @@ package-linux: GOOS:=linux
...
 package-linux:
-	@echo "> Writing descriptor file for $(GOOS)..."
-	mkdir -p $(GOOS_DIR)
-	echo "$${LIFECYCLE_DESCRIPTOR}" > $(GOOS_DIR)/lifecycle.toml
-
 	@echo "> Packaging lifecycle for $(GOOS)..."
-	tar czf $(BUILD_DIR)/$(ARCHIVE_NAME).tgz -C $(GOOS_DIR) lifecycle.toml lifecycle
+	$(GOCMD) run tools/packager/main.go -os $(GOOS) -launcherExePath $(GOOS_DIR)/lifecycle/launcher -lifecycleExePath $(GOOS_DIR)/lifecycle/lifecycle -lifecycleVersion $(LIFECYCLE_VERSION) -platformAPI $(PLATFORM_API) -buildpackAPI $(BUILDPACK_API) -outputPackagePath $(BUILD_DIR)/$(ARCHIVE_NAME).tgz
 
 package-windows: export LIFECYCLE_DESCRIPTOR:=$(LIFECYCLE_DESCRIPTOR)
```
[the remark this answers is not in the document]
ameyer-pivotal: Could you elaborate?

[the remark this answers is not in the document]
ameyer-pivotal: I see... Looks like the task is no longer using it. I'll address this.

archive/tar.go
```diff
@@ -11,7 +11,27 @@ import (
...
 )
 
-func WriteFilesToTar(dest string, uid, gid int, files ...string) (string, map[string]struct{}, error) {
+type TarWriter interface {
+	WriteHeader(hdr *tar.Header) error
+	Write(b []byte) (int, error)
+	Close() error
+}
+
+type WriterFactory interface {
```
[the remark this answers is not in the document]
ameyer-pivotal: IIRC the abstraction is needed because the tar-related helpers that receive this factory sometimes need an image-aware factory and other times need the `defaultTarWriterFactory` (when writing generic tars that aren't going to end up in image layers).

As for further refactors, unless you feel strongly, I think it'd be best to keep the factory simple at first, just to serve the primary goal of this PR.

As for having it do more, I'd prefer to keep this factory focused and simple for now. It may be a good refactor, but the PR is more focused on getting functionality

archive/tar_test.go
```diff
@@ -45,14 +46,24 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
 			{"root/readonly/readonlysub", os.ModeDir + 0500},
 		}
 
+		// Golang for Windows only implements owner permissions
+		if runtime.GOOS == "windows" {
+			pathModes = []archive.PathMode{
```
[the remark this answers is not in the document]
ameyer-pivotal: While that is true, that logic is specific to the Windows layer writer (and is [tested in imgutil](https://github.com/buildpacks/imgutil/blob/master/layer/windows_writer_test.go)).

These tests use a standard tar writer, and these tests are focused on permissions handling for different OSes.

cache/volume_cache.go
ameyer-pivotal: We considered that too. We weren't sure if it was worth the trouble/complexity for a one-time issue. Thoughts?

cmd/defaults_windows.go
```diff
@@ -0,0 +1,17 @@
+package cmd
+
+const (
+	DefaultLayersDir           = `c:\layers`
```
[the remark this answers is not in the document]
ameyer-pivotal: Shouldn't we rely on acceptance tests (right now, they're in `pack`'s repo) to catch such bugs?

archive/tar_test.go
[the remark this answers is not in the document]
ekcasey: I am asking, is this `Untar` logic wrong for windows? The cached layers should be formatted like image layers and we would need to strip out the special dirs before writing to the filesystem.

I am pretty sure that in the context of the lifecycle, every tar we work with is a layer.

cmd/defaults_windows.go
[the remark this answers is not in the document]
ekcasey: Right now, we do not acceptance test all of the flags and defaults for every phase. While we are working improving the test coverage I doubt we will ever get to 100% b/c high-level tests are costly in terms of speed and maintenance.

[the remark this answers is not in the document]
ameyer-pivotal: I see. For now there is some coverage via pack tests. Would you be open to seeing how it goes, and if it proves to be a problem, address it then? OS-specific files seem like the pattern for situations like this, looking at the standard golang libraries, and I struggle with adding complexity to the implementation when it feels like there's a better solution to the "keep in sync" problem lurking somewhere. Thoughts?

archive/tar_test.go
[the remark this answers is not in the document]
ameyer-pivotal: Ah, I see what you mean. I'll dig into this. We've been encountering errors like the following on rebuilds, so perhaps it's related:
```
ERROR: failed to export: layer 'simple/layers+cached-launch-layer' is cache=true but has no contents
```

ameyer-pivotal:
> I still need to try out the packager but I left some comments.
> 
> Also, why did we move from symlinking the `bin` dir for the test buildpacks to symlinking each individual file? It seems like we would end up with a lot more symlinks and the same behavior?

@ekcasey This was due to lack of support for directory symlinks when sharing the source code with a windows VM, or if using `docker cp` to copy into a container. It isn't a problem for `docker build` when using `COPY` in a dockerfile though...

To make the behavior more consistent, we opted to do this as dir symlinks haven't been to common in these repos anyhow.

launch/launcher_test.go
```diff
@@ -256,7 +261,7 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
 					{Type: "start", Command: "./start"},
 				}
 				launcher.Buildpacks = []launch.Buildpack{{ID: "bp.1"}, {ID: "bp.2"}}
-				launcher.Exec = syscallExecWithStdout(t, tmpDir)
+				launcher.Exec = testsyscall.SyscallExecWithStdout(t, tmpDir)
```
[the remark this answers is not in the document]
ameyer-pivotal: Will do. Moving into `testhelpers` package, as there were compilation issues with moving them into `launcher_test` (I think because the file names don't end in `_test.go`).

layers.go
```diff
@@ -98,7 +98,7 @@ func (bd *bpLayersDir) newBPLayer(name string) *bpLayer {
 	return &bpLayer{
 		layer{
 			path:       filepath.Join(bd.path, name),
-			identifier: fmt.Sprintf("%s:%s", bd.buildpack.ID, name),
+			identifier: fmt.Sprintf("%s+%s", bd.buildpack.ID, name),
```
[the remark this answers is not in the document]
ameyer-pivotal: Ultimately, these become part of a file name and the `:` is invalid for a file name on Windows. This was the simplest solution. Other than a one-time cache bust, this detail isn't visible to the end user.

env/env_test.go
```diff
@@ -65,11 +67,21 @@ func testEnv(t *testing.T, when spec.G, it spec.S) {
...
+
+			expected := []string{
 				fmt.Sprintf("LD_LIBRARY_PATH=%s/lib:some-ld", tmpDir),
 				fmt.Sprintf("LIBRARY_PATH=%s/lib:some-library", tmpDir),
 				fmt.Sprintf("PATH=%s/bin:some", tmpDir),
-			}); s != "" {
+			}
+			if runtime.GOOS == "windows" {
+				expected = []string{
+					fmt.Sprintf(`LD_LIBRARY_PATH=%s\lib;some-ld`, tmpDir),
```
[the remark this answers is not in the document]
ameyer-pivotal: Will do :)

exporter.go
```diff
@@ -377,11 +378,14 @@ func (e *Exporter) createSliceLayer(appDir, layerID string, files []string) (Sli
...
-		stat, _ := os.Stat(file)
+		stat, err := os.Stat(file)
+		if err != nil {
+			return SliceLayer{}, errors.Wrap(err, "failed to stat file")
+		}
 		if !stat.IsDir() {
 			err = os.Remove(file)
 			if err != nil {
-				e.Logger.Errorf("failed to delete file %v", err)
+				return SliceLayer{}, errors.Wrap(err, "failed to delete file")
```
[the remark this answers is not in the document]
ameyer-pivotal: That's a good point. I'll take a look.

cache/volume_cache.go
ekcasey: I think it's necessary. Busting every cache will create problems for folks using CNB at scale.

layers.go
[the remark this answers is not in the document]
ekcasey: The identifier is printed all over the place in the `analyzer`, `restorer`, and `exporter` logs. I think we should change the one place where we use the identifier to name a temporary file, instead of changing the identifier itself.

cache/volume_cache.go
ameyer-pivotal: OK, will do :)

cmd/defaults_windows.go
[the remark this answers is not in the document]
ameyer-pivotal: You could argue that a file named `defaults_unix.go` would warrant a look a the `defaults_windows.go` file which will sit next to it in most IDEs, if making any large changes.

exporter.go
```diff
@@ -377,7 +378,10 @@ func (e *Exporter) createSliceLayer(appDir, layerID string, files []string) (Sli
 	// the directories and delete if empty as a result of previous removal
 	var dirs []string
 	for file := range fileSet {
-		stat, _ := os.Stat(file)
+		stat, err := os.Stat(file)
+		if err != nil {
+			return SliceLayer{}, errors.Wrap(err, "failed to stat file")
```
ameyer-pivotal, line 383: @ekcasey I'm guessing these should follow the existing pattern and not be fatal too. I'll adjust (and revert the two below).

layers.go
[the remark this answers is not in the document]
ameyer-pivotal: Ah, that's true. OK it wasn't so bad to just change the file name. We had mistakenly thought the file name and identifier were more deeply entangled. I chose `__` instead of `+`. Logs will still show `:`.

exporter.go
```diff
@@ -276,7 +277,9 @@ func (e *Exporter) Cache(layersDir string, cacheStore Cache) error {
 }
 
 func (e *Exporter) tarLayer(layer identifiableLayer) (string, string, error) {
-	tarPath := filepath.Join(e.ArtifactsDir, launch.EscapeID(layer.Identifier())+".tar")
+	layerFilename := strings.ReplaceAll(layer.Identifier(), ":", "__")  // Replace colons, as they can be problematic for file names
```
ameyer-pivotal, line 280: @ekcasey Here's the new spot where we tweak the tar file path instead of the underlying identifier.

micahyoung:
> Also, why did we move from symlinking the `bin` dir for the test buildpacks to symlinking each individual file? It seems like we would end up with a lot more symlinks and the same behavior?

`docker cp` and `docker build` have very limited support for directory symlinks with Windows daemons but apparently complete support for file symlinks. The only directory symlinks that we could make work are ones with targets in the link's same directory or subdirectories (and even then only if the symlink path name is derived from the target name). 

This will also effect users during `pack build` if their apps have directory symlinks. This is likely a bug which would need to be fixed in Docker, though I suspect the only reason it hasn't already been fix is that these types of symlinks are uncommon in most Docker apps.

We recorded some experiments and findings [here](https://github.com/greenhouse-org/docker-windows-symlink-issues/blob/master/test.sh#L15-L32)

There's not a specific WCOW Docker/moby issue I can find for this, though some a similar issues for LCOW is present (which also would have effected `pack` LCOW users), suggesting bugs in symlink support are not uncommon: https://github.com/docker/for-win/issues/5778

archive/tar_test.go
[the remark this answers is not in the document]
ameyer-pivotal: @ekcasey This turned out to be the cause of the above error, so thanks for pointing this out! 😄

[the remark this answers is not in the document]
ameyer-pivotal: We'll be addressing it today and will hopefully have an update to the PR soon.

Makefile
ekcasey: It seems like `docker-run-linux` has some issues. It can't run `make acceptance` b/c the image is missing the `docker` CLI.

Also, when I tried `DOCKER_CMD="make unit" make docker-run-linux` I got the following error
```
docker run -v lifecycle-out:/lifecycle --rm lifecycle-img tar -cf- out | tar -xf-
tar: out: No such file or directory
```
It seems like we only want to `docker run -v lifecycle-out:/lifecycle --rm $(SOURCE_COMPILATION_IMAGE) tar -cf- out | tar -xf-` after particular tasks?

cmd/defaults_windows.go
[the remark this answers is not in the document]
ekcasey: I am not sure `pack` acceptance tests provide any coverage here b/c `pack` tends to set all of these vars explicitly rather than relying on the defaults. Using some combination of `filepath.Join` plus an OS specific root dir and executable extension feels much safer to me, and also doesn't seem like that much work? This is the kind of thing that has historically caused issues.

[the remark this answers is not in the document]
ameyer-pivotal: OK, will do :)

cmd/launcher/main.go
```diff
@@ -34,14 +36,27 @@ func runLaunch() error {
 		return cmd.FailErr(err, "read metadata")
 	}
 
+	var ex launch.ExecFunc
+	if runtime.GOOS == "windows" {
+		ex = func(argv0 string, argv []string, envv []string) error {
+			c := exec.Command(argv0, argv...)
+			c.Stdout = os.Stdout
+			c.Stderr = os.Stderr
+			c.Env = append(os.Environ(), envv...)
```
natalieparellano, line 45: Is there any danger in setting `os.Environ()` directly on the command? Should it not be `env.NewLaunchEnv(os.Environ())`? I see that we're passing `l.Env.List()` when we call `Launch`, which is appending to what's already there - but is this what we want? (I am just coming up to speed on the luancher, so apologies if this is an erroneous question...)

cmd/lifecycle/main.go
```diff
@@ -15,7 +16,7 @@ func main() {
 		cmd.Exit(err)
 	}
 
-	switch filepath.Base(os.Args[0]) {
+	switch strings.TrimSuffix(filepath.Base(os.Args[0]), ".exe") {
```
natalieparellano, line 19: I understand why we are doing this but it feels a bit odd to always trim `.exe` when it's only relevant for Windows. I wonder if a helper function could make this more clear.

cmd/launcher/main.go
ameyer-pivotal: Thanks for pointing this out. I'll look into it.

cmd/lifecycle/main.go
ameyer-pivotal: I'll be posting an update shortly that might happen to make this a tad nicer. Keep an 👁️  out for it :)

Makefile
micahyoung: > It seems like `docker-run-linux` has some issues. It can't run `make acceptance` b/c the image is missing the `docker` CLI.

Right, I believe that's due to the `make docker-run-linux` (container)-> `make acceptance` -> `go test acceptance_test.go` -> `make build-linux` -> `docker run ... go build` chain. I'm interested in how this could best work:
1.) Changing `make build-linux` to be more like - though not quite equal to - `make build-windows` (which now switches based on OS to either `build-windows-on-windows` or `build-windows-on-posix`). Maybe the targets for `build-linux` would be something like `build-linux-on-linux`/`build-linux-on-docker` depending on OS. The target `build-linux-on-docker` would essentially just do `docker-run-linux DOCKER_CMD="make build-linux"` I think builds would then be faster since we could remove the `apk add build-base` for every call and GOPATH could be cached.
2.) or, adding `docker` to the `tools/Dockerfile.linux` image to support DIND, though we'd miss out on speedup from of removing `apk add build-base` and re-using the `GOPATH` cache. Also the nesting goes one container deeper.

> It seems like we only want to `docker run -v lifecycle-out:/lifecycle --rm $(SOURCE_COMPILATION_IMAGE) tar -cf- out | tar -xf-` after particular tasks?

Good call, I feel like the way we implement `build-linux` (above) might dictate how we want to pull artifacts out of these containers (or nested containers if we go with DIND). I can do a follow-up pass to take care of both. Would we be

micahyoung: > It seems like `docker-run-linux` has some issues. It can't run `make acceptance` b/c the image is missing the `docker` CLI.

Right, I believe that's due to the `make docker-run-linux` (container)-> `make acceptance` -> `go test acceptance_test.go` -> `make build-linux` -> `docker run ... go build` chain. I'm interested in how this could best work:

1) Changing `make build-linux` to be more like - though not quite equal to - `make build-windows` (which now switches based on OS to either `build-windows-on-windows` or `build-windows-on-posix`). 

    Maybe the targets for `build-linux` could be something like `build-linux-on-linux`/`build-linux-on-docker` depending on OS. The target `build-linux-on-docker` would essentially just do `docker-run-linux DOCKER_CMD="make build-linux-on-linux"`. Repeated `build-linux-on-docker` runs would be faster since we could remove the `apk add build-base` for every call and GOPATH could be cached.
2) or, adding `docker` to the `tools/Dockerfile.linux` image to support DIND, though we'd miss out on speedup from of removing `apk add build-base` and re-using the `GOPATH` cache. Also the nesting goes one container deeper.

> It seems like we only want to `docker run -v lifecycle-out:/lifecycle --rm $(SOURCE_COMPILATION_IMAGE) tar -cf- out | tar -xf-` after particular tasks?

Good call, I feel like the way we implement `build-linux` (above) might dictate how we want to pull artifacts out of these containers (or nested containers if we go with DIND). I can do a follow-up pass to take care of both, depending on what is preferred.

cmd/defaults_windows.go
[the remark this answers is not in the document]
ameyer-pivotal: @ekcasey OK, so I took a middle-of-the-road approach and only used separate os-specific files for the parts that are not likely to ever change (the system root and executable extension string). The rest is provided through read-only values (functions) that use `filepath.Join`. :)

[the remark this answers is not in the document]
ameyer-pivotal: I'll mark this (now outdated) thread as resolved, but feel free to comment on the new code if needed.

Makefile
micahyoung: Actually, how about we delete the `docker-run-linux` target for now and I'll open a separate PR for it.

cmd/launcher/main.go
ameyer-pivotal: @natalieparellano After looking at this, I think appending to `os.Environ()` here is extraneous, and we can just set `c.Env = envv`. As you mentioned, we're passing `l.Env.List()` in as `envv`, so it should already have `os.Environ()` in it.

natalieparellano: After reading through `NewLaunchEnv` again I agree, and in fact I think we explicitly do not want to pass `os.Environ()` by itself - since variables that should be excluded are stripped out in that call: https://github.com/buildpacks/lifecycle/blob/9e53a47b65ff169c0c96fd518f006e23fcca25d7/env/launch.go#L12

Makefile
```diff
@@ -119,23 +144,19 @@ generate: install-mockgen
...
+	$(if $(shell goimports -l -w -local github.com/buildpacks/lifecycle .), @echo Fixed formatting errors. Re-run && exit 1)
 
 test: unit acceptance
 
-unit: verify-jq format lint install-yj
+unit: UNIT_PACKAGES=$(shell $(GOCMD) list ./... | grep -v acceptance)
+unit: format lint install-yj
 	@echo "> Running unit tests..."
-	$(GOTEST) -v -count=1 $$($(GOCMD) list ./... | grep -v acceptance)
+	$(GOTEST) -v -count=1 $(UNIT_PACKAGES)
```
micahyoung, line 154: Had to refactor out the inlined subcommand since the syntax doesn't work with Windows and Linux

cmd/launcher/main.go
```diff
@@ -34,14 +36,27 @@ func runLaunch() error {
 		return cmd.FailErr(err, "read metadata")
 	}
 
+	var ex launch.ExecFunc
+	if runtime.GOOS == "windows" {
+		ex = func(argv0 string, argv []string, envv []string) error {
+			c := exec.Command(argv0, argv...)
```
ekcasey, line 42: In this case `argv` will contain `argv0` and you don't want to specify it twice in `exec.Command`

archive/tar_test.go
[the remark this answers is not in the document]
aemengo: PR updated.

cmd/launcher/main.go
ameyer-pivotal: Haha, we literally just ran into this 😄  (cc: @micahyoung)

cmd/flags.go
```diff
@@ -4,24 +4,29 @@ import (
...
-	DefaultOrderPath           = "/cnb/order.toml"
-	DefaultGroupPath           = "./group.toml"
-	DefaultStackPath           = "/cnb/stack.toml"
-	DefaultAnalyzedPath        = "./analyzed.toml"
-	DefaultPlanPath            = "./plan.toml"
-	DefaultProcessType         = "web"
-	DefaultLauncherPath        = "/cnb/lifecycle/launcher"
-	DefaultLogLevel            = "info"
-	DefaultProjectMetadataPath = "./project-metadata.toml"
+func DefaultLayersDir() string     { return filepath.Join(sys.Root, "layers") }
```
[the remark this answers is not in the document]
ameyer-pivotal: I debated `var`s but wanted them read only. Happy to change to that though.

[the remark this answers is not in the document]
ekcasey: Yes please :). Sorry to be picky but I think the vars read more nicely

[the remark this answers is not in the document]
ameyer-pivotal: No problem. Will do :)

cmd/launcher/main.go
ameyer-pivotal: Fixed.

env/vars.go
```diff
@@ -0,0 +1,42 @@
+package env
+
+import "strings"
+
+type Vars struct {
+	vals       map[string]string
+	ignoreCase bool
+}
+
+func NewVars(vars map[string]string, ignoreCase bool) *Vars {
```
ameyer-pivotal, line 10: @ekcasey We debated whether to inline `ignoreCase` to `runtime.GOOS == "windows"` and eliminate it from the constructor. On the plus side of the current implementation, it does make it easy to test as a unit, and not have to switch OS targets to fully test it. Thoughts or preferences?

ameyer-pivotal: Also some background: the reason for the new `Vars` type is that case sensitivity of env vars is now an important distinction between OSes, and we wanted to eliminate potential bugs due to folks working with maps directly.

ekcasey: I like the decision to pass in an explicit `ignoreCase` or `caseSensitive` but I am wondering if inject it into `NewLaunchEnv` and `NewBuildEnv` instead of hardcoding `runtime.GOOS == "windows"` in `varsFromEnviron`. It makes sense to me from a domain modeling perspective. The entire environment is either case sensitive or case insensitive.

ameyer-pivotal: That makes sense. I can make that change. There is also one other place where the constructor is called, in `WithPlatform()`, where I don't like that we have to check the runtime. Instead, I'll just copy the original `ignoreCase` value (look for a comment in the PR). That should leave checking the runtime GOOS in only the two places you mentioned.

env/env.go
```diff
@@ -56,57 +57,46 @@ func (p *Env) AddEnvDir(envDir string) error {
...
 		return nil
 	})
 }
 
 func (p *Env) WithPlatform(platformDir string) (out []string, err error) {
-	vars := make(map[string]string)
-	for key, value := range p.Vars {
-		vars[key] = value
-	}
+	vars := NewVars(p.Vars.vals, runtime.GOOS == "windows")
```
ameyer-pivotal, line 78: I'll change this to the following, to remove the runtime check:
```
NewVars(p.Vars.vals, p.Vars.ignoreCase)
```

archive/tar_test.go
```diff
@@ -35,24 +36,40 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
...
+				{"root", os.ModeDir + 0777},
+				{"root/readonly", os.ModeDir + 0555},
+				{"root/standarddir", os.ModeDir + 0777},
+				{"root/standarddir/somefile", 0666},
+				{"root/readonly/readonlysub/somefile", 0444},
+				{"root/readonly/readonlysub", os.ModeDir + 0555},
+				{"standardfile_Files", 0666},
+				{"standardfile_Hives", 0666},
```
ekcasey, line 63: After talking with @micahyoung  in buildpacks slack it seems like this isn't a realistic test value and documents an incorrect handling actual `Hives/*` entries

> Micah:
There are five “Hive-deltas” in the Hives/* part of the layer and each one gets merged into one of 5 registry files that appear on-disk inside the container
>
>A base image has most of the entries for one of those sections, but any additional layers will add more entries
>
>The names are:
Hives/DefaultUser_Delta
Hives/Sam_Delta
Hives/Security_Delta
Hives/Software_Delta
Hives/System_Delta

Seems like we should either merge `Hive` deltas if we find them, or (if we want to punt for now) just ignore `Hives/*` files?

archive/tar_test.go
```diff
@@ -35,24 +36,40 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
...
+
+		// Golang for Windows only implements owner permissions
+		if runtime.GOOS == "windows" {
+			pathModes = []archive.PathMode{
+				{"root", os.ModeDir + 0777},
+				{"root/readonly", os.ModeDir + 0555},
+				{"root/standarddir", os.ModeDir + 0777},
+				{"root/standarddir/somefile", 0666},
+				{"root/readonly/readonlysub/somefile", 0444},
+				{"root/readonly/readonlysub", os.ModeDir + 0555},
```
ekcasey, line 61: Should tar entries with names that don't start with `Files/` get written to disk in the windows case?

ekcasey: Also, I am surprised this works with forward slashes on windows. Does `os.Stat` convert these for us on windows?

launch/launcher.go
```diff
@@ -50,10 +53,25 @@ func (l *Launcher) Launch(self string, cmd []string) error {
...
+	profileCmds, err := l.profileD()
 	if err != nil {
 		return errors.Wrap(err, "determine profile")
 	}
+
+	if runtime.GOOS == "windows" {
+		var launcher string
+		if len(profileCmds) > 0 {
+			launcher = strings.Join(profileCmds, " && ") + " && "
+		}
```
ekcasey, line 65: Joining with " && " should work for both `bash` and `cmd`

sys/sys_unix.go
```diff
@@ -0,0 +1,8 @@
+// +build linux darwin
+
+package sys
+
+const (
+	Root    = "/"
+	ExecExt = ""
+)
```
ekcasey, line 8: I don't think these two constants deserve their own package. Can we move them to `cmd`? If there are import issues I think it is okay to repeat these constants in other packages. I would prioritize keeping the import graph tidy. There may be cases like the usage in `restorer` where we could remove them entirely (we can make `UntarLayer` default to the root when given an empty destination).

Makefile
```diff
@@ -1,52 +1,59 @@
...
-[api]
-  platform = "$(PLATFORM_API)"
-  buildpack = "$(BUILDPACK_API)"
-
-[lifecycle]
-  version = "$(LIFECYCLE_VERSION)"
-endef
+BUILD_DIR?=$(PWD)$/out
+LINUX_COMPILATION_IMAGE?=golang:1.13-alpine
+WINDOWS_COMPILATION_IMAGE?=golang:1.14-windowsservercore-1809
```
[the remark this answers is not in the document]
ameyer-pivotal: Will do!

Makefile
```diff
@@ -1,52 +1,59 @@
...
+LIFECYCLE_VERSION?=$(shell type VERSION)
+else
+/:=/
+LIFECYCLE_VERSION?=$(shell cat VERSION)
+endif
+
 GOCMD?=go
 GOARCH?=amd64
 GOENV=GOARCH=$(GOARCH) CGO_ENABLED=0
-LDFLAGS=-s -w
```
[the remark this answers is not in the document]
ameyer-pivotal: I'm not sure... @micahyoung, was this intentional?

Makefile
```diff
@@ -1,52 +1,59 @@
...
 
-
 build-linux-launcher: export GOOS:=linux
 build-linux-launcher: OUT_DIR?=$(BUILD_DIR)/$(GOOS)/lifecycle
 build-linux-launcher:
 	@echo "> Building lifecycle/launcher for linux..."
 	mkdir -p $(OUT_DIR)
 	$(GOENV) $(GOBUILD) -o $(OUT_DIR)/launcher -a ./cmd/launcher
-	test $$(du -m $(OUT_DIR)/launcher|cut -f 1) -le 3
+	test $$(du -m $(OUT_DIR)/launcher|cut -f 1) -le 4
```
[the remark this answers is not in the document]
ameyer-pivotal: I'll try reverting both this and the LDFLAGS removal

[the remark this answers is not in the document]
ameyer-pivotal: Seems to work. Good catch!

archive/tar_path.go
```diff
@@ -0,0 +1,26 @@
...
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+// TarPath converts an OS path to a path suitable for TAR headers
+func TarPath(path string) string {
+	volumeName := filepath.VolumeName(path)
+	path = strings.TrimPrefix(path, volumeName)
+	return filepath.ToSlash(path)
```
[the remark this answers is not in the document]
ameyer-pivotal: Interesting question... Originally, this wasn't based on any concrete evidence other than the patterns in our code for both `pack` and `lifecycle`. We chose to "normalize" the creation of tars to simplify the code (and because somewhere along the way it was stuck in our heads that tar headers always used forward slashes).

Looking at the TAR spec seems to be consistent...

For the slashes (vs backslashes), I found some evidence for that [here](https://www.gnu.org/software/tar/manual/html_node/Standard.html):
> The `name` field is the file name of the file, with directory names (if any) preceding the file name, separated by slashes.

The volume part is harder to nail down in the TAR spec as it's not mentioned in the above quote. Based on pages like [this](https://www.gnu.org/software/tar/manual/html_node/absolute.html), it doesn't seem like volume names would be expected in the `name` field though. I'd imagine having a volume name in the header path would make untarring just have to do the same trimming.

archive/tar_test.go
```diff
@@ -35,24 +36,40 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
...
 			{"root", os.ModeDir + 0755},
 			{"root/readonly", os.ModeDir + 0500},
 			{"root/standarddir", os.ModeDir + 0755},
 			{"root/standarddir/somefile", 0644},
 			{"root/readonly/readonlysub/somefile", 0444},
 			{"root/readonly/readonlysub", os.ModeDir + 0500},
+			{"Files", os.ModeDir + 0755},
+			{"Files/standardfile_Files", 0644},
+			{"Hives", os.ModeDir + 0755},
+			{"Hives/standardfile_Hives", 0644},
```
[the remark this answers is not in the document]
ameyer-pivotal: I think these were intended for the Windows section. Will fix.

Makefile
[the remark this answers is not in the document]
ameyer-pivotal: Added it back and found no issues. It also dropped the size back down.

archive/tar_test.go
ameyer-pivotal: I think that merging Hive deltas can be saved for a later PR. I'll adjust the code to skip for now, and we can submit a PR when we want to address it.

archive/tar_test.go
ameyer-pivotal: Yeah good catch about the slashes. I'll look into this... It's probably just not trying to chmod.

archive/tar_test.go
ameyer-pivotal: OK update is pushed @ekcasey

sys/sys_unix.go
ameyer-pivotal: The benefit to having these in their own package is that it addresses two concerns:
- Avoids import cycles since this can be seen as a "leaf" node in the import tree
- This was created when addressing the concern about keeping the default paths in sync between OSes (refactored separate OS-dependent files to a single file that uses `filepath.Join`, etc. in combination with these constants)

It seems the second bullet would sort of go against repeating these constants in multiple places. I'll see if I can move them to a different package.

launch/launcher.go
ameyer-pivotal: I'm happy to do this, but we were trying to leave the Linux case rather untouched, as they were joined with newlines before.

sys/sys_unix.go
ameyer-pivotal: So tinkering a bit, `cmd` seems like too high-level of a package to move this to without causing import cycles. How important is this? It feels like one of those cases that can go either way, depending on stylistic opinions. I'd argue that the import tree is simpler to understand and use with this as a leaf node (i.e. modular, self-contained packages), but other folks might argue that simpler means fewer packages altogether.

I see a few options:
- Leave as is
- Find a lower-level package where this could live instead
- Force it into command and spend time debugging import cycles

Thoughts?

ameyer-pivotal: @ekcasey Actually, it worked pretty nicely to move into the `env` package. That feels like a reasonable compromise. Let me know what you think :)

Makefile
```diff
@@ -151,23 +170,24 @@ package-linux: GOOS:=linux
...
-	echo "$${LIFECYCLE_DESCRIPTOR}" > $(GOOS_DIR)/lifecycle.toml
-
 	@echo "> Packaging lifecycle for $(GOOS)..."
-	tar czf $(BUILD_DIR)/$(ARCHIVE_NAME).tgz -C $(GOOS_DIR) lifecycle.toml lifecycle
+	$(GOCMD) run tools/packager/main.go -os $(GOOS) -launcherExePath $(GOOS_DIR)/lifecycle/launcher -lifecycleExePath $(GOOS_DIR)/lifecycle/lifecycle -lifecycleVersion $(LIFECYCLE_VERSION) -platformAPI $(PLATFORM_API) -buildpackAPI $(BUILDPACK_API) -outputPackagePath $(BUILD_DIR)/$(ARCHIVE_NAME).tgz
 
-package-windows: export LIFECYCLE_DESCRIPTOR:=$(LIFECYCLE_DESCRIPTOR)
 package-windows: GOOS:=windows
-package-windows: GOOS_DIR:=$(BUILD_DIR)/$(GOOS)
+package-windows: GOOS_DIR:=$(BUILD_DIR)$/$(GOOS)
```
natalieparellano, line 177: What's the `$` for? Is it a typo?

ameyer-pivotal: It's not actually. It's an OS-agnostic slash. You'll see the definition near the top of the file :)

micahyoung: In this case, the `$/` allows the same target to work for both `bash` and `cmd` since the syntax is otherwise the same. My hope is that while writing cross-shell syntax may be a little trickier, by keeping most targets as identical as possible it will make it less likely for Windows and Linux coverage to diverge.

natalieparellano: Ah! That is fancy...

testdata/buildpack/bin/detect.bat
```diff
@@ -0,0 +1,2 @@
+@echo off
+bash %~dp0\detect %1 %2
```
ekcasey, line 2: I was surprised to find that I couldn't run the detect unit tests on a windows machine w/o `bash` installed. The failures give the false impression that `bash` is a dependency of the detector. However the implementation is correctly calling `detect.bat` which in turn requires `bash`. Given that we generally won't have `bash` available when the lifecycle runs on windows I am not sure I am comfortable with the dependency in the test. It makes it impossible to validate that the lifecycle can execute correctly in an environment sans `bash`. Can we reimplement the detect test logic in the `*.bat` files?

archive/tar_test.go
```diff
@@ -126,7 +195,7 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
 				defer file.Close()
 				tr := tar.NewReader(file)
 
-				tarContains(t, "directories", func() {
+				tarContains(t, func() {
```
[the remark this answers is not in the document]
ameyer-pivotal: Sure thing. Will revert :)

archive/tar_test.go
```diff
@@ -35,24 +37,46 @@ func testTar(t *testing.T, when spec.G, it spec.S) {
...
-			{"root/readonly/readonlysub", os.ModeDir + 0500},
+			{"root/symlinkdir", os.ModeDir + 0755},
+			{"root/symlinkdir/subdir", os.ModeDir + 0755},
+			{"root/symlinkdir/subdir/somefile", 0644},
+			{"root/symlinkdir/symlink", 0644},
+		}
+
+		// Golang for Windows only implements owner permissions
+		if runtime.GOOS == "windows" {
+			layerTar = "some-windows-layer.tar"
```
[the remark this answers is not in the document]
ameyer-pivotal: By:

> I expected this test fixture to have the files prefixed with the special windows Files dir

did you mean you expected them all to be prefixed? We do have the `Files` prefix in the first couple of entries, and all the files are checked for presence and permissions in the `it("extracts a tar file")` below.

That said, You could argue that the test fixture would match a more real world scenario if almost all were prefixed with `Files/`. We can leave one out to test that it doesn't get extracted (like we do for the one prefixed with `Hives/`).

I'll make this change.

sys/sys_unix.go
ekcasey: > It seems the second bullet would sort of go against repeating these constants in multiple places. I'll see if I can move them to a different package though.

I feel like this concern is less relevant for these constants than for the default input values. We will realistically change default input values at some point and we don't want linux/windows values to fall out of sync. The posix and windows fs root dirs on the other hand, should not change. This is how I would handle it:

https://github.com/buildpacks/lifecycle/commit/035a3c0909ef32ce9d63983573203d048b70a5f7

I have removed some unnecessary usage of these constants leaving us with the original motivating usage in `cmd` and one other usage in `archive` (since we already have very specific OS-dependent path logic there it doesn't feel crazy to me to redefine these in that context). Then

ekcasey: My main concern is that when these are exported they become part of the public package level interface for `env` and I don't think we want that. I would like the keep the import graph and the godoc as clean and straightforward as possible.

archive/tar_test.go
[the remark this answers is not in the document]
ameyer-pivotal: Done.

image/writer_factory.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/buildpacks/lifecycle/archive"
+)
+
+type LayerWriterFactory struct {
+}
+
+func NewLayerWriterFactory() (*LayerWriterFactory, error) {
+	return &LayerWriterFactory{}, nil
+}
```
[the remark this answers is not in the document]
ameyer-pivotal: Yeah, I agree we can remove the constructor.

testdata/buildpack/bin/detect.bat
ameyer-pivotal: We can rewrite these in pure `batch`. The makefile provides a way to run the tests in a docker container that has all the necessary dependencies, but I understand the overall concerns.

sys/sys_unix.go
ameyer-pivotal: OK. Makes sense. We'll pull in your suggested change :)

PR #315, opened 2020-06-15
Adopt race-neutral terminology

### LINKED ISSUE #314: Cloud Native Buildpacks should use race-neutral terminology allowlist & blocklist
### Summary
It's clear there are people who are hurt and made to feel excluded by the industry's use of terms such as `whitelist` and `blacklist`. The terms `allowlist` and `blocklist` describe their purpose, while the other words use metaphors to describe their purpose. We currently implement a `whitelist` for build env vars, and a `blacklist` for launch env vars, however they don't make up part of our user interface, so this appears to be a case where we can rename the variables, methods and test cases with no impact to our implementation.

---

##### Steps

1. https://github.com/buildpacks/lifecycle/blob/master/env/build.go#L3-L7
2. https://github.com/buildpacks/lifecycle/blob/master/env/launch.go#L3-L7

##### Current behavior
Use of terms `whitelist` and `blacklist`

##### Expected
Use of terms `allowlist` and `blocklist`

---

##### lifecycle version
`master`

simonjjones:
Resolves #314

Signed-off-by: Simon Jones <simonjones@vmware.com>

### DIFF
--- a/env/launch.go
+++ b/env/launch.go
@@ -1,6 +1,6 @@
 package env
 
-var LaunchEnvBlacklist = []string{
+var LaunchEnvBlocklist = []string{
 	"CNB_LAYERS_DIR",
 	"CNB_APP_DIR",
 	"CNB_PROCESS_TYPE",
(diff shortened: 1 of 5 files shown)


env/launch.go
```diff
@@ -1,6 +1,6 @@
 package env
 
-var LaunchEnvBlacklist = []string{
+var LaunchEnvBlocklist = []string{
```
ekcasey, line 3: I did a quick google search and it seems like most folks are going with `denylist` instead of `blocklist`. Might as well use the same terminology others are adopting.

E.g. https://github.com/rails/rails/issues/33677

simonjjones: Go have gone with `blocklist` https://go.googlesource.com/go/+/608cdcaede1e7133dc994b5e8894272c2dce744b

I can see the argument in favor of `denyList` in systems where we'd ever need to list `blocks` but I don't think that's the case for us.

I'd rather make the change than not have the PR accepted, but I do think blocklist reads slightly better, and makes slightly more sense.

simonjjones: Actually, both of those comparisons are slightly off, we're not describing a service that blocks/denies/allows generic things. We could be a little more specific in this case and use `include` and `exclude`

ekcasey: agree! `include` and `exclude` seem like the best/clearest option.

PR #319, opened 2020-06-23
Add analyzer acceptance tests

natalieparellano:


Makefile
```diff
@@ -135,7 +135,7 @@ unit: verify-jq format lint install-yj
 acceptance: format lint
 	@echo "> Running acceptance tests..."
 	$(GOTEST) -v -count=1 -tags=acceptance ./acceptance/...
-	
+
 acceptance-darwin: format lint
```
natalieparellano, line 139: This doesn't seem to do anything different from `make acceptance`.

acceptance/acceptance_test.go
```diff
@@ -157,23 +157,12 @@ func lifecycleCmd(binary string, args ...string) *exec.Cmd {
 	return exec.Command(filepath.Join(buildDir, runtime.GOOS, "lifecycle", binary), args...)
 }
 
-func buildBinaries(t *testing.T, dir string, goos string) {
```
natalieparellano, line 160: Moved to test helpers.

acceptance/launcher_test.go
```diff
@@ -1,3 +1,5 @@
+// +build acceptance
```
natalieparellano, line 1: This seems to be required in order to share functions such as `buildTestImage` that are defined in acceptance_test.go. I am not sure why it was missing before.

acceptance/launcher_test.go
```diff
@@ -202,20 +204,6 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
 	})
 }
 
-func buildLaunchImage(t *testing.T) {
```
natalieparellano, line 205: Made these more re-usable.

exporter_test.go
```diff
@@ -1427,7 +1427,7 @@ func assertReuseLayerLog(t *testing.T, logHandler *memory.Handler, name, sha str
 
 func assertTarFileContents(t *testing.T, tarfile, path, expected string) {
 	t.Helper()
-	exist, contents := tarFileContext(t, tarfile, path)
+	exist, contents := tarFileContents(t, tarfile, path)
```
natalieparellano, line 1430: This seemed like a typo.

testhelpers/docker.go
```diff
@@ -0,0 +1,96 @@
+package testhelpers
```
natalieparellano, line 1: Pulled docker-related test helpers into their own file for ease of discovery.

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+			)
+			output := h.Run(t, cmd)
+
+			expected := "Not restoring cached layer metadata, no cache flag specified."
+			h.AssertStringContains(t, output, expected)
+		})
+	})
+
+	when("the provided layers directory isn't writeable", func() {
+		it("recursively chowns the directory", func() {
```
natalieparellano, line 97: There is some subtlety around `canWrite()` that is likely not captured here...

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+							"--mount", "type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock",
+							"--volume", fmt.Sprintf("%s:/cache", cacheDir),
+							analyzeImage,
+							"/bin/bash", "-c",
+							fmt.Sprintf("%s -daemon -cache-dir /cache some-image; ls -alR /cache", analyzerPath),
+						)
+						output := h.Run(t, cmd)
+
+						h.AssertMatch(t, output, "2222 3333 .+ \\.")
+						h.AssertMatch(t, output, "2222 3333 .+ committed")
```
natalieparellano, line 282: The staging directory is not chowned - is that expected?

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+	})
+
+	when("layers path and analyzed path are provided", func() {
+		it("writes analyzed.toml at the provided path", func() {
+			cmd := exec.Command(
+				"docker", "run", "--rm",
+				"--env", "CNB_REGISTRY_AUTH={}",
+				analyzeImage,
+				"/bin/bash", "-c",
+				fmt.Sprintf("%s -layers /other-layers -analyzed /other-layers/other-analyzed.toml some-image; ls -al /other-layers", analyzerPath),
```
natalieparellano, line 456: I find it a little odd that `-layers` must be used together with `-analyzed` (else we might run into permissions issues trying to write `/layers/analyzed.toml`). Also we're still expecting to find the group.toml at `/layers/group.toml`. I see how this all works but it's a bit confusing...

testhelpers/testhelpers.go
```diff
@@ -343,3 +349,40 @@ func MustReadFile(t *testing.T, path string) []byte {
...
+			t.Log("WARNING: Not re-building lifecycle binaries, found existing binaries in", dir)
+			return
+		}
+	}
+
+	BuildLifecycle(t, dir, goos)
+}
+
+func BuildLifecycle(t *testing.T, dir string, goos string) {
+	cmd := exec.Command("make", "build-"+goos)
```
natalieparellano, line 370: Changed this from `runtime.GOOS`, but now that I'm looking it over again I'm not sure if that's what we want. I can change it back...

natalieparellano: Actually it seems that this function, where it was used previously (in `TestVersion`), was/is being called with `runtime.GOOS`, so this doesn't change any existing functionality.

acceptance/launcher_test.go
```diff
@@ -224,14 +176,12 @@ func buildLauncher(t *testing.T) {
 	cmd.Env = append(
 		os.Environ(),
 		"PWD="+cmd.Dir,
+		"BUILD_DIR="+launcherBinaryDir, // `make clean` should remove the launcher binary dir, not the default dir (out)
```
natalieparellano, line 179: I think removing the `launcherBinaryDir` with `make clean` is what we want and also potentially fixes a race condition with the analyzer tests relying on `out`. But maybe as mentioned above, that whole strategy needs re-working...

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+	})
+
+	when("registry case", func() {
+		it("writes analyzed.toml", func() {
+			cmd := exec.Command(
+				"docker", "run", "--rm",
+				"--env", "CNB_REGISTRY_AUTH={}",
+				analyzeImage,
+				"/bin/bash", "-c",
+				fmt.Sprintf("%s some-image; ls -al /layers", analyzerPath),
```
jromero, line 296: I was talking to @yaelharel about how we would do validation of the outcomes and looked at this PR as reference. It throw us for a loop to identify how the analyzer was being executed. After the fact, I acknowledge that it's pretty clever but wonder if it's maybe to obscure for others as well. It may be it's solely on the format (aka use of `fmt.Sprintf`) and something like a function that makes it more apparent would help.

For example:
```suggestion
				analyzeAndExec(analyzerPath, "ls -al /layers", ...),
```

An alternative being a more explicit helper struct/function.

natalieparellano: Ooh. I like this - thanks for the suggestion :)

.gitignore
```diff
@@ -6,3 +6,4 @@
 /out
 
 acceptance/testdata/*/container/cnb/lifecycle/*
+acceptance/testdata/*/**/container/cnb/lifecycle/*
```
ekcasey, line 9: I think this line covers the above case too

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+	os.Setenv("DOCKER_CONFIG", registry.DockerDirectory)
+
+	outDir, err := filepath.Abs(filepath.Join("..", "out"))
+	h.AssertNil(t, err)
+
+	info, err := h.DockerCli(t).Info(context.TODO())
+	h.AssertNil(t, err)
+	osType := info.OSType
+
+	h.ConditionallyBuildLifecycle(t, outDir, osType)
```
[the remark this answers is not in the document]
ekcasey: I think the best course of action here might be to improve our `Makefile`. A good `Makefile` would only recompile when necessary. I worry that I will be the first person to forget to rebuild the lifecycle and miss this warning :)

acceptance/launcher_test.go
ekcasey: good catch! However, if the `analyzer` test are going to use the binaries from the default `out` dir I do think we should change the `launcher` tests to be consistent with that.

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+	h.CopyLifecycle(t, filepath.Join(outDir, osType, "lifecycle"), analyzerBinaryDir)
+
+	buildTestImage(t, analyzeImage, analyzeDockerContext)
+	defer removeTestImage(t, analyzeImage)
+	spec.Run(t, "acceptance", testAnalyzer, spec.Parallel(), spec.Report(report.Terminal{}))
```
dfreilich, line 60:
```suggestion
	spec.Run(t, "acceptance-analyzer", testAnalyzer, spec.Parallel(), spec.Report(report.Terminal{}))
```

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+					h.AssertMatch(t, output, "some-layer.sha")
+					h.AssertMatch(t, output, "some-layer.toml")
+				})
+
+				when("the provided cache directory isn't writeable", func() {
+					it("recursively chowns the directory", func() {
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--mount", "type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock",
+							"--volume", fmt.Sprintf("%s:/cache", cacheDir),
```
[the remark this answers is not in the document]
ekcasey: This issue probably doesn't affect the tests as written b/c the image always running with `USER=root`(and dropping privs) but it might matter (and confuse us) if we change the container user in future test cases.

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+				analyzeImage,
+				analyzerPath,
+				"-group", "/layers/other-group.toml",
+				"some-image",
+			)
+			h.Run(t, cmd)
+		})
+	})
+
+	when("daemon case", func() {
```
dfreilich, line 126:
```suggestion
	when("mounted daemon", func() {
```

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+
+	h.ConditionallyBuildLifecycle(t, outDir, osType)
+	h.CopyLifecycle(t, filepath.Join(outDir, osType, "lifecycle"), analyzerBinaryDir)
+
+	buildTestImage(t, analyzeImage, analyzeDockerContext)
+	defer removeTestImage(t, analyzeImage)
+	spec.Run(t, "acceptance", testAnalyzer, spec.Parallel(), spec.Report(report.Terminal{}))
+}
+
+func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
```
yaelharel, line 63: Following a discussion that we had, I think we should add a test that checks a "good path" of writing the `analyzedMD` output of `analyze` (in `cmd/lifecycle/analyzer.go`).
Currently we only check the content of `analyzedMD` in the unit tests and the creation of this file. Probably it's worth checking its content at least once.

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+func TestAnalyzer(t *testing.T) {
+	dockerConfigDir, err := ioutil.TempDir("", "test.docker.config.dir")
+	h.AssertNil(t, err)
+	defer os.RemoveAll(dockerConfigDir)
+
+	registry = ih.NewDockerRegistryWithAuth(dockerConfigDir)
+	registry.Start(t)
+	defer registry.Stop(t)
+
+	os.Setenv("DOCKER_CONFIG", registry.DockerDirectory)
```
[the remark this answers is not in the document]
natalieparellano: On looking into this further it seems that this call: https://github.com/buildpacks/lifecycle/pull/319/files#diff-72171b9647f1d25c895f3dcfa8e02b44R499 to `auth.BuildEnvVar(authn.DefaultKeychain, repoName)` is going to do a `os.Getenv("DOCKER_CONFIG")` when calling `authn.DefaultKeychain.Resolve()`. I don't see an obvious way of setting that environment variable just for that function call - but it looks like that's the only place where it's required.

Given that the contents of `os.Environ()` don't actually propagate through to the docker container where analyzer is running, it seems that the potential for pollution is lower. I'm open to other ideas though about how we could make this better.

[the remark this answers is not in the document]
ekcasey: Good point! If we _really_ wanted to we could use `docker` and `ggcr` to make a Keychain from the docker config file. BUT, it's probably not worth it.

testhelpers/docker_registry.go
```diff
@@ -1,59 +0,0 @@
...
-	"fmt"
-	"testing"
-	"time"
-
-	dockertypes "github.com/docker/docker/api/types"
-	"github.com/docker/docker/api/types/container"
-	"github.com/docker/go-connections/nat"
-)
-
-type DockerRegistry struct {
```
yaelharel, line 14: Just to be sure that I didn't miss something. You deleted this file because we don't use anywhere, right? Meaning, it doesn't relate to the analyzer acceptance tests, right?

testhelpers/testhelpers.go
```diff
@@ -100,6 +103,13 @@ func AssertNil(t *testing.T, actual interface{}) {
 	}
 }
 
+func AssertNotNil(t *testing.T, actual interface{}) {
```
yaelharel, line 106: Maybe it's only me but I'm a bit confused with the name of the functions `AssertNil` and `AssertNotNil`.
`AssertNil` means that we expect the input to be `nil` and we throw "fatal" if it's `not nil`.
`AssertNotNil` means that we expect the input to be `not nil` and we throw "fatal" if it's `nil`.
What do you think about changing the names of these functions to `FatalOnNil` and `FatalOnNotNil`? But then you need to replace every line that is currently using `AssertNil` to `FatalOnNotNil`, and similarly to `AssertNotNil`.
Another idea is to change the names to be `ExpectNil` and `ExpectNotNil`.
Again, maybe it's only me, so if you think it's clear the way it is right now, I'll deal with my confusion :)

testhelpers/docker.go
```diff
@@ -0,0 +1,96 @@
...
+	"sync"
+	"testing"
+
+	dockertypes "github.com/docker/docker/api/types"
+	dockercli "github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/jsonmessage"
+	"github.com/pkg/errors"
+)
+
+var dockerCliVal *dockercli.Client
```
yaelharel, line 19: It's great that you split `testhelpers.go`!

testhelpers/testhelpers.go
natalieparellano: It looks like we already have `AssertError()` - how does that sound?

yaelharel: As far as I see, `AssertError` is a bit different because it checks that we got an error and that it contains a specific string while in `AssertNil` and `AssertNotNil` we only check whether we got an error or not.
Moreover, I think it's good to have these two functions (maybe `AssertError` can use one of them) but I would only change their names.

acceptance/launcher_test.go
```diff
@@ -31,12 +33,9 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
 		when("CNB_PROCESS_TYPE is NOT set", func() {
 			it("web is the default process-type", func() {
 				cmd := exec.Command("docker", "run", "--rm", launchImage)
-				output, err := cmd.CombinedOutput()
-				if err != nil {
-					t.Fatalf("failed to run %v\n OUTPUT: %s\n ERROR: %s\n", cmd.Args, output, err)
-				}
+				output := h.Run(t, cmd)
```
yaelharel, line 36: It is a great change! Love reducing code duplication :)
Two small things about it:
1. `cmd.CombineOutput` checks that `Stdout` and `Stderr` are `nil` before calling to `Run`. Do you think we should add it once at the beginning of `testLauncher`?
2. Do you think it's worth creating a function that combines both `Run` and checks whether the `output` contains some expected strings?

testhelpers/binaries.go
```diff
@@ -0,0 +1,76 @@
...
+		"PLATFORM_API=0.9",
+		"LIFECYCLE_VERSION=some-version",
+		"SCM_COMMIT=asdf123",
+	)
+
+	t.Log("Building binaries: ", cmd.Args)
+	Run(t, cmd)
+}
+
+func CopyLifecycle(t *testing.T, src, dst string) {
```
yaelharel, line 46: Do we really need it? Can't all tests read it from the directory we built it in?
And if we do need it, do we need to add it to `launcher_test.go`? Currently I see it only in `analyzer_test.go`.

testhelpers/binaries.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+)
+
+func ConditionallyBuildLifecycle(t *testing.T, dir string, goos string) {
```
yaelharel, line 11: Do we need to add it to launcher_test.go?

acceptance/analyzer_test.go
natalieparellano: It looks like it is being created with the wrong permissions:
```
        /cache:
        total 4
        drwxrwxr-x 4 2222 3333  128 Jun 26 19:59 .
        drwxr-xr-x 1 root root 4096 Jun 26 19:59 ..
        drwxrwxr-x 4 2222 3333  128 Aug 30  1754 committed
        drwxr-xr-x 2 root root   64 Jun 26 19:59 staging
```

What's interesting, when I add a staging directory to my fixture, it looks like that is ignored? At least, I don't see the `foo.txt` file that I inserted there. Will investigate...

acceptance/launcher_test.go
natalieparellano: For 1, I don't think it is necessary to check, as `h.Run()` calls `cmd.Output()` which also has this check for stdout, and we initialize stderr to empty. I think this is more about the internals of how the `cmd` is being run, vs. what we are testing.

For 2, I believe the windows PR that's currently up will introduce an `assertOutput`. We might take advantage of that - I didn't do it here as some of the assertions are using regex matchers which don't really pair nicely with that kind of matcher. But maybe we can have something like `assertOutputMatches`...

testhelpers/binaries.go
natalieparellano: We need to copy it into the context for the `docker build`. Dockerfiles can't refer to files outside their context.

acceptance/analyzer_test.go
natalieparellano: I was playing around with this and I came up with the following: 
```
			cmd := exec.Command("docker", append([]string{
				"run", "--rm",
				"--env", "CNB_REGISTRY_AUTH={}",
				analyzeImage,
			}, analyzeAndExec(t, analyzerPath, "some-image", "ls -al /layers")...)...)
```

Not 100% sure if it's clearer... I could do away with passing in `t`.

testhelpers/binaries.go
yaelharel: Oh, interesting!
And what about `launcher_test.go`?

acceptance/analyzer_test.go
yaelharel: I think that we set the `WORKDIR` in the Dockerfile to be `/layers` and we should give it `cnb` permissions.
I'm doing it in the detector acceptance tests with the following lines:
```
ARG cnb_uid=1234
ARG cnb_gid=1000

RUN groupadd cnb --gid ${cnb_gid} && \
  useradd --uid ${cnb_uid} --gid ${cnb_gid} -m -s /bin/bash cnb

RUN chown -R "${cnb_uid}:${cnb_gid}" "/layers"

WORKDIR /layers
```

Does it make sense?

testhelpers/binaries.go
natalieparellano: I believe launcher_test.go is currently building the binaries directly into the container context. But I believe as discussed [here](https://github.com/buildpacks/lifecycle/pull/319#discussion_r445014704) we'll want to change the launcher tests to be consistent with what we're doing here.

testhelpers/binaries.go
natalieparellano: (see above)

testhelpers/testhelpers.go
natalieparellano: Given the precedent of `AssertNil` (and its current widespread usage) I'm going to leave `AssertNotNil` for now, but we can revisit if it becomes problematic.

acceptance/launcher_test.go
natalieparellano: Making this change, but then does it make sense to remove the `make clean` part? Otherwise, running the launcher tests will cause the detector/analyzer/etc. tests to all run slower.

acceptance/analyzer_test.go
natalieparellano: Added a test for this (group can write)

acceptance/analyzer_test.go
natalieparellano: Uggh I was probably doing something wrong - seeing different results now.

acceptance/analyzer_test.go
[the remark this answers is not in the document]
natalieparellano: Latest code will copy the cache fixture dir to a docker volume and pass that as `--volume` instead.

acceptance/analyzer_test.go
natalieparellano: I still see that foo.txt is ignored, but staging has the correct permissions. I am not sure what changed other than the cache directory is now being provided as a volume vs. bind mount. What is interesting is that in the case where the user's group can write, I see that staging is chowned:

```
/cache:
total 16
drwxrwxr-x 4 9999 3333 4096 Jun 29 20:23 .
drwxr-xr-x 1 root root 4096 Jun 29 20:23 ..
drwxrwxr-x 2 9999 3333 4096 Jun 29 20:21 committed
drwxr-xr-x 2 2222 3333 4096 Jun 29 20:23 staging
```

Adding a note to investigate further. There is some weirdness here that I'm not understanding.

acceptance/analyzer_test.go
natalieparellano: I am inclined to leave this as-is for now, given that the alternative implementation doesn't seem much clearer (at least, to me...). Also, some of the latest code has bash invocations *before* analyzer is run, such as `fmt.Sprintf("chown -R 9999:3333 /cache; chmod -R 775 /cache; %s -daemon -cache-dir /cache some-image; ls -alR /cache", analyzerPath)` and I don't see a helper function adding much improvement in those cases. Feel free to reopen the conversation @jromero @yaelharel if you disagree.

jromero: I personally feel like this should be important given the loop it took us through. That said, it would always be improved after the fact. I think my main pain point is that at first glance it looks like it's bash (which it is) that is being executed instead of analyzer and therefore obfuscating the execution of analyzer as a whole. I'm sure with some more effort and thought a nice abstraction/utility could be built.

acceptance/launcher_test.go
natalieparellano: I am going to hold off on making these changes for now - for launcher_test.go, let's wait for the Windows PR. For this PR, I am not sure if the `assertOutput` matcher makes sense, as it wraps the call to `cmd.CombinedOutput()` and we really only want to call that once. We have plenty of test cases that make multiple assertions on the output - such as here: https://github.com/buildpacks/lifecycle/pull/319/files#diff-72171b9647f1d25c895f3dcfa8e02b44R208-R214 (note `AssertNotMatch` mixed with `AssertMatch`) and I am not sure how to do that cleanly; it seems the effort required might not be justified by the gains in reduced duplication and/or clarity. @yaelharel let me know if you feel strongly about it or have another suggestion.

acceptance/analyzer_test.go
natalieparellano: @dfreilich I like "daemon case" - there is slight precedent for it elsewhere in the code and RFCs (see [here](https://github.com/buildpacks/lifecycle/blob/1ef63ec5114e90962c0a4ee2dc26a278e910d84d/metadata.go#L87) and [here](https://github.com/buildpacks/rfcs/blob/main/text/0026-lifecycle-all.md)). Do you feel strongly about it?

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,503 @@
...
+				})
+			})
+		})
+
+		when("CNB_REGISTRY_AUTH is not provided", func() {
+			when("DOCKER_CONFIG is set", func() {
+				it("succeeds", func() {
+					cmd := exec.Command(
+						"docker", "run", "--rm",
+						"--env", fmt.Sprintf("DOCKER_CONFIG=%s", registry.DockerDirectory),
```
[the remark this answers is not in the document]
natalieparellano: Ah, you're right. It seems that setting `DOCKER_CONFIG` to `foo` allows analyzer to run successfully without failing to stat `/root/.docker/config.json` - but as you said the registry is not actually accessible. I updated the test setup & execution to bind mount the docker config directory and also verify that a previous image was found on the registry.

acceptance/analyzer_test.go
natalieparellano: @jromero I'll have another go at it. What do you think about the presentation above?

natalieparellano: I'm not a huge fan of wrapping the call to `analyzer` in bash. I've done it only where necessary to get the contents of the layers or cache directories after `analyzer` has run. As an alternative, I've considered letting the container run, then copying the contents of the directory that I care about to the host and inspecting it there (this is how I am checking the contents of analyzed.toml: https://github.com/buildpacks/lifecycle/pull/319/files#diff-72171b9647f1d25c895f3dcfa8e02b44R126); this approach is a bit more overhead though. @ekcasey what do you think?

natalieparellano:
The latest code should address most of the review comments except the ones WRT clarity when running analyzer with bash.

There are also some TODOs around code organization & helpers.

testhelpers/binaries.go
yaelharel: I think it's worth changing the launcher as part of this commit, since you introduce these functions, but I'm also ok with doing it in a separate commit.

acceptance/analyzer_test.go
yaelharel: Looking at the code, it looks like keeping the container will be cleaner and more readable. My only concern is the performance but I guess that since the number of acceptance tests is not too high, the effect will be minor.

acceptance/launcher_test.go
yaelharel: I'm ok with keeping the code as is for now. We can revisit it later.

acceptance/analyzer_test.go
ekcasey: Copying files out of the container is probably _slightly_ slower (fwiw `bash` take longer to launch so i wouldn't say this definitively without experimenting) and adds complexity to the test code. But, in the wild the `analyzer` won't be invoked with `bash` so the test is more realistic if we don't wrap with `bash`. With some good test helpers I think either approach could produce readable tests.

Also, in the near future the `analyzer` will support windows and we probably want to test the lifecycle in an environment w/o `bash`.

For the above reasons I lean towards copying the files out.

Makefile
```diff
@@ -30,24 +30,27 @@ all: test build package
...
-build-linux-lifecycle: export GOOS:=linux
-build-linux-lifecycle: OUT_DIR:=$(BUILD_DIR)/$(GOOS)/lifecycle
-build-linux-lifecycle: GOENV:=GOARCH=$(GOARCH) CGO_ENABLED=1
-build-linux-lifecycle: DOCKER_RUN=docker run --workdir=/lifecycle -v $(OUT_DIR):/out -v $(PWD):/lifecycle $(COMPILATION_IMAGE)
-build-linux-lifecycle:
+$(BUILD_DIR)/linux/lifecycle/lifecycle: export GOOS:=linux
+$(BUILD_DIR)/linux/lifecycle/lifecycle: OUT_DIR:=$(BUILD_DIR)/$(GOOS)/lifecycle
+$(BUILD_DIR)/linux/lifecycle/lifecycle: GOENV:=GOARCH=$(GOARCH) CGO_ENABLED=1
+$(BUILD_DIR)/linux/lifecycle/lifecycle: DOCKER_RUN=docker run --workdir=/lifecycle -v $(OUT_DIR):/out -v $(PWD):/lifecycle $(COMPILATION_IMAGE)
+$(BUILD_DIR)/linux/lifecycle/lifecycle:
```
ekcasey, line 37:
```suggestion
GOFILES := $(wildcard *.go */*.go */*/*.go)
$(BUILD_DIR)/linux/lifecycle/lifecycle: export GOOS:=linux
$(BUILD_DIR)/linux/lifecycle/lifecycle: OUT_DIR:=$(BUILD_DIR)/$(GOOS)/lifecycle
$(BUILD_DIR)/linux/lifecycle/lifecycle: GOENV:=GOARCH=$(GOARCH) CGO_ENABLED=1
$(BUILD_DIR)/linux/lifecycle/lifecycle: DOCKER_RUN=docker run --workdir=/lifecycle -v $(OUT_DIR):/out -v $(PWD):/lifecycle $(COMPILATION_IMAGE)
$(BUILD_DIR)/linux/lifecycle/lifecycle: $(GOFILES)
```
To make these target rebuild when the source files change you need to do something similar to this ^.

There might be a way to make the wildcard work for an arbitrary depth. Alternatively some `go list -json` + `jq` magic could also get you the source file list as well.

ekcasey: This could be improved to do things like exclude test files. Also, given that we want a very limited set of imports for launcher we could scope the source file list down for the launcher build targets (or we could just be lazy for now and use the same list).

acceptance/analyzer_test.go
natalieparellano: I want to say that the latest changes should address this concern. Check out `h.DockerRun` and `h.DockerRunAndCopy`. Note also that some tests still use `WithBash()` - these are the ones that deal with Unix file permissions and will possibly be skipped on Windows. Thanks @jromero for the `WithFlags()` idea that I stole from pack :)

natalieparellano:
This is odd - I am seeing "1 successful check" which is the DCO, but it doesn't look like the tests ran on Linux or Windows. Looking at https://github.com/buildpacks/lifecycle/pull/327, I see the tests running two minutes ago. Is there something uniquely wrong with my PR?

dfreilich:
@natalieparellano It looks like it doesn't run checks when there are merge conflicts. The same thing is happening in https://github.com/buildpacks/pack/pull/716

acceptance/analyzer_test.go
natalieparellano: Resolving this convo, but feel free to un-resolve or provide other feedback on the new approach.

acceptance/analyzer_test.go
natalieparellano: I am going to keep "daemon case" for now.

Makefile
natalieparellano: The latest changes use `wilcard` but the above mentioned concerns (arbitrary depth, excluding test files, improving launcher) remain (see discussion in [Slack](https://buildpacks.slack.com/archives/CSFE49VGR/p1594309841165100)). Thanks @jromero for your suggestions, maybe we can yet get `go list` to do our bidding.

natalieparellano: Thanks to @jromero this is working now with `go list`. I am going to be lazy for now and say that we should punt on optimizing the launcher.

acceptance/analyzer_test.go
natalieparellano: Looking at the code, `if cacheDir != ""` we always call `cache.NewVolumeCache(cacheDir)` which always recreates the staging directory (hence `foo.txt` was ignored). This happens after we drop privileges, so it makes sense that the uid & gid are 2222 & 3333. I have no idea why it was showing as owned by root before :/

natalieparellano:
I think this is ready for another round of review.

natalieparellano:
Uggggh nevermind, I missed that this needs to work for Windows as well (not skip). Updating...

natalieparellano:
The Windows tests are timing out trying to create the registry. It's really ODD because the registry is successfully created in the imgutil tests, and the code is nearly identical: 
  * imgutil: https://github.com/buildpacks/imgutil/blob/2281cd9b141467576e74acc76d80bddf276189de/remote/remote_test.go#L34-L40
  * analyzer acceptance: https://github.com/buildpacks/lifecycle/blob/5f7c8bc15561114dc79d7165b92e80d72acef4a2/acceptance/analyzer_test.go#L48-L54

natalieparellano:
@ekcasey I think this is ready for another round of review.

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,599 @@
...
+
+					h.AssertPathExists(t, filepath.Join(tempDir, "layers", "some-buildpack-id", "some-layer.sha"))
+					h.AssertPathExists(t, filepath.Join(tempDir, "layers", "some-buildpack-id", "some-layer.toml"))
+				})
+			})
+		})
+	})
+}
+
+func flattenMetadata(t *testing.T, path string, metadataStruct interface{}) string {
```
ekcasey, line 565: nit: maybe we can call this `minifyMetadata` instead of `flattenMetadata`? To me flatten implies something different (flattening nested arrays).

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,599 @@
...
+	when("group path is provided", func() {
+		it("uses the provided group path", func() {
+			cacheVolume := h.SeedDockerVolume(t, cacheFixtureDir)
+			defer h.DockerVolumeRemove(t, cacheVolume)
+
+			_, tempDir := h.DockerRunAndCopy(t,
+				analyzeImage,
+				"/layers",
+				h.WithFlags(
+					"--env", "CNB_REGISTRY_AUTH={}",
```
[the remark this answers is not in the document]
natalieparellano: Opened an issue: https://github.com/buildpacks/lifecycle/issues/339

acceptance/analyzer_test.go
```diff
@@ -0,0 +1,599 @@
...
+						filepath.Join("testdata", "analyzer", "cache-image"),
+					)
+					h.Run(t, cmd)
+				})
+
+				it.After(func() {
+					h.DockerImageRemove(t, cacheImage)
+				})
+
+				it("ignores the cache", func() {
```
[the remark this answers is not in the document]
natalieparellano: Clarified the test description (cache image *in a daemon*) and added a new case for cache image in a registry.

acceptance/variables/variables.go
```diff
@@ -0,0 +1,14 @@
+package variables
```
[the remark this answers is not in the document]
natalieparellano: Good question - I was copying what we're doing for pack: https://github.com/buildpacks/pack/tree/main/acceptance/variables . We can revisit, but I propose we table that deliberation for when we introduce Windows variables (the motivation behind this package).

testhelpers/testhelpers.go
```diff
@@ -293,6 +257,22 @@ func RecursiveCopy(t *testing.T, src, dst string) {
...
+	dstFile, err := os.Create(destFileName)
+	AssertNil(t, err)
+	defer dstFile.Close()
+	_, err = io.Copy(dstFile, srcFile)
+	AssertNil(t, err)
+	modifiedtime := time.Time{}
+	err = os.Chtimes(destFileName, modifiedtime, modifiedtime)
+	AssertNil(t, err)
+	err = os.Chmod(destFileName, 0664)
+	AssertNil(t, err)
```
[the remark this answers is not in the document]
natalieparellano: Ah - this logic existed before, I am just re-using the existing helper (`RecursiveCopy`). But you are right, this chmod then makes necessary the extra chmod that happens [just after](https://github.com/buildpacks/lifecycle/pull/319/files#diff-4c39693bf568afeb86919fe3c6ad2fdfR83). It's a little messy... I am open to re-working this.

testhelpers/testhelpers.go
```diff
@@ -293,6 +257,22 @@ func RecursiveCopy(t *testing.T, src, dst string) {
...
+	defer srcFile.Close()
+	dstFile, err := os.Create(destFileName)
+	AssertNil(t, err)
+	defer dstFile.Close()
+	_, err = io.Copy(dstFile, srcFile)
+	AssertNil(t, err)
+	modifiedtime := time.Time{}
+	err = os.Chtimes(destFileName, modifiedtime, modifiedtime)
+	AssertNil(t, err)
```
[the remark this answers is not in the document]
natalieparellano: See comment [here](https://github.com/buildpacks/lifecycle/pull/319#discussion_r457525894)

testhelpers/docker.go
```diff
@@ -0,0 +1,162 @@
...
+		AssertNil(t, dockerCliErr)
+	})
+	return dockerCliVal
+}
+
+func DockerBuild(t *testing.T, name, context string, ops ...DockerCmdOp) {
+	t.Helper()
+	args := formatArgs([]string{"-t", name, context}, ops...)
+	Run(t, exec.Command("docker", append([]string{"build"}, args...)...))
+}
```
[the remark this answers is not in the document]
natalieparellano: I agree! I added a TODO for it in the code. Hopefully it will become consistent over time.

testhelpers/binaries.go
```diff
@@ -0,0 +1,117 @@
...
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+func MakeAndCopyLauncher(t *testing.T, destDir string) {
+	buildDir, err := filepath.Abs(filepath.Join("..", "out"))
+	AssertNil(t, err)
+	AssertNil(t, os.MkdirAll(buildDir, 0755))
```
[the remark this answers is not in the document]
natalieparellano: I think this is an artifact of me trying to fix Windows failures. Taking it out...

testhelpers/binaries.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+func MakeAndCopyLauncher(t *testing.T, destDir string) {
```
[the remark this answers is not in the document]
natalieparellano: Changing it to be consistent. FWIW the reason this didn't take a `goos` is because in practice when we're testing the launcher it's always either Linux or Windows, so `runtime.GOOS` + some logic to convert darwin -> linux is sufficient. For the lifecycle, because we test the version on darwin the function needed more flexibility. What I have now is explicitly passing daemon OS as the goos for the launcher.

testhelpers/testhelpers.go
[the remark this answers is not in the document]
ekcasey: If it was already there I am happy to leave the fix for later.

PR #327, opened 2020-07-01
Implement process type specific env vars

jabrown85:
Include process type specific paths when processing env vars during launch.

Spec: https://github.com/buildpacks/spec/pull/98

### DIFF
--- a/launch/launcher.go
+++ b/launch/launcher.go
@@ -53,7 +53,7 @@ func (l *Launcher) Launch(self string, cmd []string) error {
 		}
 		return nil
 	}
-	profileCmds, err := l.profileD()
+	profileCmds, err := l.profileD(process)
 	if err != nil {
 		return errors.Wrap(err, "determine profile")
 	}
@@ -120,7 +120,7 @@ func (l *Launcher) env(process Process) error {
 	})
 }
 
-func (l *Launcher) profileD() ([]string, error) {
+func (l *Launcher) profileD(process Process) ([]string, error) {
 	var out []string
 
 	appendIfFile := func(path string) error {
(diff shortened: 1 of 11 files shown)


launch/launcher.go
```diff
@@ -106,7 +106,18 @@ func (l *Launcher) env() error {
 			if err := l.Env.AddEnvDir(filepath.Join(path, "env")); err != nil {
 				return err
 			}
-			return l.Env.AddEnvDir(filepath.Join(path, "env.launch"))
+			if process.Type != "" {
+				if err := l.Env.AddEnvDir(filepath.Join(path, "env", process.Type)); err != nil {
+					return err
+				}
+			}
```
ekcasey, line 113: Based on my reading of the RFC, I thought we were only accepting `env.launch/<process>/` and not accepting `env/<process>`

jabrown85: I don't see an issue with removing `env/<process>`. cc @jkutner is that your understanding as well? If so, I'll move things around.

jabrown85: As far as implementing `profile` scripts, I'll wait for that same question to be answered and follow the same decision. This PR will aim to match that particular spec PR.

jabrown85: I actually went ahead and made the change specific to `env.launch/<process>`. I think the RFC is a bit ambiguous but having `env/<process>` would only add confusion without making the feature much easier to use.

jabrown85:
@ekcasey I went ahead and pushed the `profile.d/:process` changes as discussed. Please re-review at your convenience!

jabrown85:
> Looks good! I think we should add a process-specific profile script to the acceptance tests before merging though

I was afraid you'd say that 😅  - I'll see about adding one for the new process specific path and old path. There doesn't seem to be any acceptance tests for `profile.d`.

jabrown85:
@ekcasey process specific profile.d script acceptance test is up.

Issue #312, opened 2020-06-11
Option to turn off colors in console output


v-srinivasan-sfdc:
While posting the build logs in emails or any other HTML based output, the logs looks like this.
```
�[33;1mWarning: �[0mNot restoring or caching layer data, no cache flag specified.
```
It will be great to have a `--no-color` option similar to `pack`

ekcasey:
We were going to make the default color behavior depend on the existence of a `tty` (once `pack` actually runs the lifecycle containers with a `tty`) but an explicit flag is a good idea regardless.

https://github.com/buildpacks/pack/issues/493

ekcasey:
Resolved by #328

PR #328, opened 2020-07-02
Add -no-color flag

jabrown85:
Issue: #312

I wasn't entirely sure how best to test this. I am happy to add one anywhere someone suggests tho!

### DIFF
--- a/cmd/command.go
+++ b/cmd/command.go
@@ -4,6 +4,8 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+
+	"github.com/heroku/color"
 )
 
 type Command interface {
@@ -17,11 +19,13 @@ func Run(c Command, asSubcommand bool) {
 	var (
 		printVersion bool
 		logLevel     string
+		noColor      bool
 	)
 
 	log.SetOutput(ioutil.Discard)
 	FlagVersion(&printVersion)
 	FlagLogLevel(&logLevel)
+	FlagNoColor(&noColor)
 	c.Init()
 	if asSubcommand {
 		if err := flagSet.Parse(os.Args[2:]); err != nil {
@@ -34,6 +38,7 @@ func Run(c Command, asSubcommand bool) {
 			Exit(err)
 		}
 	}
+	color.Disable(noColor)
 
 	if printVersion {
 		ExitWithVersion()
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -48,6 +48,7 @@ const (
 	EnvProcessType         = "CNB_PROCESS_TYPE"
 	EnvLogLevel            = "CNB_LOG_LEVEL"
 	EnvProjectMetadataPath = "CNB_PROJECT_METADATA_PATH"
+	EnvNoColor             = "CNB_NO_COLOR" // defaults to false
 )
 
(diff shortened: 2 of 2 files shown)


jabrown85:
I realize `pack` also strips colored output..is that something we want lifecycle to do as well? This PR gets rid of all the colored output that lifecycle emits, but it doesn't do anything with the output by the buildpacks themselves.

ekcasey:
I don't think `pack` _successfully_ strips color?
Screen Shot 2020-07-02 at 12 58 44 PM

cmd/command.go
```diff
@@ -17,11 +19,14 @@ func Run(c Command, asSubcommand bool) {
...
 		printVersion bool
 		logLevel     string
+		noColor      bool
 	)
 
 	log.SetOutput(ioutil.Discard)
 	FlagVersion(&printVersion)
 	FlagLogLevel(&logLevel)
+	FlagNoColor(&noColor)
+	color.Disable(noColor)
```
ekcasey, line 29: This need to happen after `flagSet.Parse`, `noColor` won't be set correctly at this point in the command execution.

cmd/flags.go
```diff
@@ -156,6 +156,10 @@ func FlagProcessType(processType *string) {
 	flagSet.StringVar(processType, "process-type", os.Getenv(EnvProcessType), "default process type")
 }
 
+func FlagNoColor(skip *bool) {
+	flagSet.BoolVar(skip, "no-color", false, "disable color output")
```
ekcasey, line 160: We should add an environment variable `CNB_NO_COLOR` and use it to set the default here. This is especially for things that might not be specified in the platform API, as it allows platforms to optimistically set it w/o worrying that it will cause a failure if it is unsupported by a given lifecycle.

cmd/command.go
jabrown85: ah, I moved it right after my last smoke test. Good catch.

jabrown85:
Added env support and fixed the ordering. Things look good as far as lifecycle emitted colors go.

image

I think we should handle stripping colors later, if it is even reasonable to do.

jromero:
> I don't think `pack` _successfully_ strips color?
> Screen Shot 2020-07-02 at 12 58 44 PM
> 
> If we could strip the buildpack color that would be a big win. Folks have been complaining about this for some time ([buildpacks/pack#123](https://github.com/buildpacks/pack/issues/123)) and while `pack` may have (temporarily?) fixed it, this should be supported in all platforms.

`pack` did recently ["fix"](https://github.com/buildpacks/pack/pull/696) this, to be released in 0.12.0, but having the option and functionality in lifecycle would be a huge benefit.

jromero:
@ekcasey out of curiosity is this a "platform API" change? I would be curious to know, as a platform that could support other lifecycle implementations, how we would know if this option is available if not in the API.

ekcasey:
@jromero that's a good question. I don't think we want to specify this flag I don't think a compliant lifecycle should have to have color.

I want to stake out a new philosophical principal. When using unspecified features a platform _should_ use the environment variables rather than the flags to provide inputs. There is no harm in setting an env var if the lifecycle doesn't support the feature. `pack` can just optimistically provide the env var.

PR #331, opened 2020-07-03
Bump imgutil

micahyoung:
- Uses newer Windows layer writer with validation and permissions improvements


micahyoung:
We intentionally left this out of the main Windows image work as it's an enhancement, not a blocking requirement. It got merged into imgutil just before the Windows lifecycle work was merged and we didn't want it to hold up the latter.

PR #335, opened 2020-07-09
Fix handling of symlinks in slices

ekcasey:
Fixes the following bugs with slices:

* Add the link to the slice instead of the resolved target
* Slice layers should not include files globbing through a symlink
* All entries within app dir should have normalized uid/gid

Fixing these bugs without doing terrible hacky things required a major refactor of the slice logic. This will require some effort to merge with `main`. However the code architecture in this PR is design to play nice with the windows layer writer logic.


layers/dir.go
```diff
@@ -0,0 +1,61 @@
...
+			TarPath: tarPath,
+			Digest:  sha,
+		}, nil
+	}
+	lw, err := newFileLayerWriter(tarPath)
+	if err != nil {
+		return Layer{}, err
+	}
+	defer func() {
+		err = lw.Close()
```
sclevine, line 36: 1. Name `err` return param
2. Check if nil before setting

layers/slices.go
```diff
@@ -0,0 +1,242 @@
...
+	})
+
+	tarPath := filepath.Join(f.ArtifactsDir, layerID+".tar")
+
+	lw, err := newFileLayerWriter(tarPath)
+	if err != nil {
+		return Layer{}, err
+	}
+	defer func() {
+		err = lw.Close()
```
sclevine, line 80: Check if nil before setting

layers/dir.go
```diff
@@ -0,0 +1,63 @@
...
+	}()
+	tw := archive.NewNormalizingTarWriter(tar.NewWriter(lw))
+	parents, err := parents(dir)
+	if err != nil {
+		return Layer{}, err
+	}
+	if err := archive.WriteFilesToArchive(tw, parents); err != nil {
+		return Layer{}, err
+	}
+	tw.WithUID(f.UID)
```
jabrown85, line 48: Can/Should the logic of writing parent directories without `uid`/`gid` be encapsulated in `NormalizingTarWriter`? I can see that `addParentDirs` had the same behavior.

layers/slices.go
```diff
@@ -0,0 +1,244 @@
...
+		return len(strings.Split(parentToCheck[i], string(os.PathSeparator))) > len(strings.Split(parentToCheck[j], string(os.PathSeparator)))
+	})
+
+	// if all children are slices, mark the dir as sliced
+	for _, dir := range parentToCheck {
+		allChildrenAdded := true
+		if children, ok := sd.subDirs[dir]; ok {
+			for _, child := range children {
+				if added, ok := sd.slicedFiles[child]; ok && !added {
+					allChildrenAdded = false
```
jabrown85, line 208: Should we stop iterating files at this point?

archive/tar.go
```diff
@@ -2,222 +2,71 @@ package archive
...
+		}
+	}
+	header, err := tar.FileInfoHeader(fi, target)
 	if err != nil {
 		return err
 	}
+	header.Name = path
 
-	return filepath.Walk(srcDir, func(file string, fi os.FileInfo, err error) error {
-		if _, ok := fileSet[file]; ok {
```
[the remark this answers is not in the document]
ekcasey: Yes good catch, I'll clean this up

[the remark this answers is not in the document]
ekcasey: The tracking of added files is done in the LayerFactory now, so the `archive` package can be more naive. This was necessary b/c we needed to track seen files across layers, not just within a layer.

layers/dir.go
ekcasey: I wanted to keep the rules like "don't change perms for dir parents" in the `layers` package so that the `NomalizingTarWriter` and the entire `archive` package can just include generic utilities. Encapsulating this in `NormalizingTarWriter` would require the `NormalizingTarWriter` to be context aware (it needs to know whether a header is a parent dir) and make it less simple and less reusable across contents.

layers/slices.go
ekcasey: yes, good catch

PR #340, opened 2020-07-20
Merge slice fixes into main

ekcasey:
This PR includes slices fixes for symlinks from 0.8.1 (https://github.com/buildpacks/lifecycle/pull/335).  Additional work was required to fix symlink handling in windows. The archive package now adds PAX records to windows archives headers so that we can properly make directory or file type symlinks even when the link target does not exist (for context: https://github.com/golang/go/issues/39183).

This PR runs the windows unit and acceptance tests in GHA without the `docker-run-windows` wrapper. That wrapper removes permissions from test fixtures that are required to properly test handling of directory symlinks.


archive/writer.go
```diff
@@ -0,0 +1,50 @@
...
+		hdr.Uid = uid
+		return hdr
+	})
+}
+
+func (tw *NormalizingTarWriter) WithGID(gid int) {
+	tw.headerOpts = append(tw.headerOpts, func(hdr *tar.Header) *tar.Header {
+		hdr.Gid = gid
+		return hdr
+	})
```
micahyoung, line 34: I wonder if eventually we might want to unify (or move and duplicate) this related imgutil [WindowsWriter logic](https://github.com/buildpacks/imgutil/blob/2281cd9b141467576e74acc76d80bddf276189de/layer/windows_writer.go#L54-L59) to these `WithGID` and `WithUID` methods. 

Definitely doesn't need to happen right now because the `WindowsWriter` just looks at the `hdr.Gid/Uid` and applies generic owners/groups instead of specific users via PAXRecords. But for feature parity, we may eventually need a more robust implementation where `WithUID` and `WithGID` can take specific SIDs and apply them Windows Owner and Primary group permissions. 

Nothing to do now, just something we can keep in mind.

archive/tar_windows.go
```diff
@@ -1,6 +1,52 @@
...
+	name, err := syscall.UTF16PtrFromString(hdr.Name)
+	if err != nil {
+		return err
+	}
+	target, err := syscall.UTF16PtrFromString(hdr.Linkname)
+	if err != nil {
+		return err
+	}
+	return syscall.CreateSymbolicLink(name, target, flags)
+}
```
micahyoung, line 46: It's a bummer to have to do all this to work around the divergent Windows/POSIX golang behavior. After reading that thread it sounds like a `wontfix` so `os.Symlink` won't change anytime soon. This implementation feels like it follows [one of the suggestions](https://github.com/golang/go/issues/39183#issuecomment-634251887) there:
> For the few programs that actually do need to create symlinks to nonexistent targets, windows.CreateSymbolicLink seems sufficient, and it's easy enough for users to write their own portable wrappers using +build constraints.

Makefile
```diff
@@ -187,7 +187,7 @@ docker-run-windows: docker-build-source-image-windows
 docker-run-windows:
 	@echo "> Running '$(DOCKER_CMD)' in docker windows..."
 	@docker volume rm -f lifecycle-out
-	docker run -v lifecycle-out:c:/lifecycle/out -e LIFECYCLE_VERSION -e PLATFORM_API -e BUILDPACK_API -v gopathcache:c:/gopath -v '\\.\pipe\docker_engine:\\.\pipe\docker_engine' --isolation=process --rm $(SOURCE_COMPILATION_IMAGE) $(DOCKER_CMD)
+	@docker run -v lifecycle-out:c:/lifecycle/out -e LIFECYCLE_VERSION -e PLATFORM_API -e BUILDPACK_API -v gopathcache:c:/gopath -v '\\.\pipe\docker_engine:\\.\pipe\docker_engine' --isolation=process --rm $(SOURCE_COMPILATION_IMAGE) $(DOCKER_CMD)
```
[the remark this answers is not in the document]
ekcasey: I am okay leaving it echo-y

layers/extract.go
```diff
@@ -0,0 +1,29 @@
...
+	tr.Strip("Hives")
+	tr.Strip("Files")
+	if runtime.GOOS == "windows" {
+		tr.ExcludePaths([]string{"Hives"})
+		tr.Strip(`Files`)
+		if dest == "" {
+			dest = "c:"
+		}
+	}
+	tr.PrependDir(dest)
```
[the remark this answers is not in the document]
ekcasey: Good question. I was thinking some of these helpers could probably move to `imgutil` at a later point but I wanted to incubate them here for convenience. Maybe when we move them over we can move a bunch of layer and archive utilities, not just the windows related ones.

archive/writer.go
ekcasey: Agree, that makes a lot of sense. In the lifecycle we are always creating windows layers while running in a windows environment. That is not true for `pack` so we will have to be thoughtful about that when we move them over as a generic utility.

testhelpers/testhelpers.go
```diff
@@ -285,6 +262,14 @@ func RecursiveCopy(t *testing.T, src, dst string) {
 			AssertNil(t, err)
 			RecursiveCopy(t, filepath.Join(src, fi.Name()), filepath.Join(dst, fi.Name()))
 		}
+		if fi.Mode()&os.ModeSymlink != 0 {
```
natalieparellano, line 265: Now that we have this code in `RecursiveCopy` I wonder if this code is now redundant and could be removed: https://github.com/buildpacks/lifecycle/blob/63c5c6e6a2f5c6617781ed6d42a80b1c3918730b/testhelpers/binaries.go#L94-L105

layers/layers_windows_test.go
```diff
@@ -0,0 +1,45 @@
...
+	)))
+}
+
+func assertOSSpecificFields(t *testing.T, expected *tar.Header, hdr *tar.Header) {
+	t.Helper()
+	h.AssertEq(t, hdr.Format, tar.FormatPAX)
+}
+
+func assertOSSpecificEntries(t *testing.T, tr *tar.Reader) {
+	for _, windowsEntry := range []string{"Files", "Hives"} {
```
natalieparellano, line 36: This seems to be checking that the first two entries are of type directory, but don't we want to add some sort of assertion around `header.Name == windowsEntry`?

layers/extract.go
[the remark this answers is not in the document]
micahyoung: Perfect!

layers/dir_test.go
```diff
@@ -0,0 +1,192 @@
...
+		it.Before(func() {
+			var err error
+			dirLayer, err = factory.DirLayer("some-layer-id", dir)
+			h.AssertNil(t, err)
+		})
+
+		it("creates a layer from the directory", func() {
+			// parent layers should have uid/gid matching the filesystem
+			// the dir and it's children should have normalized uid/gid
+			assertTarEntries(t, dirLayer.TarPath, append(parents(t, dir), []*tar.Header{
```
natalieparellano, line 62: I might be missing something but it doesn't seem that we test the permissions of the parent directory anywhere, no? Is this possible?

layers/layers_windows_test.go
ekcasey: Good point!

layers/dir_test.go
ekcasey: The expected parent headers have UID/GID set on them but the assertions are missing (!) and we don't do this for windows headers. I will clean this up.

https://github.com/buildpacks/lifecycle/blob/merge-slice-fix/layers/layers_unix_test.go#L18-L20

ekcasey: For now I am going to only test UID/GID for linux layers.

We currently aren't setting the SID on windows layer entries properly (looks like setting `header.Uid` doesn't do the trick) but that was true before this PR and I would like to fix it separately. The SID should be set to the value of `CNB_USER_ID` for windows layers.

Issue #306, opened 2020-05-29
Emit a special exit code when detect phase has no groups opt-in


squee1945:
If none of the groups are ultimately selected in the detect phase, the exit code is `CodeFailedDetect` (6). This is indistinguishable from an unexpected error during detect.

Please introduce a new exit code for this special non-error case where no groups have matched during detection.

ekcasey:
Resolved by #346

PR #346, opened 2020-07-28
feat: use more specific codes during detect

jabrown85:
Platforms should have the ability to tell the difference between a detector failure because of no passing buildpacks and a detector failure because of no passing buildpacks that _had errors_ during detect.

Detector will now return a `100` or `101` depending on how detection failed.

[Issue #306](https://github.com/buildpacks/lifecycle/issues/306)
[Updatd Spec](https://github.com/buildpacks/spec/blob/platform/0.4/platform.md#outputs)

### DIFF
--- a/cmd/lifecycle/detector.go
+++ b/cmd/lifecycle/detector.go
@@ -79,11 +79,15 @@ func (da detectArgs) detect() (lifecycle.BuildpackGroup, lifecycle.BuildPlan, er
 		Logger:        cmd.Logger,
 	})
 	if err != nil {
-		if err == lifecycle.ErrFail {
+		if err == lifecycle.ErrFail || err == lifecycle.ErrFailWithErrors {
 			cmd.Logger.Error("No buildpack groups passed detection.")
 			cmd.Logger.Error("Please check that you are running against the correct path.")
 		}
-		return lifecycle.BuildpackGroup{}, lifecycle.BuildPlan{}, cmd.FailErrCode(err, cmd.CodeFailedDetect, "detect")
+		exitCode := cmd.CodeFailedDetect
+		if err == lifecycle.ErrFailWithErrors {
+			exitCode = cmd.CodeFailedDetectWithErrors
+		}
+		return lifecycle.BuildpackGroup{}, lifecycle.BuildPlan{}, cmd.FailErrCode(err, exitCode, "detect")
 	}
 
 	return group, plan, nil
--- a/cmd/lifecycle/detector.go
+++ b/cmd/lifecycle/detector.go
@@ -79,15 +79,17 @@ func (da detectArgs) detect() (lifecycle.BuildpackGroup, lifecycle.BuildPlan, er
 		Logger:        cmd.Logger,
 	})
 	if err != nil {
-		if err == lifecycle.ErrFailedDetection || err == lifecycle.ErrBuildpack {
+		switch err {
+		case lifecycle.ErrFailedDetection:
 			cmd.Logger.Error("No buildpack groups passed detection.")
 			cmd.Logger.Error("Please check that you are running against the correct path.")
+			return lifecycle.BuildpackGroup{}, lifecycle.BuildPlan{}, cmd.FailErrCode(err, cmd.CodeFailedDetect, "detect")
+		case lifecycle.ErrBuildpack:
+			cmd.Logger.Error("No buildpack groups passed detection.")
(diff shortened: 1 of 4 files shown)


jabrown85:
Looking for early feedback as well as guidance on if we should make the other exit code changes in this PR or not. Since we aren't current documenting exit codes from detector, I think this is OK to ship. But we might like to have a big-bang of exit code changes instead.

detector.go
```diff
@@ -301,10 +308,14 @@ func (bo BuildpackOrder) Detect(c *DetectConfig) (BuildpackGroup, BuildPlan, err
 
 func (bo BuildpackOrder) detect(done, next []Buildpack, optional bool, wg *sync.WaitGroup, c *DetectConfig) ([]Buildpack, []BuildPlanEntry, error) {
 	ngroup := BuildpackGroup{Group: next}
+	hadErr := false
 	for _, group := range bo {
 		// FIXME: double-check slice safety here
 		found, plan, err := group.append(ngroup).detect(done, wg, c)
-		if err == ErrFail {
+		if err == ErrFailWithErrors {
```
jabrown85, line 315: I did not like having `ErrFailWithErrors` in so many places. I wonder if making a single `ErrFail ` with a flag for `DetectErr` or something would be cleaner.

detector_test.go
```diff
@@ -163,6 +164,39 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
+			}.Detect(config)
+			if err != lifecycle.ErrFailWithErrors {
+				t.Fatalf("Unexpected error:\n%s\n", err)
+			}
+
+			if runtime.GOOS == "windows" {
+				if s := allLogs(logHandler); !strings.HasSuffix(s,
+					"======== Results ========\n"+
+						"pass: A@v1\n"+
+						"err:  B@v1 (4294967295)\n",
```
ekcasey, line 185: Is this a problem with the windows test buildpack? Shouldn't it select it's exit code from `detect-status-B-v1` and match the linux case?

detector.go
```diff
@@ -21,6 +21,7 @@ const (
 )
 
 var ErrFail = errors.New("no buildpacks participating")
+var ErrFailWithErrors = errors.New("no buildpacks participating: failed with errors")
```
ekcasey, line 24: Maybe `ErrBuildpack` instead of `ErrFailWithErrors` to make clear that the error originated from the Buildpack?

I know this was already here but I also think a less generic name like`ErrFailedDetection` might be clearer than `ErrorFail` given that this variable lives in a package with many other phases.

cmd/lifecycle/detector.go
```diff
@@ -79,11 +79,15 @@ func (da detectArgs) detect() (lifecycle.BuildpackGroup, lifecycle.BuildPlan, er
 		Logger:        cmd.Logger,
 	})
 	if err != nil {
-		if err == lifecycle.ErrFail {
+		if err == lifecycle.ErrFail || err == lifecycle.ErrFailWithErrors {
 			cmd.Logger.Error("No buildpack groups passed detection.")
 			cmd.Logger.Error("Please check that you are running against the correct path.")
 		}
-		return lifecycle.BuildpackGroup{}, lifecycle.BuildPlan{}, cmd.FailErrCode(err, cmd.CodeFailedDetect, "detect")
+		exitCode := cmd.CodeFailedDetect
```
ekcasey, line 86: If we do this we are using `CodeFailedDetect` for more than just ("sucessfully" failed detection). The lifecycle itself can throw errors. I think we need a third error code to separate these from "nothing detected".

I think we should do a switch on `err` and have a `default` in addition to `ErrFail` and `ErrFailWIthErrors` cases

detector.go
ekcasey: I don't have strong opinions on this one.

Someday I could see us having an umbrella `lifecycle.Error` type with a field to denote the subspecies of error. But maybe it's unnecessary when there are only two?

jabrown85: Sounds fine to me! I like the idea of the umbrella error type in the future.

detector_test.go
jabrown85: Pretty sure this was because I was returning `-1`, which invalid. I should have caught it turning to `255` on mac, but I think the positive error code should be defined behavior and now match.

jabrown85:
New failure case (err from buildpack) exits `101`. New message is output and I removed the "Please check path" message..but I'm open for w/e on messaging.
image

Existing still exits with `100` and retains the same messaging.
image

PR #347, opened 2020-07-28
Multi api lifecycle descriptor

### LINKED ISSUE #344: Multi-API lifecycle descriptor
Implement github.com/buildpacks/rfcs/pull/92

ekcasey:
Resolves #344 

Todo

- [x] Compile with multiple supported platform and buildpack APIs (hardcoded in `api` package b/c it cannot change on the fly)
- [x] New descriptor schema
- [x] Make platform API available to the launcher (`CNB_PLATFORM_API` is set on app image)
- [x] Make buildpack APIs available to all phases that may need it (set in `group.toml` and `metadata.toml`)
- [x] New labels on lifecycle image (introduced new helper script to build image: `tools/image/main.go`)
- [x] Platform API validations/deprecation
- [x] Buildpack API validations/deprecation
- [x] Remove `CNB_PLATFORM_API` from process launch env
- ~[ ] Contributing docs explaining descriptor/images and release branching process~ This is probably easier as a separate PR


cmd/lifecycle/main.go
```diff
@@ -26,11 +27,11 @@ func main() {
 	case "builder":
 		cmd.Run(&buildCmd{}, false)
 	case "exporter":
-		cmd.Run(&exportCmd{}, false)
+		cmd.Run(&exportCmd{exportArgs: exportArgs{platformAPI: platformAPI}}, false)
```
jabrown85, line 30: Do the other stages not need `platformAPI`?

detector.go
```diff
@@ -35,7 +35,7 @@ type BuildPlanEntry struct {
 func (be BuildPlanEntry) noOpt() BuildPlanEntry {
 	var out []Buildpack
 	for _, p := range be.Providers {
-		out = append(out, p.noOpt())
+		out = append(out, p.noOpt().noAPI())
```
jabrown85, line 38: I've seen `noAPI` a few times now in this PR. I think I understand this is wiping it out so it won't get output to disk somewhere. Would it make sense to have a buildplan.buildpack writer that ignores that key or something?

cmd/lifecycle/main.go
ekcasey: They all will eventually, but I figured I would wait to inject into each stage when it is needed.

detector.go
ekcasey: Because `noOpt()` is doing a similar thing for a similar reasons (in some places, like the one above) I figured I would follow the existing pattern. I tried a different approach but things got gnarly fast and I determined it would required a larger refactor (I didn't want to make this already large PR larger)

Issue #283, opened 2020-04-03
Make buildpack detection errors info level


jabrown85:
### Summary
When `pack build` is executed no errors are reported in cases where the buildpack's detect phase produced an error. This can occur for many reasons but the most common case is running out of disk space. In this case ALL buildpacks fail with an unexpected error but the error is not displayed via `pack` without  the `--verbose` flag.

---

##### Steps

1. Fill up your docker vm disk space
2. Run `pack build my-image`
3. Observe the build fail with a no detection error with little details on why

##### Current behavior

```
$ pack build riffsquare --builder projectriff/builder:0.6.0-snapshot -e RIFF=1
0.6.0-snapshot: Pulling from projectriff/builder
Digest: sha256:9349cc12f5944982dfe41976acd734a411d87db158ed3c0e818fa3b1fd87a6fb
Status: Image is up to date for projectriff/builder:0.6.0-snapshot
base-cnb: Pulling from cloudfoundry/run
Digest: sha256:c1e5e1fdd95465525df6ecdd99be6f47d37a1e256a22c53287bad08aead92f3f
Status: Image is up to date for cloudfoundry/run:base-cnb
===> DETECTING
[detector] ERROR: No buildpack groups passed detection.
[detector] ERROR: Please check that you are running against the correct path.
[detector] ERROR: failed to detect: no buildpacks participating
ERROR: failed with status code: 6
```

##### Expected behavior

```
$ pack build riffsquare --builder projectriff/builder:0.6.0-snapshot -e RIFF=1
0.6.0-snapshot: Pulling from projectriff/builder
Digest: sha256:9349cc12f5944982dfe41976acd734a411d87db158ed3c0e818fa3b1fd87a6fb
Status: Image is up to date for projectriff/builder:0.6.0-snapshot
base-cnb: Pulling from cloudfoundry/run
Digest: sha256:c1e5e1fdd95465525df6ecdd99be6f47d37a1e256a22c53287bad08aead92f3f
Status: Image is up to date for cloudfoundry/run:base-cnb
===> DETECTING
[detector] ======== Error: org.cloudfoundry.openjdk@1.2.14 ========
[detector] mkdir /tmp/plan.422531505: no space left on device
[detector] ======== Error: org.cloudfoundry.buildsystem@1.2.15 ========
[detector] mkdir /tmp/plan.861676380: no space left on device
[detector] ======== Error: io.projectriff.java@0.3.0-BUILD-SNAPSHOT ========
[...]

PR #351, opened 2020-07-29
feat: make buildpack detection errors info level

jabrown85:
Builpacks which do not return pass (`0`) or fail (`100`) will now have their output as well as the resulting `err` message logged at info level.

[Issue](https://github.com/buildpacks/lifecycle/issues/283)

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -73,8 +73,6 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
 
 		switch run.Code {
 		case CodeDetectPass, CodeDetectFail:
-		case -1:
-			outputLogf = c.Logger.Infof
 		default:
 			outputLogf = c.Logger.Infof
 		}
--- a/detector.go
+++ b/detector.go
@@ -69,23 +69,23 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
 			return nil, nil, errors.Errorf("missing detection of '%s'", bp)
 		}
 		run := t.(detectRun)
+		outputLogf := c.Logger.Debugf
+
+		switch run.Code {
+		case CodeDetectPass, CodeDetectFail:
+		case -1:
+			outputLogf = c.Logger.Infof
+		default:
+			outputLogf = c.Logger.Infof
+		}
+
 		if len(run.Output) > 0 {
-			outputLogf := c.Logger.Debugf
-
-			switch run.Code {
-			case CodeDetectPass, CodeDetectFail:
-			case -1:
-				outputLogf = c.Logger.Infof
-			default:
-				outputLogf = c.Logger.Infof
-			}
-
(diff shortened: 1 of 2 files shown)


jabrown85:
Here is what it looks like now for the `heroku/buildpacks`, which produces errors instead of failing with a couple buildpacks.
```
heroku@65a406591997:/layers$ /cnb/lifecycle/detector -log-level info
======== Output: heroku/ruby@0.0.1 ========
no
err:  heroku/ruby@0.0.1 (1)
======== Output: heroku/maven@0.1 ========
Could not find a pom.xml file! Please check that it exists and is committed to Git.
err:  heroku/maven@0.1 (1)
ERROR: No buildpack groups passed detection.
ERROR: failed to detect: buildpack(s) failed with err
```
A normal build that fails detection without any `err` produces the same result as before.
```
cnb@e93f02d6d255:/layers$ /cnb/lifecycle/detector -log-level info
ERROR: No buildpack groups passed detection.
ERROR: Please check that you are running against the correct path.
ERROR: failed to detect: no buildpacks participating
```

jabrown85:
I re-kicked the build. The output from the linux failure was around analyzer, which I don't think I could have affected. The windows tests passed as well.

```
        Step 1/4 : ARG fromImage
        Step 2/4 : FROM $fromImage
         ---> 2eb2d388e1a2
        Step 3/4 : ARG metadata
         ---> Using cache
         ---> e97bbd5d9425
        Step 4/4 : LABEL io.buildpacks.lifecycle.metadata=$metadata
         ---> Running in 7a393bc0340f
    analyzer_test.go:193: Expected nil: failed to execute command: [docker rmi some-app-image-ljppxrxahl], exit status 1, Error: No such image: some-app-image-ljppxrxahl
```

detector.go
```diff
@@ -70,8 +70,18 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
...
+				outputLogf = c.Logger.Infof
+			default:
+				outputLogf = c.Logger.Infof
+			}
+
+			outputLogf("======== Output: %s ========", bp)
+			outputLogf(string(run.Output))
 		}
 		if run.Err != nil {
 			c.Logger.Debugf("======== Error: %s ========", bp)
```
ekcasey, line 87: Should do the same thing here for error logs that we did above with `outputLogf`?

jabrown85: I had that, and then I undid it. I actually kind of think `run.Err` being anything other than `nil` should perhaps _always_ go to `info`. If you agree, I'll make that change in this PR

ekcasey: >  I actually kind of think run.Err being anything other than nil should perhaps always go to info

Hmm, I think we should stick toggling debug/info based on the exit code. `/bin/detect` might log to stderr or invoke a tool that does, even if there is no error per se.

jabrown85: Fair enough. I pushed that change.

PR #355, opened 2020-07-30
Lifecycle returns phase specific exit codes

### LINKED ISSUE #350: Update all error codes to match platform/0.4 spec
https://github.com/buildpacks/spec/pull/115

jabrown85:
All the phases that returned exit codes before now return exit codes that match the updated spec.

Closes #350

### DIFF
--- a/cmd/exit.go
+++ b/cmd/exit.go
@@ -17,11 +17,12 @@ const (
 	CodeDetectError            = 102
 	CodeAnalyzeError           = 202
 	CodeRestoreError           = 302
-	CodeFailedBuild            = 400
+	CodeFailedBuildWithErrors  = 401
+	CodeBuildError             = 402
 	CodeLaunchError            = 702
 	// 9: CodeFailedUpdate
-	CodeExportError = 502
-	CodeRebaseError = 602
+	CodeExportError  = 502
+	CodeRebaseError  = 602
 	CodeIncompatible = 11
 )
 
--- a/cmd/exit.go
+++ b/cmd/exit.go
@@ -15,8 +15,8 @@ const (
 	CodeFailedDetect           = 100
 	CodeFailedDetectWithErrors = 101
 	CodeDetectError            = 102
-	CodeAnalyzeError           = 200
-	CodeRestoreError           = 300
+	CodeAnalyzeError           = 202
+	CodeRestoreError           = 302
 	CodeFailedBuild            = 400
 	CodeFailedLaunch           = 8
 	// 9: CodeFailedUpdate
--- a/cmd/exit.go
+++ b/cmd/exit.go
@@ -21,7 +21,7 @@ const (
 	CodeFailedLaunch           = 8
 	// 9: CodeFailedUpdate
 	CodeFailedExport = 500
-	CodeFailedSave   = 10
+	CodeFailedRebase = 600
 	CodeIncompatible = 11
 )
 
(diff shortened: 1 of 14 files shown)


cmd/exit.go
```diff
@@ -14,10 +14,14 @@ const (
 	// 5: CodeNotFound
 	CodeFailedDetect           = 100
 	CodeFailedDetectWithErrors = 101
-	CodeFailedBuild            = 7
+	CodeDetectError            = 102
```
jabrown85, line 17: @ekcasey I wasn't sure if we wanted to document this in the spec explicitly or not. If we did, I guess we would want to do that for the other generic phase failures.

cmd/exit.go
```diff
@@ -14,10 +14,14 @@ const (
...
+	CodeAnalyzeError           = 200
+	CodeRestoreError           = 300
```
ekcasey, line 19:
```suggestion
	CodeAnalyzeError           = 202
	CodeRestoreError           = 302
```
Maybe this is super OCD but given that the generic detect error is `102` I like using `x02` for the other phases.

cmd/exit.go
```diff
@@ -14,10 +14,14 @@ const (
...
+	CodeFailedBuild            = 400
 	CodeFailedLaunch           = 8
```
ekcasey, line 21:
```suggestion
	CodeFailedBuild            = 400
	CodeLaunchError            = 702
```

cmd/exit.go
```diff
@@ -14,10 +14,14 @@ const (
...
 	// 9: CodeFailedUpdate
-	CodeFailedSave   = 10
+	CodeFailedExport = 500
+	CodeFailedRebase = 600
```
ekcasey, line 24:
```suggestion
	CodeExportError = 502
	CodeRebaseError = 602
```

cmd/exit.go
```diff
@@ -14,10 +14,14 @@ const (
 	// 5: CodeNotFound
 	CodeFailedDetect           = 100
 	CodeFailedDetectWithErrors = 101
-	CodeFailedBuild            = 7
+	CodeDetectError            = 102
+	CodeAnalyzeError           = 200
+	CodeRestoreError           = 300
+	CodeFailedBuild            = 400
```
ekcasey, line 20: Here I think we need a generic `CodeBuildError` and then different `CodeFailedBuildWithError` (or something like that), set to `401` for buildpack error:

https://github.com/buildpacks/spec/blob/platform/0.4/platform.md#outputs-3

cmd/exit.go
ekcasey: I think it is sufficiently documented to the extent that it is in the detect-specific lifecycle error range (102-199).

detector.go
```diff
@@ -320,23 +320,27 @@ func (bo BuildpackOrder) detect(done, next []Buildpack, optional bool, wg *sync.
...
-			continue
+		switch err := err.(type) {
+		case *Error:
+			if err.Type == ErrTypeBuildpack {
+				buildpackErr = true
+			}
+			if err.Type == ErrTypeBuildpack || err.Type == ErrTypeFailedDetection {
+				wg = &sync.WaitGroup{}
+				continue
+			}
```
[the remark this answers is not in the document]
ekcasey: `(bg BuildpackGroup) Detect(c *DetectConfig) (BuildpackGroup, BuildPlan, error)` is also exported so we should probably do the error translation there as well. It isn't used anywhere in the lifecycle but it is a potentially useful library method.

ekcasey:
Rebase and merged here: https://github.com/buildpacks/lifecycle/pull/361

Issue #356, opened 2020-07-30
Multicall launcher


ekcasey:
[RFC 0045](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md)
RFC PR: https://github.com/buildpacks/rfcs/pull/84
Spec Issue: https://github.com/buildpacks/spec/issues/105
Spec PR: https://github.com/buildpacks/spec/pull/118

## Behavior
**When** Platform API >= `0.4`:

### `exporter`
- export layer with `/cnb/process/<type>` symlinks
- always export launcher to `/cnb/lifecycle/launcher`, not `<launcher>`
- **If** the default `process-type` matches a process type
    - set the `ENTRYPOINT` to the default process symlink
- **Else**
    - **If**  there is exactly one process type
          -  set the `ENTRYPOINT` to the only process symlink
    - **Else** set the `ENTRYPOINT` to `/cnb/lifecycle/launcher`
- `prepend` `/cnb/process` to the `PATH`
- don't set `CNB_PROCESS_TYPE` on app image

### `launcher`
- Select process type using `$0`
- If a process type is selected, treat launcher arguments as additional `args` to the process type

PR #358, opened 2020-08-03
Semantic shell args

### LINKED ISSUE #354: Semantic handle of args in direct=false processes
[RFC 0045](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md)
RFC PR: buildpacks/rfcs#84
Spec Issue: https://github.com/buildpacks/spec/issues/106
Spec PR: https://github.com/buildpacks/spec/pull/116

For linux buildpacks, when buildpack API is > `0.3` treat args for `direct=false` as part of the command instead of as arguments to `bash`.

POC Implementation: https://github.com/buildpacks/lifecycle/blob/b9862d2635c57c3f383b51e9df59b2f0bef9e531/launch/launcher.go#L141-L150

ekcasey:
Resolves #354 

Also, fixes issues found with tokenization on windows (cc @micahyoung @ameyer-pivotal).


acceptance/launcher_test.go
```diff
@@ -78,10 +77,54 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+					)
+					assertOutput(t, cmd, "'arg' 'arg with spaces'")
+				})
+			})
+		})
 
 		it("sources scripts from process specific directories", func() {
-			cmd := exec.Command("docker", "run", "--rm", launchImage, "worker")
-			expected := "sourced bp profile\nsourced bp worker profile\nsourced app profile"
+			h.SkipIf(t, runtime.GOOS == "windows", "profile scripts are broken on windows?")
```
ekcasey, line 125: @micahyoung @ameyer-pivotal

I added this test to ensure variables set in the profile scripts are available when the command runs. It doesn't pass on windows. I believe this is bug with windows profiles scripts but I would appreciate your input.

ameyer-pivotal: Very possible there's a bug, since it's so new. It seemed to be working when we implemented it, but I believe only unit tests existed at the time

launch/launcher_windows.go
```diff
@@ -3,28 +3,13 @@ package launch
...
 	c.Stdout = os.Stdout
 	c.Stderr = os.Stderr
 	return c.Run()
 }
-
-func OSExecCmd(argv []string, envv []string) *exec.Cmd {
-	var c *exec.Cmd
-	arg0 := filepath.Base(argv[0])
-	if arg0 == "cmd" || arg0 == "cmd.exe" {
-		c = exec.Command(argv[0])
```
ameyer-pivotal, line 23: @ekcasey Are you sure it's safe to remove this branching? I forget the details but using `SysProcAttr` was necessary for proper parsing in some situations. Are you sure this change isn't related to the test breakage you noted on Windows?

ekcasey: Working backwards from the code as I found it: I believe problem this branching was motivated by the fact that profile script invocations were being added as a single argument: e.g "call profile.bat" instead of "call" "profile.bat" and therefore quotes needed to be stripped and the default argument escaping in the go stdlib had to be circumvented to undo the incorrect tokenization of the profile invocations. However, that workaround removed _all_ tokenization.

If we do nothing, go will escape arguments correctly for us (adding quotes only when an arg contains a " " or tab charcter). We just need to ensure that the arguments we pass in are properly tokenized.

ekcasey: https://github.com/buildpacks/lifecycle/blob/main/launch/launcher.go#L136

The linux implementation does something similar but the linux implementation is making a bash script which makes it slightly different from the windows shell logic (`cmd /c` can't take accept a batch script as a string afaik)

launch/launcher_test.go
```diff
@@ -233,15 +260,197 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+								"bash", "-c",
+								`exec bash -c "$@"`,
+								"/path/to/launcher", "some-command",
+							})
+						}
+					})
+				})
+			})
+
+			when("buildpacks have provided profile.d scripts", func() {
```
natalieparellano, line 343: Is this just a nesting change?

acceptance/launcher_test.go
```diff
@@ -78,10 +77,54 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+					"--env", "VAR2=val with space",
+					launchImage, "indirect-process-with-script",
+				)
+				assertOutput(t, cmd, "'val1' 'val with space'")
+			})
+		})
+
+		when("the process type has args", func() {
+			when("buildpack API 0.4", func() {
+				it("command and args become shell-parsed tokens in a script", func() {
```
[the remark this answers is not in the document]
yaelharel: +1 on the idea to show the failure scenario

acceptance/launcher_test.go
```diff
@@ -78,10 +77,54 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+					"--env", "VAR1=val1",
+					"--env", "VAR2=val with space",
+					launchImage, "indirect-process-with-script",
+				)
+				assertOutput(t, cmd, "'val1' 'val with space'")
+			})
+		})
+
+		when("the process type has args", func() {
+			when("buildpack API 0.4", func() {
```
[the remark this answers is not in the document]
ekcasey: It didn't, that issue is wrong. Good catch! I will fix it.

acceptance/testdata/launcher/posix/container/layers/config/metadata.toml
```diff
@@ -1,20 +1,51 @@
...
+
+[[buildpacks]]
+  id = "older/buildpack"
+  version = "0.0.1"
+  api = "0.3"
 
 [[processes]]
   type = "web"
-  command = "echo \"Executing web process-type\""
-  direct = false
```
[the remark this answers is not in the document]
ekcasey: We were inconsistent about when we explicitly added `direct = false` and when we let it default. I wanted to make it consistent but I am happy to make it consistently explicit if you think that is better.

launch/launcher_test.go
ekcasey: Yes, I moved it under `when("a start command is NOT marked as direct", func() {}` to make it clear that profile scripts are only executed for `direct=false` processes.

acceptance/testdata/launcher/posix/container/layers/config/metadata.toml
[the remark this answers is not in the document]
natalieparellano: Ah, I see. I'd lean toward being as explicit as possible given the level of complexity we are dealing with. As a newcomer, I always appreciate it :)

launch/launcher_test.go
```diff
@@ -233,15 +260,197 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
 					h.AssertEq(t, syscallExecArgsColl[0].argv0, "/bin/sh")
-					h.AssertEq(t, syscallExecArgsColl[0].argv, []string{"sh", "arg1", "arg2"})
+					h.AssertEq(t, syscallExecArgsColl[0].argv, []string{"sh", "arg1", "arg with space"})
 				}
 				h.AssertEq(t, syscallExecArgsColl[0].envv, envList)
 			})
 		})
 
+		when("a start command is NOT marked as direct", func() {
+			when("the providing buildpack has api >= 0.4", func() {
```
natalieparellano, line 273: https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md mentions using `CNB_PLATFORM_API` to toggle the launcher process resolution. I'm assuming that we don't need to inspect the platform API version for the arg handling, but please correct me if I'm wrong!

Also, I see that we have a `when` for buildpack API >= 0.4 but would it be worth also having one for lower API versions?

launch/launcher_test.go
```diff
@@ -233,15 +260,197 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+						if runtime.GOOS == "windows" {
+							h.AssertEq(t, syscallExecArgsColl[0].argv0, "cmd")
+							h.AssertEq(t, syscallExecArgsColl[0].argv, []string{
+								"cmd", "/q", "/c", `some-command`, `arg1`, "arg with space",
+							})
+						} else {
+							h.AssertEq(t, syscallExecArgsColl[0].argv0, "/bin/bash")
+							h.AssertEq(t, syscallExecArgsColl[0].argv, []string{
+								"bash", "-c",
+								`exec bash -c '"$(eval echo \"$0\")" "$(eval echo \"$1\")" "$(eval echo \"$2\")"' "${@:1}"`,
```
natalieparellano, line 309: I'll admit that this syntax still leaves me feeling slightly confused. Can we document in some detail how this works? 

Where I'm getting lost: there are three evals, but four "arguments" including the path to launcher. I see that `@:1` will trim off the first argument but then how does bash know what to exec? I imagine others might also have this problem...

launch/shell_unix.go
```diff
@@ -0,0 +1,71 @@
...
+		launcher += fmt.Sprintf("source \"%s\"\n", profile)
+	}
+	bashCommand, err := l.bashCommand(process)
+	if err != nil {
+		return errors.Wrap(err, "determine bash command")
+	}
+	launcher += bashCommand
+	if err := l.Exec("/bin/bash", append([]string{
+		"bash", "-c",
+		launcher, self, process.Command,
```
natalieparellano, line 31: I recall there is some complexity around the use of `self` here and why it's needed. Is it possible we could document that as well, maybe in a comment?

launch/shell_unix.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		if isLegacyProcess(bpAPI) {
+			return bashCommandWithScript, nil
+		}
+		return bashCommandWithTokens(len(process.Args) + 1), nil
+	}
+	return "", fmt.Errorf("process type '%s' provided by unknown buildpack '%s'", process.Type, process.BuildpackID)
+}
+
+func isLegacyProcess(bpAPI *api.Version) bool {
```
natalieparellano, line 61: I notice we don't do this check in shell_windows.go, is it because Windows support is so new? While I doubt we'd get buildpacks supporting an older API, it might be worthwhile to have this check, if it's considered supported.

launch/shell_unix.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/buildpacks/lifecycle/api"
+)
+
+var (
+	bashCommandWithScript = `exec bash -c "$@"` // for processes w/o argsument
```
natalieparellano, line 14:
```suggestion
	bashCommandWithScript = `exec bash -c "$@"` // for processes w/o arguments
```

launch/launcher_test.go
ekcasey: Adding a test case for buildpack API < 0.4.

The platform API will come into the picture in the implementation of https://github.com/buildpacks/lifecycle/issues/356

I intend to do some more `launcher_test.go` cleanup at that point as well.

launch/shell_unix.go
ekcasey: Windows buildpack support was introduced in buildpack API `0.4`. Adding an analogous legacy execution to windows would require some complicated additional work and it probably isn't worth doing (considering we want to move to the new API). We may want to add a check to ensuring that windows buildpacks never have API < 0.4. But, we probably want do that a build time so folks aren't surprised at launch, and therefore I think we should punt it to a separate PR.

launch/launcher_test.go
```diff
@@ -314,7 +314,7 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
 				})
 
 				when("there are zero args", func() {
-					it("treats the command as a bash script", func() {
+					it("treats the command as a bash script on linux", func() {
```
natalieparellano, line 317:
```suggestion
					it("treats the command as a bash script on linux or invokes with cmd on windows", func() {
```

natalieparellano: (having `on linux` in the description makes me expect to see an `h.SkipIf(runtime.GOOS == "windows")`)

launch/launcher_test.go
natalieparellano: To summarize from elsewhere - there is an outer bash, and an inner bash. /path/to/launcher is the 0th arg to the outer bash, and some-command is the 0th arg to the inner bash (thanks to `${@:1}`).

PR #363, opened 2020-08-04
packaging fixes

ekcasey:
* Uses Normalizing tar writer to fix paths when running on windows
* Move lifecycle.toml to top-level in archive
* Zeros UID/GID

### DIFF
--- a/layers/writer.go
+++ b/layers/writer.go
@@ -35,8 +35,12 @@ func (lw *layerWriter) Digest() string {
 }
 
 func tarWriter(lw *layerWriter) *archive.NormalizingTarWriter {
+	var tw *archive.NormalizingTarWriter
 	if runtime.GOOS == "windows" {
-		return archive.NewNormalizingTarWriter(layer.NewWindowsWriter(lw))
+		tw = archive.NewNormalizingTarWriter(layer.NewWindowsWriter(lw))
+	} else {
+		tw = archive.NewNormalizingTarWriter(tar.NewWriter(lw))
 	}
-	return archive.NewNormalizingTarWriter(tar.NewWriter(lw))
+	tw.WithModTime(archive.NormalizedModTime)
+	return tw
 }
--- a/tools/packager/main.go
+++ b/tools/packager/main.go
@@ -48,7 +48,7 @@ func main() {
 		os.Exit(3)
 	}
 	if err := archive.AddFileToArchive(tw, "lifecycle.toml", descriptorInfo); err != nil {
-		fmt.Printf("Failed to write descriptor to arichive: %s", err)
+		fmt.Printf("Failed to write descriptor to archive: %s", err)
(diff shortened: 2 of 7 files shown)


tools/packager/main.go
```diff
@@ -35,16 +37,28 @@ func main() {
...
+	tw.WithGID(0)
 	defer tw.Close()
 
+	descriptorInfo, err := os.Stat(descriptorPath)
+	if err != nil {
+		fmt.Printf("Failed stat descriptor file at path %s: %s", descriptorPath, err)
+		os.Exit(3)
+	}
+	if err := archive.AddFileToArchive(tw, "lifecycle.toml", descriptorInfo); err != nil {
+		fmt.Printf("Failed to write descriptor to arichive: %s", err)
```
ameyer-pivotal, line 51: Typo in `arichive`

tools/packager/main.go
```diff
@@ -35,16 +37,28 @@ func main() {
...
+		os.Exit(4)
+	}
+
 	if err := os.Chdir(filepath.Dir(inputDir)); err != nil {
 		fmt.Printf("Failed to switch directories to %s: %s", filepath.Dir(inputDir), err)
-		os.Exit(3)
```
nebhale, line 43: Are these being changed to spec or just incremented because some other code has changed?

layers/writer.go
```diff
@@ -35,8 +35,12 @@ func (lw *layerWriter) Digest() string {
 }
 
 func tarWriter(lw *layerWriter) *archive.NormalizingTarWriter {
+	var tw *archive.NormalizingTarWriter
 	if runtime.GOOS == "windows" {
-		return archive.NewNormalizingTarWriter(layer.NewWindowsWriter(lw))
+		tw = archive.NewNormalizingTarWriter(layer.NewWindowsWriter(lw))
+	} else {
+		tw = archive.NewNormalizingTarWriter(tar.NewWriter(lw))
```
nebhale, line 42: Are you comfortable with the lack of symmetry here?  That there's a custom windows writer, but not a custom *nix writer?

ekcasey: I'll admit it does bother me a little. However, windows layers needs special entries and prefixes and linux layers don't need anything special. If we moved some of generic layer helpers to `imgutil` it would look symmetric to the consumers. This  would be a nice change, but not one I am willing to make right now.

tools/packager/main.go
ekcasey: This is essentially just a script for our internal use while packaging the lifecycle, totally unspecified. `tools` is actually it's own module right now, although I think an internal package is probably more appropriate in the future.

ekcasey:
> It seems the creator.exe binary is missing for Windows lifecycle builds

Good catch, seems like we missed that in the original PR adding windows support

Issue #357, opened 2020-07-30
Remove top-level version from BOM


ekcasey:
[RFC 0043](https://github.com/buildpacks/rfcs/blob/main/text/0043-increase-build-plan-flexibility.md)
RFC PR: https://github.com/buildpacks/rfcs/pull/82
Spec PR: https://github.com/buildpacks/spec/pull/117
Spec Issue: https://github.com/buildpacks/spec/issues/107

## Behavior
If platform API is `0.3` Bill-of-Materials entries should include a top-level version key and any `metadata.version` fields should be converted to `version`
If platform API is `0.4` Bill-of-Materials entries should not include a top-level version key and `version` provided by the buildpack should be converted to `metadata.version`

ekcasey:
Updated requirements to preserve metadata.version in the platform 0.3 BOM. Some buildpacks are already using metadata.version with platform API 0.3  and we should not remove keys that are currently used/expected.

PR #367, opened 2020-08-05
Remove top-level version from BOM

yaelharel:
resolves: #357

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -2,11 +2,6 @@ package lifecycle
 
 import (
 	"fmt"
-
-	"github.com/BurntSushi/toml"
-
-	"github.com/buildpacks/lifecycle/api"
-
 	"io/ioutil"
 	"log"
 	"os"
@@ -14,6 +9,9 @@ import (
 	"path/filepath"
 	"sort"
 
+	"github.com/BurntSushi/toml"
+
+	"github.com/buildpacks/lifecycle/api"
 	"github.com/buildpacks/lifecycle/launch"
 	"github.com/buildpacks/lifecycle/layers"
 )
--- a/builder.go
+++ b/builder.go
@@ -1,14 +1,16 @@
 package lifecycle
 
 import (
+	"fmt"
+	"github.com/BurntSushi/toml"
+	"github.com/buildpacks/lifecycle/api"
+
 	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"sort"
-
-	"github.com/BurntSushi/toml"
 
 	"github.com/buildpacks/lifecycle/launch"
 	"github.com/buildpacks/lifecycle/layers"
@@ -19,6 +21,7 @@ type Builder struct {
 	LayersDir     string
 	PlatformDir   string
 	BuildpacksDir string
+	PlatformAPI   *api.Version
 	Env           BuildEnv
 	Group         BuildpackGroup
 	Plan          BuildPlan
@@ -143,6 +146,18 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 		labels = append(labels, launch.Labels...)
 	}
 
+   	comparisonVersion := api.MustParse("0.4")
+   	if b.PlatformAPI.Compare(comparisonVersion) < 0 {
+		//plaformApiVersion is less than comparisonVersion
+   		for _, bomEntry := range bom {
+			bomEntry.convertMetadataToVersion()
+		}
+	} else {
+		for _, bomEntry := range bom {
+			bomEntry.convertVersionToMetadata()
+		}
+	}
+
 	return &BuildMetadata{
 		BOM:        bom,
 		Buildpacks: b.Group.Group,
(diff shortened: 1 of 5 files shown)


builder.go
```diff
@@ -191,6 +208,23 @@ func (p buildpackPlan) has(entry BuildPlanEntry) bool {
 	return false
 }
 
+func (bom *BOMEntry) convertMetadataToVersion() {
+	if version, ok := bom.Metadata["version"]; ok {
+		bom.Version = fmt.Sprintf("%v", version)
```
natalieparellano, line 213: Do we want to check that `bom.Version == ""`. It seems like we shouldn't override a value that's already there.

builder.go
```diff
@@ -191,6 +208,23 @@ func (p buildpackPlan) has(entry BuildPlanEntry) bool {
...
+		delete(bom.Metadata, "version")
+	}
+}
+
+func (bom *BOMEntry) convertVersionToMetadata() {
+	if bom.Version != "" {
+		if bom.Metadata == nil {
+			bom.Metadata = make(map[string]interface{})
+		}
+		bom.Metadata["version"] = bom.Version
```
natalieparellano, line 223: Same comment as above...

builder.go
yaelharel: According to the [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0043-increase-build-plan-flexibility.md#what-it-is):
```
If both a version metadata entry and the top-level version key are both set, the lifecycle should fail with a descriptive error.
```
So I thought we should have only one parameter with the value.

builder.go
yaelharel: Same answer as above :)

builder.go
yaelharel: @ekcasey, @natalieparellano and I agree that it doesn't really matter whether we make this change or not.
What do you think?

builder.go
```diff
@@ -1,15 +1,19 @@
 package lifecycle
 
 import (
+	"fmt"
+
+	"github.com/BurntSushi/toml"
+
+	"github.com/buildpacks/lifecycle/api"
+
```
ekcasey, line 9: Fix import grouping. `fmt` should be group with other stdlib imports at the top of the `import` block. `github.com/buildpacks/lifecycle/api` should be group with other `"github.com/buildpacks/lifecycle/*` imports at the bottom and `github.com/BurntSushi/toml` should be in a separate group in the middle.

builder.go
ekcasey: Although we are checking this during detect in https://github.com/buildpacks/lifecycle/pull/369/files, I think we need to check again after build. We should check whether the destination value is set and fail if both are set and the values don't match.

builder_test.go
```diff
@@ -11,6 +11,10 @@ import (
 	"strings"
 	"testing"
 
+	h "github.com/buildpacks/lifecycle/testhelpers"
+
+	"github.com/buildpacks/lifecycle/api"
```
ekcasey, line 16: fix imports

cmd/lifecycle/builder.go
```diff
@@ -5,6 +5,8 @@ import (
 	"log"
 	"os"
 
+	"github.com/buildpacks/lifecycle/api"
```
ekcasey, line 8: imports

builder.go
yaelharel: Just to make sure - you think we should keep both values, right?
I find it a bit weird to have both values set. Especially due to the quote from the RFC ([here](https://github.com/buildpacks/lifecycle/pull/367#discussion_r466647047)) but I'm ok with doing so.

builder.go
yaelharel: Sorry, I thought that `make all` is doing that for me.
Why don't we do it?
Anyway, I fixed the import order in all of the files, please let me know if I missed something.
Thanks!

builder.go
yaelharel: Also, when you're saying that we need to check it again after build, do you mean at the end of the `Build` function or somewhere else?

yaelharel: I'm sorry, I read your comment and thought about something different.
I'll make these  changes.

builder.go
```diff
@@ -143,6 +146,18 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 		labels = append(labels, launch.Labels...)
 	}
 
+	comparisonVersion := api.MustParse("0.4")
+	if b.PlatformAPI.Compare(comparisonVersion) < 0 {
```
natalieparellano, line 150: We can probably inline `comparisonVersion`

builder.go
yaelharel: Hi @ekcasey, just want to double check:
It's clear that if both values are set and they don't match, we want to return an error, but is it ok that both values will be set to the same value? If yes, why is the quote I wrote [here](https://github.com/buildpacks/lifecycle/pull/367#discussion_r466647047 written in the RFC?

yaelharel: Following our conversation, we decided that:
* If both values are set with the same value, we aren't returning an error.
* We want to delete the original value after the conversion.

builder.go
ekcasey: I made sure stdlib, external, and internal imports were all in separate groups, but not correctly (it created three new groups instead of having only three groups total)... It has some unfortunately edge cases depending on the state of the input :/

Adding similar settings to your IDEA can help to avoid getting into the edge cases

Issue #330, opened 2020-07-02
Remove top-level version in build plan


ekcasey:
RFC Text: https://github.com/buildpacks/rfcs/blob/main/text/0043-increase-build-plan-flexibility.md
RFC PR: https://github.com/buildpacks/rfcs/pull/82
Spec PR: https://github.com/buildpacks/spec/pull/97
Spec Issue: https://github.com/buildpacks/spec/issues/91

### Behavior
Given a buildpack declaring buildpack API `0.3+` The `version` key should be removed from the build plan and buildpack plan.

IF an `0.3` buildpack sets `requires.version` THEN fail.

#### Passing `0.2` Requirements to `0.3` Providers
**Given** A `version` requirement from an `0.2` buildpack
```toml
[[requires]]
name = "<dependency name>"
version = "<dependency version>"

[requires.metadata]
some-key = "some-val"
```
**If** This requirement is met by an `0.3` buildpack
**Then** `version` should be converted to `metadata.version` in the buildpack plan, leaving the original  top-level `version`
```toml
[[entries]]
name = "<dependency name>"
version = "<dependency version>"

[entries.metadata]
some-key = "some-val"
version = "<dependency version>"
```

#### Passing `0.3` requirements to `0.2` Providers
**Given** A `version` metadata requirement from an `0.3` buildpack
```toml
[[requires]]
name = "<dependency name>"

[requires.metadata]
version = "<dependency version>"
some-key = "some-val"
```
**If** This requirement is met by an `0.2` buildpack
**Then** `metadata.version` should be converted to `version` in the buildpack plan
```toml
[[entries]]
name = "<dependency name>"
version = "<dependency version>"

[entries.metadata]
some-key = "some-val"
```
#### Error cases
**If** an `0.3` buildpack sets `requires.version`
**Then** the lifecycle should warn

**If** an `0.2` buildpack sets `requires.version` and `requires.metadata.version`
**And** these values are not equal to each other.
**Then** then the lifecycle should fail

### Issue History
**Updated** to account for https://github.com/buildpacks/spec/pull/110

natalieparellano:
@ekcasey I see

IF an 0.3 buildpack sets requires.version THEN fail.

and 

If an 0.3 buildpack sets requires.version
Then the lifecycle should warn

Should we warn or fail?

natalieparellano:
Also there is 

If an 0.2 buildpack sets requires.version and requires.metadata.version
And these values are not equal to each other.
[...]

PR #369, opened 2020-08-06
Remove top level version

natalieparellano:
Fixes https://github.com/buildpacks/lifecycle/issues/330

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -26,7 +26,7 @@ var errFailedDetection = errors.New("no buildpacks participating")
 var errBuildpack = errors.New("buildpack(s) failed with err")
 var errInconsistentVersion = errors.New("top level version does not match metadata version")
 var errDoublySpecifiedVersions = errors.New("top level version cannot be specified along with metadata version; use metadata version instead")
-var warnTopLevelVersion = "Warning: top level version is deprecated in buildpack API 0.3"
+var warnTopLevelVersion = `Warning: the "version" key is deprecated in build plan requirements in buildpack API 0.3. "metadata.version" should be used instead`
 
 type BuildPlan struct {
 	Entries []BuildPlanEntry `toml:"entries"`
@@ -329,12 +329,10 @@ func (bg BuildpackGroup) Detect(c *DetectConfig) (BuildpackGroup, BuildPlan, err
 	} else if err == errFailedDetection {
 		err = NewLifecycleError(err, ErrTypeFailedDetection)
 	}
-	for i, entry := range entries {
-		for j, req := range entry.Requires {
-			req.convertVersionToMetadata()
-			entry.Requires[j] = req
+	for i := range entries {
+		for j := range entries[i].Requires {
+			entries[i].Requires[j].convertVersionToMetadata()
 		}
-		entries[i] = entry
 	}
 	return BuildpackGroup{Group: bps}, BuildPlan{Entries: entries}, err
 }
@@ -389,12 +387,10 @@ func (bo BuildpackOrder) Detect(c *DetectConfig) (BuildpackGroup, BuildPlan, err
 	} else if err == errFailedDetection {
 		err = NewLifecycleError(err, ErrTypeFailedDetection)
 	}
-	for i, entry := range entries {
-		for j, req := range entry.Requires {
-			req.convertVersionToMetadata()
-			entry.Requires[j] = req
+	for i := range entries {
+		for j := range entries[i].Requires {
+			entries[i].Requires[j].convertVersionToMetadata()
 		}
-		entries[i] = entry
 	}
 	return BuildpackGroup{Group: bps}, BuildPlan{Entries: entries}, err
 }
--- a/detector.go
+++ b/detector.go
@@ -26,7 +26,7 @@ var errFailedDetection = errors.New("no buildpacks participating")
 var errBuildpack = errors.New("buildpack(s) failed with err")
 var errInconsistentVersion = errors.New("top level version does not match metadata version")
 var errDoublySpecifiedVersions = errors.New("top level version cannot be specified along with metadata version; use metadata version instead")
-var warnTopLevelVersion = `Warning: the "version" key is deprecated in build plan requirements in buildpack API 0.3. "metadata.version" should be used instead`
+var warnTopLevelVersion = `Warning: buildpack %s has a "version" key. This key is deprecated in build plan requirements in buildpack API 0.3. "metadata.version" should be used instead`
 
 type BuildPlan struct {
 	Entries []BuildPlanEntry `toml:"entries"`
@@ -308,7 +308,7 @@ func (bp *BuildpackTOML) detect(c *DetectConfig) detectRun {
 	}
 	if api.MustParse(bp.API).Compare(api.MustParse("0.3")) >= 0 {
 		if t.hasTopLevelVersions() || t.Or.hasTopLevelVersions() {
-			c.Logger.Warn(warnTopLevelVersion)
+			c.Logger.Warnf(warnTopLevelVersion, bp.Buildpack.Name)
 		}
 	}
 	t.Output = out.Bytes()
--- a/detector.go
+++ b/detector.go
@@ -101,13 +101,13 @@ type DetectConfig struct {
 }
 
 func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry, error) {
-	var runs []detectRun
+	var runs []DetectRun
 	for _, bp := range done {
 		t, ok := c.runs.Load(bp.String())
 		if !ok {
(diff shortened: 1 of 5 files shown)


detector.go
```diff
@@ -22,6 +22,9 @@ const (
 
 var errFailedDetection = errors.New("no buildpacks participating")
 var errBuildpack = errors.New("buildpack(s) failed with err")
+var errInconsistentVersion = errors.New("top level version does not match metadata version")
+var errDoublySpecifiedVersions = errors.New("top level version cannot be specified along with metadata version; use metadata version instead")
+var warnTopLevelVersion = "Warning: top level version is deprecated in buildpack API 0.3"
```
ekcasey, line 27:
```suggestion
var warnTopLevelVersion = `Warning: the "version" key is deprecated in build plan requirements in buildpack API 0.3. "metadata.version" should be used instead`
```

detector.go
```diff
@@ -251,6 +294,21 @@ func (bp *BuildpackTOML) detect(c *DetectConfig) detectRun {
...
+		}
+	}
+	if bp.API == "0.3" {
+		if t.hasDoublySpecifiedVersions() /*|| t.Or.hasDoublySpecifiedVersions()*/ { // TODO: figure out how to construct test case
+			t.Err = errDoublySpecifiedVersions
+		}
+	}
+	if bp.API == "0.3" {
+		if t.hasTopLevelVersions() /*|| t.Or.hasTopLevelVersions()*/ { // TODO: figure out how to construct test case
+			c.Logger.Warn(warnTopLevelVersion)
```
ekcasey, line 309: We should mention which buildpack this warning is for

detector_test.go
```diff
@@ -465,9 +466,11 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
 				toappfile("\n[[or]]", "detect-plan-A-v1.toml")
 				toappfile("\n[[or.provides]]\n name = \"dep1-present\"", "detect-plan-A-v1.toml")
 
-				toappfile("\n[[requires]]\n name = \"dep3-missing\"", "detect-plan-B-v1.toml")
+				toappfile("\n[[requires]]\n name = \"dep3-missing\"\n version=\"some-version\"", "detect-plan-B-v1.toml")
+				toappfile("\n[requires.metadata]\n version=\"some-version\"", "detect-plan-B-v1.toml")
 				toappfile("\n[[or]]", "detect-plan-B-v1.toml")
-				toappfile("\n[[or.requires]]\n name = \"dep1-present\"", "detect-plan-B-v1.toml")
+				toappfile("\n[[or.requires]]\n name = \"dep1-present\"\n version=\"some-version\"", "detect-plan-B-v1.toml")
+				toappfile("\n[or.requires.metadata]\n version=\"some-version\"", "detect-plan-B-v1.toml")
```
natalieparellano, line 473: Verify happy path for buildpack with buildpack api 0.2 and doubly specified (but matching) versions.

detector.go
```diff
@@ -251,6 +294,21 @@ func (bp *BuildpackTOML) detect(c *DetectConfig) detectRun {
 	if _, err := toml.DecodeFile(planPath, &t); err != nil {
 		return detectRun{Code: -1, Err: err}
 	}
+	if bp.API == "0.2" {
+		if t.hasInconsistentVersions() || t.Or.hasInconsistentVersions() { // TODO: should tests demonstrate that doubly specified (but consistent) versions work?
```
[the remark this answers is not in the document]
natalieparellano: Sure, I'll have a look. For now, [this change](https://github.com/buildpacks/lifecycle/pull/369/files#r467102699) is meant to address the TODO mentioned here.

detector.go
```diff
@@ -22,6 +24,9 @@ const (
 
 var errFailedDetection = errors.New("no buildpacks participating")
 var errBuildpack = errors.New("buildpack(s) failed with err")
+var errInconsistentVersion = errors.New("top level version does not match metadata version")
+var errDoublySpecifiedVersions = errors.New("top level version cannot be specified along with metadata version; use metadata version instead")
```
natalieparellano, line 28: Should we also mention the buildpack name when these errors are printed, and update the language to look more like `warnTopLevelVersion`? @ekcasey

detector.go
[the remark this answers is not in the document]
natalieparellano: @yaelharel and I did this refactor and the tests are SO much easier to understand now!

detector.go
natalieparellano: We assumed that this would be needed - latest code has this change. We needed to set the condition code along with the error, since the error message will now vary according to the buildpack name.

detector_test.go
```diff
@@ -528,6 +531,233 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
+							{Name: "dep2", Metadata: map[string]interface{}{"version": "some-version"}},
+							{Name: "dep2", Metadata: map[string]interface{}{"version": "some-version"}},
+						},
+					},
+				}) {
+					t.Fatalf("Unexpected entries:\n%+v\n", plan.Entries)
+				}
+			})
+
+			it("should fail if buildpacks with buildpack api 0.2 have a top level version and a metadata version that are different", func() {
```
natalieparellano, line 598: Should we nest this in a `when` or something, to make clear that it's the buildpack TOML `Detect`?

detector.go
```diff
@@ -298,12 +295,14 @@ func (bp *BuildpackTOML) Detect(c *DetectConfig) DetectRun {
 	}
 	if api.MustParse(bp.API).Equal(api.MustParse("0.2")) {
 		if t.hasInconsistentVersions() || t.Or.hasInconsistentVersions() {
-			t.Err = errInconsistentVersion
+			t.Err = errors.Errorf(errInconsistentVersion, bp.Buildpack.Name)
+			t.Code = CodeVersionFail
```
ekcasey, line 299: I think we should using `t.Code = -1` here. We need to use a code that it would be impossible for the buildpack itself to return (negative error codes aren't valid).

detector.go
```diff
@@ -298,12 +295,14 @@ func (bp *BuildpackTOML) Detect(c *DetectConfig) DetectRun {
...
 		}
 	}
 	if api.MustParse(bp.API).Compare(api.MustParse("0.3")) >= 0 {
 		if t.hasDoublySpecifiedVersions() || t.Or.hasDoublySpecifiedVersions() {
-			t.Err = errDoublySpecifiedVersions
+			t.Err = errors.Errorf(errDoublySpecifiedVersions, bp.Buildpack.Name)
+			t.Code = CodeVersionFail
```
ekcasey, line 305: Same as above: code -1

detector.go
```diff
@@ -62,16 +102,18 @@ type DetectConfig struct {
...
 		t, ok := c.runs.Load(bp.String())
 		if !ok {
 			return nil, nil, errors.Errorf("missing detection of '%s'", bp)
 		}
-		run := t.(detectRun)
+		run := t.(DetectRun)
 		outputLogf := c.Logger.Debugf
 
 		switch run.Code {
+		case CodeVersionFail:
```
ekcasey, line 115: Instead of returning here I think we should be in the `default:` case so that this is fatal to group detection for required buildpacks but not for optional buildpacks

detector.go
```diff
@@ -242,14 +281,31 @@ func (bp *BuildpackTOML) detect(c *DetectConfig) detectRun {
...
 	}
-	var t detectRun
+	var t DetectRun
 	if _, err := toml.DecodeFile(planPath, &t); err != nil {
-		return detectRun{Code: -1, Err: err}
+		return DetectRun{Code: -1, Err: err}
+	}
+	if api.MustParse(bp.API).Equal(api.MustParse("0.2")) {
+		if t.hasInconsistentVersions() || t.Or.hasInconsistentVersions() {
+			t.Err = errors.Errorf(errInconsistentVersion, bp.Buildpack.Name)
```
ekcasey, line 295:
```suggestion
			t.Err = errors.Errorf(errInconsistentVersion, bp.Buildpack.ID)
```

`ID` is required but `Name` is optional

detector.go
```diff
@@ -242,14 +281,31 @@ func (bp *BuildpackTOML) detect(c *DetectConfig) detectRun {
...
+			t.Code = -1
+		}
+	}
+	if api.MustParse(bp.API).Compare(api.MustParse("0.3")) >= 0 {
+		if t.hasDoublySpecifiedVersions() || t.Or.hasDoublySpecifiedVersions() {
+			t.Err = errors.Errorf(errDoublySpecifiedVersions, bp.Buildpack.Name)
```
ekcasey, line 301:
```suggestion
			t.Err = errors.Errorf(errInconsistentVersion, bp.Buildpack.ID)
```

`ID` is required but `Name` is optional

PR #370, opened 2020-08-06
Multicall launcher

ekcasey:
Multicall launcher in platform 0.4


launch/process_test.go
```diff
@@ -0,0 +1,227 @@
...
+							h.AssertNil(t, err)
+							h.AssertEq(t, proc, launch.Process{
+								Command: "user-command",
+								Args:    []string{"user-arg1", "user-arg2"},
+								Direct:  true,
+							})
+						})
+					})
+
+					when("cmd contains more than one arg and does not start with --", func() {
```
natalieparellano, line 147: I'm a little confused... what happens when the `cmd` contains exactly one arg? Is that covered by these tests?

launch/process_test.go
```diff
@@ -0,0 +1,227 @@
...
+							h.AssertNil(t, err)
+							h.AssertEq(t, proc, launch.Process{
+								Command: "user-command",
+								Args:    []string{"user-arg1", "user-arg2"},
+							})
+						})
+					})
+				})
+
+				when("doesn't exit", func() {
```
natalieparellano, line 214:
```suggestion
				when("doesn't exist", func() {
```

launch/process_test.go
```diff
@@ -0,0 +1,227 @@
...
+						h.AssertEq(t, proc, launch.Process{
+							Type:        "some-type",
+							Command:     "some-command",
+							Args:        []string{"some-arg1", "some-arg2", "user-arg1", "user-arg1"},
+							BuildpackID: "some-buildpack",
+						})
+					})
+				})
+
+				when("doesn't exit", func() {
```
natalieparellano, line 95:
```suggestion
				when("doesn't exits", func() {
```

launch/launcher_test.go
```diff
@@ -219,351 +131,409 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+					})
+
+					when("has no args", func() {
+						it.Before(func() {
+							process.Args = []string{}
+						})
+
+						when("linux", func() {
+							it.Before(func() {
+								h.SkipIf(t, runtime.GOOS == "windows", "windows test")
```
natalieparellano, line 423:
```suggestion
								h.SkipIf(t, runtime.GOOS == "windows", "linux test")
```

launch/launcher_test.go
```diff
@@ -219,351 +131,409 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+							}
+						})
+					})
+
+					when("linux", func() {
+						it.Before(func() {
+							h.SkipIf(t, runtime.GOOS == "windows", "linux test")
+						})
+
+						it("is script", func() {
```
natalieparellano, line 478: I think I missed what it is that makes the process a script (it has args)... is that the default?

natalieparellano: I actually don't see what makes this test case different from https://github.com/buildpacks/lifecycle/pull/370/files#diff-3841e5f60214a915376943a789c0747cR492

launch/process_test.go
natalieparellano, line 95:
```suggestion
				when("doesn't exist", func() {
```

layers/factory.go
```diff
@@ -36,6 +36,45 @@ type Logger interface {
...
+			Digest:  sha,
+		}, nil
+	}
+	lw, err := newFileLayerWriter(tarPath)
+	if err != nil {
+		return Layer{}, err
+	}
+	defer func() {
+		if closeErr := lw.Close(); err == nil {
+			err = closeErr
```
sclevine, line 58: Needs named return param.

acceptance/launcher_test.go
```diff
@@ -42,217 +42,292 @@ func TestLauncher(t *testing.T) {
...
 
 func testLauncher(t *testing.T, when spec.G, it spec.S) {
-	when("there is no CMD provided", func() {
-		when("CNB_PROCESS_TYPE is NOT set", func() {
-			it("web is the default process-type", func() {
-				cmd := exec.Command("docker", "run", "--rm", launchImage)
+	when("Platform API >= 0.4", func() {
+		when("entrypoint is a process", func() {
+			it("launches that process", func() {
+				entrypoint := "/cnb/process/web"
```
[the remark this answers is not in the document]
ekcasey: Good point, I'll add `/cnb/process` to the path in the `Dockerfile`

acceptance/launcher_test.go
```diff
@@ -42,217 +42,292 @@ func TestLauncher(t *testing.T) {
...
+				)
+			})
 		})
-	})
 
-	when("process is direct=false", func() {
-		when("the process type has no args", func() {
-			it("runs command as script", func() {
-				h.SkipIf(t, runtime.GOOS == "windows", "scripts are unsupported on windows")
+		when("CNB_PROCESS_TYPE is set", func() {
```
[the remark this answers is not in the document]
ekcasey: It shouldn't work, and I didn't necessary want to write an "it fails" test

acceptance/launcher_test.go
```diff
@@ -42,217 +42,292 @@ func TestLauncher(t *testing.T) {
...
-				expected = "Usage: ping"
-			}
-			assertOutput(t, cmd, expected)
-		})
+		when("entrypoint is a not a process", func() {
+			it("builds a process from the arguments", func() {
+				cmd := exec.Command("docker", "run", "--rm",
+					"--entrypoint=/cnb/lifecycle/launcher",
+					"--env=CNB_PLATFORM_API=0.4",
+					launchImage, "--", "env")
```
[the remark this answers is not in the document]
ekcasey: To avoid testing the full combinatorial explosion of APIs/args/direct (we would have a lot of tests) I was going to let the platform 0.3 test handle that, knowing that we should be in the same code path at that point.

launch/launcher_back_test.go
```diff
@@ -0,0 +1,592 @@
+package launch_test
```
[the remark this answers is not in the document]
ekcasey: yes! lol

thanks

env/launch_test.go
```diff
@@ -26,35 +29,33 @@ func testLaunchEnv(t *testing.T, when spec.G, it spec.S) {
...
-					t.Skip("This test only applies to Windows launches")
-				}
-			})
-
-			it("ignores case when initializing", func() {
-				benv := env.NewBuildEnv([]string{
-					"Path=some-path",
+		when("#NewLaunchEnv", func() {
+			when("process dir is the first element in PATH", func() {
+				it("strips it", func() {
```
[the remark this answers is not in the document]
ekcasey: Good catch! missed that one

launch/bash_test.go
```diff
@@ -0,0 +1,209 @@
...
+					}
+					h.AssertStringContains(t, stdout, "profile env: 'some-profile-var'")
+				})
+			})
+
+			it("sets env", func() {
+				process = launch.ShellProcess{
+					Script:  true,
+					Command: `printf "SOME_VAR: '%s'" "$SOME_VAR"`,
+					Caller:  "some-profile-argv0",
```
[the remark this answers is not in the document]
ekcasey: At first I called it `ProfileArgv0` but that isn't true in `cmd`. I think it is no more confusing than `self` but tbf that was already a confusing name

launch/bash_test.go
```diff
@@ -0,0 +1,209 @@
...
+					t.Fatalf("stdout was empty: stderr: %s\n", stderr)
+				}
+				h.AssertStringContains(t, stdout, "SOME_VAR: 'some-val'")
+			})
+
+			it("provides args to bash", func() {
+				process = launch.ShellProcess{
+					Script:  true,
+					Command: `printf "SOME_ARG: '%s'" "$1"`,
+					Args:    []string{"", "some arg1"},
```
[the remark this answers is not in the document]
ekcasey: Functionally it doesn't matter to this test we could have printed `$0` and dropped the empty first arg. But because `$0` is supposed to represent the process being invoked it feels weird to put an actual argument there. We could change `$0` to `bash` to make it more realistic/readable.

launch/process_test.go
ekcasey: one vs many doesn't matter to `cmd` right now. I'll update the `when`s

metadata.go
```diff
@@ -61,12 +65,13 @@ func (cm *CacheMetadata) MetadataForBuildpack(id string) BuildpackLayersMetadata
 
 // NOTE: This struct MUST be kept in sync with `LayersMetadataCompat`
```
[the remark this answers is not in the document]
ekcasey: I suppose it does say `MUST`! :)

exporter_test.go
```diff
@@ -912,43 +957,134 @@ type = "Apache-2.0"
...
+						_, err := exporter.Export(opts)
+						h.AssertNil(t, err)
+
+						val, err := fakeAppImage.Env("CNB_PROCESS_TYPE")
+						h.AssertNil(t, err)
+						h.AssertEq(t, val, "some-process-type")
+					})
+
+					it("sets ENTRYPOINT to launcher", func() {
+						_, err := exporter.Export(opts)
```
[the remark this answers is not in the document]
ekcasey: For API < 0.4 the launcher point should be launcher even when there is no default process. I implemented your suggestion to move the default process setup to the before (fixing the test  and the inaccurate description here)

launch/launcher_test.go
ekcasey: Hmmm the short `when`'s might be confusing in the code but I wanted to make the `go test` output a bit more readable on the command line

Should hopefully tell you it is a script if it has no args and is linux:
`LaunchProcess/Direct=false/buildpack-provided/buildpack_API_>=_0.4/has_no_args/linux/is_script`

The link above didn't take me to a different test.

PR #375, opened 2020-08-09
Sets working dir to /layers in lifecycle image

ekcasey:
I manually released the `buildpacksio/lifecycle:0.9.0` and `buildpacksio/lifecycle:latest` images to fix this.

In the future I think we need a way configure the lifecycle image in `pack` and in the `pack` acceptance tests to enable better testing of lifecycle images.


natalieparellano:
Oof, I regret not having caught this. I confirmed a Linux image built from this branch will run with pack.

PR #379, opened 2020-08-12
Exporter: when no default process is specified, don't set a default process (even when there is only one process)

### LINKED ISSUE #378: Exporter: when no default process is specified, don't set a default process (even when there is only one process)
With platform API 0.4, when no default process is specified and there is exactly one process, the exporter will set the default process (entrypoint) to the only process.

As discussed in WG: the lifecycle should be as transparent as possible and have no opinions. The current behavior could be surprising to some users.

Given: a build with only one process type using platform API 0.4
When: no process type is specified (`-process-type` not passed)
Then: the exporter should set the entrypoint to `/cnb/lifecycle/launcher` (or the Windows equivalent)

natalieparellano:
Fixes #378 

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -353,14 +353,7 @@ func (e *Exporter) setEnv(opts ExportOptions, launchMD launch.Metadata) error {
 }
 
 func (e *Exporter) entrypoint(launchMD launch.Metadata, defaultProcessType string) (string, error) {
-	if !e.supportsMulticallLauncher() {
-		return launch.LauncherPath, nil
-	}
-	if defaultProcessType == "" {
-		if len(launchMD.Processes) == 1 {
-			e.Logger.Infof("Setting default process type '%s'", launchMD.Processes[0].Type)
-			return launch.ProcessPath(launchMD.Processes[0].Type), nil
-		}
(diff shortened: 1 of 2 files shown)


natalieparellano:
So it occurred to me to check the spec, and in fact the current behavior is in the spec:

> If <process-type> does not match a buildpack-provided process:
> If there is exactly one buildpack-provided process:
> MUST have ENTRYPOINT=/cnb/process/<type> where <type> matches the type of the process
> Else:
> MUST have ENTRYPOINT set to /cnb/lifecycle/launcher

I believe we should hold off on merging this one.

nebhale:
Concur with the hold.  It’s not critical of 0.9.1 and we might actually want a spec change.

natalieparellano:
I will close this PR (so that we don't accidentally merge it) and we can open another one when we've decided on the implementation strategy.

yaelharel:
@natalieparellano @ekcasey 
Do we want to keep this PR open?
Issue #378 was closed.

kadirselcuk (approved):

natalieparellano:
Closing, we will do this work as part of https://github.com/buildpacks/lifecycle/issues/457

PR #381, opened 2020-08-13
Pass through http(s) proxy variables to buildpacks environment

### LINKED ISSUE #380: Includelist breaks HTTP(S) proxy support
### Summary
The following variables may be set by the platform to enable proxy support: `HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`, `https_proxy`, `http_proxy`, `no_proxy`.

These variables are not present in the include list that the lifecycle uses when constructing the build environment: https://github.com/buildpacks/lifecycle/blob/e93dc331478b0b4b02bad2749f1e332e7fabfba9/env/build.go#L8

As a result, http(s) requests made by buildpacks may fail.

---

##### Steps

Using pack, but there may be other repros:

1. `HTTPS_PROXY=some-proxy pack build my-app` <- may be run in an environment that requires requests to go through a proxy
2. Observe that the build does not behave as expected

##### Current behavior
`HTTPS_PROXY`, etc. is stripped from the build environment.

##### Expected
`HTTPS_PROXY`, etc. should be present in the build environment if set in the lifecycle's environment.

---

##### lifecycle version
This was likely introduced with lifecycle version 0.7.0.

##### platform version(s)
pack as of version 0.12.0 passes these variables to the lifecycle in the container env: https://github.com/buildpacks/lifecycle/blob/e93dc331478b0b4b02bad2749f1e332e7fabfba9/env/build.go#L8

natalieparellano:
Fixes #380 


natalieparellano:
I removed and then re-added the VERSION file update, it can hop along with this PR, rather than having its own separate PR.

env/build_test.go
```diff
@@ -31,14 +39,28 @@ func testBuildEnv(t *testing.T, when spec.G, it spec.S) {
...
+				"HTTP_PROXY=some-http-proxy",
 				"LD_LIBRARY_PATH=some-ld-library-path",
 				"LIBRARY_PATH=some-library-path",
+				"NO_PROXY=some-no-proxy",
 				"PATH=some-path",
 				"PKG_CONFIG_PATH=some-pkg-config-path",
-			}); s != "" {
+			}
+			if runtime.GOOS != "windows" {
+				expectedVars = append(
```
jromero, line 56: It took me a while to figure out why/how this expectation works. Case-sensitivity right? Might be worth calling out in a comment.

natalieparellano: Sounds good!

PR #391, opened 2020-08-17
Add Makefile target for local lifecycle image

natalieparellano:


natalieparellano:
This could be useful not only for trying things out locally but also when running the acceptance tests - our helpers that currently wrap `make build` could wrap this target instead.

I'm not quite sure how the tagging should work - right now it just uses `dev`. I thought about using the current commit, but if you have uncommitted changes that could get confusing.

ekcasey:
> I'm not quite sure how the tagging should work - right now it just uses dev. I thought about using the current commit, but if you have uncommitted changes that could get confusing.

Maybe something like `git describe --tags --dirty`?

Example: `v0.9.0-5-gc5adecf` or `v0.9.0-5-gc5adecf-dirty`

It could be nice b/c you get the most recent tag, commits since the tag, and an indication of whether there were uncommitted changes?

tools/image/main.go
```diff
@@ -242,3 +274,18 @@ func lifecycleLayer() string {
...
+		// Retry
+		rc, err = dockerCli.ImagePull(context.Background(), ref, dockertypes.ImagePullOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	if _, err := io.Copy(ioutil.Discard, rc); err != nil {
+		return err
+	}
+	return rc.Close()
```
ekcasey, line 290: I think we want to `defer rc.Close()` before we copy. If there is an error during copy the reader will not be closed properly.

PR #400, opened 2020-08-22
Stack buildpacks

jkutner:
This change introduces a stack buildpack that is identified by a privileged key in the `buildpack.toml`

When the privileged key is set to true, the builder will ignore layers
created in the layers directory and instead will take a snapshot of the
changes made during execution of the stack buildpack.

See: https://github.com/buildpacks/rfcs/pull/87


cmd/lifecycle/builder.go
```diff
@@ -48,39 +58,96 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
 func (b *buildCmd) Privileges() error {
-	// builder should never be run with privileges
-	if priv.IsPrivileged() {
+	_, stackGroup, _, err := b.readData()
+	if err != nil {
+		return err
+	}
+
+	// builder should never be run with privileges if there aren't any stack buildpacks
+	if len(stackGroup.Group) == 0 && priv.IsPrivileged() {
```
jabrown85, line 67: Does this mean the caller must know if the contents of the file specified at `stackGroupPath`? It seems like this should check for the existence of the flag instead of the contents of the file.

jkutner: Yea, I think I agree, but I was also worried about the case where:
1. the flag is not provided
2. the default file (`stack-group.toml`) is used
3. the default file is empty

PR #401, opened 2020-08-23
Bugfix/windows cmd syntax

### LINKED ISSUE #359: Environment variables set in windows profile scripts are not set in the process environment
On the Linux side, we have an "outer bash" and an "inner bash" to accomplish this. On Windows, this is complicated by the fact that `cmd` does not accept a `.bat` script from stdin - it needs to be written to a file.

Possible way forward - we write a temporary file and pass that to `cmd`. This would introduce file i/o and potentially slow down the launch process, but may be the "correct" thing to do.

Are there other options? This may require investigation.

micahyoung:
1. Enables `!VAR!`-style env vars in process `args` with otherwise no change to existing behavior

    Sets `/v:on` on `cmd` to enable usage of env vars set in profile scripts, in the same scope, known as Delayed Expansion

     From Microsoft Docs: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd
     > If you enable delayed environment variable expansion, you can use the exclamation point character to substitute the value of an environment variable at run time.

    From SS64.com: https://ss64.com/nt/delayedexpansion.html
    > Delayed Expansion will cause variables within a batch file to be expanded at execution time rather than at parse time

    Fixes #359

2. Also, makes cmd use consistent command quote interpretation 

    Sets `/s` on `cmd` to use consistent quote interpolation, matching current Docker `RUN` implementation and Microsoft-recommended `SHELL` syntax:

    From Docker Docs on `RUN` behavior: https://docs.docker.com/engine/reference/builder/#run
    > RUN <command> (shell form, the command is run in a shell, which by default is /bin/sh -c on Linux or cmd /S /C on Windows)

    From Microsoft Docs on recommended usage: https://docs.microsoft.com/en-us/visualstudio/install/advanced-build-tools-container#dockerfile
    > `# Restore the default Windows shell for correct batch processing.`
    > `SHELL ["cmd", "/S", "/C"]`

    From Microsoft Docs on /s behavior and the legacy corner case it avoids: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd#remarks

    For brevity, I will not quote inline but my understanding of the 5 conditions of the corner case are: When not using /s, a user can have an ambiguous command when using a binary with spaces in the name followed by space-separated arguments. Cmd then parses any quotes, and if it finds a binary with a space-separated name matching a subset of the args, it will call it.
    
    Instead of allowing this ambiguous syntax, we should use `/s` and always require users to split commands as expected using `command` and `args` array. This also follows existing recommended Docker syntax anyways (`cmd /s /c`) so would be more likely to match existing commands being ported to CNBs.

   This should have been in the original implementation and will have no change on existing behavior (existing test case `indirect-process-with-args` using `command = "test tokens.bat"` works before and after this change), but this should prevent unintentionally supporting ambiguous syntax without proper escaping.

### DIFF
--- a/acceptance/launcher_test.go
+++ b/acceptance/launcher_test.go
@@ -198,7 +198,6 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
 			})
 
 			it("sources scripts from process specific directories", func() {
-				h.SkipIf(t, runtime.GOOS == "windows", "profile scripts are broken on windows?")
 				cmd := exec.Command("docker", "run", "--rm", launchImage, "profile-checker")
 				expected := "sourced bp profile\nsourced bp profile-checker profile\nsourced app profile\nval-from-profile"
 				assertOutput(t, cmd, expected)
--- a/acceptance/testdata/launcher/windows/container/layers/config/metadata.toml
+++ b/acceptance/testdata/launcher/windows/container/layers/config/metadata.toml
@@ -27,7 +27,7 @@
 [[processes]]
   type = "profile-checker"
   command = "echo"
-  args = ["%VAR_FROM_PROFILE%"]
+  args = ["!VAR_FROM_PROFILE!"]
   direct = false
   buildpack-id = "some/buildpack"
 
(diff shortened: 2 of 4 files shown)


micahyoung:
@ekcasey Thanks, missed that one. All `Pend`s are gone and the test is fixed. I think it's good to go.

Issue #384, opened 2020-08-13
Exporter: fix handling for run images with null Config.Env.PATH


micahyoung:
### Summary
Platform 0.4 exposed a bug with Windows images where an exported image with `ENTRYPOINT` set to `c:\cnb\lifecycle\launcher.exe` (`pack`'s default) are unrunnable because `launcher` tries to run subcommands but fails to find `cmd` in the `PATH` at `c:\Windows\system32\cmd.exe`.

```
docker run --rm sample-hello-world-windows-app:nanoserver-1809
ERROR: failed to launch: cmd execute: exec: "cmd": executable file not found in %PATH%
```

Example app image: `micahyoung/sample-hello-world-windows-app:nanoserver-1809`

---

##### Steps

1. run `make build-windows build-windows-app` from this samples [PR commit](https://github.com/buildpacks/samples/pull/85/commits/ec6a4acec140637f083b6bfe17c87b1e5c8ade24)

##### Current behavior
1. attempt to run app
```
docker run --rm sample-hello-world-windows-app:nanoserver-1809
ERROR: failed to launch: cmd execute: exec: "cmd": executable file not found in %PATH%
```

1. inspect 
```
docker inspect sample-hello-world-windows-app:nanoserver-1809
...
[
    {
        ...
        "Config": {
            "Env": [
                "PATH=c:\\cnb\\process;c:\\cnb\\lifecycle;",
                 ...
            ],
            ...
    }
]
...
```

##### Expected
1. attempt to run app
```
docker run --rm sample-hello-world-windows-app:nanoserver-1809
# successful run message
```
2. Unsure what the inspected image config should look like (see **potential solutions** at bottom)

---

### Context
Pack issue: https://github.com/buildpacks/pack/issues/800
Related moby issues/unimplemented PRs: 
* https://github.com/moby/moby/issues/22017
* https://github.com/moby/moby/issues/30033
* https://github.com/moby/moby/pull/29048

This appears to happen because the exported image config's `Config.Env.PATH` only has CNB paths (`c:\cnb\process;c:\cnb\lifecycle`) and no default Windows paths (`c:\Windows\system32;C:\Windows`), making it impossible for `lifecycle.exe` to find `cmd.exe`.

[...]

PR #402, opened 2020-08-24
Fix inconsistent Config.Env.PATH behavior for Windows images

micahyoung:
- When launcher or lifecycle are executed, immediately sync env vars from the Windows registry and runtime environment (no-op for Linux)
- Works around Docker only using Windows registry PATH when image Config.Env.PATH is not set (same is true for other env vars):
  Moby issue: https://github.com/moby/moby/issues/22017
  BuildKit issue: https://github.com/moby/buildkit/issues/74
  PR attempt: https://github.com/moby/moby/pull/29048

Fixes: https://github.com/buildpacks/lifecycle/issues/384

### DIFF
--- a/acceptance/launcher_test.go
+++ b/acceptance/launcher_test.go
@@ -282,6 +282,22 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
 				)
 			})
 
+			it("merges env vars from system, user provided and CNB", func() {
+				if runtime.GOOS == "windows" {
+					cmd := exec.Command("docker", "run", "--rm",
+						`--env=PATH=c:\Windows\system32;c:\extra`,
+						launchImage, "--",
+						`cmd`, `/c`, `echo %PATH%`)
+					assertOutput(t, cmd, `c:\layers\some_buildpack\some_layer\bin;C:\Windows\system32;C:\Windows;c:\extra`)
+				} else {
+					cmd := exec.Command("docker", "run", "--rm",
+						"--env=PATH=/bin:/extra",
+						launchImage, "--",
+						"sh", "-c", "echo $PATH")
+					assertOutput(t, cmd, "/layers/some_buildpack/some_layer/bin:/bin:/extra")
+				}
+			})
+
 			it("passes through env vars from user, excluding excluded vars", func() {
 				cmd := exec.Command("docker", "run", "--rm",
 					"--env", "CNB_APP_DIR=/workspace",
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -5,7 +5,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
+
+	"github.com/buildpacks/lifecycle/env"
 )
 
 var (
@@ -205,4 +208,17 @@ func EnvOrDefault(key string, defaultVal string) string {
 		return envVal
 	}
 	return defaultVal
+}
+
+func PrepareOSEnv() error {
+	if runtime.GOOS != "windows" {
+		return nil
+	}
+
+	regEnvMap, err := env.WindowsRegistryEnvMap()
+	if err != nil {
+		return err
+	}
+
+	return env.PrepareWindowsOSEnv(regEnvMap, os.Getenv, os.Setenv)
 }
--- a/cmd/lifecycle/main.go
+++ b/cmd/lifecycle/main.go
@@ -12,6 +12,10 @@ import (
 )
 
 func main() {
+	if err := cmd.PrepareOSEnv(); err != nil {
+		cmd.Exit(err)
+	}
+
 	platformAPI := cmd.EnvOrDefault(cmd.EnvPlatformAPI, cmd.DefaultPlatformAPI)
 	if err := cmd.VerifyPlatformAPI(platformAPI); err != nil {
 		cmd.Exit(err)
--- a/env/prepare.go
+++ b/env/prepare.go
@@ -1,0 +1,65 @@
+package env
+
+import (
+	"os"
+	"path/filepath"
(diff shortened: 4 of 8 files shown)


acceptance/launcher_test.go
```diff
@@ -282,6 +282,22 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+					assertOutput(t, cmd, `c:\layers\some_buildpack\some_layer\bin;C:\Windows\system32;C:\Windows;c:\extra`)
+				} else {
+					cmd := exec.Command("docker", "run", "--rm",
+						"--env=PATH=/bin:/extra",
+						launchImage, "--",
+						"sh", "-c", "echo $PATH")
+					assertOutput(t, cmd, "/layers/some_buildpack/some_layer/bin:/bin:/extra")
+				}
+			})
+
```
micahyoung, line 300: I tried to include some acceptance-level coverage. Not 100% it conveys the purpose but I feel like it would catch any regressions.

cmd/flags.go
```diff
@@ -206,3 +209,16 @@ func EnvOrDefault(key string, defaultVal string) string {
...
+		return nil
+	}
+
+	regEnvMap, err := env.WindowsRegistryEnvMap()
+	if err != nil {
+		return err
+	}
+
+	return env.PrepareWindowsOSEnv(regEnvMap, os.Getenv, os.Setenv)
+}
```
micahyoung, line 224: I feel like an a os-agnostic name `PrepareOSEnv` is better for `main` but wanted to keep windows-specific names below that.

I split up windows-build only `WindowsRegistryEnvMap` from the business logic unit `PrepareWindowsOSEnv` so unit tests for the latter could run on any OS.

env/prepare.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// PrepareWindowsOSEnv merges map of Windows registry environment variables with in-memory environment vars
+// Injects os.Getenv an os.Setenv
+func PrepareWindowsOSEnv(regEnvMap map[string]string, getenv func(string) string, setenv func(string, string) error) error {
```
micahyoung, line 11: Injects `setenv` to avoid test pollution and `getenv` to allow testing for the former.

env/prepare.go
```diff
@@ -0,0 +1,65 @@
...
+		envValue := getenv(regKey)
+
+		// exit early if values match
+		if regValue == envValue {
+			continue
+		}
+
+		newValue := ""
+
+		if regKey == "PATH" || regKey == "PATHEXT" {
```
micahyoung, line 22: Reg keys will always be upcase, thanks to `WindowsRegistryEnvMap`

env/prepare.go
```diff
@@ -0,0 +1,65 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// PrepareWindowsOSEnv merges map of Windows registry environment variables with in-memory environment vars
+// Injects os.Getenv an os.Setenv
+func PrepareWindowsOSEnv(regEnvMap map[string]string, getenv func(string) string, setenv func(string, string) error) error {
+	for regKey, regValue := range regEnvMap {
+		envValue := getenv(regKey)
```
micahyoung, line 13: `os.Getenv` on windows is case-insensitive so it will return the same for `PATH` or `Path`

env/prepare.go
```diff
@@ -0,0 +1,65 @@
...
+	newValueMap := map[string]bool{}
+
+	for _, regValuePart := range filepath.SplitList(regValue) {
+		key := strings.ToLower(regValuePart)
+		if regValuePart == "" || newValueMap[key] {
+			continue
+		}
+
+		newValueMap[key] = true
+		newValueParts = append(newValueParts, regValuePart)
```
micahyoung, line 53: Use separate lowercase key for uniqueness checking but preserve original case in appended value

cmd/lifecycle/main.go
```diff
@@ -12,6 +12,10 @@ import (
 )
 
 func main() {
+	if err := cmd.PrepareOSEnv(); err != nil {
+		cmd.Exit(err)
+	}
```
ekcasey, line 17: I am not convinced we should be doing any special env logic in phases other than launch? It seems to me that we are doing something slightly unorthodox at launch time (merging `PATH` from `Config.Env` with `PATH` from registry environment) to get around a specific issue that we introduced.

Generally values in `Config.Env` are expected to override registry values in windows containers. To avoid surprising I think we should deviate from the default behavior only where necessary. Therefore if someone set `PATH` in `Config.Env` on a builder image I would expect it to override the registry value.

cmd/flags.go
```diff
@@ -206,3 +209,16 @@ func EnvOrDefault(key string, defaultVal string) string {
 	}
 	return defaultVal
 }
+
+func PrepareOSEnv() error {
```
ekcasey, line 213: Why is this a separate step instead of simply being part of `NewLaunchEnv` (and `NewBuildEnv` depending on what decisions we make about the build environment). I am inferring that the idea is to ensure registry values are set in the environment before we read environment config (e.g `CNB_PLATFORM_API`), but isn't this already handled by the container runtime?

micahyoung:
I'm going to close this PR based on an alternate approach we discussed at the [sub-team sync](https://docs.google.com/document/d/1zBYJsBwcwLZ5huG4nt7t7kYqaL1W_J12WuLsr2a9mAo/edit#heading=h.hs9uvsn4ks2c) (TL;DR make an RFC like "config.env.PATH MUST be set on run/build images"). Hoping this will clean up the my force-push-spam I created on other repos but if not, I may request deletion of this PR.

PR #405, opened 2020-08-26
Fallback to anonymous auth if default keychain fails to resolve

### LINKED ISSUE #339: Failure to set `CNB_REGISTRY_AUTH` results in failure to initialize cache in some scenarios
### Summary
Today if you run `analyzer some-image` with no flags you will get the following error:
```
ERROR: failed to get previous image: /root/.docker/config.json: stat /root/.docker/config.json: permission denied
```

The following succeeds: `analyzer --env CNB_REGISTRY_AUTH={} some-image`

---

##### Expected
`analyzer some-image` without any flags should succeed.

---

### Context

There are two places where this error can occur:

* `analyzeArgs.analyze()` (registry case)
  * When looking up the previous image - failure to set `CNB_REGISTRY_AUTH` causes us to use the default keychain, we then to proceed to stat `/root/.docker/config.json` after dropping privileges
* `analyzeCmd.Exec()` -> `cache.NewImageCacheFromName()` (cache image case)
  * Same situation as above but occurs when setting up the cache

This doesn't crop up in the wild because in `pack` we always set `CNB_REGISTRY_AUTH` when using the registry. In Tekton we override the docker config directory. But I am sure it will confuse some people. We should fix this.

##### lifecycle version
It is an issue as of v0.8.1

natalieparellano:
Fixes #339

cc @yaelharel

### DIFF
--- a/auth/env_keychain_test.go
+++ b/auth/env_keychain_test.go
@@ -51,92 +51,92 @@ func testEnvKeychain(t *testing.T, when spec.G, it spec.S) {
 				h.AssertEq(t, envKeyChain, authn.DefaultKeychain)
 			})
 		})
+	})
 
-		when("#EnvKeychain", func() {
-			var envKeyChain authn.Keychain
+	when("#EnvKeychain", func() {
+		var envKeyChain authn.Keychain
 
-			when("#Resolve", func() {
+		when("#Resolve", func() {
+			it.Before(func() {
+				envKeyChain = &auth.EnvKeychain{EnvVar: "CNB_REGISTRY_AUTH"}
+			})
+
+			it.After(func() {
+				err := os.Unsetenv("CNB_REGISTRY_AUTH")
+				h.AssertNil(t, err)
+			})
+
+			when("valid auth env variable is set", func() {
 				it.Before(func() {
-					envKeyChain = &auth.EnvKeychain{EnvVar: "CNB_REGISTRY_AUTH"}
-				})
-
-				it.After(func() {
-					err := os.Unsetenv("CNB_REGISTRY_AUTH")
+					err := os.Setenv(
+						"CNB_REGISTRY_AUTH",
+						`{"basic-registry.com": "Basic asdf=", "bearer-registry.com": "Bearer qwerty="}`,
+					)
 					h.AssertNil(t, err)
 				})
 
-				when("valid auth env variable is set", func() {
-					it.Before(func() {
-						err := os.Setenv(
-							"CNB_REGISTRY_AUTH",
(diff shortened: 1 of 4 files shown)


auth/env_keychain_test.go
```diff
@@ -46,97 +46,154 @@ func testEnvKeychain(t *testing.T, when spec.G, it spec.S) {
...
+				if ok != true {
+					t.Fatalf("expected *auth.FallbackKeychain, got %s", reflect.TypeOf(keychain))
+				}
 			})
 		})
+	})
 
-		when("#EnvKeychain", func() {
-			var envKeyChain authn.Keychain
+	when("#EnvKeychain", func() {
```
natalieparellano, line 59: Moved this context out of #NewKeychain (no other changes)

auth/env_keychain_test.go
```diff
@@ -46,97 +46,151 @@ func testEnvKeychain(t *testing.T, when spec.G, it spec.S) {
...
+				var dockerConfigEnv string
+				it.Before(func() {
+					dockerConfigEnv = os.Getenv("DOCKER_CONFIG")
+					// authn.DefaultKeychain will error if DOCKER_CONFIG directory is not readable
+					h.AssertNil(t, os.Setenv("DOCKER_CONFIG", "does-not-exist"))
+				})
+				it.After(func() {
+					h.AssertNil(t, os.Setenv("DOCKER_CONFIG", dockerConfigEnv))
+				})
+				it("returns an anonymous authenticator", func() {
```
ekcasey, line 185: This test doesn't actually require `auth.FallbackKeychain` to succeed. `authn.DefaultKeychain` already return `authn.Anonymous` rather than an error when the docker config directory does not exist. Rather, I believe this issue occurs when we cannot determine whether or not the dir exists because it is located in `/home/root` and we do not have permissions to read `/home/root` after dropping privileges.

https://github.com/google/go-containerregistry/blob/4e3ad4635a12aa559961ef1c8cab4af8054d04b6/vendor/github.com/docker/cli/cli/config/config.go#L102-L106

natalieparellano: Our use of `DOCKER_CONFIG` here is a bit lazy. We were not concerned with how `DefaultKeychain.Resolve()` errored, only that an error occurred. We considered updating `FakeKeychain.Resolve()` to return an error on demand, which also would've satisfied this test. I think what this is exposing is that our rescuing from error cases is too broad. This should go away once we address your [other comment](https://github.com/buildpacks/lifecycle/pull/405#pullrequestreview-475844005).

natalieparellano: Oh, and now that I'm reading your comment again, I see this test is not actually testing what it is supposed to...

yaelharel:
Closing since we'll fix #339 as part of #409.

Issue #390, opened 2020-08-17
Update stack metadata during rebase 


ekcasey:
[RFC 0050](https://github.com/buildpacks/rfcs/blob/main/text/0050-stack-metadata.md)
[RFC PR](https://github.com/buildpacks/rfcs/pull/78)
https://github.com/buildpacks/spec/pull/89

During rebase, update/add/remove the following labels as required to ensure the rebased app image labels match those on the new run image:

* `io.buildpacks.stack.maintainer`
* `io.buildpacks.stack.homepage`
* `io.buildpacks.stack.distro.name`
* `io.buildpacks.stack.distro.version`
* `io.buildpacks.stack.version`
* `io.buildpacks.stack.release_date`
* `io.buildpacks.stack.description`
* `io.buildpacks.stack.metadata`

ekcasey:
Resolved by https://github.com/buildpacks/lifecycle/pull/407

PR #407, opened 2020-08-28
Sync io.buildpacks.stack.* labels from run image during rebase

jabrown85:
- Update matching labels present on both images to value on run image
- Remove matching labels no long present on run image
- Add matching labels that are newly added on run image
- Ignore non-matching labels on both working and run image

Ref #390

### DIFF
--- a/rebaser.go
+++ b/rebaser.go
@@ -72,7 +72,7 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 
 	hasPrefix := func(l string) bool { return strings.HasPrefix(l, "io.buildpacks.stack.") }
 	if err := SyncLabels(newBaseImage, workingImage, hasPrefix); err != nil {
-		return RebaseReport{}, errors.Wrap(err, "set stack metadata label")
+		return RebaseReport{}, errors.Wrap(err, "set stack labels")
 	}
 
 	report := RebaseReport{}
--- a/rebaser.go
+++ b/rebaser.go
@@ -71,7 +71,7 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 	}
 
 	hasPrefix := func(l string) bool { return strings.HasPrefix(l, "io.buildpacks.stack.") }
-	if err := SyncLabels(newBaseImage, workingImage, hasPrefix); err != nil {
+	if err := syncLabels(newBaseImage, workingImage, hasPrefix); err != nil {
 		return RebaseReport{}, errors.Wrap(err, "set stack labels")
 	}
 
--- a/rebaser.go
+++ b/rebaser.go
@@ -3,6 +3,7 @@ package lifecycle
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/buildpacks/imgutil"
 	"github.com/pkg/errors"
@@ -67,6 +68,11 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 
 	if err := workingImage.SetLabel(LayerMetadataLabel, string(data)); err != nil {
 		return RebaseReport{}, errors.Wrap(err, "set app image metadata label")
+	}
+
+	hasPrefix := func(l string) bool { return strings.HasPrefix(l, "io.buildpacks.stack.") }
+	if err := SyncLabels(newBaseImage, workingImage, hasPrefix); err != nil {
+		return RebaseReport{}, errors.Wrap(err, "set stack metadata label")
 	}
 
 	report := RebaseReport{}
--- a/utils.go
+++ b/utils.go
@@ -61,3 +61,45 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
 	}
 	return nil
 }
+
+// SyncLabels adds/updates/removes labels matching the predicate and a source image
+func SyncLabels(sourceImg imgutil.Image, destImage imgutil.Image, test func(string) bool) error {
+	if err := RemoveLabels(destImage, test); err != nil {
+		return err
+	}
+	return CopyLabels(sourceImg, destImage, test)
+}
+
+// RemoveLabels removes labels from an image matching the predicate
+func RemoveLabels(image imgutil.Image, test func(string) bool) error {
+	labels, err := image.Labels()
+	if err != nil {
+		return err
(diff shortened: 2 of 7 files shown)


jabrown85:
Should we `remove(image.config.lablels, key)` instead of setting empty string? I think this will require a change to `imgutil` to add a new `RemoveLabel` function. In go it doesn't matter since the value will be empty string in both situations, but I think removing the label from the config would mean the resulting image does not have the label.

utils.go
```diff
@@ -61,3 +61,45 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
...
+// RemoveLabels removes labels from an image matching the predicate
+func RemoveLabels(image imgutil.Image, test func(string) bool) error {
+	labels, err := image.Labels()
+	if err != nil {
+		return err
+	}
+
+	for label := range labels {
+		if test(label) {
+			if err := image.SetLabel(label, ""); err != nil {
```
yaelharel, line 82: `Labels` is a map from string to string, right?
So why don't we create a `RemoveLabel` function that use `delete` in order to remove a label from the map?

jabrown85: I went ahead and updated it.

utils.go
```diff
@@ -61,3 +61,43 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
...
+// RemoveLabels removes labels from an image matching the predicate
+func RemoveLabels(image imgutil.Image, test func(string) bool) error {
+	labels, err := image.Labels()
+	if err != nil {
+		return err
+	}
+
+	for label := range labels {
+		if test(label) {
+			delete(labels, label)
```
yaelharel, line 82: Thank you for making this change.
Two additional comments:
1. I think it's better that there will be a `RemoveLabel` function (like `SetLabel` on line 97) instead of calling the `delete` function directly in order to hide the implementation of `Labels`.
2. I'm not sure how golang behaves if you delete a `label` for `labels` during a for loop. I'm not sure if it's "smart" enough or we should create a list of "toDelete" labels during the for loop and then call the `RemoveLabel` on each one of them.

jabrown85: 2. In go, it is safe to remove in the loop. [Docs](https://golang.org/ref/spec#For_statements)

As someone who came from C#/Ruby, this feels weird but is totally OK :) 

I'll work on getting a PR up for `imgutil` to add the corresponding `RemoveLabel`. That is the only reason it wasn't done here, the change needs to be done in `imgutil`.

rebaser.go
```diff
@@ -69,6 +70,11 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 		return RebaseReport{}, errors.Wrap(err, "set app image metadata label")
 	}
 
+	hasPrefix := func(l string) bool { return strings.HasPrefix(l, "io.buildpacks.stack.") }
+	if err := SyncLabels(newBaseImage, workingImage, hasPrefix); err != nil {
+		return RebaseReport{}, errors.Wrap(err, "set stack metadata label")
```
ekcasey, line 75:
```suggestion
		return RebaseReport{}, errors.Wrap(err, "set stack labels")
```

utils.go
ekcasey: I don't think this implementation of `RemoveLabels` will actually work for a `remote.Image`. When you call `Labels` on the fake or local image you get a pointer to the actual labels map that will be used to create the image config during save. However, the remote image is backed by a `ggcr` image and the config must be explicitly mutated (Example: https://github.com/buildpacks/imgutil/blob/main/remote/remote.go#L277).

We could make make the behavior of the fake and local image consistent with remote (in a way that would catch errors like this) by returning a copy of the `Labels` map from `image.Labels()` instead of returning the exact map used by the image.

jabrown85:
Opened https://github.com/buildpacks/imgutil/pull/61 to add `RemoveLabel`, which also makes `Labels()` return a copy. This should normalize the API and prevent consumers from thinking `delete()` will produce the image regardless of remote/local.

utils.go
yaelharel: @ekcasey, maybe I missed something but are you saying that we can't add `RemoveLabel` to the `Image` interface?

jabrown85: @yaelharel see https://github.com/buildpacks/imgutil/pull/61 - the implementation is specific to local/remote..so I think Emily is saying returning the actual map for `Labels()` for local/fake images could produce code that relies on that, which would fail on a remote image.

ekcasey: @jabrown85 Yep that is what I meant. Thanks for restating it more clearly.

I think this code as originally PR'ed (without incorporating https://github.com/buildpacks/imgutil/pull/61) falls into that trap. Tests that assert on label removal pass here with the fake image, but would probably not pass with a remote image.

utils.go
```diff
@@ -61,3 +61,43 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
...
+
+// SyncLabels adds/updates/removes labels matching the predicate and a source image
+func SyncLabels(sourceImg imgutil.Image, destImage imgutil.Image, test func(string) bool) error {
+	if err := RemoveLabels(destImage, test); err != nil {
+		return err
+	}
+	return CopyLabels(sourceImg, destImage, test)
+}
+
+// RemoveLabels removes labels from an image matching the predicate
```
natalieparellano, line 73:
```suggestion
// RemoveLabels removes labels matching the predicate from an image 
```

utils.go
```diff
@@ -61,3 +61,43 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
 	}
 	return nil
 }
+
+// SyncLabels adds/updates/removes labels matching the predicate and a source image
+func SyncLabels(sourceImg imgutil.Image, destImage imgutil.Image, test func(string) bool) error {
```
natalieparellano, line 66: Is it possible to add unit tests for these functions?

ekcasey: Agree, or alternatively, if we don't think these are likely to be reusable/useful library functions, we should unexport them.

utils.go
yaelharel: Thank you both! Now it's much clearer!

utils.go
jabrown85: I went ahead and unexported them. I don't think they'll be used outside of the library, so I rather not export them at this time. The test I did add, I think, is probably enough testing given most of the heavy lifting lives in `imgutil`.

PR #409, opened 2020-08-31
Resolve auth before dropping privileges

### LINKED ISSUE #339: Failure to set `CNB_REGISTRY_AUTH` results in failure to initialize cache in some scenarios

natalieparellano:
Fixes #339

Is an alternative to #405

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -54,6 +54,10 @@ func TestAnalyzer(t *testing.T) {
 	defer registry.Stop(t)
 
 	os.Setenv("DOCKER_CONFIG", registry.DockerDirectory)
+	// Copy docker config directory to analyze-image container
+	targetDockerConfig := filepath.Join("testdata", "analyzer", "analyze-image", "container", "docker-config")
+	h.AssertNil(t, os.RemoveAll(filepath.Join(targetDockerConfig, "config.json")))
+	h.RecursiveCopy(t, registry.DockerDirectory, targetDockerConfig)
 
 	// Setup test container
 
@@ -92,7 +96,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		it("errors", func() {
 			cmd := exec.Command(
 				"docker", "run", "--rm",
-				"--env", "CNB_REGISTRY_AUTH={}",
 				analyzeImage,
 				analyzerPath,
 			)
@@ -108,7 +111,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		it("warns", func() {
 			output := h.DockerRun(t,
 				analyzeImage,
-				h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
 				h.WithArgs(analyzerPath, "some-image"),
 			)
 
@@ -123,7 +125,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 
 			output := h.DockerRun(t,
 				analyzeImage,
-				h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
 				h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s some-image; ls -al /layers", analyzerPath)),
 			)
 
@@ -142,7 +143,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				analyzeImage,
 				"/layers",
 				h.WithFlags(
-					"--env", "CNB_REGISTRY_AUTH={}",
 					"--volume", cacheVolume+":"+"/cache", // use a cache so that we can observe the effect of other-group.toml on /layers
 				),
 				h.WithArgs(
@@ -165,9 +165,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				copyDir,
 				analyzeImage,
 				"/some-dir/some-analyzed.toml",
-				h.WithFlags(
-					"--env", "CNB_REGISTRY_AUTH={}",
-				),
 				h.WithArgs(
 					analyzerPath,
 					"-analyzed", "/some-dir/some-analyzed.toml",
@@ -193,7 +190,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 			assertAnalyzedMetadata(t, filepath.Join(copyDir, "analyzed.toml"))
 		})
 
-		when("app image is found", func() {
+		when("app image exists", func() {
 			var appImage string
 
 			it.Before(func() {
@@ -215,8 +212,8 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				h.DockerImageRemove(t, appImage)
 			})
 
-			it("restores layer metadata", func() {
-				h.DockerRunAndCopy(t,
+			it("restores app metadata", func() {
+				output := h.DockerRunAndCopy(t,
 					containerName,
 					copyDir,
 					analyzeImage,
@@ -225,23 +222,12 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					h.WithArgs(analyzerPath, "-daemon", appImage),
 				)
 
-				layerFilenames := []string{
-					"launch-build-cache-layer.sha",
-					"launch-build-cache-layer.toml",
-					"launch-cache-layer.sha",
-					"launch-cache-layer.toml",
-					"launch-layer.sha",
-					"launch-layer.toml",
-					"store.toml",
-				}
-				for _, filename := range layerFilenames {
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", filename))
-				}
+				assertLogsAndRestoresAppMetadata(t, copyDir, output)
 			})
 
 			when("skip layers is provided", func() {
 				it("writes analyzed.toml and does not write buildpack layer metadata", func() {
-					h.DockerRunAndCopy(t,
+					output := h.DockerRunAndCopy(t,
 						containerName,
 						copyDir,
 						analyzeImage,
@@ -256,18 +242,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					)
 
 					assertAnalyzedMetadata(t, filepath.Join(copyDir, "layers", "analyzed.toml"))
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "store.toml"))
-					layerFilenames := []string{
-						"launch-build-cache-layer.sha",
-						"launch-build-cache-layer.toml",
-						"launch-cache-layer.sha",
-						"launch-cache-layer.toml",
-						"launch-layer.sha",
-						"launch-layer.toml",
-					}
-					for _, filename := range layerFilenames {
-						h.AssertPathDoesNotExist(t, filepath.Join(copyDir, "layers", "some-buildpack-id", filename))
-					}
+					assertWritesStoreTomlOnly(t, copyDir, output)
 				})
 			})
 		})
@@ -302,10 +277,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 							copyDir,
 							analyzeImage,
 							"/layers",
-							h.WithFlags(append(
-								variables.DockerSocketMount,
-								"--env", "CNB_REGISTRY_AUTH={}", // In practice, we never set this variable in the daemon case. Setting to avoid failure to stat docker config directory when initializing cache
-							)...),
+							h.WithFlags(variables.DockerSocketMount...),
 							h.WithArgs(
 								analyzerPath,
 								"-daemon",
@@ -337,8 +309,8 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 						h.DockerImageRemove(t, cacheImage)
 					})
 
-					it("uses the provided cache", func() {
-						h.DockerRunAndCopy(t,
+					it("restores cache metadata", func() {
+						output := h.DockerRunAndCopy(t,
 							containerName,
 							copyDir,
 							analyzeImage,
@@ -356,17 +328,16 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 							),
 						)
 
-						h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.sha"))
-						h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.toml"))
+						assertLogsAndRestoresCacheMetadata(t, copyDir, output)
 					})
 				})
 			})
 
 			when("cache directory case", func() {
-				it("uses the provided cache", func() {
+				it("restores cache metadata", func() {
 					cacheVolume = h.SeedDockerVolume(t, cacheFixtureDir)
 
-					h.DockerRunAndCopy(t,
+					output := h.DockerRunAndCopy(t,
 						containerName,
 						copyDir,
 						analyzeImage,
@@ -383,8 +354,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 						),
 					)
 
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.sha"))
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.toml"))
+					assertLogsAndRestoresCacheMetadata(t, copyDir, output)
 				})
 
 				when("the provided cache directory isn't writeable by the CNB user's group", func() {
@@ -462,44 +432,55 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				copyDir,
 				analyzeImage,
 				"/layers/analyzed.toml",
-				h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
 				h.WithArgs(analyzerPath, "some-image"),
 			)
 
 			assertAnalyzedMetadata(t, filepath.Join(copyDir, "analyzed.toml"))
 		})
 
-		when("app image is found", func() {
-			it("restores layer metadata", func() {
-				h.DockerRunAndCopy(t,
-					containerName,
-					copyDir,
-					analyzeImage,
-					"/layers",
-					h.WithFlags(
-						"--network", "host",
-						"--env", "CNB_REGISTRY_AUTH="+appAuthConfig,
-					),
-					h.WithArgs(analyzerPath, appImage),
-				)
+		when("app image exists", func() {
+			when("registry creds are provided in CNB_REGISTRY_AUTH", func() {
+				it("restores app metadata", func() {
+					output := h.DockerRunAndCopy(t,
+						containerName,
+						copyDir,
+						analyzeImage,
+						"/layers",
+						h.WithFlags(
+							"--network", "host",
+							"--env", "CNB_REGISTRY_AUTH="+appAuthConfig,
+						),
+						h.WithArgs(analyzerPath, appImage),
+					)
 
-				layerFilenames := []string{
-					"launch-build-cache-layer.sha",
-					"launch-build-cache-layer.toml",
-					"launch-cache-layer.sha",
-					"launch-cache-layer.toml",
-					"launch-layer.sha",
-					"launch-layer.toml",
-					"store.toml",
-				}
-				for _, filename := range layerFilenames {
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", filename))
-				}
+					assertLogsAndRestoresAppMetadata(t, copyDir, output)
+				})
+			})
+
+			when("registry creds are provided in the docker config.json", func() {
+				it("restores app metadata", func() {
+					output := h.DockerRunAndCopy(t,
+						containerName,
+						copyDir,
+						analyzeImage,
+						"/layers",
+						h.WithFlags(
+							"--env", "DOCKER_CONFIG=/docker-config",
+							"--network", "host",
+						),
+						h.WithArgs(
+							analyzerPath,
+							appImage,
+						),
+					)
+
+					assertLogsAndRestoresAppMetadata(t, copyDir, output)
+				})
 			})
 
 			when("skip layers is provided", func() {
 				it("writes analyzed.toml and does not write buildpack layer metadata", func() {
-					h.DockerRunAndCopy(t,
+					output := h.DockerRunAndCopy(t,
 						containerName,
 						copyDir,
 						analyzeImage,
@@ -516,53 +497,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					)
 
 					assertAnalyzedMetadata(t, filepath.Join(copyDir, "layers", "analyzed.toml"))
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "store.toml"))
-					layerFilenames := []string{
-						"launch-build-cache-layer.sha",
-						"launch-build-cache-layer.toml",
-						"launch-cache-layer.sha",
-						"launch-cache-layer.toml",
-						"launch-layer.sha",
-						"launch-layer.toml",
-					}
-					for _, filename := range layerFilenames {
-						h.AssertPathDoesNotExist(t, filepath.Join(copyDir, "layers", "some-buildpack-id", filename))
-					}
-				})
-			})
-		})
-
-		when("CNB_REGISTRY_AUTH is not provided", func() {
-			when("DOCKER_CONFIG is set", func() {
-				it("succeeds", func() {
-					h.SkipIf(t, runtime.GOOS == "windows", "Can't run bash on Windows")
-
-					// Copy docker config directory to a temp directory to avoid tampering with permissions for other tests.
-					dockerConfig, err := ioutil.TempDir("", "test-docker-config-")
-					defer os.RemoveAll(dockerConfig)
-					h.AssertNil(t, err)
-					h.RecursiveCopy(t, registry.DockerDirectory, dockerConfig)
-					// Some OSes (like macOS) use symlinks for the standard temp dir.
-					// Resolve it so it can be properly mounted by the Docker daemon.
-					dockerConfig, err = filepath.EvalSymlinks(dockerConfig)
-					h.AssertNil(t, err)
-
-					h.DockerRunAndCopy(t,
-						containerName,
-						copyDir,
-						analyzeImage,
-						"/layers",
-						h.WithFlags(
-							"--mount", fmt.Sprintf("type=bind,source=%s,target=/mounted-docker-config", dockerConfig),
-							"--env", "DOCKER_CONFIG=/mounted-docker-config",
-							"--network", "host",
-						),
-						h.WithBash(
-							fmt.Sprintf("chown -R 2222:3333 /mounted-docker-config; %s %s; ls -alR /layers", analyzerPath, appImage), // provide a real app image, so that we can test that the registry is accessible
-						),
-					)
-
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id"))
+					assertWritesStoreTomlOnly(t, copyDir, output)
 				})
 			})
 		})
@@ -586,39 +521,62 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					h.DockerImageRemove(t, cacheImage)
 				})
 
-				it("uses the provided cache", func() {
-					h.DockerRunAndCopy(t,
-						containerName,
-						copyDir,
-						analyzeImage,
-						"/layers",
-						h.WithFlags(
-							"--network", "host",
-							"--env", "CNB_REGISTRY_AUTH="+cacheAuthConfig,
-						),
-						h.WithArgs(
-							analyzerPath,
-							"-cache-image", cacheImage,
-							"some-image",
-						),
-					)
+				when("registry creds are provided in CNB_REGISTRY_AUTH", func() {
+					it("restores cache metadata", func() {
+						output := h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							analyzeImage,
+							"/layers",
+							h.WithFlags(
+								"--network", "host",
+								"--env", "CNB_REGISTRY_AUTH="+cacheAuthConfig,
+							),
+							h.WithArgs(
+								analyzerPath,
+								"-cache-image", cacheImage,
+								"some-image",
+							),
+						)
 
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.sha"))
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.toml"))
+						assertLogsAndRestoresCacheMetadata(t, copyDir, output)
+					})
+				})
+
+				when("registry creds are provided in the docker config.json", func() {
+					it("restores cache metadata", func() {
+						output := h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							analyzeImage,
+							"/layers",
+							h.WithFlags(
+								"--env", "DOCKER_CONFIG=/docker-config",
+								"--network", "host",
+							),
+							h.WithArgs(
+								analyzerPath,
+								"-cache-image",
+								cacheImage,
+								appImage,
+							),
+						)
+
+						assertLogsAndRestoresCacheMetadata(t, copyDir, output)
+					})
 				})
 			})
 
 			when("cache directory case", func() {
-				it("uses the provided cache", func() {
+				it("restores cache metadata", func() {
 					cacheVolume = h.SeedDockerVolume(t, cacheFixtureDir)
 
-					h.DockerRunAndCopy(t,
+					output := h.DockerRunAndCopy(t,
 						containerName,
 						copyDir,
 						analyzeImage,
 						"/layers",
 						h.WithFlags(
-							"--env", "CNB_REGISTRY_AUTH={}",
 							"--volume", cacheVolume+":"+"/cache",
 						),
 						h.WithArgs(
@@ -628,8 +586,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 						),
 					)
 
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.sha"))
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.toml"))
+					assertLogsAndRestoresCacheMetadata(t, copyDir, output)
 				})
 			})
 		})
@@ -648,7 +605,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					analyzeImage,
 					"/",
 					h.WithFlags(
-						"--env", "CNB_REGISTRY_AUTH={}",
 						"--env", "CNB_PLATFORM_API=0.4",
 						"--volume", cacheVolume+":"+"/cache",
 					),
@@ -677,7 +633,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					analyzeImage,
 					"/some-other-layers",
 					h.WithFlags(
-						"--env", "CNB_REGISTRY_AUTH={}",
 						"--env", "CNB_PLATFORM_API=0.5",
 						"--volume", cacheVolume+":"+"/cache",
 					),
@@ -730,4 +685,49 @@ func assertAnalyzedMetadata(t *testing.T, path string) {
 	var analyzedMd lifecycle.AnalyzedMetadata
 	_, err := toml.Decode(string(contents), &analyzedMd)
 	h.AssertNil(t, err)
+}
+
+func assertLogsAndRestoresAppMetadata(t *testing.T, dir, output string) {
+	layerFilenames := []string{
+		"launch-build-cache-layer.sha",
+		"launch-build-cache-layer.toml",
+		"launch-cache-layer.sha",
+		"launch-cache-layer.toml",
+		"launch-layer.sha",
+		"launch-layer.toml",
+		"store.toml",
+	}
+	for _, filename := range layerFilenames {
+		h.AssertPathExists(t, filepath.Join(dir, "layers", "some-buildpack-id", filename))
+	}
+	layerNames := []string{
+		"launch-build-cache-layer",
+		"launch-cache-layer",
+		"launch-layer",
+	}
+	for _, layerName := range layerNames {
+		h.AssertStringContains(t, output, fmt.Sprintf("Restoring metadata for \"some-buildpack-id:%s\"", layerName))
+	}
+}
+
+func assertLogsAndRestoresCacheMetadata(t *testing.T, dir, output string) {
+	h.AssertPathExists(t, filepath.Join(dir, "layers", "some-buildpack-id", "some-layer.sha"))
+	h.AssertPathExists(t, filepath.Join(dir, "layers", "some-buildpack-id", "some-layer.toml"))
+	h.AssertStringContains(t, output, "Restoring metadata for \"some-buildpack-id:some-layer\" from cache")
+}
+
+func assertWritesStoreTomlOnly(t *testing.T, dir, output string) {
+	h.AssertPathExists(t, filepath.Join(dir, "layers", "some-buildpack-id", "store.toml"))
+	layerFilenames := []string{
+		"launch-build-cache-layer.sha",
+		"launch-build-cache-layer.toml",
+		"launch-cache-layer.sha",
+		"launch-cache-layer.toml",
+		"launch-layer.sha",
+		"launch-layer.toml",
+	}
+	for _, filename := range layerFilenames {
+		h.AssertPathDoesNotExist(t, filepath.Join(dir, "layers", "some-buildpack-id", filename))
+	}
+	h.AssertStringContains(t, output, "Skipping buildpack layer analysis")
 }
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -478,21 +478,6 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 	when("registry case", func() {
 		var appImage, appAuthConfig string
 
-		it.Before(func() {
-			metadata := minifyMetadata(t, filepath.Join("testdata", "analyzer", "app_image_metadata.json"), lifecycle.LayersMetadata{})
-			appImage, appAuthConfig = buildRegistryImage(
-				t,
-				"some-app-image-"+h.RandString(10),
-				filepath.Join("testdata", "analyzer", "app-image"),
-				"--build-arg", "fromImage="+variables.ContainerBaseImage,
-				"--build-arg", "metadata="+metadata,
-			)
-		})
-
-		it.After(func() {
-			h.DockerImageRemove(t, appImage)
-		})
-
 		it("writes analyzed.toml", func() {
 			h.DockerRunAndCopy(t,
 				containerName,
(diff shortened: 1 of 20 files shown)


acceptance/analyzer_test.go
```diff
@@ -54,6 +54,10 @@ func TestAnalyzer(t *testing.T) {
 	defer registry.Stop(t)
 
 	os.Setenv("DOCKER_CONFIG", registry.DockerDirectory)
+	// Copy registry docker directory to analyze-image container
```
natalieparellano, line 57: Before, we were mounting the docker config, and (on Mac) Docker Desktop was translating the uid and gid of the directory to be the CNB user after dropping privileges.

acceptance/analyzer_test.go
```diff
@@ -529,39 +533,31 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+								fmt.Sprintf(
+									"chown -R root /registry-docker-config; chmod -R 700 /registry-docker-config; %s %s",
+									analyzerPath,
+									appImage,
+								),
+							),
+						)
 
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id"))
+						h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id")) // the presence of some-buildpack-id indicates that we were able to retrieve the app image from the registry
```
natalieparellano, line 559: This test might be more intelligible if we asserted on the log output, e.g., it could assert that the output contains something like `Restoring metadata for "some-buildpack-id:launch-layer" from app image`.

acceptance/analyzer_test.go
```diff
@@ -605,6 +601,35 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+									fmt.Sprintf(
+										"chown -R root /registry-docker-config; chmod -R 700 /registry-docker-config; %s -cache-image %s %s",
+										analyzerPath,
+										cacheImage,
+										appImage,
+									),
+								),
+							)
+
+							h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.sha"))
```
natalieparellano, line 628: Same comment as [here](https://github.com/buildpacks/lifecycle/pull/409/files#r480407637) but for the cache.

auth/env_keychain.go
```diff
@@ -12,15 +12,30 @@ import (
 	"github.com/pkg/errors"
 )
 
-// NewKeychain returns either a EnvKeychain or a authn.DefaultKeychain depending on whether the provided environment variable is set
-func NewKeychain(envVar string) authn.Keychain {
+// NewKeychain returns an EnvKeychain based on the provided environment variable.
+// If the provided environment variable is not set, NewKeychain will build an auth config
+// from authn.DefaultKeychain and set the variable in the environment.
+// If any errors are encountered, NewKeychain will return an anonymous keychain.
+func NewKeychain(envVar string, images ...string) authn.Keychain {
```
natalieparellano, line 19: We are wondering if `NewKeychain` is still a good name for this function. It seems like `GetKeychain` might be better. We are assuming that we can't change it, since it's exported, but maybe we're wrong.

auth/env_keychain_test.go
```diff
@@ -36,107 +38,191 @@ func testEnvKeychain(t *testing.T, when spec.G, it spec.S) {
...
+			it.Before(func() {
+				h.AssertNil(t, os.Unsetenv("CNB_REGISTRY_AUTH"))
+
+				tmpDir, err = ioutil.TempDir("", "")
+				h.AssertNil(t, err)
+				h.AssertNil(t, os.Setenv("DOCKER_CONFIG", tmpDir))
 			})
-		})
 
-		when("#EnvKeychain", func() {
```
natalieparellano, line 55: We pulled `#EnvKeychain` out of `#NewKeychain`'s scope in 42b9d21.

cmd/lifecycle/analyzer.go
```diff
@@ -65,6 +65,7 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
 }
 
 func (a *analyzeCmd) Privileges() error {
+	a.setEnvRegistryAuth()
```
natalieparellano, line 68: We considered exporting this function and making it part of the interface, but not every phase needs it, so we did it this way.

cmd/lifecycle/analyzer.go
```diff
@@ -107,6 +108,17 @@ func (a *analyzeCmd) Exec() error {
 	return nil
 }
 
+func (a *analyzeCmd) setEnvRegistryAuth() {
```
natalieparellano, line 111: The addition of this function is covered in the analyzer acceptance tests. All other phases are untested...

auth/env_keychain.go
```diff
@@ -12,15 +12,30 @@ import (
...
 	_, ok := os.LookupEnv(envVar)
 	if !ok {
-		return authn.DefaultKeychain
+		authConfig, err := BuildEnvVar(authn.DefaultKeychain, images...)
```
natalieparellano, line 22: If `BuildEnvVar` fails to read the docker config, it will return an empty auth config (instead of erroring). This makes it difficult to log at this stage.

natalieparellano: Note that we already warn when we fail to find the app image.

acceptance/analyzer_test.go
ekcasey: Can we do both? Also, is there a reason we check for specific layer file names when using env auth but not when using docker config auth? (e.g. `launch-layer.toml`)

auth/env_keychain.go
natalieparellano: On second thought, maybe it's better if we don't modify the behavior of `NewKeychain` and create a separate function (say, `SetEnvVar`) that we can call before dropping privileges.

acceptance/analyzer_test.go
```diff
@@ -605,6 +601,35 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.sha"))
 					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.toml"))
 				})
+
```
ekcasey, line 604: We don't necessarily need to block this PR on it (given it might require some significant testhelper work) but it would be good to have an acceptance tests that uses anonymous auth, given that anonymous auth has been a source of bugs.

acceptance/analyzer_test.go
```diff
@@ -605,6 +601,35 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+				when("CNB_REGISTRY_AUTH is not provided", func() {
+					when("DOCKER_CONFIG is set", func() {
```
ekcasey, line 606: The `CNB_REGISTRY_AUTH is not provided` context feel strange given that we don't have a `CNB_REGISTRY_AUTH is provided` context. Maybe we can get rid of the wrapping "not provided" context and change this `when` and the `when` of the above test to describe how credentials are provided.

Maybe something like... `when("registry creds are provided in the docker config.json"` vs `when("registry creds  are provided in CNB_REGISTRY_AUTH"`?

auth/env_keychain.go
ekcasey: I am okay omitting the warning entirely for now. I only really wanted to warn if something strange has happened (e.g. there is a docker `config.json` but we can't read it).

In addition I agree that we should add a log message for a missing cache image, as you suggested. However, I think it should be an info message rather than a warning because the cache image would normally be non-existent on a first build and we shouldn't produce warnings in a happy path situation.

acceptance/analyzer_test.go
```diff
@@ -54,6 +54,10 @@ func TestAnalyzer(t *testing.T) {
 	defer registry.Stop(t)
 
 	os.Setenv("DOCKER_CONFIG", registry.DockerDirectory)
+	// Copy registry docker directory to analyze-image container
+	targetDockerConfig := filepath.Join("testdata", "analyzer", "analyze-image", "container", "registry-docker-config")
```
ekcasey, line 58: Tests appear to be failing b/c the `registry-docker-config` directory does not exist in the image. Missing `.gitkeep`?

cmd/lifecycle/analyzer.go
ekcasey: I like it as is. `Privileges` still feels like the right high-level abstraction to me.

acceptance/analyzer_test.go
natalieparellano: Nope, we can make it consistent.

auth/env_keychain.go
natalieparellano: We went ahead and made this change (`SetEnvVar`), it seemed like the right thing to do and allowed us to delete some code.

acceptance/analyzer_test.go
```diff
@@ -529,39 +533,31 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+							analyzeImage,
+							"/layers",
+							h.WithFlags(
+								"--env", "DOCKER_CONFIG=/registry-docker-config",
+								"--network", "host",
+							),
+							h.WithBash(
+								// ensure docker config directory is root owned and NOT world readable
+								fmt.Sprintf(
+									"chown -R root /registry-docker-config; chmod -R 700 /registry-docker-config; %s %s",
```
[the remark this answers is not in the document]
natalieparellano: YES!

auth/env_keychain.go
```diff
@@ -12,15 +12,30 @@ import (
...
+		if err != nil {
+			return &AnonymousKeychain{}
+		}
+		if err = os.Setenv(envVar, authConfig); err != nil {
+			return &AnonymousKeychain{}
+		}
```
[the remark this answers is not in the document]
natalieparellano: > putting credentials in the environment when the user doesn't expect it could create security risks
Agree on this point. How bad is it really (just wondering...)?

We initially thought about doing it this way but it requires a lot more changes to the existing code as we'll now need `analyzeCmd`, etc. to hold on to a `keychain` if I'm not mistaken.

acceptance/analyzer_test.go
natalieparellano: Actually on reviewing this again we think that for tests that are only verifying that we successfully connected to the registry (either for the app or the cache) it isn't important to check the specific files. For tests that are verifying the usage of the registry image, it is important. We can make things consistent with this philosophy, or we can just check specific files everywhere.

auth/env_keychain.go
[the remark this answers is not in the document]
ekcasey: > How bad is it really (just wondering...)?

As part of lifecycle binaries like `analyzer`, `exporter`, and `creator`? Not that bad b/c we are in control over how the environment is used by the lifecycle and we explicitly strip `CNB_REGISTRY_AUTH` from the buildpack environment when `creator` is used.

However, `NewKeychain` is part of the library interface meaning that folks could be pulling in this code in unpredictable situations. This behavior would likely be surprising to any potential Consumers of `NewKeychain` (I know I would be surprised) and may open folks up to vulnerabilities.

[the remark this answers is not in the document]
natalieparellano: Ah, then our latest commit (fff589e) may be helpful (in response to this [comment](https://github.com/buildpacks/lifecycle/pull/409#discussion_r481211308)). We ultimately realized that changing the library method's behavior was not a good idea. Do you think the latest incarnation is okay? Or should we proceed with re-working this to store the auth in memory?

acceptance/analyzer_test.go
ekcasey: IMO seems worth checking at least one launch layer metadata file to verify that we connected to the registry. Practically, you are right that this test will ensure that our analyzer connects successfully, but reading through the test I found it momentarily confusing that nothing being asserted upon actually required data from the image.

Given that all of the buildpack IDs are provided in the `group.toml` an implementation of `analyzer` could hypothetically decided to create `<layers>/<buildpack ID>` directories even when there is no app image.

natalieparellano: We're going with consistency :)

auth/env_keychain.go
natalieparellano: We looked into this a bit more - the line that I linked to was incorrect because we don't have a logger at that point in the code and furthermore, `DecodeLabel` does not error if the image is not found. We extended the cache interface with an `Exists()` method to get this information at the point where we have a logger.

auth/env_keychain.go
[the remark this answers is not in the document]
ekcasey: TBH setting the env var does bother me a little. However this change goes a long way towards addressing any concrete problems I can foresee arising. Because I agree the changing the library method behavior in a patch is probably a bad idea I think this solution makes sense for `v0.9.2` as is.

However, before we merge this fix into v0.10.0 I think we should:

1. Make `NewKeychain` return a `MultiKeychain` where
   * The `EnvKeychain` or a refactored equivalent is the first keychain
  * A keychain that stores pre-resolved auth-headers from the docker config.json in memory is the second keychain

ekcasey:
@natalieparellano @yaelharel I created an `release/0.9.x` branch from which we can cut `v0.9.2` pointed this PR at that branch.

acceptance/analyzer_test.go
natalieparellano: Yes, I .gitignore'd it :/

ekcasey:
Moving to `v0.10.0` after discussion. The best solution involves behavior changes that go beyond what we would typically include in a patch.

testhelpers/docker.go
```diff
@@ -49,12 +49,13 @@ func DockerRun(t *testing.T, image string, ops ...DockerCmdOp) string {
 	return Run(t, exec.Command("docker", append([]string{"run", "--rm"}, args...)...))
 }
 
-func DockerRunAndCopy(t *testing.T, containerName, copyDir, image, path string, ops ...DockerCmdOp) {
+func DockerRunAndCopy(t *testing.T, containerName, copyDir, image, path string, ops ...DockerCmdOp) string {
```
yaelharel, line 52: Why did we make this change?
I don't see that we're using the output of this function anywhere.

natalieparellano: IIRC we were using it for debugging...

cmd/lifecycle/analyzer.go
```diff
@@ -75,6 +78,12 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
 }
 
 func (a *analyzeCmd) Privileges() error {
+	keychain, err := a.resolveKeychain() // TODO: call resolveKeychain only if needed (useDaemon==false or using a cache image)
```
natalieparellano, line 81: Thinking about this more, I don't think we want to gate calling `a.resolveKeychain()` on this logic as it'd essentially be duplicating the logic in `resolveKeychain()` itself. Maybe things are okay as they are - `ResolveKeychain` will not even try to read the `DOCKER_CONFIG` directory if `CNB_REGISTRY_AUTH` is unset.

acceptance/analyzer_test.go
```diff
@@ -54,6 +54,10 @@ func TestAnalyzer(t *testing.T) {
 	defer registry.Stop(t)
 
 	os.Setenv("DOCKER_CONFIG", registry.DockerDirectory)
+	// Copy docker config directory to analyze-image container
+	targetDockerConfig := filepath.Join("testdata", "analyzer", "analyze-image", "container", "docker-config")
+	h.AssertNil(t, os.RemoveAll(filepath.Join(targetDockerConfig, "config.json")))
+	h.RecursiveCopy(t, registry.DockerDirectory, targetDockerConfig)
```
natalieparellano, line 60: Now we can delete all occurrences of `"--env", "CNB_REGISTRY_AUTH={}",` in this file : )

auth/env_keychain.go
```diff
@@ -12,27 +12,33 @@ import (
...
 )
 
-// NewKeychain returns either a EnvKeychain or a authn.DefaultKeychain depending on whether the provided environment variable is set
-func NewKeychain(envVar string) authn.Keychain {
+// ResolveKeychain returns either:
+// * a resolved keychain from the ggcr DefaultKeychain or
+// * a multi-keychain with a resolved keychain from the provided environment variable, and a resolved keychain from the ggcr DefaultKeychain
+// depending on whether the provided environment variable is set
+func ResolveKeychain(envVar string, images ...string) (authn.Keychain, error) {
+	defaultKeychain := resolvedKeychain{Auths: buildEnvMap(authn.DefaultKeychain, images...)}
```
natalieparellano, line 20: Summarizing other conversation - since `buildEnvMap` accepts an interface, we could set a package level variable to designate the "default keychain" that will be passed here, and set it to `authn.DefaultKeychain` in our implementation. Then in the tests, we could set that variable to be a fake keychain that implements `authn.Keychain` (this will avoid having main care about this variable).

To simulate the error that occurs when the `DOCKER_CONFIG` directory is not readable, we can have our fake keychain return an error when `Resolve()` is called ([here](https://github.com/buildpacks/lifecycle/blob/ca47d3defd627d7ebb16e5e3b98bb9cc9b2d7ca4/auth/env_keychain.go#L159)). Then we can remove all of the test setup that tried to modify the `DOCKER_CONFIG` directory. Our tests don't need to care at all about `DOCKER_CONFIG`, only that the "default keychain" returned an error.

natalieparellano: We actually already have a fake keychain for other tests that implements the interface: https://github.com/buildpacks/lifecycle/blob/ca47d3defd627d7ebb16e5e3b98bb9cc9b2d7ca4/auth/env_keychain_test.go#L195

I feel like a quick way to get these tests going would be to add a `returnsForResolve` field on the struct (though it's a little awkward because `Resolve()` returns two values...). Then the `Resolve()` method could check that field, and if set return whatever was provided.

acceptance/analyzer_test.go
yaelharel: I thought that we already did it.
I'll delete it!

cmd/lifecycle/analyzer.go
```diff
@@ -25,6 +26,9 @@ type analyzeCmd struct {
 
 	//flags: paths to write data
 	analyzedPath string
+
+	//construct before dropping privileges
+	keychain authn.Keychain
```
yaelharel, line 31: I'm not sure whether it's better to have the `keychain` here or as part of the `analyzeArgs` struct.
We thought about putting it as part of `analyzeCmd` to be consistent with other commands that don't have the `args` structure (such as the `creator`), but then we need to pass the `keychain` as an argument to the `analyze` function which feels a bit weird.
@ekcasey , what do you think?

cmd/lifecycle/exporter.go
```diff
@@ -33,6 +34,9 @@ type exportCmd struct {
 
 	//flags: paths to write outputs
 	analyzedPath string
+
+	//construct before dropping privileges
+	keychain authn.Keychain
```
yaelharel, line 39: Same comment as https://github.com/buildpacks/lifecycle/pull/409/files#r540433488
But here we need to pass the keychain as an argument to two functions: `export` and `initRemoteImage`.

cmd/lifecycle/analyzer.go
yaelharel: Maybe we should add an `args` struct to the other functions as well? It seems like an overkill...

yaelharel: See https://github.com/buildpacks/lifecycle/pull/409/files#r540434516 for another case.

acceptance/analyzer_test.go
yaelharel: @ekcasey @natalieparellano given that we want to cut a release tomorrow (or on Monday), should we just create an issue for that?

cmd/lifecycle/restorer.go
```diff
@@ -61,13 +73,20 @@ func (r *restoreCmd) Exec() error {
...
+	cacheStore, err := initCache(r.cacheImageTag, r.cacheDir, r.keychain)
 	if err != nil {
 		return err
 	}
 	return restore(r.layersDir, group, cacheStore)
 }
 
+func (r *restoreCmd) resolveKeychain() (authn.Keychain, error) {
+	if r.cacheImageTag == "" {
+		return authn.DefaultKeychain, nil // fix the keychain
```
yaelharel, line 85: Don't we need to fix it?

auth/env_keychain.go
yaelharel: @natalieparellano, I mark this conversation as resolved.
Feel free to reopen it if you think that I missed something.

cmd/lifecycle/analyzer.go
yaelharel: @natalieparellano, I'm marking this conversation as resolved.
Feel free to reopen it if you think that I missed something.

auth/env_keychain.go
yaelharel: @natalieparellano, marking this conversation as resolved.
Feel free to reopen if needed.

auth/env_keychain.go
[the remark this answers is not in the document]
yaelharel: @ekcasey @natalieparellano, I think we addressed all the comments here so I mark this conversation as resolved.
Please reopen if I missed something.

auth/env_keychain.go
```diff
@@ -12,15 +12,30 @@ import (
 	"github.com/pkg/errors"
 )
 
-// NewKeychain returns either a EnvKeychain or a authn.DefaultKeychain depending on whether the provided environment variable is set
-func NewKeychain(envVar string) authn.Keychain {
+// NewKeychain returns an EnvKeychain based on the provided environment variable.
+// If the provided environment variable is not set, NewKeychain will build an auth config
+// from authn.DefaultKeychain and set the variable in the environment.
+// If any errors are encountered, NewKeychain will return an anonymous keychain.
```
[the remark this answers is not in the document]
yaelharel: @ekcasey, marking this conversation as resolved.
Feel free to reopen if you think it's still relevant.

auth/env_keychain.go
```diff
@@ -12,27 +12,37 @@ import (
 	"github.com/pkg/errors"
 )
 
-// NewKeychain returns either a EnvKeychain or a authn.DefaultKeychain depending on whether the provided environment variable is set
-func NewKeychain(envVar string) authn.Keychain {
+// ResolveKeychain returns either:
+// * a resolved keychain from the ggcr DefaultKeychain or
+// * a multi-keychain with a resolved keychain from the provided environment variable, and a resolved keychain from the ggcr DefaultKeychain
+// depending on whether the provided environment variable is set
+func ResolveKeychain(envVar string, images []string, overrideKeychain ...authn.Keychain) (authn.Keychain, error) {
```
jromero, line 19: This seems like an awkward way of providing an option argument. From the interface alone I may infer that I can pass multiple keychain and assume that they would be "merged" somehow but that doesn't appear to be the case.

As an example:

```
auth.ResolveKeychain("SOME_VAR", myImage, keychain1, keychain2);
```

natalieparellano: I think we should refactor this function to take "ops" functions, like [here](https://github.com/buildpacks/lifecycle/blob/63c5c6e6a2f5c6617781ed6d42a80b1c3918730b/testhelpers/docker_cmd.go#L28).

Then when this function is called from `main` it could look like `auth. ResolveKeychain(cmd.EnvRegistryAuth, auth.WithImages(registryImages))` and in the tests it could look like `auth.ResolveKeychain("CNB_REGISTRY_AUTH", auth.WithDefault(fakeKeychain))`.

cmd/lifecycle/restorer.go
natalieparellano: It still feels weird that we call `resolveKeychain()` when we don't actually need a keychain. Maybe it's better for each command to have a helper function like `registryImages()` that returns the images being used and if this is non-empty then we resolve the auth.

So something like 
```
	registryImages := a.registryImages()
	if len(registryImages) > 0 {
		var err error
		a.keychain, err = auth.ResolveKeychain(cmd.EnvRegistryAuth, registryImages)
		if err != nil {
			return cmd.FailErr(err, "resolve keychain")
		}
	}
```
in the `Privileges()` function (very similar to how we initialize the docker client).

auth/env_keychain_test.go
```diff
@@ -17,125 +17,112 @@ import (
...
-					it("returns an Anonymous authenticator when the environment does not have a auth header", func() {
-						envKeyChain = auth.NewKeychain("CNB_REGISTRY_AUTH")
-						registry, err := name.NewRegistry("no-env-auth-registry.com", name.WeakValidation)
-						h.AssertNil(t, err)
+					when("the environment does not have an auth header", func() {
+						it("uses the fallback keychain", func() {
+							useFallbackKeychain(t)
+						})
+
+						when("the fallback keychain does not have an auth header", func() {
```
natalieparellano, line 80: IMO our use of `fallback keychain` here is a bit confusing, that is just a name that we thought of because we had too many variables in our function but it doesn't have any real semantic meaning... I think "default keychain" would be clearer.

auth/env_keychain_test.go
```diff
@@ -208,3 +200,76 @@ type providedAuth struct {
 func (p *providedAuth) Authorization() (*authn.AuthConfig, error) {
 	return p.config, nil
 }
+
+func useFallbackKeychain(t *testing.T) {
```
natalieparellano, line 204: To mirror what we did in the acceptance tests, maybe something like `assertUsesDefaultKeychain`?

auth/env_keychain_test.go
```diff
@@ -208,3 +200,76 @@ type providedAuth struct {
...
+	authenticator, err := resolvedKeychain.Resolve(registry)
+	h.AssertNil(t, err)
+
+	header, err := authenticator.Authorization()
+	h.AssertNil(t, err)
+
+	h.AssertEq(t, header, &authn.AuthConfig{Auth: "asdf="})
+}
+
+func fallbackKeychainNoAuthHeader(t *testing.T) {
```
natalieparellano, line 231:
```suggestion
func assertDefaultKeychainWithoutHeaderReturnsAnonymous(t *testing.T) {
```
...this function name is starting to feel very long :/

auth/env_keychain_test.go
```diff
@@ -208,3 +200,76 @@ type providedAuth struct {
...
+	authenticator, err := resolvedKeychain.Resolve(registry)
+	h.AssertNil(t, err)
+
+	_, err = authenticator.Authorization()
+	h.AssertNil(t, err)
+
+	h.AssertEq(t, authenticator, authn.Anonymous)
+}
+
+func fallbackKeychainNotResolved(t *testing.T) {
```
natalieparellano, line 254:
```suggestion
func assertDefaultKeychainErrorReturnsAnonymous(t *testing.T) {
```

auth/env_keychain_test.go
yaelharel: I'm fine with the `fallback keychain` term since we're falling to this keychain when the environment keychain doesn't work.
The problem that I have with the `default keychain` term is that there is a `DefaultKeychain` struct in `authn` and we're not using it here so I think it might be confusing.
What do you think?

auth/env_keychain_test.go
yaelharel: I don't have problems with long function names if they make the code more readable

acceptance/analyzer_test.go
ekcasey: I assume this is tricky b/c the registry requires auth by default?

auth/env_keychain.go
```diff
@@ -12,27 +12,37 @@ import (
 	"github.com/pkg/errors"
 )
 
-// NewKeychain returns either a EnvKeychain or a authn.DefaultKeychain depending on whether the provided environment variable is set
-func NewKeychain(envVar string) authn.Keychain {
+// ResolveKeychain returns either:
+// * a resolved keychain from the ggcr DefaultKeychain or
+// * a multi-keychain with a resolved keychain from the provided environment variable, and a resolved keychain from the ggcr DefaultKeychain
+// depending on whether the provided environment variable is set
```
ekcasey, line 18: 1) I previously used `*` bullet points in doc comments, but I recently realized it looks _terrible_ https://godoc.org/github.com/buildpacks/lifecycle/layers#Factory.SliceLayers.

2) I found this comment very hard to understand the first time I read it. I think if we leave out some of the if/then logic here it is easier to understand. Whether or not the env keychain is technically included in the multi keychain doesn't matter to the consumer? It only matter that if the env var is set that credential source takes precedence.

Maybe something like the following?
```suggestion
// ResolveKeychain returns a keychain containing authentication configuration for the given images from the following sources, if they exist, in order of precedence: the provided environment variable and the docker config.json file.
```

auth/env_keychain.go
```diff
@@ -77,32 +87,38 @@ func ReadEnvVar(envVar string) (map[string]string, error) {
 	return authMap, nil
 }
 
-// BuildEnvVar creates the contents to use for authentication environment variable.
-//
-// Complementary to `ReadEnvVar`.
-func BuildEnvVar(keychain authn.Keychain, images ...string) (string, error) {
+func buildEnvMap(keychain authn.Keychain, images ...string) map[string]string {
```
ekcasey, line 90: `buildEnvMap` -> `buildAuthMap`? `ResolveKeychain` uses this in ways that have nothing to do with the env var.

acceptance/analyzer_test.go
natalieparellano: Hmmm... actually we have `NewDockerRegistry()` instead of `NewDockerRegistryWithAuth()` - we'll just need to deploy two registries.

auth/env_keychain_test.go
```diff
@@ -208,3 +200,76 @@ type providedAuth struct {
...
+		},
+	}}
+
+	resolvedKeychain, err := auth.ResolveKeychain(
+		"CNB_REGISTRY_AUTH",
+		[]string{"no-env-auth-registry.com/some-image"},
+		&fakeKeychain,
+	)
+	h.AssertNil(t, err)
+	h.AssertNil(t, os.Unsetenv("CNB_REGISTRY_AUTH")) // unset variable to prove that our test is using a pre-resolved keychain
```
ekcasey, line 217: instead of moving all the test logic into these helpers could we reduce duplicated by instead, moving shared chunks like this into an `it.Before`? This seems like a `when("there is a fallback keychain", func(){}` to me and that organization would be more in keeping with the conventions in our other tests.

cmd/lifecycle/analyzer.go
ekcasey: In my mind this seems most analogous to the pre-connected `docker` client which we put on the `analyzeArgs` I think it makes sense to move it there.

cmd/lifecycle/exporter.go
```diff
@@ -153,15 +163,27 @@ func (e *exportCmd) Exec() error {
...
+}
+
+func (e *exportCmd) resolveKeychain() (authn.Keychain, error) {
+	var registryImages []string
+	if e.cacheImageTag != "" {
+		registryImages = append(registryImages, e.cacheImageTag)
+	}
+	if !e.useDaemon {
+		registryImages = append(registryImages, e.imageNames...)
+		registryImages = append(registryImages, e.runImageRef)
```
ekcasey, line 181: I think we have a small problem in that, at this point, e.runImageRef might be empty and the best run image is not located until https://github.com/buildpacks/lifecycle/pull/409/files#diff-92c6ed289675df1e2edf598694dfc99f22a591496c130d10a43e86830cbd3686R193

We may need to reorganize the run image logic a bit.

cmd/lifecycle/exporter.go
```diff
@@ -249,7 +268,7 @@ func (ea exportArgs) export(group lifecycle.BuildpackGroup, cacheStore lifecycle
 	return nil
 }
 
-func initDaemonImage(imagName string, runImageRef string, analyzedMD lifecycle.AnalyzedMetadata, launchCacheDir string, docker client.CommonAPIClient) (imgutil.Image, string, error) {
+func (ea exportArgs) initDaemonImage(runImageRef string, analyzedMD lifecycle.AnalyzedMetadata) (imgutil.Image, string, error) {
```
ekcasey, line 271:
```suggestion
func (ea exportArgs) initDaemonAppImage(runImageRef string, analyzedMD lifecycle.AnalyzedMetadata) (imgutil.Image, string, error) {
```
While we are here? Just to make it clear that it isn't completely generic.

cmd/lifecycle/exporter.go
```diff
@@ -283,7 +302,7 @@ func initDaemonImage(imagName string, runImageRef string, analyzedMD lifecycle.A
 	return appImage, runImageID.String(), nil
 }
 
-func initRemoteImage(imageName string, runImageRef string, analyzedMD lifecycle.AnalyzedMetadata, registry string) (imgutil.Image, string, error) {
+func (ea exportArgs) initRemoteImage(runImageRef string, analyzedMD lifecycle.AnalyzedMetadata, registry string, keychain authn.Keychain) (imgutil.Image, string, error) {
```
ekcasey, line 305:
```suggestion
func (ea exportArgs) initRemoteAppImage(runImageRef string, analyzedMD lifecycle.AnalyzedMetadata, registry string, keychain authn.Keychain) (imgutil.Image, string, error) {
```

cmd/lifecycle/rebaser.go
```diff
@@ -128,7 +136,7 @@ func (r *rebaseCmd) Exec() error {
 	} else {
 		newBaseImage, err = remote.NewImage(
 			r.imageNames[0],
```
ekcasey, line 138:
```suggestion
			r.runImageRef,
```
Unrelated bug I think.

cmd/lifecycle/rebaser.go
```diff
@@ -148,3 +156,7 @@ func (r *rebaseCmd) Exec() error {
 	}
 	return nil
 }
+
+func (r *rebaseCmd) resolveKeychain() (authn.Keychain, error) {
+	return auth.ResolveKeychain(cmd.EnvRegistryAuth, r.imageNames)
```
ekcasey, line 161: This needs to include `r.runImageRef`.

cmd/lifecycle/rebaser.go
ekcasey: same here https://github.com/buildpacks/lifecycle/pull/409/files#diff-92c6ed289675df1e2edf598694dfc99f22a591496c130d10a43e86830cbd3686R193

cmd/lifecycle/rebaser.go
ekcasey: This is another case where we will need to ensure we have resolved the best run image first.

analyzer.go
```diff
@@ -56,6 +56,9 @@ func (a *Analyzer) analyzeLayers(appMeta LayersMetadata, cache Cache) error {
 	var cacheMeta CacheMetadata
 	if cache != nil {
 		var err error
+		if !cache.Exists() {
+			a.Logger.Info("cache does not exist")
```
ekcasey, line 60: What is the motivation behind adding this? I worry slightly that this may look like a _bad thing_ if it is expected on a first build. Also we should capitalize log statement e.g. `Cache does not exist`.

If we really want this I am inclined to say we should check whether the metadata is empty (or maybe return a bool in addition to the metadata?) and then then the users the cache is empty or unpopulated. "Exists" feels a bit confusing to me because, in the case of a cache volume for example the directory/volume may exist, even if it has no stored data.

cmd/lifecycle/restorer.go
ekcasey: tbh, I don't mind passing in the empty images or getting an essentially empty keychain back.

ekcasey (changes requested):

auth/env_keychain.go
```diff
@@ -12,27 +12,37 @@ import (
...
-func NewKeychain(envVar string) authn.Keychain {
+// ResolveKeychain returns either:
+// * a resolved keychain from the ggcr DefaultKeychain or
+// * a multi-keychain with a resolved keychain from the provided environment variable, and a resolved keychain from the ggcr DefaultKeychain
+// depending on whether the provided environment variable is set
+func ResolveKeychain(envVar string, images []string, overrideKeychain ...authn.Keychain) (authn.Keychain, error) {
+	fallbackKeychain := authn.DefaultKeychain
+	if len(overrideKeychain) == 1 {
+		fallbackKeychain = overrideKeychain[0]
+	}
```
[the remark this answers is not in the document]
yaelharel: Please see the last comment that I've pushed a few minutes ago. We changed it.

analyzer.go
natalieparellano: This came up in https://github.com/buildpacks/lifecycle/pull/409#discussion_r481233694 and just rode along with this PR.

cmd/lifecycle/restorer.go
natalieparellano: Sounds good. In this case I think we should add a test for `ResolveKeychain()` that verifies empty images produces a valid keychain.

acceptance/analyzer_test.go
natalieparellano: Turns out this is pretty easy to do - will add it.

auth/env_keychain_test.go
natalieparellano: We looked into this - it would mean re-ordering / re-grouping the tests such that the outer `when` is something like `when("the default keychain does not have an auth header", ...` followed by `when("valid auth env variable is set", ...` and then `when("auth env variable is not set", ...`. But to us, this ordering is less intuitive because it doesn't follow the order that the function itself is following.

cmd/lifecycle/restorer.go
natalieparellano: We added said test...

auth/env_keychain_test.go
natalieparellano: After talking about this more we are going with default keychain for reasons.

auth/env_keychain.go
natalieparellano: Latest changes should resolve all this weirdness. See https://github.com/buildpacks/lifecycle/pull/409#discussion_r541151115

auth/env_keychain.go
```diff
@@ -12,27 +12,37 @@ import (
...
-func (k *EnvKeychain) Resolve(resource authn.Resource) (authn.Authenticator, error) {
-	authHeaders, err := ReadEnvVar(k.EnvVar)
+	authHeaders, err := ReadEnvVar(envVar)
 	if err != nil {
-		return nil, errors.Wrap(err, "reading auth env var")
+		return nil, err
 	}
 
-	header, ok := authHeaders[resource.RegistryStr()]
+	envKeychain := resolvedKeychain{Auths: authHeaders}
```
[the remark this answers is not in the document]
natalieparellano: Thank you, @ekcasey ! This is much better.

auth/env_keychain_test.go
natalieparellano: Latest changes totally solve this problem, see https://github.com/buildpacks/lifecycle/pull/409#discussion_r541151115

cmd/lifecycle/exporter.go
```diff
@@ -305,7 +300,7 @@ func (ea exportArgs) initDaemonImage(runImageRef string, analyzedMD lifecycle.An
 	return appImage, runImageID.String(), nil
 }
 
-func (ea exportArgs) initRemoteImage(runImageRef string, analyzedMD lifecycle.AnalyzedMetadata, registry string, keychain authn.Keychain) (imgutil.Image, string, error) {
+func (ea exportArgs) initRemoteAppImage(runImageRef string, analyzedMD lifecycle.AnalyzedMetadata, registry string, keychain authn.Keychain) (imgutil.Image, string, error) {
```
yaelharel, line 303: We don't need to pass the `keychain`. We can access it through `ea exportArgs`

auth/env_keychain.go
```diff
@@ -77,32 +98,38 @@ func ReadEnvVar(envVar string) (map[string]string, error) {
...
-// Complementary to `ReadEnvVar`.
-func BuildEnvVar(keychain authn.Keychain, images ...string) (string, error) {
+func buildAuthMap(keychain authn.Keychain, images ...string) map[string]string {
 	registryAuths := map[string]string{}
 
 	for _, image := range images {
 		reference, authenticator, err := ReferenceForRepoName(keychain, image)
 		if err != nil {
-			return "", nil
+			return map[string]string{}
```
ekcasey, line 107: This problem existed before this PR but do we want to `continue` here instead of returning empty, in case this succeeds for other images in the set? Same on lines 115 and 120?

auth/env_keychain.go
```diff
@@ -77,32 +98,38 @@ func ReadEnvVar(envVar string) (map[string]string, error) {
...
 
-// BuildEnvVar creates the contents to use for authentication environment variable.
-//
-// Complementary to `ReadEnvVar`.
-func BuildEnvVar(keychain authn.Keychain, images ...string) (string, error) {
+func buildAuthMap(keychain authn.Keychain, images ...string) map[string]string {
 	registryAuths := map[string]string{}
 
 	for _, image := range images {
 		reference, authenticator, err := ReferenceForRepoName(keychain, image)
```
ekcasey, line 105: I don't think `ReferenceForRepoName` needs to or should be exported. We could potentially inline it here?

cmd/lifecycle/rebaser.go
natalieparellano: @ekcasey I'm having trouble following your link - could you show us the other line of code that should be changed? Also, just trying to follow what the code is doing here - what is the impact of this bug?

auth/env_keychain.go
```diff
@@ -12,27 +12,48 @@ import (
...
-// EnvKeychain uses the contents of an environment variable to resolve auth for a registry
-type EnvKeychain struct {
-	EnvVar string
+type ResolvedKeychain struct {
+	Auths map[string]string
 }
 
-func (k *EnvKeychain) Resolve(resource authn.Resource) (authn.Authenticator, error) {
-	authHeaders, err := ReadEnvVar(k.EnvVar)
+func EnvKeychain(envVar string) (authn.Keychain, error) {
```
ekcasey, line 34:
```suggestion
// EnvKeychain returns an authn.Keychain that uses the provided environment variable as a source of credentials.
// The value of the environment variable should be a JSON object that maps OCI registry hostnames to Authorization headers.
func EnvKeychain(envVar string) (authn.Keychain, error) {
```

auth/env_keychain.go
```diff
@@ -12,27 +12,48 @@ import (
...
+	authHeaders, err := ReadEnvVar(envVar)
 	if err != nil {
-		return nil, errors.Wrap(err, "reading auth env var")
+		return nil, err
 	}
+	return &ResolvedKeychain{Auths: authHeaders}, nil
+}
 
-	header, ok := authHeaders[resource.RegistryStr()]
+func InMemoryKeychain(keychain authn.Keychain, images ...string) authn.Keychain {
```
ekcasey, line 42:
```suggestion
// InMemoryKeychain resolves credentials for the given images from the given keychain and returns a new keychain
// that stores the pre-resolved credentials in memory and returns them on demand. This is useful in cases where the
// backing credential store may become inaccessible in the the future.
func InMemoryKeychain(keychain authn.Keychain, images ...string) authn.Keychain {
```

auth/env_keychain.go
```diff
@@ -12,27 +12,48 @@ import (
...
+	return authn.NewMultiKeychain(
+		envKeychain,
+		InMemoryKeychain(authn.DefaultKeychain, images...),
+	), nil
 }
 
-// EnvKeychain uses the contents of an environment variable to resolve auth for a registry
-type EnvKeychain struct {
-	EnvVar string
+type ResolvedKeychain struct {
```
ekcasey, line 30:
```suggestion
// ResolvedKeychain is an implementation of authn.Keychain that stores credentials in memory.
type ResolvedKeychain struct {
```

auth/env_keychain.go
```diff
@@ -12,27 +12,48 @@ import (
...
 
-// NewKeychain returns either a EnvKeychain or a authn.DefaultKeychain depending on whether the provided environment variable is set
-func NewKeychain(envVar string) authn.Keychain {
-	_, ok := os.LookupEnv(envVar)
-	if !ok {
-		return authn.DefaultKeychain
+// DefaultKeychain returns a keychain containing authentication configuration for the given images
+// from the following sources, if they exist, in order of precedence:
+// the provided environment variable and the docker config.json file.
+func DefaultKeychain(envVar string, images ...string) (authn.Keychain, error) {
```
ekcasey, line 18: What do you think about moving the `EnvRegistryAuth` `const` into this package and using this value directly in the `DefaultKeychain` impl, instead of passing in the `envVar` parameter?

cmd/lifecycle/exporter.go
natalieparellano: Moved this resolution up into `Args()`

analyzer.go
ekcasey, line 60: I still think we should tweak this slightly. `does not exist` seems potentially misleading given this could happen if we do not have the correct auth for example. Or maybe the cache volume directory exists but doesn't have metadata/layers in it. Maybe:
```suggestion
			a.Logger.Info("Layer cache not found")
```

cache.go
```diff
@@ -6,6 +6,9 @@ import (
 
 func (e *Exporter) Cache(layersDir string, cacheStore Cache) error {
 	var err error
+	if !cacheStore.Exists() {
+		e.Logger.Info("cache does not exist")
```
ekcasey, line 10:
```suggestion
		e.Logger.Info("Layer cache not found")
```

auth/env_keychain.go
natalieparellano: Great! With this change, we no longer need to strip out empty image names, so it makes the other code simpler too.

auth/env_keychain.go
natalieparellano: Sounds good. We kept the function to avoid making `buildAuthMap` too long, but we un-exported it.

auth/env_keychain.go
natalieparellano: Sounds good!

auth/env_keychain.go
natalieparellano: On looking into this further, it seems that `pack` is using this function. We exported it again, but we are wondering - what are the circumstances wherein we allow ourselves to make breaking changes to this package?

ekcasey: Someday I expect we might have more stability at the library interface level but right now I think we can force consumers to deal with changes on upgrade as we move towards a more intentional interface that will be easier to keep stable in the long run. I am happy to leave it for now. But the logic exposed here isn't very extensive (it's basically just two ggcr function calls that would be easy to inline in pack) and isn't really specific to the lifecycle and therefore I don't see this as the kind of utility we need to expose/maintain.

All that said, happy to leave it for now in the interest of time, if you would rather.

cmd/lifecycle/exporter.go
```diff
@@ -352,20 +358,34 @@ func parseOptionalAnalyzedMD(logger lifecycle.Logger, path string) (lifecycle.An
...
+				cmd.CodeInvalidArgs,
+				"parse arguments",
+			)
 		}
 		runImageRef, err = stackMD.BestRunImageMirror(registry)
 		if err != nil {
-			return lifecycle.StackMetadata{}, "", err
+			return lifecycle.StackMetadata{}, "", "", err
 		}
 	}
```
ekcasey, line 388:
```suggestion
	} else {
		runImageRef = runImageRefOrig
	}
```

cmd/lifecycle/exporter.go
```diff
@@ -153,26 +169,27 @@ func (e *exportCmd) Exec() error {
...
+		registryImages = append(registryImages, e.cacheImageTag)
 	}
-	registry := ref.Context().RegistryStr()
-
-	stackMD, runImageRef, err := resolveStack(ea.stackPath, ea.runImageRef, registry)
-	if err != nil {
-		return err
+	if !e.useDaemon {
+		registryImages = append(registryImages, e.imageNames...)
+		registryImages = append(registryImages, e.runImageRef)
```
ekcasey, line 187: Worth noting that the previous (analyzed) image is also a registry image that we access. However, this currently won't break because we ensure the previous image and the destination tag are on the same registry and our `DefaultKeychain` actually only cares about registries not images.

cmd/lifecycle/rebaser.go
```diff
@@ -79,32 +92,72 @@ func (r *rebaseCmd) Privileges() error {
...
+	if err := lifecycle.WriteTOML(r.reportPath, &report); err != nil {
+		return cmd.FailErrCode(err, cmd.CodeRebaseError, "write rebase report")
+	}
+	return nil
+}
+
+func (r *rebaseCmd) registryImages() []string {
+	registryImages := r.imageNames
+	if r.runImageRef != "" {
+		registryImages = append(registryImages, r.runImageRef)
```
ekcasey, line 130: This is another case where we need to resolve the best mirror from the stack metadata first. I see that we are actually doing this because we call `setAppImage` first. However `setAppImage` actually needs the `keychain` to be initialized.

cmd/lifecycle/rebaser.go
ekcasey: https://github.com/buildpacks/lifecycle/pull/409/files#diff-b19694987281b87997e43d8b402d163180ec5b9763cd3567079168459c0fec56R97-R107

re impact: small but if there was ever an error getting data from the newBaseImage, the image would return a misleading error that includes the wrong image name. E.g if https://github.com/buildpacks/lifecycle/blob/main/rebaser.go#L57 errored you would get a bad error string here https://github.com/buildpacks/imgutil/blob/main/remote/remote.go#L390

cmd/lifecycle/exporter.go
natalieparellano: We'll pass in previous now. In the future, we could send in registries instead of images.

cmd/lifecycle/creator.go
```diff
@@ -172,16 +191,31 @@ func (c *createCmd) Exec() error {
...
 }
+
+func (c *createCmd) registryImages() []string {
+	var registryImages []string
+	if c.cacheImageTag != "" {
+		registryImages = append(registryImages, c.cacheImageTag)
+	}
+	if !c.useDaemon {
+		registryImages = append(registryImages, append([]string{c.imageName}, c.additionalTags...)...)
+		registryImages = append(registryImages, c.runImageRef)
```
ekcasey, line 218:
```suggestion
		registryImages = append(registryImages, c.runImageRef)
		registryImages = append(registryImages, c.previousImage)
```

ekcasey (approved):

PR #410, opened 2020-09-09
Stack Buildpacks

jkutner:
This change introduces a stack buildpack that is identified by a privileged key in the `buildpack.toml`

When the privileged key is set to true, the builder will ignore layers
created in the layers directory and instead will take a snapshot of the
changes made during execution of the stack buildpack.

See: https://github.com/buildpacks/rfcs/pull/111


jabrown85:
### Items for the future:
- static `buildpack.toml` parsing and validation for stackpacks + buildpacks
  - this only exists in pack today, nothing in lifecycle parses that part..so we'd have to do that for the first time
  - we would need to parse a specific set of mixins for the current stack using CNB_STACK_ID
- static run-mixins.toml as optional toml in detector
  - would be used to pre-populate the run dependencies so that detection doesn't fail
- A way for stackpacks to disable cache on rebuilds
  - some stackpacks may not want/need to produce snapshots that persist in subsequent builds
  - the cost of applying the snapshot or the fact that a buildpack may not realize it restored would be reasons to not restore
  - we could probably use `cache = false` to achieve this if we have stackpacks write a `layer.toml`
- A way for stackpacks to skip taking new snapshots on subsequent re-runs
  - Check something was restored from cache and then communicate that you don't need a snapshot taken
- A way for stackpacks to exclude/include parts of the snapshot
  - documented in rfc as a part of `layer.toml` - I have questions on what it would mean to have multiple layer directories for a stackpack.
- `extender` binary/phase. Right now we can at least get stackpacks in that are needed for `build:`.

### Open Qs:
- Should stackpacks **require** writing a `layer.toml`?
- How should we log the detect phase for build+run phase logs to not be confusing?
  - We need to communicate that `Buildpack X@v1 [build,run]` or that it was skipped entirely.
  - Right now we are just logging the build phase buildpacks.
- Should stackpacks be dynamically executed based on the folder alone?
  - I like the idea of declaring in `buildpack.toml` or `stack-order.toml` in `/layers` by default. Similar to `order.toml`. The platform may wish to drive the order/exclusion programmatically.

acceptance/testdata/builder/image/container/cnb/stack/buildpacks/example_stack/1.0.0/bin/build
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+echo "hello" > /file-to-snapshot
+echo "hello" > /bin/exe-to-snapshot
+chmod +x /bin/exe-to-snapshot
+
+# this should be ignored
+mkdir -p /cnb
+echo "hello" > /cnb/file-to-ignore
```
natalieparellano, line 9: Does it also make sense to check that the stack buildpack can't write to `/layers` and `/platform`?

jabrown85: That could make sense, especially for `/platform`. It really isn't a test that they can't write somewhere, because they can. It is more a test of what we capture in the snapshot. Because they run as root, they'll be able to do anything to either directory and whatever changes will remain changed for the duration of the userspace buildpacks but will not be captured in the resulting snapshot.

natalieparellano (commented):
This is an impressive amount of work! I am still going through it, but wanted to offer some comments and questions in the meantime.

acceptance/testdata/builder/image/container/cnb/stack/buildpacks/example_stack/1.0.0/bin/build
natalieparellano: Ah, makes sense. I see now that `/layers` is tested in snapshot/kaniko_snapshot_unix_test.go...

snapshot/kaniko_snapshot_unix_test.go
```diff
@@ -0,0 +1,157 @@
...
+		if err != nil {
+			t.Fatalf("Error: %s\n", err)
+		}
+		tmpDir, err = ioutil.TempDir(filepath.Join(cwd, "..", "tmp"), "kaniko")
+		if err != nil {
+			t.Fatalf("Error: %s\n", err)
+		}
+
+		createTestFile(t, filepath.Join(tmpDir, "cnb", "privatefile"))
+		createTestFile(t, filepath.Join(tmpDir, "layers", "privatefile"))
```
natalieparellano, line 45: Does it make sense to add `/platform`?

snapshot/kaniko_snapshot_unix.go
```diff
@@ -0,0 +1,107 @@
...
+	snapshotter                *ksnap.Snapshotter
+}
+
+func NewKanikoSnapshotter(rootDir string) (lifecycle.LayerSnapshotter, error) {
+	ls := KanikoSnapshotter{
+		RootDir:                    rootDir,
+		DetectFilesystemIgnoreList: true,
+		IgnoredPaths: []string{
+			"/tmp",
+			"/layers",
```
natalieparellano, line 32: Thinking about this more... does it make sense to hard code `/layers`, since it can change? (If a different value is provided for `-layers` or `CNB_LAYERS_DIR`) 

Same for `/platform`...

detector_test.go
```diff
@@ -44,12 +44,13 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
-			AppDir:        appDir,
-			PlatformDir:   platformDir,
-			BuildpacksDir: buildpacksDir,
-			Logger:        &log.Logger{Handler: logHandler},
+			FullEnv:            append(os.Environ(), "ENV_TYPE=full"),
+			ClearEnv:           append(os.Environ(), "ENV_TYPE=clear"),
+			AppDir:             appDir,
+			PlatformDir:        platformDir,
+			BuildpacksDir:      buildpacksDir,
+			StackBuildpacksDir: buildpacksDir,
```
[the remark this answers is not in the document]
jabrown85: Yeah, the stackpack used to be in its own folder..then I started slowly needing all the things the other buildpacks were doing for testing and just created a new buildpack from another. I suppose I could move them to a folder and keep the linked files the same. 

It is true the default would be in another directory, but AFAIK nothing says they can not just live in `/buildpacks` either.

detector_test.go
```diff
@@ -591,7 +592,840 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
+						"pass: B@v1\n"+
+						"Resolving plan... (try #1)\n"+
+						"A v1\n"+
+						"B v1\n",
+				) {
+					t.Fatalf("Unexpected log:\n%s\n", s)
+				}
+			})
+
+			it("should produce build plans for both standard and privileged that provide the same dep", func() {
```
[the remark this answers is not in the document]
jabrown85: Yeah, the naming is still confusing. I've gone back and forth. The only thing today that makes a buildpack a "stack buildpack" is `privileged` essentially. So the logic around producing build plans and groups to run is currently built around that naming. The tests I sort of left in "user terminology" of stack buildpacks..if that makes any sense.

detector_test.go
```diff
@@ -591,7 +592,840 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
+						"Resolving plan... (try #1)\n"+
+						"skip: X@1.0.0[run] not required\n"+
+						"X 1.0.0\n"+
+						"B v1\n",
+				) {
+					t.Fatalf("Unexpected log:\n%s\n", s)
+				}
+			})
+
+			it("should fail if all requires are not met by mixins", func() {
```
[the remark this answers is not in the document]
jabrown85: Today, static mixins are _only_ validated by `pack` (and other platforms, I suppose). I do think lifecycle needs to do this, but I wasn't ready to include that as part of this. Lifecycle currently only parses that `stack.toml` to figure out the run image and mirrors IIRC.

builder.go
```diff
@@ -57,128 +66,210 @@ type BuildpackPlan struct {
...
+		procMap.add(launchData.Processes)
+		slices = append(slices, launchData.Slices...)
+		labels = append(labels, launchData.Labels...)
+	}
+
+	if err := b.convertMetadataToVersion(bom); err != nil {
+		return nil, err
+	}
+
+	return &BuildMetadata{
```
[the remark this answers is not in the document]
jabrown85: It is not currently used by the stack buildpacks process at all. It's thrown away in `cmd/builder.go`. I figure it probably does need to show up in BOM eventually..but I'm not familiar with that process and it hasn't been mentioned yet in any stackpack meetings that I recall.

snapshot/kaniko_snapshot_unix.go
jabrown85: ah, yeah. Good call. I think we may need to thread through the configurable bits here and make this more dynamic.

detector_test.go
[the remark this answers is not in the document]
natalieparellano: Just summarizing here in case it would help others - 

Eventually, every combination of static (via buildpack.toml) or dynamic (via the build plan) "requires" and static (via the stack) or dynamic (via stack buildpacks) "provides" will be supported. Today, we have static -> static support with `pack` validating mixins up front. This PR introduces support for dynamic -> dynamic. The others will be supported in the future.

snapshot/kaniko_snapshot_unix.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/buildpacks/lifecycle"
+
+	kconfig "github.com/GoogleContainerTools/kaniko/pkg/config"
+	ksnap "github.com/GoogleContainerTools/kaniko/pkg/snapshot"
+	kutil "github.com/GoogleContainerTools/kaniko/pkg/util"
+)
+
+type KanikoSnapshotter struct {
+	RootDir                    string
+	DetectFilesystemIgnoreList bool
```
natalieparellano, line 21: What's this for? It seems always to be `true` except for the tests?

snapshot/kaniko_snapshot_unix_test.go
```diff
@@ -0,0 +1,157 @@
...
+				case "dir-with-ignored-files/":
+				case "my-space/":
+				case strings.Trim(filepath.Join(snapshotter.RootDir, ".wh.file-to-delete"), "/"):
+					continue
+				case "newfile":
+				case "my-space/newfile-in-dir":
+					assertSnapshotFile(t, tr, "hello\n")
+				case "file-to-change":
+					assertSnapshotFile(t, tr, "hola\n")
+				default:
```
natalieparellano, line 117: This test might be a bit more readable if you had a `case` for `file-to-ignore` and then fall through to any other unexpected files.

snapshot/kaniko_snapshot_unix_test.go
```diff
@@ -0,0 +1,157 @@
...
+		var (
+			snapshotFile string
+		)
+
+		it.Before(func() {
+			h.AssertNil(t, snapshotter.Init())
+
+			os.Remove(filepath.Join(snapshotter.RootDir, "file-to-delete"))
+			createTestFileWithContent(t, filepath.Join(snapshotter.RootDir, "file-to-change"), "hola\n")
+			createTestFile(t, filepath.Join(snapshotter.RootDir, "my-space", "newfile-in-dir"))
```
natalieparellano, line 72: If you change this to `createTestFileWithContent` that could make the assertion more understandable (don't have to check the definition of `createTestFile` to know why we're expecting `hello`).

snapshot/kaniko_snapshot_unix_test.go
```diff
@@ -0,0 +1,157 @@
...
+			DetectFilesystemIgnoreList: false,
+		}
+		snapshotter.IgnoredPaths = []string{filepath.Join(snapshotter.RootDir, "dir-with-ignored-files")}
+	})
+
+	it.After(func() {
+		os.RemoveAll(tmpDir)
+	})
+
+	when("files are added and modified", func() {
```
natalieparellano, line 62: Maybe it makes sense to add a test that when a snapshot is applied, `snapshotter.TakeSnapshot` captures the changes as if the changes were made manually.

cmd/lifecycle/detector.go
```diff
@@ -53,34 +63,75 @@ func (d *detectCmd) Privileges() error {
...
+	dr, err := d.detect()
 	if err != nil {
 		return err
 	}
-	return d.writeData(group, plan)
+	return d.writeData(dr)
 }
 
-func (da detectArgs) detect() (lifecycle.BuildpackGroup, lifecycle.BuildPlan, error) {
+func (da detectArgs) mergeOrderWithStackBuildpacks(order lifecycle.BuildpackOrder) (lifecycle.BuildpackOrder, error) {
```
natalieparellano, line 73: For readability, it might be nice to call this function something like `prependOrderWithStackBuildpacks` to make it clear that stack buildpacks will come first.

cmd/flags.go
```diff
@@ -24,7 +24,12 @@ var (
 	DefaultProcessType         = "web"
 	DefaultProjectMetadataPath = filepath.Join(".", "project-metadata.toml")
 	DefaultReportPath          = filepath.Join(".", "report.toml")
+	DefaultStackOrderPath      = filepath.Join(rootDir, "cnb", "stack", "order.toml")
```
natalieparellano, line 27: The RFC has this path as `/cnb/stack/buildpacks/order.toml`, I believe. Not sure if the RFC or the PR should be updated, but probably it's worth keeping things consistent.

cmd/lifecycle/detector.go
```diff
@@ -53,34 +63,75 @@ func (d *detectCmd) Privileges() error {
...
+	stackBuildPacksOrder, err := lifecycle.ReadOrder(da.stackOrderPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(stackBuildPacksOrder) == 0 {
+		return order, nil
+	}
+
+	if len(stackBuildPacksOrder) > 1 {
```
natalieparellano, line 89: Ah, interesting - what is the motivation for this requirement? I don't see it in the RFC, but maybe I missed it.

cmd/lifecycle/detector.go
```diff
@@ -119,13 +174,43 @@ func (da detectArgs) verifyBuildpackApis(order lifecycle.BuildpackOrder) error {
 	return nil
 }
 
-func (d *detectCmd) writeData(group lifecycle.BuildpackGroup, plan lifecycle.BuildPlan) error {
-	if err := lifecycle.WriteTOML(d.groupPath, group); err != nil {
+func (da detectArgs) validateBuildpacks(order lifecycle.BuildpackOrder) error {
```
natalieparellano, line 177: It might be nice for this to have a more specific name, perhaps `verifyBuildpackPrivileges` ...or something. `validateBuildpacks` is very vague :)

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+
+	if len(privGroup.Group) == 0 {
+		return b.build(group, plan)
+	}
+	return b.buildWithReexec(group, privGroup, plan)
+}
+
+func (ba buildArgs) buildAll(group, privGroup lifecycle.BuildpackGroup, plan lifecycle.BuildPlan, rootuid, rootgid, uid, gid int) error {
+	if len(privGroup.Group) > 0 {
+		if err := priv.RunAsEffective(rootuid, rootgid); err != nil {
```
natalieparellano, line 97: Didn't we already drop privileges by this point when running the creator?

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+	if len(group.Group) > 0 {
+		if err = ba.buildAsSubProcess(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (ba buildArgs) buildAsSubProcess() error {
+	if err := priv.RunAs(ba.uid, ba.gid); err != nil {
```
natalieparellano, line 166: I know there was conversation at some point (though I can't find it) around whether a "stack build" should share the same `Exec` as a regular build. It feels weird to me that we're calling `priv.RunAs` here whereas everywhere else in the code it's being called in `Privileges()`. It makes me wonder if we need to pull the two flows apart at a higher level of abstraction.

snapshot/kaniko_snapshot_unix_test.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func testKanikoSnapshotter(t *testing.T, when spec.G, it spec.S) {
+	var (
+		snapshotter *snapshot.KanikoSnapshotter
+		tmpDir      string
+	)
+
+	it.Before(func() {
+		// Using the default tmp dir causes kaniko to go haywire for some reason
```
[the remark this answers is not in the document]
jabrown85: I actually don't know. @jkutner might remember (he originally wrote this bit)

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+
+	builder, err := ba.createBuilder(privGroup, plan)
+	if err != nil {
 		return err
 	}
-	return b.build(group, plan)
+
+	builder.Snapshotter = stackSnapshotter
+	builder.AppDir = workspaceDir
+	builder.BuildpacksDir = ba.stackBuildpacksDir
```
natalieparellano, line 129: This could be folded into `createBuilder` if you passed a boolean flag like `stack`. I don't know if that's more readable or not...

cmd/flags.go
jabrown85: @jkutner thoughts?

snapshot/kaniko_snapshot_unix.go
jabrown85: So that bool triggers `DetectFilesystemIgnoreList()`. The reason it is off for tests is that we don't want it detecting the test runner's ignore list. In a real container, this is safe to run and produces additional ignore list items we don't want in the snapshot.

https://github.com/GoogleContainerTools/kaniko/blob/f20f495051627b988c7449e9b163a4f9f30cc044/pkg/util/fs_util.go#L410

cmd/lifecycle/detector.go
jabrown85: Yeah, I think it was just added or is going to be added. 

The reason we did this is because we would have to explode the combined list of stackpacks+buildpacks. Right now if you have 5 buildpack groups, we end up still testing just 5 groups by prepending the stackpacks. If you had 3 groups of stackpacks, I think we would need to create a group for each stackpack group combined with each buildpack group. Maybe that isn't bad and there is a use case, but we couldn't think of one.

cmd/lifecycle/builder.go
jabrown85: Not anymore. The creator was updated to use `RunAsEffective`, which now allows us to regain privileges after dropping them. I did this work prior to discussing creator can't really ever fully do stackpacks...but it CAN do stackpacks as long as they are `build:` phase only requirements.

cmd/lifecycle/builder.go
jabrown85: This part is a bit strange. I'd be down for figuring out a cleaner way to go about this. Ideally without re-exec entirely. That bit is gross anyway. I'm not convinced it is buying us anything over only dropping privileges.

snapshot/kaniko_snapshot_unix_test.go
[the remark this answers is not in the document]
jabrown85: I took a quick look just now. The problem is that the the test times out while getting the initial kaniko snapshot when give `""` for the directory to create the temporary directory. I believe the reason this happens is because macOS uses symlinks for the default tmp directory path. 

```
» echo $TMPDIR
/var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/
» ls -al /var
lrwxr-xr-x 1 root admin 11 Jan 23  2020 /var -> private/var
```

Kaniko has some code when walking the filesystem to check ancestors for symlinks. I think kaniko can't get out of the loop of it trying to resolve the symlinks properly. Maybe a bug in kaniko?

```
time="2020-10-07T12:58:58-05:00" level=info msg="Taking snapshot of full filesystem..."
time="2020-10-07T12:58:59-05:00" level=trace msg="Resolving paths [/var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310 /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/cnb /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/cnb/privatefile /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/layers /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/layers/privatefile /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/bin /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/bin/file-not-to-change /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/file-not-to-change /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/file-to-change /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/file-to-delete]"
time="2020-10-07T12:58:59-05:00" level=trace msg="updated link /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/cnb/privatefile to /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/cnb"
time="2020-10-07T12:58:59-05:00" level=trace msg="updated link /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/layers/privatefile to /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/layers"
time="2020-10-07T12:58:59-05:00" level=trace msg="updated link /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/bin/file-not-to-change to /var/folders/cx/xmrdk7sx6x5gcv0gl951jhx00000gp/T/kaniko-snapshotter883391310/bin"
panic: test timed out after 30s
```

cmd/lifecycle/builder.go
natalieparellano: Ohhh I missed that part. Thanks for explaining :)

builder.go
```diff
@@ -75,110 +90,186 @@ func (b *Builder) Build() (*BuildMetadata, error) {
...
-		if bpInfo.Buildpack.ClearEnv {
-			cmd.Env = b.Env.List()
-		} else {
-			cmd.Env, err = b.Env.WithPlatform(platformDir)
-			if err != nil {
+			snapshotPath := filepath.Join(layerDir, "snapshot.tgz")
+			if err := b.Snapshotter.TakeSnapshot(snapshotPath); err != nil {
 				return nil, err
 			}
+			layerTOMLPath := filepath.Join(bpLayersDir, "snapshot.toml")
```
natalieparellano, line 128: I noticed `snapshot.toml` isn't captured in the RFC, would that be worth adding?

builder.go
```diff
@@ -75,110 +90,186 @@ func (b *Builder) Build() (*BuildMetadata, error) {
...
+		Labels:     labels,
+		Processes:  procMap.list(),
+		Slices:     slices,
+	}, nil
+}
 
-		var launch LaunchTOML
-		tomlPath := filepath.Join(bpLayersDir, "launch.toml")
-		if _, err := toml.DecodeFile(tomlPath, &launch); os.IsNotExist(err) {
+func (bp Buildpack) ApplyLayerSnapshots(snapshotter LayerSnapshotter, layersDir string) error {
```
natalieparellano, line 159: This function name is odd to me because while there may be multiple layers, we only expect to apply one snapshot... maybe `ApplyLayersSnapshot`? Also just for my understanding, when would you expect a stack buildpack to have multiple layers? I thought it could only ever create one layer (the snapshot)...

snapshot/kaniko_snapshot_unix_test.go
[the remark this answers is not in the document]
natalieparellano: Ah... over in `pack` we've used `filepath.EvalSymlinks(tmpDir)` to resolve issues with volume mounts that I think stem from the same problem. Might be worth giving it a try...

snapshot/kaniko_snapshot_unix_test.go
natalieparellano: Hmm I was thinking it would look like 

```
case "layers/file-to-ignore":
  t.Fatalf("Unexpected file: %s\n", hdr.Name)
case "cnb/file-to-ignore":
  t.Fatalf("Unexpected file: %s\n", hdr.Name)
case "platform/file-to-ignore":
  t.Fatalf("Unexpected file: %s\n", hdr.Name)
default:
  t.Fatalf("Unexpected file: %s\n", hdr.Name)
```

I think we would never expect to hit `case "file-to-ignore"` because the test setup never creates that file at the root. But maybe I'm missing something...

builder.go
jabrown85: `snapshot.toml` is actually just a `layer.toml` for the generated `snapshot` layer. 🤯

builder.go
jabrown85: I left that intentionally open. We could give the stackpack developer a way to potentially create multiple layers. I was imagining a situation where instead of generating a single `layer.toml` for a `snapshot` layer. The buildpack developer could instead be responsible for creating `layer.toml` but with different `exclude` or `include` paths (along with `cache=false` potentially. Then we could split the kaniko generated snapshot into X distinct tars. This might be useful for some advanced cache scenarios where you don't want to restore or store _all_ of the snapshot on subsequent re-runs.

detector.go
```diff
@@ -27,6 +28,8 @@ var errBuildpack = errors.New("buildpack(s) failed with err")
 var errInconsistentVersion = `buildpack %s has a "version" key that does not match "metadata.version"`
 var errDoublySpecifiedVersions = `buildpack %s has a "version" key and a "metadata.version" which cannot be specified together. "metadata.version" should be used instead`
 var warnTopLevelVersion = `Warning: buildpack %s has a "version" key. This key is deprecated in build plan requirements in buildpack API 0.3. "metadata.version" should be used instead`
+var errInvalidRequirementsBuildpack = `priviledged buildpack %s has defined "requires", which is not allowed.`
+var errInvalidProvidesBuildpack = `buildpack %s has defined "provide" with "mixin = true", which is not allowed.`
```
natalieparellano, line 32: Check out https://github.com/buildpacks/lifecycle/commit/2f74559b2af29ad09e24cbae5a812c0fe78dcb51 which introduces a different pattern for the custom errors.

snapshot/kaniko_snapshot_unix_test.go
jabrown85: I cleaned this up a bit more. I realized the case statement also didn't really handle the absence of files and added some assertions around that.

detector.go
```diff
@@ -86,26 +186,35 @@ func (r *Require) hasTopLevelVersions() bool {
 	return r.Version != ""
 }
 
+type depKey struct {
+	name  string
+	mixin bool
+	any   bool
```
natalieparellano, line 192: Can these both be defined at the same time (and what would that mean if so)? Just curious...

jabrown85: Yes, a stackpack could present itself as being able to install _any_ mixin. This would be what an `apt-buildpack` might do. This way it can pass detection and resolve any other mixin request from downstream buildpacks without having to statically (or even dynamically) present every single option it could do.

detector.go
```diff
@@ -155,92 +264,184 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
 			detected = detected && bp.Optional
 		}
 	}
+
+	if len(results) > 0 {
+		detected = detected && anyUserBuildpacks(results)
```
natalieparellano, line 269: This is clever :)

detector.go
```diff
@@ -155,92 +264,184 @@ func (c *DetectConfig) process(done []Buildpack) ([]Buildpack, []BuildPlanEntry,
...
+	PrivOptions  []detectOption
+	RunDeps      depMap
+	RunOptions   []detectOption
+}
+
+func (c *DetectConfig) runTrial(i int, trial detectTrial) (*trialResult, error) {
 	c.Logger.Debugf("Resolving plan... (try #%d)", i)
 
-	var deps depMap
+	buildTrial := append([]detectOption{}, trial...)
```
natalieparellano, line 356: What's the purpose of this line?

detector.go
```diff
@@ -373,47 +571,66 @@ func (bg BuildpackGroup) append(group ...BuildpackGroup) BuildpackGroup {
...
 	if err == errBuildpack {
 		err = NewLifecycleError(err, ErrTypeBuildpack)
 	} else if err == errFailedDetection {
 		err = NewLifecycleError(err, ErrTypeFailedDetection)
 	}
-	for i := range entries {
-		for j := range entries[i].Requires {
-			entries[i].Requires[j].convertVersionToMetadata()
+
+	if dr != nil {
```
natalieparellano, line 594: When might `dr` be nil?

detector.go
```diff
@@ -556,45 +772,82 @@ type depEntry struct {
...
 }
 
-func (m depMap) provide(bp Buildpack, provide Provide) {
-	entry := m[provide.Name]
+func (m depMap) provide(stage stageName, bp Buildpack, provide Provide) {
+	if !bp.Privileged && provide.stage() != stage {
+		return
+	}
+
+	// if it is not a match, use the full name to not match when required
```
natalieparellano, line 805: Is it possible you could elaborate on this comment?

detector.go
jabrown85: Creating a copy of the `trial`. One for `build:` and one for `run:`. The methods following this modify the `trial` that is passed in and I want to ensure the same `trial` can satisfy both phases.

detector.go
jabrown85: `dr` is now null in case of `errFailedDetection` or `errBuildpack` (and probably other situations in the recursion). `detect` is returning a pointer of `DetectResult`

natalieparellano:
Really impressive work! I am still trying to wrap my head around some of the code in detector.go ...but that was true even before this PR :)

I was thinking, would there be value in creating some test stacks / buildpacks that exercise this functionality? Like maybe working off of the samples repo (or a branch there)? That might make it easier to play around with this stuff and get a feel for the end-user experience. I'd be happy to help out with this if you think it's something worth doing.

detector.go
natalieparellano: Ohhh... I see... so if `any` is true then we'd expect `mixin` to always be true, right? (Do we do any check for this?)

detector_test.go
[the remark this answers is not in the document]
natalieparellano: Sounds good! The tests are fairly readable in my opinion :)

cmd/lifecycle/detector.go
natalieparellano: I am all for not exploding the number of scenarios we need to consider, if it's not necessary!

builder.go
natalieparellano: Ohh right... I recall that was part of the RFC that is unimplemented for now.

detector.go
jabrown85: This will probably be hard to explain and may take me a few attempts. Let's set the stage a bit. `detect` passing AND providing a build plan means allows buildpacks to skip in addition to pass/fail via exit code alone. If a buildpack is optional and the buildpack writes a `provides` of `libpq` - we skip the buildpack but don't fail detection.

This concept becomes tricker in stackpacks. Because a single detection may produce a buildplan like this:
```
[[provides]]
name = run:libpq
mixin = true
```
Let's say a user buildpack has
```
[[requires]]
name = run:libpq
mixin = true
```
The mere _existence_ of a build plan that `provides` something means we should NOT solely use the detect exit code. In this case I want this stackpack to be `skipped` during `build` phase and to not be skipped during `run` phase. 

Visualized, the logs might look something like...
```
...
Resolving plan... (try #1)
skip: Stackpack/A@v1 [build] - not required
pass: Stackpack/A@v1 [run]
pass: userbuildpack/z@v1
...
```

If I were to simply remove the non-stage matching `provide`/`requires` entirely when doing the dependency resolution per-stage, the stackpack would end up executing during `build` phase. This is because if the dependency map doesn't contain anything, it is assumed that there was no build plan essentially and it relies on the exit code alone. The reverse of this problem is more costly. I 100% don't want to try and extend the run image if there is nothing to be done on the image.

So anyway...there is probably a better way to fix this. The way I chose to do it for now is to allow stage specific mixins to exist in the map for stages not being processed. It's a silly way to fix not having an empty depMap.

`should not require extending run image for build stage mixins` is one of the few tests that maybe help explain this a bit more.

detector.go
natalieparellano: Hmm... then couldn't you do something like 

```
	if err == errBuildpack {
		return nil, NewLifecycleError(err, ErrTypeBuildpack)
```

etc.?

Unless there are cases when `dr` can be nil without an error having occurred?

jabrown85:
> I was thinking, would there be value in creating some test stacks / buildpacks that exercise this functionality? Like maybe working off of the samples repo (or a branch there)? That might make it easier to play around with this stuff and get a feel for the end-user experience. I'd be happy to help out with this if you think it's something worth doing.

I do have a rough workbench of sorts [here](https://github.com/jkutner/stackpack-example/tree/jab/stackpack-tester). Make sure you are on the `jab/stackpack-tester` branch. The `Makefile` has some machine-specific paths to lifecycle and a test workspace, but otherwise is a pretty good way to test things out. It uses a lot of volume mounts to make iterating and changing buildpacks/stackpacks, order of buildpack and more without requiring rebuilding/re-packaging everything over and over. I'd be happy to work with anyone who wants to set it up or if we thought having a fully built builder with stackpacks published somewhere makes sense I could build this into an actual builder.

yaelharel (commented):
I've just finished going over your PR for the first time. Your work is very impressive!
Overall it looks really good.
I was looking for some content that exists in the RFC and isn't part of this PR (such as the `extend` phase), but then I saw [Jesse's comment](https://github.com/buildpacks/lifecycle/pull/410#issuecomment-700296077) (appeared under the hidden items in Github) listing things that will be implemented in the future.
Following [this slack thread](https://buildpacks.slack.com/archives/CSFE49VGR/p1602772499067300), I hope that the new PR/s will be ready for review soon.

tools/go.mod
```diff
@@ -6,12 +6,17 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/buildpacks/imgutil v0.0.0-20200831154319-afd98bd2f655
 	github.com/buildpacks/lifecycle v0.8.1
-	github.com/docker/docker v1.4.2-0.20200221181110-62bd5a33f707
+	github.com/docker/docker v1.14.0-0.20190319215453-e7b5f7dbe98c
 	github.com/golang/mock v1.4.4
 	github.com/golangci/golangci-lint v1.30.0
-	github.com/google/go-containerregistry v0.0.0-20200313165449-955bf358a3d8
+	github.com/google/go-containerregistry v0.1.2-0.20200804170047-b0d31a182cf0
```
ekcasey, line 12: To avoid dependency hell we should test early to see if this code has problems compiling with ggcr `v0.1.4` and the associated version of `docker` -> https://github.com/buildpacks/imgutil/issues/70

cmd/flags.go
```diff
@@ -24,7 +24,12 @@ var (
...
+	DefaultStackBuildpacksDir  = filepath.Join(rootDir, "cnb", "stack", "buildpacks")
+	DefaultPrivilegedGroupPath = filepath.Join(".", "privileged-group.toml")
```
ekcasey, line 29: RFC has `-stack-group.toml` which I think I prefer.

cmd/lifecycle/analyzer.go
```diff
@@ -130,7 +131,7 @@ func (aa analyzeArgs) analyze(group lifecycle.BuildpackGroup, cacheStore lifecyc
 	}
 
 	analyzedMD, err := (&lifecycle.Analyzer{
-		Buildpacks: group.Group,
+		Buildpacks: append(privGroup.Group, group.Group...),
```
ekcasey, line 134: What happens with cached launch layers, it seems like we are only analyzing/restoring layers created on the build image?

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+		return err
+	}
+
+	hasPrivGroups := len(privGroup.Group) > 0
+
+	// builder should never be run with privileges if there aren't any stack buildpacks
+	if !hasPrivGroups && priv.IsPrivileged() {
 		return cmd.FailErr(errors.New("refusing to run as root"), "build")
+	} else if hasPrivGroups && !priv.IsPrivileged() {
+		return cmd.FailErr(errors.New("must run as root"), "build")
```
ekcasey, line 74: We probably need to explain more in this error message with you must run as root (you passed the privileged flag).

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+	_, privGroup, _, err := b.readData()
+	if err != nil {
+		return err
+	}
+
+	hasPrivGroups := len(privGroup.Group) > 0
+
+	// builder should never be run with privileges if there aren't any stack buildpacks
+	if !hasPrivGroups && priv.IsPrivileged() {
 		return cmd.FailErr(errors.New("refusing to run as root"), "build")
```
ekcasey, line 72: I am not sure this should actually be a failure. Maybe we can just drop privileges in this case. As a user I would be annoyed one build command succeeds and another (identical except for a single flag) created this error.

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+		}
+		if err := priv.RunAsEffective(uid, gid); err != nil {
+			return err
+		}
+	}
+
+	return ba.build(group, plan)
+}
+
+func (ba buildArgs) stackBuild(privGroup lifecycle.BuildpackGroup, plan lifecycle.BuildPlan) error {
```
ekcasey, line 111: In general I want to move logic/complexity out of this untested command files and PR is moving a lot of logic into this file. I think a lot of this logic should move to builder.

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+	}
+
+	builder, err := ba.createBuilder(privGroup, plan)
+	if err != nil {
 		return err
 	}
-	return b.build(group, plan)
+
+	builder.Snapshotter = stackSnapshotter
+	builder.AppDir = workspaceDir
```
ekcasey, line 128: This feel very backdoor to me. I don't let setting a field called `AppDir` to something that isn't the app dir. It seems like we are doing it to change as little as possible in builder but in the process we may be creating a bad interface. As a user, if I read the godoc for the `lifecycle` package it would be very hard to figure out how to run a stack build. We would need comments that said things like "To run a stack build set the AppDir to a temporary directory and supply a snapshotter." which doesn't seem like a very natural interface.

cmd/lifecycle/builder.go
```diff
@@ -48,26 +60,160 @@ func (b *buildCmd) Args(nargs int, args []string) error {
...
+	builder.BuildpacksDir = ba.stackBuildpacksDir
+
+	_, err = ba.execBuild(builder)
+	return err
 }
 
 func (ba buildArgs) build(group lifecycle.BuildpackGroup, plan lifecycle.BuildPlan) error {
-	builder := &lifecycle.Builder{
+	builder, err := ba.createBuilder(group, plan)
```
ekcasey, line 136: Instead of creating two separate builders for stack and regular builds I think we should be able to execute both parts of the build with the same builder struct.

ekcasey (commented):
Leaving this partially finished review for now

builder.go
```diff
@@ -75,110 +90,186 @@ func (b *Builder) Build() (*BuildMetadata, error) {
...
 			continue
-		} else if err != nil {
-			return nil, err
 		}
-		for i := range launch.Processes {
-			launch.Processes[i].BuildpackID = bp.ID
+		_, err := layer.read()
+		if err != nil {
+			continue
 		}
```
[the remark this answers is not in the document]
jabrown85: The `read()` is used to ensure the directory was a valid layer (with a `<layer>.toml`) that is readable.

cmd/lifecycle/analyzer.go
jabrown85: This first draft doesn't have the `extend` phase at all, so the layers being restored here for stackpacks are JUST the ones that ran on build. Does that make sense? I might be misunderstanding the question though.

cmd/lifecycle/builder.go
jabrown85: That's fair and for sure the direction I was going (changing as little as possible). I agree that we should create a better interface, so I'll put something up around this.

kadirselcuk (approved):

kadirselcuk (approved):

kadirselcuk (approved):

natalieparellano:
@jkutner @jabrown85 would you be okay closing this one? I know we wanted to keep it around as a reference, but we could still reference it if it's closed. It would be nice to declutter our PRs page for mental organization :)

Issue #389, opened 2020-08-17
Move release automation to GHA


ekcasey:
Now that github actions support manually triggered workflows we should move the lifecycle release process from concourse to GHA

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

natalieparellano:
As part of this work, we should change the way we're publishing lifecycle images. Instead of building a new image when releases are cut, we should build a new remote image with every push to `main` and re-tag the lifecycle image on release e.g., if the release SHA is 101f194 then the published `lifecycle:101f194` would be re-tagged as `lifecycle:0.X.0`.

natalieparellano:
@yaelharel and I are thinking of picking this up

natalieparellano:
After some experimentation and discussion we think this is the way to proceed:

* `pack` already has automation around the release process with GHA. We want to copy most of what is being done, namely:
  * Derive the release version from the branch name -> if branch is `release/0.12.0`, the release version will be `0.12.0`
    * We should therefore remove the hard coded VERSION file and the hard coded version in lifecycle.toml. We'll need to template lifecycle.toml and fill in the right version in the uploaded artifact. 
  * Auto generated changelog. We'll create a separate issue for this.
* However with `pack` a new draft release is created every time there are new commits on the branch. We want to avoid this. We should look into manually triggered workflows: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

Other needed things:
* We'll need to ensure that tests are run against PRs to `release/*`
* `release/*` should probably be protected branches 
* The workflow that creates the draft release should also build a new lifecycle image 
* The existing lifecycle-image.yml workflow should be changed to re-tag an existing image with the release version 
* Add logic to appropriately describe the release in the release notes as minor / patch

ekcasey:
> * Add logic to appropriately describe the release in the release notes as minor / patch

We could also just scrap this from the release notes text. Folks can already see from the version number whether it is a release or a patch.

yaelharel:
In order to protect the `release/*` branches, an owner or an admin should follow the steps in [this link](https://docs.github.com/en/enterprise/2.13/user/articles/enabling-branch-restrictions).
@ekcasey, I guess that you'll be able to do so before we close this issue.

ekcasey:
hey @yaelharel, I am pretty sure the release branches are already protected.

natalieparellano:
[...]

PR #415, opened 2020-09-24
Move release automation to Github actions

yaelharel:
For local development and commits to main, we derive the lifecycle version from `git describe --tags`.
For new commits to release branches, we derive the lifecycle version from the branch name.
New commits to main and release branches will trigger a new lifecycle image.
After publishing new releases, the lifecycle image will be re-tagged.
PRs to any branch won't trigger a new lifecycle image.

This PR should fix issue #389


tools/version/main.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func main() {
+
+	// if the current branch is a release branch, parse the version number from the branch
+
+	cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		fmt.Println("0.0.0")
```
yaelharel, line 16: Do we want exit with an error? Have a different "error version" on every error flow? Anything else?

tools/version/main.go
```diff
@@ -0,0 +1,41 @@
...
+		fmt.Println(matches[1])
+		return
+	}
+
+	// otherwise parse the version from git history (remove leading 'v' and the 'g' in front of commit sha)
+
+	cmd = exec.Command("git", "describe", "--tags")
+	output, err = cmd.CombinedOutput()
+	if err != nil {
+		fmt.Println("0.0.0")
```
yaelharel, line 31: See https://github.com/buildpacks/lifecycle/pull/415/files#r494575956

tools/version/main.go
```diff
@@ -0,0 +1,41 @@
...
+		return
+	}
+	re = regexp.MustCompile("v(?P<version>.+)-(?P<commits>.+)-g(?P<sha>.+)")
+	matches = re.FindStringSubmatch(string(output))
+	if len(matches) != 4 {
+		fmt.Println("0.0.0")
```
yaelharel, line 37: See https://github.com/buildpacks/lifecycle/pull/415/files#r494575956

Makefile
```diff
@@ -4,10 +4,15 @@ PWD?=$(subst /,\,${CURDIR})
...
 /:=\$(BLANK)
-LIFECYCLE_VERSION?=$(shell type VERSION)
 else
 /:=/
-LIFECYCLE_VERSION?=$(shell cat VERSION)
+endif
+
+ifeq ($(LIFECYCLE_VERSION),)
+LIFECYCLE_IMAGE_TAG?=$(shell git describe --always --dirty)
+LIFECYCLE_VERSION:=$(shell go run tools/version/main.go)
```
yaelharel, line 13: Do we want to add `-dirty` to `LIFECYCLE_VERSION` as we do in `LIFECYCLE_IMAGE_TAG`?

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,187 @@
...
+                  owner: "buildpacks",
+                  repo: "lifecycle",
+                  workflow_id: workflow_id,
+                  branch: "release/${{ env.LIFECYCLE_VERSION }}",
+                  event: "push"
+                })
+              })
+              .then(workflow_runs_result => {
+                let workflow_runs = workflow_runs_result.data.workflow_runs
+                  .filter(run => run.conclusion === "success");
```
yaelharel, line 106: We plan to additionally add a check that the SHA is correct.

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,187 @@
...
+          asset_content_type: application/gzip
+      - name: Upload Release Asset - windows
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ${{ env.ARTIFACT_WINDOWS_PATH }}
+          asset_name: lifecycle-v${{ env.LIFECYCLE_VERSION }}+windows.x86-64.tgz
+          asset_content_type: application/gzip
```
dfreilich, line 187: In https://github.com/buildpacks/pack/pull/807, we actually moved from using `actions/create-release` and `actions-upload-release-assets` to using the https://github.com/softprops/action-gh-release action, which allows you to create a release and upload assets at the same time. It may not be worth moving from the defaults, given that you only upload 2 releases, but you can look into it and see.

dfreilich (approved):
Party

.github/workflows/build.yml
```diff
@@ -4,15 +4,19 @@ on:
...
+      - 'release/**'
 
 jobs:
-  test-and-build:
+  test-and-build-linux:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
yaelharel: We had a problem getting the tags (`git describe --tags`) on our fork because the default value of `fetch-depth` is 1, but when setting it to 0, everything works as expected.
For more details, please see https://buildpacks.slack.com/archives/CSFE49VGR/p1600804742079700.

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,187 @@
...
+          [[ $GITHUB_REF =~ ^refs\/heads\/release/(.*)$ ]] && version=${BASH_REMATCH[1]}
+          if [[ -z "${version}" ]]; then
+            echo "lifecycle version not detected."
+            exit 1
+          fi
+          echo "::set-env name=LIFECYCLE_VERSION::$version"
+        shell: bash
+      - name: Determine download url for linux
+        id: artifact-url-linux
+        uses: actions/github-script@0.8.0
```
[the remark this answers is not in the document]
natalieparellano: Thank you, will update!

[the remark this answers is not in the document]
natalieparellano: No reason except this code was copy-pasted from an old pack workflow..

.github/workflows/build.yml
```diff
@@ -67,3 +77,24 @@ jobs:
         with:
           name: lifecycle-windows-x86-64
           path: out/lifecycle-v*+windows.x86-64.tgz
+      - name: Publish lifecycle image
+        run: |
+          If ("${{ github.event_name }}" -Ne 'push') { exit }
+          echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+          $LIFECYCLE_IMAGE_TAG=$(git describe --always --dirty)
+          $LIFECYCLE_VERSION=$(go run ./tools/version/main.go)
```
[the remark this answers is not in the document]
yaelharel: As you can see on line 50, on linux we're able to use `*` (`go run ./tools/image/main.go -lifecyclePath ./out/lifecycle-v*+linux.x86-64.tgz -tag buildpacksio/lifecycle:${LIFECYCLE_IMAGE_TAG}-linux`) while on windows we need to specify the exact version.

.github/workflows/build.yml
```diff
@@ -38,6 +42,12 @@ jobs:
         with:
           name: lifecycle-linux-x86-64
           path: out/lifecycle-v*+linux.x86-64.tgz
+      - name: Publish lifecycle image
+        run: |
+          if [[ ${{ github.event_name }} != 'push' ]]; then exit 0; fi
+          echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+          LIFECYCLE_IMAGE_TAG=$(git describe --always --dirty)
+          go run ./tools/image/main.go -lifecyclePath ./out/lifecycle-v*+linux.x86-64.tgz -tag buildpacksio/lifecycle:${LIFECYCLE_IMAGE_TAG}-linux
   test-and-build-windows:
```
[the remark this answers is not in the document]
natalieparellano: Nope, just linux and windows.

.github/workflows/build.yml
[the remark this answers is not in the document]
natalieparellano: For sure! We run `git describe --tags` to assign a version when on the `main` branch, so we need the repo to be checked out with history. From [actions/checkout](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches): 

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set `fetch-depth: 0` to fetch all history for all branches and tags.

.github/workflows/build.yml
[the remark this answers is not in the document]
ekcasey: Yeah, the darwin version is just for local testing. There is no such thing as a macos container and therefore there is no macos lifecycle.

.github/workflows/build.yml
```diff
@@ -67,3 +77,24 @@ jobs:
         with:
           name: lifecycle-windows-x86-64
           path: out/lifecycle-v*+windows.x86-64.tgz
+      - name: Publish lifecycle image
+        run: |
+          If ("${{ github.event_name }}" -Ne 'push') { exit }
+          echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+          $LIFECYCLE_IMAGE_TAG=$(git describe --always --dirty)
+          $LIFECYCLE_VERSION=$(go run ./tools/version/main.go)
+          go run ./tools/image/main.go -lifecyclePath ./out/lifecycle-v${LIFECYCLE_VERSION}+windows.x86-64.tgz -tag buildpacksio/lifecycle:${LIFECYCLE_IMAGE_TAG}-windows
```
ekcasey, line 86: If we didn't want to maintain both a linux and windows version of this logic we could move all image creation to the `create-lifecycle-image-manifest-list`. Windows image creation doesn't actually need to happen in a windows container. We could invoke `./tools/image/main.go` twice, once with the `-os` flag and generate both images in the linux environment.

This has the added benefit that no images will be published if either linux or windows tests fail.

tools/packager/main.go
```diff
@@ -15,50 +18,88 @@ var (
...
+	return templatedContent.Bytes(), nil
+}
+
+func handle(err error, msg string) {
+	if err != nil {
+		fmt.Println(msg)
+		os.Exit(exitCode)
 	}
+	exitCode++
 }
```
ekcasey, line 105: It not important to preserve the different error codes here. I think the most natural way to reduce duplication in this file would be to pull the bulk of the logic into a function and return and error from that function. Then in `main`, if `err != nil` print the error and exit with a nonzero code.

tools/version/main.go
ekcasey: I think I might go with "unknown"

Makefile
yaelharel: @ekcasey, we would love to know your opinion

ekcasey: I honestly don't really have an opinion here. If pressed, I would say leave if off.

tools/version/main.go
yaelharel: We don't think we can do that because semver will fail parsing a string that isn't in the "version format".

.github/workflows/draft-release.yml
natalieparellano: Ah, thanks for the tip! It's good to know that it exists. We want to hold off on changing it for now, but perhaps we'll revisit it later.

natalieparellano:
@ekcasey we think this **should** be good to go - but we'd like to run the acceptance on our fork one more time (likely on Monday).

ekcasey (approved):

PR #418, opened 2020-09-29
Add buildpack homepage to metadata during build phase

### LINKED ISSUE #349: Add buildpack homepage information to the "io.buildpacks.build.metadata" label on app images produced by lifecycle
Currently, the metadata located under the `io.buildpacks.build.metadata` label contains buildpack `id` and `version` on app images produced by the lifecycle. We reference this metadata to show users what buildpacks were used to produce their images. However, users don't always know where to go to get more information on a specific buildpack. Therefore, we'd like to add `homepage` to that metadata so we can show users where to look to get more information.

djoyahoy:
Closes #349


buildpack.go
```diff
@@ -29,6 +30,11 @@ func (bp Buildpack) noAPI() Buildpack {
 	return bp
 }
 
+func (bp Buildpack) withHomepage(bpInfo BuildpackInfo) Buildpack {
+	bp.Homepage = bpInfo.Homepage
```
jromero, line 34:
```suggestion
func (bp Buildpack) withHomepage(string url) Buildpack {
	bp.Homepage = bpInfo.Homepage
```

It seems unnecessary to pass a whole object in a method that only cares about a single value.

buildpack.go
```diff
@@ -29,6 +30,11 @@ func (bp Buildpack) noAPI() Buildpack {
 	return bp
 }
 
+func (bp Buildpack) withHomepage(bpInfo BuildpackInfo) Buildpack {
```
yaelharel, line 33:
```suggestion
func (bp Buildpack) setHomepage(bpInfo BuildpackInfo) Buildpack {
```

natalieparellano: I kind of like `withHomepage` ...it's more clear that the `bp` of the calling fucntion isn't actually being modified.

buildpack.go
djoyahoy: Ah, good call!

buildpack.go
yaelharel: @natalieparellano, I agree. I missed this part.
Resolving this conversation.

builder.go
```diff
@@ -174,7 +176,7 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 
 	return &BuildMetadata{
 		BOM:        bom,
-		Buildpacks: b.Group.Group,
```
ekcasey, line 177: 1. Instead of replacing the group the `group.toml` with an enhanced group here I think we actually want to add `homepage` to `group.toml` -https://github.com/buildpacks/spec/pull/127/files

2. The merged spec change doesn't actually add `homepage` to `metadata.toml`. I'll grant that may be unexpected in that this is the first time `metadata.toml` and the `io.buildpacks.build.metadata` label are diverging. However, the label currently contains a lot of information we don't need/want inside the contianer filesystem (e.g. the bom) and I expect these to diverge further in the future (I have a RFC planned on this topic).

ekcasey (changes requested):

builder_test.go
```diff
@@ -372,17 +372,17 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
 					Buildpacks: []lifecycle.Buildpack{
-						{ID: "A", Version: "v1", API: "0.3"},
+						{ID: "A", Version: "v1", API: "0.3", Homepage: "Buildpack A Homepage"},
 						{ID: "B", Version: "v2", API: "0.2"},
 					},
 					BOM: []lifecycle.BOMEntry{
 						{
 							Require:   lifecycle.Require{Name: "dep1", Version: "v1"},
-							Buildpack: lifecycle.Buildpack{ID: "A", Version: "v1"},
+							Buildpack: lifecycle.Buildpack{ID: "A", Version: "v1", Homepage: "Buildpack A Homepage"},
```
ekcasey, line 381: I think it is okay to have the `homepage` in the `buidpacks` section of `metadata.toml`, but I think we should leave it out of the BOM entries like we do with API https://github.com/buildpacks/lifecycle/blob/main/builder.go#L207

ekcasey (changes requested):
Can we test `homepage` in the `exporter` unit tests. It appears to be working, but we should ensure that the `homepage` value ends up in the final label as that was the motivation for this change.

relevant exporter tests:
https://github.com/buildpacks/lifecycle/blob/main/exporter_test.go#L460-L469

builder_test.go
djoyahoy: I'm surprised that we would want to leave the homepage out of the BOM, but I also don't know how the BOM will be used or displayed to users.

ekcasey: I think given that the image metadata now contains the homepage for each buildpack we don't need to repeat it in each BOM entry. The BOM entry only needs enough data to uniquely identify the buildpack. The homepage can be looked up using the buildpack `id` and `version`.

ekcasey (approved):

PR #422, opened 2020-10-02
Update windows base image to use LTSC Server Core 2019

TisVictress:
The lifecycle currently uses a Windows base image called `nanoserver`, which is going out of support soon.  This change updates the base image to `servercore:ltsc2019`

### DIFF
--- a/acceptance/testdata/launcher/windows/Dockerfile
+++ b/acceptance/testdata/launcher/windows/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/windows/nanoserver:1809
+FROM mcr.microsoft.com/windows/servercore:ltsc2019
 
 COPY container /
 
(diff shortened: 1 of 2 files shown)


tools/go.mod
```diff
@@ -3,15 +3,41 @@ module github.com/buildpacks/lifecycle/tools
 go 1.14
 
 require (
```
ekcasey, line 5: Why are the tools `go.mod` and `go.sum` modified? It doesn't seem like windows base image change should require any dependency additions/upgrades?

ekcasey (changes requested):

tools/go.mod
TisVictress: @ekcasey  That's a good question...I'm not sure either.  When I ran `make test`, these files were updated.  Should I undo these changes?

ekcasey (approved):

PR #431, opened 2020-10-06
Remove double parsing of commit sha in Makefile

natalieparellano:
This requires extra processing in cmd/version.go because Version when it includes the commit will not contain
'-dirty'.

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -28,13 +28,11 @@ const (
 // buildVersion is a display format of the version and build metadata in compliance with semver.
 func buildVersion() string {
 	// noinspection GoBoolExpressions
-
-	sha := strings.TrimSuffix(SCMCommit, "-dirty")
-	if SCMCommit == "" || strings.Contains(Version, sha) {
+	if SCMCommit == "" || strings.Contains(Version, SCMCommit) {
 		return Version
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/version.go
```diff
@@ -28,11 +28,13 @@ const (
 // buildVersion is a display format of the version and build metadata in compliance with semver.
 func buildVersion() string {
 	// noinspection GoBoolExpressions
-	if SCMCommit == "" || strings.Contains(Version, SCMCommit) {
+
+	sha := strings.TrimSuffix(SCMCommit, "-dirty")
```
ekcasey, line 32: Maybe we should just use the same commit in both the version output and the lifecycle tag, either with or without dirty seems fine. Do we have a reason to prefer/maintain this difference?

ekcasey: should we just use the version without `-dirty` in both places then?

yaelharel: I think it's a good idea to have consistency here.
Maybe we can remove `-dirty` for now and open a new issue for adding it in all places.

natalieparellano: I am fine with this. `-dirty` provides a little extra information for local development but it seems not worth the extra confusion in having to keep track of all these versions!

ekcasey (approved):

PR #432, opened 2020-10-06
Remove all unsafe usage of filepath.Glob

ekcasey:
Fixes 2 issues:
**Issue 1** Unsafe use of `filepath.Glob`

We previous used filepath.Glob for four purposes:

1. Globbing slice files - This was updated to use a combination of filepath.Walk and filepath.Match to match children of the app dir by relative path. We already had logic to unsure that no files outside of the app dir were included. A test was added to confirm this and prevent regressions

2. Globbing profile scripts - This commit changes the implementation so that we no longer use filepath.Glob here

3. Globbing layer.toml files - This commit changes the implementation so that we no longer use filepath.Glob here

4. Test logic - unsafe use of filepath.Glob is unchanged in the test code

**Issue 2** Slice globbing on Windows

According to the buildpack API in the CNB Speciication slice path globs should follow the pattern syntac defined in the Go standard library (Follow the pattern syntax defined in the [Go standard library](https://golang.org/pkg/path/filepath/#Match).). We were previously translating unix path separators in slice patterns into windows slice separators before matching. This contradicts the wording of the spec. Instead windows path separators should be notated with '\\' as described in the stdlib godoc.

### DIFF
--- a/launch/launcher.go
+++ b/launch/launcher.go
@@ -90,7 +90,7 @@ func (l *Launcher) env(process Process) error {
 	if err != nil {
 		return errors.Wrap(err, "find app directory")
 	}
-	return l.eachBuildpack(l.LayersDir, func(path string) error {
+	return l.eachBuildpack(func(path string) error {
 		bpInfo, err := os.Stat(path)
 		if err != nil {
 			if os.IsNotExist(err) {
@@ -142,9 +142,13 @@ func eachDir(dir string, fn func(path string) error) error {
 	return nil
 }
 
-func (l *Launcher) eachBuildpack(dir string, fn func(path string) error) error {
+func (l *Launcher) eachBuildpack(fn func(path string) error) error {
+	layersDir, err := filepath.Abs(l.LayersDir)
+	if err != nil {
+		return errors.Wrap(err, "failed to compute absolute path of layers directory")
+	}
 	for _, bp := range l.Buildpacks {
-		if err := fn(filepath.Join(l.LayersDir, EscapeID(bp.ID))); err != nil {
+		if err := fn(filepath.Join(layersDir, EscapeID(bp.ID))); err != nil {
 			return err
 		}
 	}
--- a/launch/launcher.go
+++ b/launch/launcher.go
@@ -143,12 +143,8 @@ func eachDir(dir string, fn func(path string) error) error {
 }
 
 func (l *Launcher) eachBuildpack(fn func(path string) error) error {
-	layersDir, err := filepath.Abs(l.LayersDir)
-	if err != nil {
-		return errors.Wrap(err, "failed to compute absolute path of layers directory")
-	}
 	for _, bp := range l.Buildpacks {
-		if err := fn(filepath.Join(layersDir, EscapeID(bp.ID))); err != nil {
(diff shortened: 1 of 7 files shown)


layers/slices_test.go
```diff
@@ -312,5 +312,32 @@ func testSlices(t *testing.T, when spec.G, it spec.S) {
 				}...))
 			})
 		})
+
+		when("the dir has special characters", func() {
```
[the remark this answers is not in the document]
ekcasey: FYI - I just pushed an update to make this test excercise the globbing functionality as originally intended.

In this test the app dir has a name `test-di[r]` that, if included in a glob would be interpreted as a pattern that would match the adjacent director `test-dir` (see [bash pattern matching](https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html)).

You can see this experimentally with `ls`
```
ls -l ./layers/testdata/target-di[r]
total 0
lrwxr-xr-x  1 ekcasey  staff   17 Oct  6 16:34 dir-link -> ../dir-to-exclude
lrwxr-xr-x  1 ekcasey  staff   31 Oct  6 16:34 file-link.txt -> ../dir-to-exclude/some-file.txt
-rw-r--r--  1 ekcasey  staff    0 Oct  6 16:34 file.txt
drwxr-xr-x  4 ekcasey  staff  128 Oct  6 16:34 other-dir
drwxr-xr-x  4 ekcasey  staff  128 Oct  6 16:34 some-dir
```
The files printed above are the contents of `target-dir`, not the contents of `target-di[r]`.

By adding a test that uses an app dir that includes these special characters we are ensuring that the app dir is treated as a literal name and not as a pattern when we create slice layers.

launch/launcher.go
```diff
@@ -142,9 +142,13 @@ func eachDir(dir string, fn func(path string) error) error {
 	return nil
 }
 
-func (l *Launcher) eachBuildpack(dir string, fn func(path string) error) error {
+func (l *Launcher) eachBuildpack(fn func(path string) error) error {
+	layersDir, err := filepath.Abs(l.LayersDir)
```
yaelharel, line 146: Does this change relate to subject of this PR - "Remove all unsafe usage of filepath.Glob", or did you just do it on the way?
I think it's good that we're not passing the `dir` to `eachBuildpack` anymore, but I don't understand why we need `filepath.Abs`.

ekcasey: I just made this refactor along the way. We were previous doing this for profile scripts here https://github.com/buildpacks/lifecycle/pull/432/files#diff-470373c3e39c585a84421cca78d2d178L62-L65. We weren't previously doing this for env files (the other comsumers of `eachBuildpack`), but it doesn't hurt and I like the consistent interface. If I call `eachBuildpack`, I don't have to worry about absolute vs relative paths.

yaelharel: This is what I thought. Great! Thanks!

yaelharel (approved):

launch/launcher.go
ekcasey: After thinking more about your question I ended up moving this to make it profile script specific again to make this less confusing and avoid changing behavior unnecessarily.

PR #433, opened 2020-10-06
Enable errcheck

ekcasey:
Enables `errcheck` linter in golangci-lint. Fixes missing error checks!

### DIFF
--- a/analyzer_test.go
+++ b/analyzer_test.go
@@ -59,7 +59,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		}
 		if testing.Verbose() {
 			analyzer.Logger = cmd.DefaultLogger
-			cmd.SetLogLevel("debug")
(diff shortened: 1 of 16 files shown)


natalieparellano:
Looks like we found some real errors on Windows :) I think if we move `h.AssertNil(t, os.RemoveAll(tmpDir))` to the end of the `it.After` that should do the trick.

jabrown85 (approved):
🥳

ekcasey:
@natalieparellano 

> Looks like we found some real errors on Windows :) I think if we move h.AssertNil(t, os.RemoveAll(tmpDir)) to the end of the it.After that should do the trick.

Instead of removing the cleanup I tracked down why it was failing. Fixing cleanup requires some fixes to the fake image: https://github.com/buildpacks/imgutil/pull/65

yaelharel (approved):

PR #434, opened 2020-10-08
-no-color strips color from buidpack logs

### LINKED ISSUE #329: Strip color from buildpack output when -no-color is true
`-no-color` should strip color from buildpack logs.

See example implementation in `pack`: https://github.com/buildpacks/pack/pull/696

Adding this functionality to lifecycle would enable a better UX on all platforms.

ekcasey:
* Resolves #329

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -2,8 +2,8 @@ package lifecycle
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
-	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -26,7 +26,7 @@ type Builder struct {
 	Env           BuildEnv
 	Group         BuildpackGroup
 	Plan          BuildPlan
-	Out, Err      *log.Logger
+	Out, Err      io.Writer
 }
 
 type BuildEnv interface {
@@ -115,8 +115,8 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 			bpPlanPath,
 		)
 		cmd.Dir = appDir
-		cmd.Stdout = b.Out.Writer()
-		cmd.Stderr = b.Err.Writer()
+		cmd.Stdout = b.Out
+		cmd.Stderr = b.Err
 
 		if bpInfo.Buildpack.ClearEnv {
 			cmd.Env = b.Env.List()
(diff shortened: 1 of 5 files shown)


yaelharel:
Hi @ekcasey, I have a few questions about this PR:

- Regarding [this PR](https://github.com/buildpacks/lifecycle/pull/328). Does the current commit apply to the buildpacks logs while the previous one apply to other some other type of logs?
- I think it's great to give users the ability to disable the color in their output if they're willing to do so, but I'm trying to understand why would they want to do that. Is it because of what appears in [this message](https://github.com/buildpacks/lifecycle/issues/312#issue-637256336)?
- Are you planning to add tests to this new commit (similar to [this PR in pack](https://github.com/buildpacks/pack/pull/696/files))?
- I saw that [this issue in pack](https://github.com/buildpacks/pack/issues/123) was resolved. Does it mean that we don't need this commit if we're using pack?

ekcasey:
@yaelharel 
> Regarding this PR. Does the current commit apply to the buildpacks logs while the previous one apply to other some other type of logs?

Yes the previous commit made `-no-color` work for logs written directly by the lifecycle, but it did not apply to buildpack logs. This PR fixes `-no-color`, to make it work for all logs.

> I think it's great to give users the ability to disable the color in their output if they're willing to do so, but I'm trying to understand why would they want to do that. Is it because of what appears in this message?

Yes. The control characters that create colors in a TTY look like garbage when printed anywhere other than a terminal.

> Are you planning to add tests to this new commit (similar to this PR in pack)?

Uhh... I wasn't, given that this part of the code was previously untested. But I would be happy to add a test.

> I saw that this issue in pack was resolved. Does it mean that we don't need this commit if we're using pack?

Yes, `pack` users won't care about this change because `pack` with `--no-color` already strips color from the lifecycle output. But other platforms should be able to turn off color easily.

jabrown85 (approved):
This looks great!

yaelharel (approved):
@ekcasey, thanks for your answers! It's really helpful!
The changes looks good to me.
Unless you would like to add some tests, I think you can merge the PR.

ekcasey:
@yaelharel If you'll forgive me I am going to punt on the test for now. I think an acceptance test would make the most sense but the only phase that exercises this logic (the `builder` phase) doesn't yet have any acceptance tests.

PR #438, opened 2020-10-13
Faster linux builds

ekcasey:
* caches go modules in a volumes
* uses docker build to create linux compilation image


jabrown85 (approved):
💯

PR #442, opened 2020-10-19
Builds windows artifacts on linux

### LINKED ISSUE #441: pack create-builder creates broken windows builders when archive was packaged on windows
### Summary
When lifecycle archives are packaged on `windows` and a builder is created using the archive with `pack create-builder`. The created builder fails when run as a non-admin user.

---

##### Steps

1. Download and unzip an `0.9.2` release candidate windows artifact that was built in GHA on a windows worker: https://github.com/buildpacks/lifecycle/suites/1338822592/artifacts/21560675
2. Add the following lines to the nanoserver sample `builder.toml`
```
[lifecycle]
uri = "/path/to/lifecycle-v0.9.2+windows.x86-64.tgz"
```
3. Create a nanoserver sample builder
```
pack create-builder -c ./builders/nanoserver-1809/builder.toml bldr/nanoserver:0.9.2
```
4. Pull and tag matching lifecycle image
```
docker pull buildpacksio/lifecycle:b069455
docker tag buildpacksio/lifecycle:b069455 buildpacksio/lifecycle:0.9.2
```
5. Build a sample app with the builder
```
pack build -p ./apps/batch-script --builder bldr/nanoserver:0.9.2 samples/batch
```

##### Current behavior
The following error occurs:
```
===> DETECTING
ERROR: failed to build: executing lifecycle. This may be the result of using an untrusted builder: container start: Error response from daemon: container 5f5228da392b2cb411a11564300070680c60ecb408c9e10a613d12aa515e447f encountered an error during hcsshim::System::CreateProcess: failure in a Windows system call: Access is denied. (0x5) extra info: {"CommandLine":"/cnb/lifecycle/detector -app c:\\workspace -platform c:\\platform","User":"ContainerUser","WorkingDirectory":"/layers","Environment":{"CNB_GROUP_ID":"1","CNB_PLATFORM_API":"0.4","CNB_STACK_ID":"io.buildpacks.samples.stacks.nanoserver-1809","CNB_USER_ID":"1","PATH":"c:\\Windows\\system32;C:\\Windows"},"CreateStdInPipe":true,"CreateStdOutPipe":true,"CreateStdErrPipe":true,"ConsoleSize":[0,0]}
```

##### Expected
A successful build

##### Cause

This happens because the lifecycle executables in the generated windows builder image are missing a needed `MSWINDOWS.rawsd` PAX header.

This happens because:
1. The lifecycle, when packaging an archive on windows adds a `MSWINDOWS.fileattr` PAX header to each tar entry containing windows file attributes.
2. The version of `imgutil` imported in `pack` version `1.14.1` does not include this imgutil bug fix and therefore does not add the missing head if any other PAX header is set - https://github.com/buildpacks/imgutil/pull/55

##### Solution

This issue did not arise in lifecycle `v0.9.1` because previously released windows archives were built in a linux environment and therefore the MSWINDOWS.fileattr` was not added.

Although this issue is new to release `0.9.2` the underlying cause appears to be an issue in `pack`. For now, we can build and package all lifecycle artifacts on a linux machine, and therefore sidestep the issue above.

---

##### lifecycle version
Current head of `release/0.9.2` branch -> commit `b069455`
##### platform version(s)
```
pack --version
0.14.1+git-b630309.build-1353
```

ekcasey:
* Fixes #441

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,12 +70,6 @@ jobs:
           echo "::set-env name=GOPATH::$(go env GOPATH)"
           echo "::add-path::$(go env GOPATH)/bin"
         shell: bash
-      - name: Install jq
-        run: |
-          mkdir -p deps/bin
-          curl -s -L -o deps/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
-          chmod +x deps/bin/jq
-          echo "::add-path::${PWD}/deps/bin"
       - name: Build
         run: |
           make clean
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ on:
       - 'release/**'
 
 jobs:
-  test-linux-and-build:
+  test-linux:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -34,19 +34,6 @@ jobs:
           echo "::add-path::${PWD}/deps/bin"
       - name: Test
         run: make test
-      - name: Build
-        run: |
-          make clean
-          make build
-          make package
-      - uses: actions/upload-artifact@v2
-        with:
-          name: lifecycle-linux-x86-64
-          path: out/lifecycle-v*+linux.x86-64.tgz
-      - uses: actions/upload-artifact@v2
-        with:
-          name: lifecycle-windows-x86-64
-          path: out/lifecycle-v*+windows.x86-64.tgz
   test-windows:
     runs-on: windows-latest
     steps:
@@ -67,10 +54,9 @@ jobs:
       - name: Test
         run: make test
         shell: cmd
-  create-lifecycle-image-manifest-list:
-    if: github.event_name == 'push'
+  build:
     needs:
-      - test-linux-and-build
+      - test-linux
       - test-windows
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -37,13 +37,17 @@ jobs:
...
-          make package-linux
+          make build
+          make package
       - uses: actions/upload-artifact@v2
         with:
           name: lifecycle-linux-x86-64
           path: out/lifecycle-v*+linux.x86-64.tgz
-  test-and-build-windows:
+      - uses: actions/upload-artifact@v2
+        with:
```
yaelharel, line 47: Is it possible to upload multiple artifacts under one `uses: actions/upload-artiface@v2`?

.github/workflows/build.yml
```diff
@@ -37,13 +37,17 @@ jobs:
       - name: Build
         run: |
           make clean
-          make build-linux
-          make package-linux
+          make build
+          make package
```
yaelharel, line 41: Maybe it's better to `build` after both `test-linux` and `test-windows` finish successfully.
@ekcasey, what do you think?

.github/workflows/build.yml
ekcasey: Skimming the docs, it appears the answer is no, not if we want them attached as separate artifacts, rather than a single zip containing both artifacts.

.github/workflows/build.yml
ekcasey: sure thing!

yaelharel (approved):

ekcasey:
GH appears to be running outdated checks. Closing and reopening in an attempt to fix.

PR #444, opened 2020-10-19
Fix artifact paths in GHA

ekcasey:
* Removes unecessary 'Set up go env' steps

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,17 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
       - name: Set up go
         uses: actions/setup-go@v2-beta
         with:
           go-version: '1.15'
-      - name: Set up go env
-        run: |
-          echo "::set-env name=GOPATH::$(go env GOPATH)"
-          echo "::add-path::$(go env GOPATH)/bin"
-        shell: bash
       - name: Install jq
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -15,17 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
```
yaelharel, line 19: What was changed since https://github.com/buildpacks/lifecycle/pull/415/files#r494993607 ?

.github/workflows/build.yml
```diff
@@ -43,8 +36,6 @@ jobs:
           git config --global core.eol lf
           git config --global core.symlinks true
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
```
yaelharel, line 47: Same as https://github.com/buildpacks/lifecycle/pull/444/files#r508181188

.github/workflows/build.yml
ekcasey: We don't need the history to run tests but we do need it to build and package, so I moved these lines to the `build` job https://github.com/buildpacks/lifecycle/pull/444/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R54-R55

yaelharel: You're right. I missed it. Thanks for the clarification!

PR #449, opened 2020-10-22
Change defaults that depend on the working dir to the layers dir

### LINKED ISSUE #393: Change defaults that depend on the working dir
https://github.com/buildpacks/spec/pull/128

In platform API 0.5, all default file paths that are currently relative to the working directory should changed to be relative to the layers directory.

yaelharel:

Fixes #393

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -6,25 +6,27 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+
+	"github.com/buildpacks/lifecycle/api"
 )
 
 var (
-	DefaultAnalyzedPath        = filepath.Join(".", "analyzed.toml")
-	DefaultAppDir              = filepath.Join(rootDir, "workspace")
-	DefaultBuildpacksDir       = filepath.Join(rootDir, "cnb", "buildpacks")
-	DefaultDeprecationMode     = DeprecationModeWarn
-	DefaultGroupPath           = filepath.Join(".", "group.toml")
-	DefaultLauncherPath        = filepath.Join(rootDir, "cnb", "lifecycle", "launcher"+execExt)
-	DefaultLayersDir           = filepath.Join(rootDir, "layers")
-	DefaultLogLevel            = "info"
-	DefaultOrderPath           = filepath.Join(rootDir, "cnb", "order.toml")
-	DefaultPlanPath            = filepath.Join(".", "plan.toml")
-	DefaultPlatformAPI         = "0.3"
-	DefaultPlatformDir         = filepath.Join(rootDir, "platform")
-	DefaultProcessType         = "web"
-	DefaultProjectMetadataPath = filepath.Join(".", "project-metadata.toml")
-	DefaultReportPath          = filepath.Join(".", "report.toml")
-	DefaultStackPath           = filepath.Join(rootDir, "cnb", "stack.toml")
+	DefaultAnalyzedFileName        = "analyzed.toml"
+	DefaultAppDir                  = filepath.Join(rootDir, "workspace")
+	DefaultBuildpacksDir           = filepath.Join(rootDir, "cnb", "buildpacks")
+	DefaultDeprecationMode         = DeprecationModeWarn
+	DefaultGroupFileName           = "group.toml"
+	DefaultLauncherPath            = filepath.Join(rootDir, "cnb", "lifecycle", "launcher"+execExt)
+	DefaultLayersDir               = filepath.Join(rootDir, "layers")
+	DefaultLogLevel                = "info"
+	DefaultOrderPath               = filepath.Join(rootDir, "cnb", "order.toml")
+	DefaultPlanFileName            = "plan.toml"
+	DefaultPlatformAPI             = "0.3"
+	DefaultPlatformDir             = filepath.Join(rootDir, "platform")
+	DefaultProcessType             = "web"
+	DefaultProjectMetadataFileName = "project-metadata.toml"
+	DefaultReportFileName          = "report.toml"
+	DefaultStackPath               = filepath.Join(rootDir, "cnb", "stack.toml")
 )
 
 const (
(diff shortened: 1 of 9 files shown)


cmd/flags.go
```diff
@@ -206,3 +213,14 @@ func EnvOrDefault(key string, defaultVal string) string {
 	}
 	return defaultVal
 }
+
+func isVersionLessThan05(version string) bool {
+	return api.MustParse(version).Compare(api.MustParse("0.5")) < 0
+}
+
+func getRelativePath(version, layersDir string) string {
```
yaelharel, line 221: Does anyone have a better name for this function?

cmd/lifecycle/detector.go
```diff
@@ -33,8 +34,8 @@ func (d *detectCmd) Init() {
 	cmd.FlagAppDir(&d.appDir)
 	cmd.FlagPlatformDir(&d.platformDir)
 	cmd.FlagOrderPath(&d.orderPath)
-	cmd.FlagGroupPath(&d.groupPath)
-	cmd.FlagPlanPath(&d.planPath)
+	cmd.FlagGroupPath(d.platformAPI, "", &d.groupPath)
+	cmd.FlagPlanPath(d.platformAPI, "", &d.planPath)
```
yaelharel, line 38: Should we allow the user to set the `layers` directory in this command or just keep it empty as is now and then the default `groupPath` and `planPath` will be under `.`?

cmd/lifecycle/rebaser.go
```diff
@@ -32,7 +33,7 @@ type rebaseCmd struct {
 
 func (r *rebaseCmd) Init() {
 	cmd.FlagGID(&r.gid)
-	cmd.FlagReportPath(&r.reportPath)
+	cmd.FlagReportPath(r.platformAPI, "", &r.reportPath)
```
yaelharel, line 36: See https://github.com/buildpacks/lifecycle/pull/449/files#r510449985

yaelharel:
Two notes regarding this PR (in addition to those I added in the code):
* I didn't add any new tests. Do you think it's needed?
* I changed the order of the calls in the `Init` functions so `cmd.FlagLayersDir` is being called before the functions that set a path that is based on the `layers` directory (such as `groupPath`). I'm wondering whether we should add a check that this is the calls order.

cmd/flags.go
```diff
@@ -83,8 +86,9 @@ func FlagGID(gid *int) {
 	flagSet.IntVar(gid, "gid", intEnv(EnvGID), "GID of user's group in the stack's build and run images")
 }
 
-func FlagGroupPath(path *string) {
-	flagSet.StringVar(path, "group", EnvOrDefault(EnvGroupPath, DefaultGroupPath), "path to group.toml")
+func FlagGroupPath(version, layersDir string, path *string) {
+	groupDir := getRelativePath(version, layersDir)
+	flagSet.StringVar(path, "group", EnvOrDefault(EnvGroupPath, filepath.Join(groupDir, DefaultGroupFileName)), "path to group.toml")
```
ekcasey, line 91: Unfortunately this implementation doesn't work because at the the time `FlagGroupPath` is called, the the flags have not been parsed yet (see https://github.com/buildpacks/lifecycle/blob/main/cmd/command.go#L27-L38). There for `layerDir` string passed here and use to set the default is _always_ the default layers dir regardless of whether the user passes `-layers`. I think we need to either:
1. brute force this by not setting a default value for the flag and applying the computed default later if it is unset.
or
2. do something clever that I haven't through all the way with our command object that involve multiple rounds of parsing.

ekcasey (changes requested):
Since this is the first Platform API 0.5 feature we need to add Platform API 0.5 to https://github.com/buildpacks/lifecycle/blob/main/api/apis.go#L8 to be able to use try this feature. We should probably add it to `lifecycle.toml` while we are at it.

Unfortunately this implementation doesn't produce the correct results
```
CNB_PLATFORM_API=0.5 ./out/darwin/lifecycle/analyzer -layers /tmp some-image
Warning: Not restoring cached layer metadata, no cache flag specified.
ERROR: failed to read buildpack group: open /layers/group.toml: no such file or directory
```
see

cmd/lifecycle/rebaser.go
yaelharel: https://github.com/buildpacks/spec/issues/156

cmd/lifecycle/main.go
```diff
@@ -19,17 +19,17 @@ func main() {
...
 	case "restorer":
-		cmd.Run(&restoreCmd{}, false)
+		cmd.Run(&restoreCmd{platformAPI: platformAPI}, false)
 	case "builder":
 		cmd.Run(&buildCmd{buildArgs: buildArgs{platformAPI: platformAPI}}, false)
 	case "exporter":
 		cmd.Run(&exportCmd{exportArgs: exportArgs{platformAPI: platformAPI}}, false)
 	case "rebaser":
-		cmd.Run(&rebaseCmd{}, false)
+		cmd.Run(&rebaseCmd{platformAPI: platformAPI}, false)
```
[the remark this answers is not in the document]
yaelharel: When I added `platformAPI` to the commands, I was wondering whether I should add them as they're right now in the PR or I make them part of `*Args`. I thought that my idea is better because I didn't want to pass the value to functions that don't care about it (although it'll be consistent with `buildArgs` and `exportArgs`. But if you think that we'll need it in the **near** future, I'll make this change.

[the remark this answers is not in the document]
yaelharel: I'm not sure I understand. Should I make any changes to the link you indicated in the `creator` (https://github.com/buildpacks/lifecycle/blob/main/cmd/lifecycle/creator.go#L167)?

yaelharel:
I'm going to close this PR following https://github.com/buildpacks/lifecycle/pull/449#discussion_r510946726 and submit a new PR with a new approach to this problem.

Issue #425, opened 2020-10-05
Rebase should validate mixins


ekcasey:
## Summary
`rebaser` should ensure that that new run-image has the same mixins or a superset of the mixins present on the app image.

This validation should use the relaxed mixin contract described in https://github.com/buildpacks/rfcs/pull/109

### Case 1 - Identical Mixins
**Given** `<image>` has following mixins `mixin-1` `run:mixin-2`
**Given** a `<run-image>` has mixings `mixin-1` `run:mixin-2`
**When** `rebaser -run-image <run-image> <image> `
**Then** the rebase should **succeed**

### Case 2 - Extra Mixins
**Given** `<image>` has following mixins `mixin-1` `run:mixin-2`
**Given** a `<run-image>` has mixings `mixin-1` `run:mixin-2` `mixin-3`
**When** `rebaser -run-image <run-image> <image> `
**Then** the rebase should **succeed**

### Case 3 - No stage specifier
**Given** `<image>` has following mixins `mixin-1` `run:mixin-2`
**Given** a `<run-image>` has mixings `mixin-1` `mixin-2`
**When** `rebaser -run-image <run-image> <image> `
**Then** the rebase should **succeed**

### Case 4 - Extra stage specifier
**Given** `<image>` has following mixins `mixin-1` `run:mixin-2`
**Given** a `<run-image>` has mixings `run:mixin-1` `run:mixin-2`
**When** `rebaser -run-image <run-image> <image> `
**Then** the rebase should **succeed**

### Case 3 - Missing Mixins
**Given** `<image>` has following mixins `mixin-1` `run:mixin-2`
**Given** a `<run-image>` has mixings `run:mixin-2`
**When** `rebaser -run-image <run-image> <image> `
**Then** the rebase should **fail** with an informative error message

ekcasey:
Resolved in #450

PR #450, opened 2020-10-22
feature: validate mixins during rebase

jabrown85:
Implemented as discussed [here](https://github.com/buildpacks/lifecycle/issues/425). Using the `io.buildpacks.stack.mixins` label of the working and new run image, lifecycle will now verify that mixins are present on the new base image prior to the rebase operation.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -62,8 +62,7 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
 	return nil
 }
 
-// RemoveStagePrefixes creates a slice with the build: or run: style prefixes removed.
-func RemoveStagePrefixes(mixins []string) []string {
+func removeStagePrefixes(mixins []string) []string {
 	var result []string
 	for _, m := range mixins {
 		s := strings.SplitN(m, ":", 2)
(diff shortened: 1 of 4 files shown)


utils.go
```diff
@@ -62,6 +62,58 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
...
+			result = append(result, s[1])
+		}
+	}
+	return result
+}
+
+// FromSlice converts the given slice to a set in the form of unique keys in a map.
+// The value associated with each key should not be relied upon. A value is present
+// in the set if its key is present in the map, regardless of the key's value.
+// from: https://github.com/buildpacks/pack/blob/main/internal/stringset/stringset.go
```
jabrown85, line 82: The 2 methods with this comment were pulled from pack. I wasn't sure if there is a better place for these to exist. I didn't think we'd want to reference pack as a library :)

utils.go
```diff
@@ -62,6 +62,58 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
 	return nil
 }
 
+// RemoveStagePrefixes creates a slice with the build: or run: style prefixes removed.
+func RemoveStagePrefixes(mixins []string) []string {
```
ekcasey, line 66: I don't think `lifecycle.RemoveStagePrefixes` should be exported from the lifecycle package.

If `pack` or another platform wants to consume this functionality from lifecycle as a lifecycle I think we should do either of the following instead
1. export a higher level method instead along the lines of `ValidateRebaseMixins`.
2. move utitlities into a mixins package so that the interface is more like`mixins.RemoveStagePrefixes`. In this case I would want to take the time to design what we think is the best library interface for mixins rather than taking this wholesale.

I inclined to say we can just leave this unexported for now.

utils.go
ekcasey: I think we can leave them here but we shouldn't export them.

Honestly I am not sure we need to 
a) describe the `FromSlice` behavior in quite as much detail in the comment here; It is common to use a `map[T]struct{}` to represent a set in golang.
b) have a `FromSlice` func at all? I am happy to leave it if we don't export it but I don't necessarily see this standard `for` loop as the type of duplication that needs to be DRYed up.

utils.go
```diff
@@ -62,6 +62,58 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
...
+
+// FromSlice converts the given slice to a set in the form of unique keys in a map.
+// The value associated with each key should not be relied upon. A value is present
+// in the set if its key is present in the map, regardless of the key's value.
+// from: https://github.com/buildpacks/pack/blob/main/internal/stringset/stringset.go
+func FromSlice(strings []string) map[string]interface{} {
+	set := map[string]interface{}{}
+	for _, s := range strings {
+		set[s] = nil
+	}
```
ekcasey, line 87:
```suggestion
func FromSlice(strings []string) map[string]struct{} {
	set := map[string]struct{}{}
	for _, s := range strings {
		set[s] = struct{}
	}
```
It is typical to use a struct for this since it takes zero bytes but some people also use `map[T]bool` https://www.davidkaya.com/sets-in-golang/.

ekcasey (commented):
Nice work!

I only left a few nitpicky stylistic suggestions.

ekcasey (approved):

jabrown85:
@ekcasey do I need to do anything else with this? Or are we holding for a release branch to merge into?

I am not in need, just don't want it to be left because it is waiting on something I wasn't aware of. 😄

ekcasey:
@jabrown85 apologies! For some reason I thought I had merged it. Merging now

PR #454, opened 2020-10-29
Change defaults that are depending on the working dir to be under <layers>

### LINKED ISSUE #393: Change defaults that depend on the working dir

yaelharel:
Notes:
* This commit introduces platfromAPI 0.5 as a supported API.
* The changes apply only for platformAPI >= 0.5.
* The path of the relevant files is being changed as part of Args after parsing the flags.
* Allow the user to set the layers flag in the detector.

Fixes #393

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -88,174 +88,143 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		os.RemoveAll(copyDir)
 	})
 
-	when("called without an app image", func() {
-		it("errors", func() {
-			cmd := exec.Command(
-				"docker", "run", "--rm",
-				"--env", "CNB_REGISTRY_AUTH={}",
-				analyzeImage,
-				analyzerPath,
-			)
-			output, err := cmd.CombinedOutput()
-
-			h.AssertNotNil(t, err)
-			expected := "failed to parse arguments: received 0 arguments, but expected 1"
-			h.AssertStringContains(t, string(output), expected)
-		})
-	})
-
-	when("cache image tag and cache directory are both blank", func() {
-		it("warns", func() {
-			output := h.DockerRun(t,
-				analyzeImage,
-				h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
-				h.WithArgs(analyzerPath, "some-image"),
-			)
-
-			expected := "Not restoring cached layer metadata, no cache flag specified."
-			h.AssertStringContains(t, output, expected)
-		})
-	})
-
-	when("the provided layers directory isn't writeable", func() {
-		it("recursively chowns the directory", func() {
-			h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
-
-			output := h.DockerRun(t,
-				analyzeImage,
-				h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
-				h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s some-image; ls -al /layers", analyzerPath)),
-			)
-
-			h.AssertMatch(t, output, "2222 3333 .+ \\.")
-			h.AssertMatch(t, output, "2222 3333 .+ group.toml")
-		})
-	})
-
-	when("group path is provided", func() {
-		it("uses the provided group path", func() {
-			cacheVolume = h.SeedDockerVolume(t, cacheFixtureDir)
-
-			h.DockerRunAndCopy(t,
-				containerName,
-				copyDir,
-				analyzeImage,
-				"/layers",
-				h.WithFlags(
-					"--env", "CNB_REGISTRY_AUTH={}",
-					"--volume", cacheVolume+":"+"/cache", // use a cache so that we can observe the effect of other-group.toml on /layers
-				),
-				h.WithArgs(
-					analyzerPath,
-					"-cache-dir", "/cache",
-					"-group", "/layers/other-group.toml",
-					"some-image",
-				),
-			)
-
-			h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-other-buildpack-id"))
-			h.AssertPathDoesNotExist(t, filepath.Join(copyDir, "layers", "some-buildpack-id"))
-		})
-	})
-
-	when("layers path and analyzed path are provided", func() {
-		it("writes analyzed.toml at the provided path", func() {
-			h.DockerRunAndCopy(t,
-				containerName,
-				copyDir,
-				analyzeImage,
-				"/other-layers/other-analyzed.toml",
-				h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
-				h.WithArgs(
-					analyzerPath,
-					"-layers", "/other-layers",
-					"-analyzed", "/other-layers/other-analyzed.toml",
-					"some-image",
-				),
-			)
-
-			assertAnalyzedMetadata(t, filepath.Join(copyDir, "other-analyzed.toml"))
-		})
-	})
-
-	when("daemon case", func() {
-		it("writes analyzed.toml", func() {
-			h.DockerRunAndCopy(t,
-				containerName,
-				copyDir,
-				analyzeImage,
-				"/layers/analyzed.toml",
-				h.WithFlags(variables.DockerSocketMount...),
-				h.WithArgs(analyzerPath, "-daemon", "some-image"),
-			)
-
-			assertAnalyzedMetadata(t, filepath.Join(copyDir, "analyzed.toml"))
-		})
-
-		when("app image is found", func() {
-			var appImage string
-
-			it.Before(func() {
-				appImage = "some-app-image-" + h.RandString(10)
-				metadata := minifyMetadata(t, filepath.Join("testdata", "analyzer", "app_image_metadata.json"), lifecycle.LayersMetadata{})
-
+	when("Platform API < 0.5", func() {
+		when("called without an app image", func() {
+			it("errors", func() {
 				cmd := exec.Command(
-					"docker",
-					"build",
-					"-t", appImage,
-					"--build-arg", "fromImage="+variables.ContainerBaseImage,
-					"--build-arg", "metadata="+metadata,
-					filepath.Join("testdata", "analyzer", "app-image"),
+					"docker", "run", "--rm",
+					"--env", "CNB_REGISTRY_AUTH={}",
+					analyzeImage,
+					analyzerPath,
 				)
-				h.Run(t, cmd)
-			})
+				output, err := cmd.CombinedOutput()
 
-			it.After(func() {
-				h.DockerImageRemove(t, appImage)
+				h.AssertNotNil(t, err)
+				expected := "failed to parse arguments: received 0 arguments, but expected 1"
+				h.AssertStringContains(t, string(output), expected)
 			})
+		})
 
-			it("restores layer metadata", func() {
+		when("cache image tag and cache directory are both blank", func() {
+			it("warns", func() {
+				output := h.DockerRun(t,
+					analyzeImage,
+					h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
+					h.WithArgs(analyzerPath, "some-image"),
+				)
+
+				expected := "Not restoring cached layer metadata, no cache flag specified."
+				h.AssertStringContains(t, output, expected)
+			})
+		})
+
+		when("the provided layers directory isn't writeable", func() {
+			it("recursively chowns the directory", func() {
+				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
+
+				output := h.DockerRun(t,
+					analyzeImage,
+					h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
+					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s some-image; ls -al /layers", analyzerPath)),
+				)
+
+				h.AssertMatch(t, output, "2222 3333 .+ \\.")
+				h.AssertMatch(t, output, "2222 3333 .+ group.toml")
+			})
+		})
+
+		when("group path is provided", func() {
+			it("uses the provided group path", func() {
+				cacheVolume = h.SeedDockerVolume(t, cacheFixtureDir)
+
 				h.DockerRunAndCopy(t,
 					containerName,
 					copyDir,
 					analyzeImage,
 					"/layers",
-					h.WithFlags(variables.DockerSocketMount...),
-					h.WithArgs(analyzerPath, "-daemon", appImage),
+					h.WithFlags(
+						"--env", "CNB_REGISTRY_AUTH={}",
+						"--volume", cacheVolume+":"+"/cache", // use a cache so that we can observe the effect of other-group.toml on /layers
+					),
+					h.WithArgs(
+						analyzerPath,
+						"-cache-dir", "/cache",
+						"-group", "/layers/other-group.toml",
+						"some-image",
+					),
 				)
 
-				layerFilenames := []string{
-					"launch-build-cache-layer.sha",
-					"launch-build-cache-layer.toml",
-					"launch-cache-layer.sha",
-					"launch-cache-layer.toml",
-					"launch-layer.sha",
-					"launch-layer.toml",
-					"store.toml",
-				}
-				for _, filename := range layerFilenames {
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", filename))
-				}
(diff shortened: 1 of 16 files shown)


cmd/command.go
```diff
@@ -6,8 +6,13 @@ import (
 	"os"
 )
 
+// PreInit defines all the flags that are going to be used.
+// If the default value is not going to be used,
+// the flags will be set as part of the flags.Parse function.
+// In Args, several paths will be changed to be under the updated layers directory (if the user didn't set them using a flag)
+// TODO: should we add more documentation?
```
yaelharel, line 13: Would love to know your opinion on that.

cmd/command.go
```diff
@@ -6,8 +6,13 @@ import (
 	"os"
 )
 
+// PreInit defines all the flags that are going to be used.
+// If the default value is not going to be used,
+// the flags will be set as part of the flags.Parse function.
+// In Args, several paths will be changed to be under the updated layers directory (if the user didn't set them using a flag)
```
yaelharel, line 12: Do you think the update should be part of the Args function, or it's better to add a new function to the interface and make the update over there?
And if we should keep the update as part of the Args function, maybe its name should be changed?

cmd/flags.go
```diff
@@ -163,6 +180,10 @@ func FlagProjectMetadataPath(projectMetadataPath *string) {
 	flagSet.StringVar(projectMetadataPath, "project-metadata", EnvOrDefault(EnvProjectMetadataPath, DefaultProjectMetadataPath), "path to project-metadata.toml")
 }
 
+func UpdateProjectMetadataPath(projectMetadataPath *string, platformAPI, layersDir string) {
```
ekcasey, line 183: Instead of having a function that updates the value referenced by the pointer it might be more straightforward/readable to have a `DefaultProjectMetadataPath(platformAPI, layersDir string) string` function that simply returns the computed default. Then in places where we are currently doing things like 
```
cmd.UpdatePlanPath(&b.planPath, b.buildArgs.platformAPI, b.layersDir)
```
we could instead do something like
```
if b.planPath == cmd.PlaceholderPlanPath {
   b.planPath = cmd.DefaultPlanPath(b.platformAPI, b.layersDir)
}
```
In my opinion, the above usage makes it very clear to someone reading a given implementation of `Args` what exactly is happening and why, without a lot of digging.

cmd/flags.go
```diff
@@ -206,3 +227,20 @@ func EnvOrDefault(key string, defaultVal string) string {
...
+		// layersDir is unset when this call comes from the rebaser - will be fixed as part of https://github.com/buildpacks/spec/issues/156
+		*pathToUpdate = filepath.Join(".", fileName)
+	} else {
+		*pathToUpdate = filepath.Join(layersDir, fileName)
+	}
+}
+
+func isPlatformAPILessThan05(platformAPI string) bool {
+	return api.MustParse(platformAPI).Compare(api.MustParse("0.5")) < 0
+}
```
ekcasey, line 246: Given that `api.MustParse(platformAPI).Compare(api.MustParse("0.5")) < 0` is already a one-liner I don't see the value of this function unless we rename it to something like `shouldUseWorkingDirRelativeDefaults`.

cmd/lifecycle/analyzer.go
```diff
@@ -60,6 +61,9 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
 	if a.cacheImageTag == "" && a.cacheDir == "" {
 		cmd.DefaultLogger.Warn("Not restoring cached layer metadata, no cache flag specified.")
 	}
+	cmd.UpdateAnalyzedPath(&a.analyzedPath, a.analyzeArgs.platformAPI, a.layersDir)
+	cmd.UpdateGroupPath(&a.groupPath, a.analyzeArgs.platformAPI, a.layersDir)
```
ekcasey, line 65:
```suggestion
	cmd.UpdateAnalyzedPath(&a.analyzedPath, a.platformAPI, a.layersDir)
	cmd.UpdateGroupPath(&a.groupPath, a.platformAPI, a.layersDir)
```
Since `analyzeArgs` is embedded we can reference `a.platformAPI` just like we reference `a.layersDir`.

cmd/lifecycle/builder.go
```diff
@@ -43,6 +43,10 @@ func (b *buildCmd) Args(nargs int, args []string) error {
 	if nargs != 0 {
 		return cmd.FailErrCode(errors.New("received unexpected arguments"), cmd.CodeInvalidArgs, "parse arguments")
 	}
+
+	cmd.UpdateGroupPath(&b.groupPath, b.buildArgs.platformAPI, b.layersDir)
+	cmd.UpdatePlanPath(&b.planPath, b.buildArgs.platformAPI, b.layersDir)
```
ekcasey, line 48:
```suggestion
	cmd.UpdateGroupPath(&b.groupPath, b.platformAPI, b.layersDir)
	cmd.UpdatePlanPath(&b.planPath, b.platformAPI, b.layersDir)
```

cmd/lifecycle/detector.go
```diff
@@ -41,6 +44,10 @@ func (d *detectCmd) Args(nargs int, args []string) error {
 	if nargs != 0 {
 		return cmd.FailErrCode(errors.New("received unexpected arguments"), cmd.CodeInvalidArgs, "parse arguments")
 	}
+
+	cmd.UpdateGroupPath(&d.groupPath, d.detectArgs.platformAPI, d.layersDir)
+	cmd.UpdatePlanPath(&d.planPath, d.detectArgs.platformAPI, d.layersDir)
```
ekcasey, line 49:
```suggestion
	cmd.UpdateGroupPath(&d.groupPath, d.platformAPI, d.layersDir)
	cmd.UpdatePlanPath(&d.planPath, d.platformAPI, d.layersDir)
```

cmd/lifecycle/exporter.go
```diff
@@ -99,6 +99,11 @@ func (e *exportCmd) Args(nargs int, args []string) error {
 		return cmd.FailErrCode(errors.New("supply only one of -run-image or (deprecated) -image"), cmd.CodeInvalidArgs, "parse arguments")
 	}
 
+	cmd.UpdateAnalyzedPath(&e.analyzedPath, e.exportArgs.platformAPI, e.layersDir)
+	cmd.UpdateGroupPath(&e.groupPath, e.exportArgs.platformAPI, e.layersDir)
+	cmd.UpdateProjectMetadataPath(&e.projectMetadataPath, e.exportArgs.platformAPI, e.layersDir)
+	cmd.UpdateReportPath(&e.projectMetadataPath, e.exportArgs.platformAPI, e.layersDir)
```
ekcasey, line 105:
```suggestion
	cmd.UpdateAnalyzedPath(&e.analyzedPath, e.platformAPI, e.layersDir)
	cmd.UpdateGroupPath(&e.groupPath, e.platformAPI, e.layersDir)
	cmd.UpdateProjectMetadataPath(&e.projectMetadataPath, e.platformAPI, e.layersDir)
	cmd.UpdateReportPath(&e.projectMetadataPath, e.platformAPI, e.layersDir)
```

ekcasey (changes requested):
Although we don't think we need to acceptance test every affected flag I think having at least one acceptance test that uses a default relative to layers dir would be good. 

This analyzer test seems like a good candidate https://github.com/buildpacks/lifecycle/blob/main/acceptance/analyzer_test.go#L162-L178

It should pass now even if you remove the `-analyzed` flag. If we alter that test we should add another test that sets `-analyzed` but not `-layers`, to ensure both cases work.

cmd/command.go
```diff
@@ -6,8 +6,13 @@ import (
...
+// TODO: should we add more documentation?
 type Command interface {
-	Init()
+	PreInit()
```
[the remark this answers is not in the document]
yaelharel: I agree that `PreInit` isn't the best name since you expect to have an `Init` step after it. But I think that `Init` is also pretty confusing because the flags aren't being initialized as part of this function. If I'm not mistaken, in this function we only define the flags that we're going to use, so maybe `SetNeededFlags` is a better name?

cmd/flags.go
yaelharel: You're right. When I created this function, I called it multiple times around the code but after some refactoring I did (before making the PR), it's being called only once.
I'll delete it.
Thanks!

cmd/command.go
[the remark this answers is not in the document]
ekcasey: Haha I feel like my brain interprets this the opposite way. The command, including the flagset, is being initialized but the values aren't actually _set_ until later. In the original PR it just called `Flags` but we changed it on what I thought was a solid suggestion in the code review https://github.com/buildpacks/lifecycle/pull/232#discussion_r373076090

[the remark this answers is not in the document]
yaelharel: If the `Parse` function was being called inside it, I think it would be very clear to call it `Init`.
What about moving [these lines](https://github.com/buildpacks/lifecycle/blob/b2ce634c61ee84c57d1980c2d80b732311d200c4/cmd/command.go#L28-L38) to be part of the `Init` function?
Then in the `Run` function, we will have:
```
if err := c.Init(asSubcommand); err != nil {
    Exit(err)
}
```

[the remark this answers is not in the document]
yaelharel: After talking with @ekcasey, we agreed on just changing the `Init` function name to `DefineFlags`.

acceptance/analyzer_test.go
```diff
@@ -507,131 +480,212 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
 
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.sha"))
-					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-buildpack-id", "some-layer.toml"))
-				})
+				assertAnalyzedMetadata(t, filepath.Join(copyDir, "some-other-layers", "analyzed.toml"))
+				h.AssertPathExists(t, filepath.Join(copyDir, "some-other-layers", "another-buildpack-id"))
 			})
 		})
+
+		// TODO: fix!!!
```
yaelharel, line 667: Currently I'm getting the following error:
```
ERROR: write analyzed.toml: open /other-layers/other-analyzed.toml: permission denied
```

yaelharel:
@ekcasey, I pushed a commit with some fixes following your feedback.
I added two new tests to `analyzer_test.go` and one of them is failing due to
```
ERROR: write analyzed.toml: open /other-layers/other-analyzed.toml: permission denied
```
Please ignore it for now and I'll push a fix as soon as possible.

acceptance/testdata/analyzer/analyze-image/Dockerfile
```diff
@@ -9,3 +9,7 @@ WORKDIR /layers
 ENV CNB_USER_ID=2222
 
 ENV CNB_GROUP_ID=3333
+
+RUN chmod -R 777 /some-dir
+
+RUN chmod -R 777 /layers
```
yaelharel, line 15: This line is in order to make the test on line 647 work.
I'm not sure if it's ok to add it, although we're actually doing something similar as part of the `Privileges` function of the `analyzer`.

acceptance/analyzer_test.go
```diff
@@ -158,23 +158,31 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+
+			cacheVolume = h.SeedDockerVolume(t, cacheFixtureDir)
+
 			h.DockerRunAndCopy(t,
 				containerName,
 				copyDir,
 				analyzeImage,
-				"/other-layers/other-analyzed.toml",
-				h.WithFlags("--env", "CNB_REGISTRY_AUTH={}"),
+				"/",
```
ekcasey, line 171: Copying the entire filesystem out seems like overkill potentially. Seems like we potentially need to refactor the helpers here to accept multiple files to copy out.

acceptance/analyzer_test.go
```diff
@@ -158,23 +158,31 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 
-	when("layers path and analyzed path are provided", func() {
-		it("writes analyzed.toml at the provided path", func() {
+	when("analyzed path is provided", func() {
+		it("writes analyzed.toml at the provided path and uses group.toml from the working directory", func() {
```
ekcasey, line 162: I found the inclusion of "and uses group.toml from the working directory" given that we aren't in one of the "Platform API ..." contexts. Do we even need to mention `group.toml` here, whether we use the given `group.toml` is from the `layers` or working directories given that they are the same in this case?

ekcasey (changes requested):

acceptance/analyzer_test.go
yaelharel: Actually you're right. It's a leftover from the test I changed but we don't really need to test `group.toml` in this scenario.
I'll change it.

acceptance/analyzer_test.go
yaelharel: Following https://github.com/buildpacks/lifecycle/pull/454#discussion_r520737124, we don't need to change `DockerRunAndCopy`.

acceptance/testdata/analyzer/analyze-image/Dockerfile
yaelharel: I agree that it's not a great idea to run `chmod -R` on both directories.
I'm not sure that I fully understand what you wrote. In the commit I'll push in 2 minutes, I've replaced `RUN chmod -R` in the dockerfile to `RUN chown -R $CNB_USER_ID:$CNB_GROUP_ID` and everything works.
Do you want me to make any other changes?

cmd/flags.go
ekcasey: @yaelharel thoughts on this suggestion? I still think the suggested change would make the commands more readable.

yaelharel: I'm really sorry, I missed this comment. I'll review it today.

acceptance/testdata/analyzer/analyze-image/Dockerfile
ekcasey: I was suggesting that we change the default user from `root` and only run explicitly as `root` when testing behavior that requires it (e.g. connecting to the daemon socket, chowning the layers dir). But as you point out that isn't necessary to test this feature and I think we can make that change at a later point if we desire.

cmd/flags.go
yaelharel: Your idea is great! I'll go and make this change.
Sorry again for missing your comment for so long...

cmd/flags.go
```diff
@@ -206,3 +229,14 @@ func EnvOrDefault(key string, defaultVal string) string {
...
+
+func defaultPath(placeholderPath, platformAPI, layersDir string) string {
+	fileName := filepath.Base(placeholderPath)
+	path := filepath.Join(layersDir, fileName) // starting from platform api 0.5, the default directory is the layers dir.
+	if (api.MustParse(platformAPI).Compare(api.MustParse("0.5")) < 0) || (layersDir == "") {
+		// prior to platform api 0.5, the default directory was the working dir.
+		// layersDir is unset when this call comes from the rebaser - will be fixed as part of https://github.com/buildpacks/spec/issues/156
+		path = filepath.Join(".", fileName)
+	}
+	return path
```
yaelharel, line 241: When possible, I prefer to have one `return` statement from a function (or have:
```
if (api.MustParse(platformAPI).Compare(api.MustParse("0.5")) < 0) || (layersDir == "") {
  return filepath.Join(".", fileName)
} else {
  return filepath.Join(layersDir, fileName)
}
```
but the linter doesn't allow this).

But if the convention is:
```
if (api.MustParse(platformAPI).Compare(api.MustParse("0.5")) < 0) || (layersDir == "") {
  return filepath.Join(".", fileName)
}
return filepath.Join(layersDir, fileName)
```
I can make this change.

ekcasey: The latter is the typical convention.

cmd/flags.go
```diff
@@ -206,3 +229,14 @@ func EnvOrDefault(key string, defaultVal string) string {
 	}
 	return defaultVal
 }
+
+func defaultPath(placeholderPath, platformAPI, layersDir string) string {
+	fileName := filepath.Base(placeholderPath)
```
ekcasey, line 234: It might be simpler just to make more constants along the lines of `DefaultGroupFile` rather than splitting it out from the Placeholder. You could even use those constants to create the placeholders
```
PlaceholderGroupPath           = filepath.Join("<layers>", DefaultGroupFile)
```

ekcasey (approved):

cmd/flags.go
yaelharel: Haha, no problem. I'll change it.

cmd/flags.go
yaelharel: I don't have strong opinion about any of these approaches so I'll make this change.
Thanks!

Issue #455, opened 2020-11-02
Slices No Longer Handle Trailing Slashes Properly


nebhale:
### Summary
Previously, the Spring Boot Layer Index was properly interpreted and created layers with metadata that had trailing slashes in the paths.  The [recent change](https://github.com/buildpacks/lifecycle/commit/23f1c9745a042cd4c9434512a157400514fdc20e) to stop using `filepath.Glob` has broken this (new matching algorithm).  Credit to @lzimd for [diagnosing](https://github.com/paketo-buildpacks/spring-boot/issues/1#issuecomment-719908485) this issue.

---

### Reproduction
Reproduction can be accomplished with @jonashackt's [repository](https://github.com/jonashackt/spring-boot-buildpack).

#### Steps
1. `git clone https://github.com/jonashackt/spring-boot-buildpack`
2. `mvn spring-boot:build-image`
3. `dive docker.io/library/spring-boot-buildpack:0.0.1-SNAPSHOT`

nebhale:
Found in https://github.com/paketo-buildpacks/spring-boot/issues/1

ekcasey:
Resolved by #456

PR #456, opened 2020-11-02
Slices: Clean pattern before matching

natalieparellano:
I took a stab at fixing #455 but I have some questions.

`filepath.Clean` seemed like the easiest fix, but it would also have the effect of making other patterns match - e.g., `some-dir//some-other-dir` would match `some-dir/some-other-dir` and I am not sure this is what we want.

Also it seems important to validate this change against the linked repro (spring-boot-buildpack) but having limited knowledge of Spring Boot, I am not sure how to do that.

### DIFF
--- a/layers/slices.go
+++ b/layers/slices.go
@@ -65,6 +65,7 @@ func (f *Factory) createLayerFromSlice(slice Slice, sdir *sliceableDir, layerID 
 }
 
 func glob(sdir *sliceableDir, pattern string) ([]string, error) {
+	pattern = filepath.Clean(pattern)
 	var matches []string
 	if err := filepath.Walk(sdir.path, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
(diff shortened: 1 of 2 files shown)


layers/slices.go
```diff
@@ -65,6 +65,7 @@ func (f *Factory) createLayerFromSlice(slice Slice, sdir *sliceableDir, layerID
 }
 
 func glob(sdir *sliceableDir, pattern string) ([]string, error) {
+	pattern = filepath.Clean(pattern)
```
natalieparellano, line 68: Maybe it would be better to clean the slices when `metadata.toml` is read (or written). But this seemed like the smallest & safest change.

ekcasey: thanks!

ekcasey (approved):

ekcasey:
I think we will want to ship a `v0.9.3` with this fix so I am going to create a new release branch and repoint this PR

ekcasey:
@natalieparellano would you mind updating this so it merges onto `release/0.9.3` (branched from `v0.9.2`) without bringing partial `v0.10.0` changes?

natalieparellano:
@ekcasey this should be good to merge into the release branch

PR #461, opened 2020-11-10
Decouple buildpack plan and bom

### LINKED ISSUE #396: [RFC 0053] Decouple Buildpack Plan and BOM
[RFC 0053](https://github.com/buildpacks/rfcs/blob/main/text/0053-decouple-buildpack-plan-and-bom.md)
https://github.com/buildpacks/rfcs/pull/100
https://github.com/buildpacks/spec/issues/133
https://github.com/buildpacks/spec/issues/134

Buildpacks using API >= 0.5 should contribute launch and build BOM entries to `launch.toml` and `build.toml` respectively. Entries to be passed to subsequent buildpacks should be added to an `entries` section in `build.toml`.

Entries in `build.toml` should be added to a `bom` in `report.toml`.

natalieparellano:
Fixes #396


builder.go
```diff
@@ -181,6 +206,15 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 	}, nil
 }
 
+func withBuildpack(bp Buildpack, bom []BOMEntry) []BOMEntry {
```
natalieparellano, line 209: Maybe this should be a function on `[]BOMEntry` so that it reads `bom.withBuildpack(bp)`

builder_test.go
```diff
@@ -88,44 +88,6 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
-					)
-					buildMetadata, err := builder.Build()
-					if err != nil {
-						t.Fatalf("Unexpected error:\n%s\n", err)
-					}
-					h.AssertEq(t, buildMetadata.BOM[0].Version, "")
-					h.AssertEq(t, buildMetadata.BOM[0].Metadata["version"], "v1")
-				})
-			})
-
```
natalieparellano, line 128: Moved these tests to a different context

builder_test.go
```diff
@@ -167,60 +129,6 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
-					Processes: []launch.Process{},
-					Buildpacks: []lifecycle.Buildpack{
-						{ID: "A", Version: "v1", API: "0.3", Homepage: "Buildpack A Homepage"},
-						{ID: "B", Version: "v2", API: "0.2"},
-					},
-				}); s != "" {
-					t.Fatalf("Unexpected:\n%s\n", s)
-				}
-			})
-
```
natalieparellano, line 223: Moved these tests to a different context

builder_test.go
```diff
@@ -282,163 +190,426 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								`name = "dep1"`+"\n"+
+								"[entries.metadata]\n"+
+								`version = "v1"`+"\n",
+							filepath.Join(appDir, "build-plan-out-A-v1.toml"),
+						)
+						metadata, err := builder.Build()
+						if err != nil {
+							t.Fatalf("Unexpected error:\n%s\n", err)
+						}
+						if s := cmp.Diff(metadata.BOM, []lifecycle.BOMEntry{
```
natalieparellano, line 474: Only asserting on the `BOM` here, whereas previously we also looked at `Processes` and `Buildpacks`. I think this makes it clearer but maybe the other way is safer.

builder_test.go
```diff
@@ -282,163 +190,426 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
-							Buildpack: lifecycle.Buildpack{ID: "A", Version: "v1"},
-						},
-						{
-							Require:   lifecycle.Require{Name: "dep1-replace", Version: "v7"},
-							Buildpack: lifecycle.Buildpack{ID: "A", Version: "v1"},
-						},
-						{
-							Require:   lifecycle.Require{Name: "dep1-next", Version: "v9"},
-							Buildpack: lifecycle.Buildpack{ID: "B", Version: "v2"},
+			when("build metadata", func() {
```
natalieparellano, line 400: "build metadata" seems like an overloaded term, maybe build.toml would be better

exporter_test.go
```diff
@@ -1114,6 +1114,37 @@ version = "4.5.6"
 				h.AssertNil(t, err)
 				h.AssertContains(t, fakeAppImage.SavedNames(), append(opts.AdditionalNames, fakeAppImage.Name())...)
 			})
+
+			when("there is build metadata", func() {
```
natalieparellano, line 1118: "build metadata" seems like an overloaded term, perhaps "build.toml" would be better

natalieparellano:
After re-reading the spec PR I think this logic still needs to be added: https://github.com/buildpacks/spec/pull/139/files#diff-d5ffdcac80bcacec34b3a8a1ba9da8efbbc4df5ad118ec581201736330f73183R825-R833 i.e., name is required and must match an entry in the buildpack plan.

builder_test.go
```diff
@@ -633,6 +802,52 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 					})
 				})
 			})
+
+			when("invalid unmet entries", func() {
+				it.Before(func() {
```
natalieparellano, line 807: Should this also have a `when("platform API >= 0.5", ...`?

ekcasey: Probably, just to be consistent. In fact we should probably change the implementation so that we actually aren't validating `build.toml` when it is written by an older buildpack.

exporter.go
```diff
@@ -422,3 +431,21 @@ func (e *Exporter) addOrReuseLayer(image imgutil.Image, layer layers.Layer, prev
...
+func (e *Exporter) makeBuildReport(layersDir string) (BuildReport, error) {
+	var out []BOMEntry
+	for _, bp := range e.Buildpacks {
+		var bpBuildReport BuildReport
+		bpBuildTOML := filepath.Join(layersDir, bp.ID, "build.toml")
+		if _, err := toml.DecodeFile(bpBuildTOML, &bpBuildReport); os.IsNotExist(err) {
+			// nop
+		} else if err != nil {
+			return BuildReport{}, err
+		}
```
ekcasey, line 444:
```suggestion
		if _, err := toml.DecodeFile(bpBuildTOML, &bpBuildReport); err != nil && !os.IsNotExist(err) {
			return BuildReport{}, err
		}
```

exporter.go
```diff
@@ -422,3 +431,21 @@ func (e *Exporter) addOrReuseLayer(image imgutil.Image, layer layers.Layer, prev
 	e.Logger.Debugf("Layer '%s' SHA: %s\n", layer.ID, layer.Digest)
 	return layer.Digest, image.AddLayerWithDiffID(layer.TarPath, layer.Digest)
 }
+
+func (e *Exporter) makeBuildReport(layersDir string) (BuildReport, error) {
+	var out []BOMEntry
+	for _, bp := range e.Buildpacks {
+		var bpBuildReport BuildReport
```
ekcasey, line 438: We should check the buildpack API and continue to the next buildpack if it is < 0.5. There is an extreme edge case here where an older buildpack may have had a layer with id `build` which is no longer valid. We don't want to try to decode that `layer.toml` file (with name `build.toml`) as a build report. Although it probably wouldn't actually error (?)... its weird and it's safer just to be strict about it.

exporter_test.go
ekcasey: agree. unlike other files with configurable names, `build.toml` is always called `build.toml` so we might as well stick with that instead of wasting precious terms here that we might want to (or already) use for other concepts.

exporter.go
```diff
@@ -422,3 +431,21 @@ func (e *Exporter) addOrReuseLayer(image imgutil.Image, layer layers.Layer, prev
...
+		bpBuildTOML := filepath.Join(layersDir, bp.ID, "build.toml")
+		if _, err := toml.DecodeFile(bpBuildTOML, &bpBuildReport); os.IsNotExist(err) {
+			// nop
+		} else if err != nil {
+			return BuildReport{}, err
+		}
+		for _, entry := range bpBuildReport.BOM {
+			entry.Buildpack = bp.noAPI().noHomepage()
```
ekcasey, line 446: should we use `withBuildpack`? https://github.com/buildpacks/lifecycle/pull/461/files#diff-5dce70063c4b1b07e611b3fcd32da0445a0b95970028c8992beb9c955e38320fR361

ekcasey (changes requested):

builder_test.go
natalieparellano: Sounds good. I should have said `buildpack API` above but same idea :)

builder_test.go
```diff
@@ -282,163 +190,424 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								{
+									Require:   lifecycle.Require{Name: "dep1-next", Version: "v9"},
+									Buildpack: lifecycle.Buildpack{ID: "B", Version: "v2"},
+								},
+								{
+									Require:   lifecycle.Require{Name: "dep2", Version: "v4"},
+									Buildpack: lifecycle.Buildpack{ID: "B", Version: "v2"},
+								},
+								{
+									Require:   lifecycle.Require{Name: "dep2-replace", Version: "v8"},
```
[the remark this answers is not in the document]
natalieparellano: Sounds good. I took a shot at it.

builder.go
```diff
@@ -82,83 +100,39 @@ func (b *Builder) Build() (*BuildMetadata, error) {
...
-		cmd.Stderr = b.Err
-
-		if bpInfo.Buildpack.ClearEnv {
-			cmd.Env = b.Env.List()
-		} else {
-			cmd.Env, err = b.Env.WithPlatform(platformDir)
-			if err != nil {
-				return nil, err
-			}
+		if err := b.runBuildCmd(bp, appDir, bpLayersDir, platformDir, bpPlanPath); err != nil {
```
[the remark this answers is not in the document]
natalieparellano: Sounds good... while we're at it, what's the system that we want to be using to organize code within a file? Sometimes we have the structs at the top, sometimes structs grouped with methods, etc.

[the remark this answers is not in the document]
natalieparellano: On second thought... what would be the advantage of making `BuildpackTOML` the receiver given that we already have `[]Buildpack` in `Builder.Group.Group`?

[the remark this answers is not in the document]
natalieparellano: After experimenting with this approach I think I like how it reads... probably it makes sense to unit-test `BuildpackTOML.Build()` as part of this PR, will add in the next round.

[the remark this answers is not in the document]
natalieparellano: There are some differences in how `BuildpackTOML.Detect()` behaves vs `BuildpackTOML.Build()` which may or may not be significant - notably the detect config takes `FullEnv` and `ClearEnv` whereas build config has `Env` (this is because of `setupEnv(config.Env, bpLayersDir)`). Also detect passes the output back to the caller whereas build takes a writer.

natalieparellano:
@ekcasey there is still work to be done for the tests (unit testing `BuildpackTOML.Build()`) but if it's possible to provide feedback on the direction, that would be awesome.

builder_test.go
ekcasey: I support the decision to prioritize clarity.

builder.go
```diff
@@ -181,73 +128,185 @@ func (b *Builder) Build() (*BuildMetadata, error) {
...
+	if _, err := toml.DecodeFile(launchPath, &launch); err != nil && !os.IsNotExist(err) {
+		return BuildResult{}, err
+	}
+	// setup result
+	var br BuildResult
+	if api.MustParse(bp.API).Compare(api.MustParse("0.5")) < 0 { // bp.API <= 0.4
+		br.BOM = withBuildpack(bp, bpPlanOut.toBOM())
+		br.Met = bpPlanOut.Entries
+	} else {
+		br.BOM = withBuildpack(bp, launch.BOM)
```
ekcasey, line 269: We should be validating that the top-level version is not set in the launch.BOM entries. We don't want to carry-over the deprecated usage to this new file.

exporter.go
```diff
@@ -422,3 +431,22 @@ func (e *Exporter) addOrReuseLayer(image imgutil.Image, layer layers.Layer, prev
...
+	if e.PlatformAPI.Compare(api.MustParse("0.5")) < 0 { // platform API < 0.5
+		return BuildReport{}, nil
+	}
+	var out []BOMEntry
+	for _, bp := range e.Buildpacks {
+		if api.MustParse(bp.API).Compare(api.MustParse("0.5")) < 0 { // buildpack API < 0.5
+			continue
+		}
+		var bpBuildReport BuildReport
+		bpBuildTOML := filepath.Join(layersDir, bp.ID, "build.toml")
```
ekcasey, line 445:
```suggestion
		bpBuildTOML := filepath.Join(layersDir,  launch.EscapeID(bp.ID), "build.toml")
```
we can't use `bp.ID` directly here, we need to use the escaped version instead.

exporter.go
```diff
@@ -422,3 +431,22 @@ func (e *Exporter) addOrReuseLayer(image imgutil.Image, layer layers.Layer, prev
...
+		if _, err := toml.DecodeFile(bpBuildTOML, &bpBuildReport); err != nil && !os.IsNotExist(err) {
+			return BuildReport{}, err
+		}
+		out = append(out, withBuildpack(bp, bpBuildReport.BOM)...)
```
ekcasey, line 449: I think we are accidentally allowing top-level versions here as well. I think we should either be validating that they aren't set or using a different struct types in different situations to avoid these gotchas.

builder.go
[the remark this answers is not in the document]
ekcasey: I think the work you have done so far is a major improvement. We work on tackling all of the inconsistencies over time but this feels like a big step in the right direction.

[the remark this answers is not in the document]
ekcasey: > Sounds good... while we're at it, what's the system that we want to be using to organize code within a file? Sometimes we have the structs at the top, sometimes structs grouped with methods, etc.

One of my default opinions is that the definition should be directly above any function with that type as a receiver (if the definition and method are in the same file).
```
type Foo struct {} // struct with methods
func (*Foo) Method1() {}
func (*Foo) Method2() {}
type Bar struct {} // struct with methods
func (*Bar) Method1() {}
```
stdlib example: https://golang.org/src/net/iprawsock.go#L99

Not sure there is one right answer to this question. If there is one right answer, I don't know it. Whatever feels organized and easy to navigate? In the future I think a new `buildpack` package might help us organize some of our existing types. In this top-level package a lot of things end up in `build.go` that we would probably organize in a larger number of smaller files if they lived in a separate package.

exporter.go
natalieparellano: If a buildpack tries to write a top-level version to the BOM in either build.toml or launch.toml, is that a failure case? That would require a "validator" in order to detect (I believe).

At first I was leaning toward separate structs, but that would require some sort of duplication of both `Require` and `BOMEntry` (and converting of structs during `Build()`) which started to feel more confusing than just adding the logic at the time of reading the files.

ekcasey: I lean towards erroring even though we don't do that for unrecognized keys just to ensure everyone migrate the version to the metadata section in this new API. At the very least a top-level version shouldn't end up in `report.toml`.

builder.go
```diff
@@ -82,83 +88,24 @@ func (b *Builder) Build() (*BuildMetadata, error) {
...
+		bom = append(bom, br.BOM...)
+		labels = append(labels, br.Labels...)
+		plan = plan.filter(br.Met)
+		procMap.add(br.Processes)
+		slices = append(slices, br.Slices...)
 	}
 
-	if b.PlatformAPI.Compare(api.MustParse("0.4")) < 0 {
-		//plaformApiVersion is less than comparisonVersion
+	if b.PlatformAPI.Compare(api.MustParse("0.4")) < 0 { // PlatformAPI <= 0.3
```
natalieparellano, line 108: As I'm reading this, it seems like top level version must match metadata version, regardless of the platform API. Maybe it makes sense to validate this up front, so that the `convert` methods don't have to return an error.

builder.go
```diff
@@ -5,75 +5,100 @@ import (
...
-	LayersDir     string
-	PlatformDir   string
-	BuildpacksDir string
-	PlatformAPI   *api.Version
-	Env           BuildEnv
-	Group         BuildpackGroup
-	Plan          BuildPlan
-	Out, Err      io.Writer
-}
-
```
natalieparellano, line 31: Re-ordered some things in this file to bring struct definitions closer to their methods.

builder_test.go
```diff
@@ -2,6 +2,7 @@ package lifecycle_test
 
 import (
 	"bytes"
+	"errors"
```
natalieparellano, line 5: I think I accidentally changed this from `github.com/pkg/errors` - I can change it back.

buildpacktoml_test.go
```diff
@@ -0,0 +1,729 @@
...
+							},
+						},
+					}
+
+					mkfile(t,
+						"[[entries]]\n"+
+							`name = "some-deprecated-bp-dep"`+"\n"+
+							`version = "v1"`+"\n"+
+							"[[entries]]\n"+
+							`name = "some-deprecated-bp-replace-version-dep"`+"\n"+
```
natalieparellano, line 610: This is onerous... even as I was writing this code I felt that I was going too far. It may be enough to assert something like "BOM is an exact reflection of output buildpack plan" or something like that.

buildpacktoml_test.go
```diff
@@ -0,0 +1,729 @@
...
+						"[[entries]]\n"+
+							`name = "dep1"`+"\n"+
+							`version = "v2"`+"\n"+
+							"[entries.metadata]\n"+
+							`version = "v1"`+"\n",
+						filepath.Join(appDir, "build-plan-out-A-v1.toml"),
+					)
+					_, err := bpTOML.Build(lifecycle.BuildpackPlan{}, config)
+					h.AssertNotNil(t, err)
+					expected := "top level version does not match metadata version"
```
natalieparellano, line 723: Note that this conflates two previous errors, whereas before we also had `metadata version does not match top level version`. But I thought that this was worth it to keep platform API logic all in builder.go and do the BOM validation here.

tools/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/docker/docker v1.4.2-0.20190924003213-a8608b5b67c7
 	github.com/golang/mock v1.4.4
 	github.com/golangci/golangci-lint v1.30.0
-	github.com/google/go-containerregistry v0.2.0
+	github.com/google/go-containerregistry v0.2.1
```
natalieparellano, line 12: I'm not sure why this changed...

tools/go.sum
```diff
@@ -301,9 +301,8 @@ github.com/google/go-cmp v0.5.2 h1:X2ev0eStA3AbceY54o37/0PQ/UWqKEiiO2dKL5OPaFM=
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.4 h1:L8R9j+yAqZuZjsqh/z+F1NCffTKKLShY6zXTItVIZ8M=
 github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
-github.com/google/go-containerregistry v0.1.4 h1:fZm+V2pYnvb8NMPM1YOsyxr31XKfpHTun5oVTRnG8qc=
-github.com/google/go-containerregistry v0.1.4/go.mod h1:6EGiuQp36pL82lX6rFN0s9AJOVL0Mlgx/DAsYZW5X3s=
```
natalieparellano, line 305: I think all these deletions came from a weirdly resolved merge conflict...

natalieparellano:
This is probably ready for some more feedback, though you'll note there are some `TODO`s that still need to be addressed.

buildpacktoml_test.go
```diff
@@ -0,0 +1,729 @@
...
+	"github.com/sclevine/spec/report"
+
+	"github.com/buildpacks/lifecycle"
+	"github.com/buildpacks/lifecycle/launch"
+	"github.com/buildpacks/lifecycle/layers"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+	"github.com/buildpacks/lifecycle/testmock"
+)
+
+const latestBuildpackAPI = "0.5" // TODO: is there a good way to ensure this is kept up to date?
```
ekcasey, line 24: Instead of defining this constant you could use `api.Buildpack` and pull the last element or a helper func to the `api` package to return the latest api (if you have conerns about relying on the order), as it seems like it could be a generally useful.

buildpacktoml.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+		if err := validateUnmet(bpBuild.Unmet, bpPlanIn); err != nil {
+			return BuildResult{}, err
+		}
+		br.BOM = withBuildpack(bpFromBpInfo, launchTOML.BOM)
+		br.Met = names(bpPlanIn.filter(bpBuild.Unmet).Entries)
+	}
+
+	br.Labels = append([]Label{}, launchTOML.Labels...)
+	for i := range launchTOML.Processes {
```
jabrown85, line 204: I _think_ this behavior changed a bit. Previously, if there was no `launch.toml` it bailed out and never set processes, slices, or labels. This iteration feels better than what was happening before, but this loop of `launchTOML .Processes` for a file that does not exist file caught my attention.

buildpacktoml_test.go
jabrown85: I like the `BOM is an exact reflection of output buildpack plan` idea

cmd/lifecycle/builder.go
```diff
@@ -76,16 +76,17 @@ func (b *buildCmd) Exec() error {
...
+		LayersDir:       ba.layersDir,
+		PlatformDir:     ba.platformDir,
+		BuildpacksDir:   ba.buildpacksDir,
+		PlatformAPI:     api.MustParse(ba.platformAPI),
+		Env:             env.NewBuildEnv(os.Environ()),
+		Group:           group,
+		Plan:            plan,
+		Out:             cmd.Stdout,
+		Err:             cmd.Stderr,
+		BuildpackFinder: &lifecycle.DefaultBuildpackFinder{},
```
jabrown85, line 89: I would love to see `ba.buildpacksDir` be put into `lifecycle.DefaultBuildpackFinder{}` here to simplify the signatures of the find function. I think Emily mentioned the same thing.

builder.go
```diff
@@ -5,75 +5,100 @@ import (
...
+	Find(bpID, bpVersion, buildpacksDir string) (BuildpackTOML, error)
 }
 
-type Label struct {
-	Key   string `toml:"key"`
-	Value string `toml:"value"`
+type DefaultBuildpackFinder struct{}
+
+// TODO: this duplicates code in buildpack.go
+func (f *DefaultBuildpackFinder) Find(bpID, bpVersion, buildpacksDir string) (BuildpackTOML, error) {
```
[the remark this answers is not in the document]
natalieparellano: Ah, I tried calling the interface `BuildpackStore` but metadata.go already has a `BuildpackStore` struct - maybe this could become `BuildpackStoreTOML`?

buildpacktoml.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+type LaunchTOML struct {
+	BOM       []BOMEntry
+	Labels    []Label
+	Processes []launch.Process `toml:"processes"`
+	Slices    []layers.Slice   `toml:"slices"`
+}
+
+type DefaultBuildpackTOML struct {
```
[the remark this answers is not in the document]
natalieparellano: Made an issue to track this: https://github.com/buildpacks/lifecycle/issues/486

builder.go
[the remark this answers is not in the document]
natalieparellano: Trying out `BuildpackStore` -> `StoreTOML` for the existing struct. Then we can use `BuildpackStore` and `DirBuildpackStore` here.

buildpacktoml.go
```diff
@@ -0,0 +1,294 @@
...
+	Buildpack BuildpackInfo  `toml:"buildpack"`
+	Order     BuildpackOrder `toml:"order"`
+	Path      string         `toml:"-"`
+}
+
+func (b DefaultBuildpackTOML) String() string {
+	return b.Buildpack.Name + " " + b.Buildpack.Version
+}
+
+func (b *DefaultBuildpackTOML) Build(bpPlan BuildpackPlan, config BuildConfig) (BuildResult, error) {
```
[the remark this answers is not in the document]
natalieparellano: This seemed to make things easier to test - since `config` is the same for every buildpack, it only needs to be created once. Then the tests can assert on the specific value of `bpPlan` without having to make further modifications. I can change it though, if you think it's better to have the plan nested.

builder.go
```diff
@@ -5,75 +5,100 @@ import (
...
+	LayersDir   string
+	PlanDir     string
+	Out         io.Writer
+	Err         io.Writer
+}
+
+type BuildResult struct {
+	BOM       []BOMEntry
+	Labels    []Label
+	Met       []string
```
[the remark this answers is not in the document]
natalieparellano: Changing to `MetRequires`. I originally chose `Met` for parity with `Unmet` (`unmet` being a list in build.toml). But in this case probably it's better to have clarity over consistency!

buildpacktoml.go
natalieparellano: Ah, good call... looking for a way to improve this flow.

natalieparellano: Latest changes have a bail out if launch.toml does not exist. It makes the function even more verbose but could probably be made more readable with some helper functions.

buildpack.go
```diff
@@ -35,8 +35,8 @@ func (bp Buildpack) noHomepage() Buildpack {
 	return bp
 }
 
-func (bp Buildpack) Lookup(buildpacksDir string) (*BuildpackTOML, error) {
-	bpTOML := BuildpackTOML{}
+func (bp Buildpack) Lookup(buildpacksDir string) (*DefaultBuildpackTOML, error) {
```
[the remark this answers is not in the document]
natalieparellano: As I'm looking into this, it's slightly more challenging than expected. For build, it's sufficient that the `BuildpackStore.Lookup` method returns something that satisfies the `Buildpack` interface i.e., it can `Build()`. For detect, the thing returned by `Lookup` also needs to have `API`, `Buildpack`, `Order`, etc. Should I add those to the interface? Or is there a cleaner way to do this that I'm not thinking of?

buildpacktoml_test.go
```diff
@@ -0,0 +1,745 @@
...
+						filepath.Join(appDir, "launch-A-v1.toml"),
+					)
+
+					br, err := bpTOML.Build(lifecycle.BuildpackPlan{}, config)
+					if err != nil {
+						t.Fatalf("Unexpected error:\n%s\n", err)
+					}
+
+					if s := cmp.Diff(br, lifecycle.BuildResult{
+						BOM: nil,
```
natalieparellano, line 272: It feels a little weird to be returning `nil` here instead of an empty array (and also for `MetRequires`). But this is somewhat consistent with the behavior when there is no launch.toml - we bail out early without ever setting `Labels`, `Processes`, or `Slices`.

buildpacktoml_test.go
```diff
@@ -0,0 +1,745 @@
...
+				); s != "" {
+					t.Fatalf("Unexpected info:\n%s\n", s)
+				}
+			})
+
+			it("should set CNB_BUILDPACK_DIR", func() {
+				if _, err := bpTOML.Build(lifecycle.BuildpackPlan{}, config); err != nil {
+					t.Fatalf("Error: %s\n", err)
+				}
+				bpsDir, err := filepath.Abs(buildpacksDir) // TODO: this should be passed in somehow
```
natalieparellano, line 362: This feels kind of odd, it would be nice to derive this from `bpTOML.Path` instead of `buildpacksDir` which is even higher in the test setup. But that seemed to introduce more complexity than I'd like to deal with at the moment...

buildpacktoml_test.go
```diff
@@ -0,0 +1,745 @@
...
+			})
+		})
+
+		when("building succeeds with a clear env", func() {
+			it.Before(func() {
+				env.EXPECT().List().Return(append(os.Environ(), "TEST_ENV=cleared"))
+
+				bpTOML.Buildpack.Version = "v1.clear"
+				bpTOML.Path = filepath.Join(buildpacksDir, "A", "v1.clear")
+				bpTOML.Buildpack.ClearEnv = true // TODO: it is weird that all three of these conditions need to be specified
```
natalieparellano, line 344: It feels a bit weird that all three of these conditions need to be specified in order to get the test to behave correctly. I am not sure how to improve it...

buildpacktoml_test.go
```diff
@@ -0,0 +1,745 @@
...
+						`version = "some-version"`+"\n",
+					filepath.Join(appDir, "launch-A-v1.toml"),
+				)
+				_, err := bpTOML.Build(lifecycle.BuildpackPlan{}, config)
+				h.AssertNotNil(t, err)
+				expected := "top level version which is deprecated"
+				h.AssertStringContains(t, err.Error(), expected)
+			})
+
+			it("should error when the build bom has a top level version", func() {
```
natalieparellano, line 482: It might make more sense to fail out during export, but it's very convenient to test here :/

buildpacktoml_test.go
natalieparellano: On thinking about this a little bit more, this is not the behavior we had before (before the refactor, when everything was in one function). We used to just move on to processing the next buildpack. I think we should initialize all these fields to valid values.

natalieparellano:
Oy, I also need to fix merge conflicts with main...

builder_test.go
```diff
@@ -84,553 +93,348 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
 
-			it("should convert metadata version to top level version for buildpacks with buildpack api 0.2", func() {
-				builder.Plan = lifecycle.BuildPlan{
-					Entries: []lifecycle.BuildPlanEntry{
-						{
-							Providers: []lifecycle.Buildpack{
-								{ID: "B", Version: "v2"},
+			when("build metadata", func() {
+				when("bom", func() {
+					it("should convert top level version to metadata.version", func() {
```
ekcasey, line 152: Should we need to do a conversion in the default case here? It might be nice if the buildpack always returned the data in a consistent way regardless of the bp API (never top-level) and the the builder only converted provide labels in the legacy format.

builder_test.go
```diff
@@ -84,553 +93,348 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								Direct:      true,
+								BuildpackID: "A",
+							},
+						}); s != "" {
+							t.Fatalf("Unexpected:\n%s\n", s)
+						}
+					})
+				})
+
+				when("slices", func() {
```
ekcasey, line 325: Love the additional coverage and rigor here.

buildpacktoml_test.go
```diff
@@ -0,0 +1,773 @@
...
+						},
+					},
+				}
+
+				_, err := bpTOML.Build(bpPlan, config)
+				if err != nil {
+					t.Fatalf("Unexpected error:\n%s\n", err)
+				}
+
+				testPlan(t,
```
ekcasey, line 550: `testPlan` is only used in the this test file but is still defined in `builder_test.go`. Maybe we should audit the test functions in `builder_test.go` and move specific ones like this to the file they are used and move generic ones to the `testhelpers` package if appropriate?

buildpacktoml.go
```diff
@@ -0,0 +1,336 @@
...
+				metadataVersion := fmt.Sprintf("%v", version)
+				if entry.Version != "" && entry.Version != metadataVersion {
+					return errors.New("top level version does not match metadata version")
+				}
+			}
+		}
+	} else {
+		for _, entry := range bom {
+			if entry.Version != "" {
+				return fmt.Errorf("bom entry '%s' has a top level version which is deprecated", entry.Name)
```
ekcasey, line 268: This is an error so we shouldn't call it "deprecated".
```suggestion
				return fmt.Errorf("bom entry '%s' has a top level version which is not allowed. The buildpack should instead set metadata.version", entry.Name)
```

buildpacktoml_test.go
```diff
@@ -0,0 +1,773 @@
...
+		if err != nil {
+			t.Fatalf("Error: %s\n", err)
+		}
+		stdout, stderr = &bytes.Buffer{}, &bytes.Buffer{}
+		platformDir = filepath.Join(tmpDir, "platform")
+		layersDir = filepath.Join(tmpDir, "launch")
+		appDir = filepath.Join(layersDir, "app")
+		mkdir(t, layersDir, appDir, filepath.Join(platformDir, "env"))
+
+		buildpacksDir, err = filepath.Abs(filepath.Join("testdata", "by-id"))
```
ekcasey, line 63: I was going to suggest a cleanup of the testdata here given we don't need the full symlinked buildpacks dir for this test, but I am realizing it is probably needed for detect tests. This is something we could likely simplify in the future when we refactor detector to use the new patterns.

builder_test.go
```diff
@@ -18,33 +18,39 @@ import (
...
 	"github.com/buildpacks/lifecycle"
 	"github.com/buildpacks/lifecycle/api"
-	"github.com/buildpacks/lifecycle/env"
 	"github.com/buildpacks/lifecycle/launch"
-	h "github.com/buildpacks/lifecycle/testhelpers"
+	"github.com/buildpacks/lifecycle/layers"
 	"github.com/buildpacks/lifecycle/testmock"
 )
 
+var latestPlatformAPI = api.Platform.Latest()
```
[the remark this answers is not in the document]
natalieparellano: I think this is a remnant of an earlier approach... I'll make a change.

buildpack.go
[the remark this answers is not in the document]
ekcasey: let's punt this for now

buildpacktoml_test.go
ekcasey: Yeah, I had that thought while reading. I think we could fix it with some refactoring of the test buildpacks but it's probably better to defer until we complete the package extraction refactor.

buildpacktoml_test.go
ekcasey: Might as well fail early and save the user some time, especially if it is convenient :)

utils_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/sclevine/spec/report"
 
 	"github.com/buildpacks/lifecycle"
+	"github.com/buildpacks/lifecycle/testhelpers"
```
ekcasey, line 14:
```suggestion
	h "github.com/buildpacks/lifecycle/testhelpers"
```

utils_test.go
```diff
@@ -34,7 +35,7 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("should return an ordering of buildpacks", func() {
-			mkfile(t,
+			testhelpers.Mkfile(t,
```
ekcasey, line 38:
```suggestion
			h.Mkfile(t,
```

utils_test.go
```diff
@@ -70,15 +71,15 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("should return a group of buildpacks", func() {
-			mkfile(t, `group = [{id = "A", version = "v1"}, {id = "B", optional = true}]`,
+			testhelpers.Mkfile(t, `group = [{id = "A", version = "v1"}, {id = "B", optional = true}]`,
```
ekcasey, line 74:
```suggestion
			h.Mkfile(t, `group = [{id = "A", version = "v1"}, {id = "B", optional = true}]`,
```

utils_test.go
```diff
@@ -104,11 +105,11 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("should write TOML", func() {
-			group := lifecycle.BuildpackGroup{Group: []lifecycle.Buildpack{{ID: "A", Version: "v1"}}}
+			group := lifecycle.BuildpackGroup{Group: []lifecycle.GroupBuildpack{{ID: "A", Version: "v1"}}}
 			if err := lifecycle.WriteTOML(filepath.Join(tmpDir, "subdir", "group.toml"), group); err != nil {
 				t.Fatal(err)
 			}
-			b := rdfile(t, filepath.Join(tmpDir, "subdir", "group.toml"))
+			b := testhelpers.Rdfile(t, filepath.Join(tmpDir, "subdir", "group.toml"))
```
ekcasey, line 112:
```suggestion
			b := h.Rdfile(t, filepath.Join(tmpDir, "subdir", "group.toml"))
```

builder_test.go
```diff
@@ -18,9 +16,9 @@ import (
 
 	"github.com/buildpacks/lifecycle"
 	"github.com/buildpacks/lifecycle/api"
-	"github.com/buildpacks/lifecycle/env"
 	"github.com/buildpacks/lifecycle/launch"
-	h "github.com/buildpacks/lifecycle/testhelpers"
+	"github.com/buildpacks/lifecycle/layers"
+	"github.com/buildpacks/lifecycle/testhelpers"
```
ekcasey, line 21:
```suggestion
	h "github.com/buildpacks/lifecycle/testhelpers"
```

builder_test.go
```diff
@@ -55,25 +58,28 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 		platformDir = filepath.Join(tmpDir, "platform")
 		layersDir = filepath.Join(tmpDir, "launch")
 		appDir = filepath.Join(layersDir, "app")
-		mkdir(t, layersDir, appDir, filepath.Join(platformDir, "env"))
-
-		buildpacksDir := filepath.Join("testdata", "by-id")
+		testhelpers.Mkdir(t, layersDir, appDir, filepath.Join(platformDir, "env"))
```
ekcasey, line 61:
```suggestion
		h.Mkdir(t, layersDir, appDir, filepath.Join(platformDir, "env"))
```

ekcasey (approved):

PR #462, opened 2020-11-12
Update GitHub Actions workflows to use environment files

### LINKED ISSUE #424: Update GitHub Actions workflows to use environment files
As of 10/1/2020, `set-env` and `add-path` are deprecated: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

natalieparellano:
Fixes #424

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -21,7 +21,7 @@ jobs:
             echo "lifecycle version not detected."
             exit 1
           fi
-          echo "name=LIFECYCLE_VERSION::$version" >> $GITHUB_ENV
+          echo "LIFECYCLE_VERSION=$version" >> $GITHUB_ENV
         shell: bash
       - name: Determine download urls for linux and windows
         id: artifact-urls
@@ -97,7 +97,7 @@ jobs:
           mkdir artifact-linux
           unzip -d artifact-linux artifact-linux.zip
           lifecycle_path=$(ls artifact-linux/lifecycle-*linux.x86-64.tgz)
-          echo "name=ARTIFACT_LINUX_PATH::$PWD/$lifecycle_path" >> $GITHUB_ENV
+          echo "ARTIFACT_LINUX_PATH=$PWD/$lifecycle_path" >> $GITHUB_ENV
       - name: Download windows artifact
         run: |
           url=$(echo '${{ steps.artifact-urls.outputs.result }}' | jq -r .windows )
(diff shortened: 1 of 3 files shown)


.github/workflows/draft-release.yml
```diff
@@ -13,15 +13,15 @@ jobs:
...
+          echo "${PWD}/deps/bin" >> $GITHUB_PATH
       - name: Derive lifecycle version from branch name
         run: |
           [[ $GITHUB_REF =~ ^refs\/heads\/release/(.*)$ ]] && version=${BASH_REMATCH[1]}
           if [[ -z "${version}" ]]; then
             echo "lifecycle version not detected."
             exit 1
           fi
-          echo "::set-env name=LIFECYCLE_VERSION::$version"
+          echo "name=LIFECYCLE_VERSION::$version" >> $GITHUB_ENV
```
ekcasey, line 24: The way I read the [documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable), I would have expected these replacements for `set-env` to look like
```
echo "LIFECYCLE_VERSION=$version" >> $GITHUB_ENV
```
am I missing something?

.github/workflows/post-release.yml
```diff
@@ -33,8 +33,8 @@ jobs:
           echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
       - name: Set env
         run: |
-          echo "::set-env name=LIFECYCLE_VERSION::$(echo ${{ github.event.release.tag_name }} | cut -d "v" -f2)"
-          echo "::set-env name=LIFECYCLE_IMAGE_TAG::$(git describe --always --dirty)"
+          echo "name=LIFECYCLE_VERSION::$(echo ${{ github.event.release.tag_name }} | cut -d "v" -f2)" >> $GITHUB_ENV
+          echo "name=LIFECYCLE_IMAGE_TAG::$(git describe --always --dirty)" >> $GITHUB_ENV
```
ekcasey, line 37: Found this in docs, seems like something we might want to pay attention to:

> Warning: Powershell does not use UTF-8 by default. Make sure you write files using the correct encoding. For > example, you need to set UTF-8 encoding when you set the path:
>
> steps:
>   - run: echo "mypath" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append

.github/workflows/draft-release.yml
natalieparellano: Ohhh I think I must have been very careless. Taking a look.

ekcasey (approved):

PR #473, opened 2020-11-18
Exec.d

ekcasey:
Skipping windows pending further discussion: https://github.com/buildpacks/spec/issues/164

Introduces Buildpack API 0.5


launch/launcher_test.go
```diff
@@ -229,13 +239,126 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+							filepath.Join(tmpDir, "launch", "0.4_buildpack", "layer4", "exec.d", "exec_d_1"),
+							filepath.Join(tmpDir, "launch", "0.4_buildpack", "layer4", "exec.d", "start", "exec_d_1"),
+							filepath.Join(tmpDir, "launch", "0.5_buildpack", "layer5", "exec.d", "exec_d_1"),
+							filepath.Join(tmpDir, "launch", "0.5_buildpack", "layer5", "exec.d", "exec_d_2"),
+							filepath.Join(tmpDir, "launch", "0.5_buildpack", "layer5", "exec.d", "start", "exec_d_1"),
+							filepath.Join(tmpDir, "launch", "0.5_buildpack", "layer5", "exec.d", "start", "exec_d_2"),
+						)
+					})
+
+					it("should run exec.d binaries after apply the static env files", func() {
```
jabrown85, line 273:
```suggestion
					it("should run exec.d binaries after static env files", func() {
```

jabrown85 (approved):

launch/exec_d.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	errChan := make(chan error, 1)
+	go func() {
+		defer pw.Close()
+		cmd := exec.Command(path)
+		cmd.Stdout = e.Out
+		cmd.Stderr = e.Err
+		cmd.ExtraFiles = []*os.File{pw}
+		cmd.Env = env.List()
+		errChan <- cmd.Run()
```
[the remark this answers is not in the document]
ekcasey: Yeah, I thought about that. We currently don't do that for `profile.d` (which is analogous). It seems like something we would need to specify out explicitly in the `exec.d` interface if we were to enforce it.

launch/launcher.go
```diff
@@ -124,6 +146,85 @@ func (l *Launcher) env(process Process) error {
...
+	}
+	return files, nil
+}
+
+func (l *Launcher) eachBuildpack(fn func(path string) error, filters ...func(bp Buildpack) bool) error {
+Loop:
+	for _, bp := range l.Buildpacks {
+		for _, filter := range filters {
+			if !filter(bp) {
+				continue Loop
```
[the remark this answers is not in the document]
ekcasey: Yeah, that might have been a little lazy. I'll refactor it.

launch/launcher.go
```diff
@@ -42,15 +60,19 @@ func (l *Launcher) Launch(self string, cmd []string) error {
...
-// LaunchProcess launches the provided process
-//   For direct=false processes, self is used to set argv0 during profile script execution
+// LaunchProcess launches the provided process.
+// For direct=false processes, self is used to set argv0 during profile script execution
 func (l *Launcher) LaunchProcess(self string, process Process) error {
+	if err := os.Chdir(l.AppDir); err != nil {
+		return errors.Wrap(err, "change to app directory")
+	}
 	if err := l.env(process); err != nil {
 		return errors.Wrap(err, "modify env")
```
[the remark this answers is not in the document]
ekcasey: This error comes from applying the static env files, not from the `profile.d` scripts.

launch/launcher.go
```diff
@@ -124,6 +146,85 @@ func (l *Launcher) env(process Process) error {
...
+			if err := appendFilesInDir(filepath.Join(path, dirName)); err != nil {
+				return err
+			}
+			if process.Type != "" {
+				return appendFilesInDir(filepath.Join(path, dirName, process.Type))
+			}
+			return nil
+		})
+	}, filters...); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
ekcasey: Because of this line https://github.com/buildpacks/lifecycle/pull/473/files#diff-dcbaa95f60b04e4252f4de7f9e6718e4f63b3f5e516a8871b6e9da20d2c51461R152, the error would end up saying `failed to find all exec.d executables in layers dir` twice if I accepted this suggestion. Also this helper method is used for profile scripts as well.

acceptance/launcher_test.go
```diff
@@ -42,6 +42,28 @@ func TestLauncher(t *testing.T) {
...
+				cmd := exec.Command("docker", "run", "--rm",
+					"--env=VAR_FROM_EXEC_D=ORIG_VAL",
+					launchImage, "exec.d-checker")
+				expected := "/layers/0.5_buildpack/some_layer/exec.d/helper was executed\n"
+				expected += "Exec.d Working Dir: /workspace\n"
+				expected += "/layers/0.5_buildpack/some_layer/exec.d/exec.d-checker/helper was executed\n"
+				expected += "Exec.d Working Dir: /workspace\n"
+				expected += "sourced bp profile\n"
+				expected += "sourced app profile\n"
+				expected += "VAR_FROM_EXEC_D: ORIG_VAL:VAL_FROM_EXEC_D:VAL_FROM_EXEC_D"
```
natalieparellano, line 61: Why is `VAL_FROM_EXEC_D` printed twice?

launch/exec_d_test.go
```diff
@@ -0,0 +1,114 @@
...
+				env.EXPECT().List().Return([]string{})
+				env.EXPECT().Set(gomock.Any(), gomock.Any()).AnyTimes()
+				h.AssertNil(t, runner.ExecD(path, env))
+				h.AssertEq(t, errOut.String(), "stderr from execd\n")
+			})
+		})
+
+		when("windows", func() {
+			it.Before(func() {
+				h.SkipIf(t, runtime.GOOS != "windows", "exec.d is not yet supported on windows")
```
natalieparellano, line 106:
```suggestion
				h.SkipIf(t, runtime.GOOS != "windows", "windows test")
```

launch/launcher_test.go
```diff
@@ -229,13 +239,126 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
 							t.Fatalf("expected syscall.Exec to be called once: actual %v\n", syscallExecArgsColl)
 						}
 					})
 				})
+
+				when("buildpack API supports exec.d", func() {
+					it.Before(func() {
+						mkdir(t,
+							filepath.Join(tmpDir, "launch", "0.4_buildpack", "layer4", "exec.d"),
+							filepath.Join(tmpDir, "launch", "0.4_buildpack", "layer4", "exec.d", "start"),
```
natalieparellano, line 259: Do we expect these to be ignored?

launch/launcher_test.go
```diff
@@ -229,13 +239,126 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+							filepath.Join(tmpDir, "launch", "0.5_buildpack", "layer5", "exec.d"),
+							filepath.Join(tmpDir, "launch", "0.5_buildpack", "layer5", "exec.d", "start"),
+						)
+						mkfile(t, "",
+							filepath.Join(tmpDir, "launch", "0.4_buildpack", "layer4", "exec.d", "exec_d_1"),
+							filepath.Join(tmpDir, "launch", "0.4_buildpack", "layer4", "exec.d", "start", "exec_d_1"),
```
natalieparellano, line 265: Same Q as above...

launch/launcher_test.go
```diff
@@ -229,13 +239,126 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
...
+						)
+						h.AssertNil(t, launcher.LaunchProcess("", process))
+						if len(syscallExecArgsColl) != 1 {
+							t.Fatalf("expected syscall.Exec to be called once: actual %v\n", syscallExecArgsColl)
+						}
+					})
+
+					when("process is buildpack-provided", func() {
+						it.Before(func() {
+							process.Type = "start"
```
natalieparellano, line 310: In reading through these tests at first I thought "start" had some special meaning - maybe it's clearer to call it `some-process`

launch/launcher.go
```diff
@@ -124,6 +146,86 @@ func (l *Launcher) env(process Process) error {
...
+			return err
+		}
+	}
+	return nil
+}
+
+func supportsExecD(bp Buildpack) bool {
+	if bp.API == "" {
+		return false
+	}
```
natalieparellano, line 165: I gather that if a buildpack implementing an earlier API writes an `exec.d`, it will be ignored. I am wondering why we don't give Windows buildpacks the same treatment i.e., can we check the runtime.GOOS here, and not fail if we find a Windows `exec.d`?

launch/launcher.go
```diff
@@ -124,6 +146,86 @@ func (l *Launcher) env(process Process) error {
...
+			return errors.Wrapf(err, "failed to list files in dir '%s'", path)
+		}
+
+		for _, fi := range fis {
+			appendIfFile(filepath.Join(path, fi.Name()), fi)
+		}
+		return nil
+	}
+
+	if err := l.eachBuildpack(func(path string) error {
```
natalieparellano, line 193: This code (and elsewhere where `eachBuildpack` is called) might be easier to understand if we give the function a name, similar to how `buildpackFiles` is called with `supportsExecD`

acceptance/launcher_test.go
ekcasey: The first `VAL_FROM_EXEC_D` comes from `exec.d/helper` and the second comes from `exec.d/<process-type>/helper` I am happy to add a comment here or I change the actual exec.d logic to make the appended values different in these two cases. Do you have a preference?

launch/launcher.go
ekcasey: I thought about that but decided to go this direction b/c buildpack spec 0.5 implies that `exec.d` should be available on windows. Therefore, if the lifecycle is not going to support exec.d on windows, I think it should be very explicit about it so buildpack authors aren't surprised to learn that they were silently omitted.

I would be happy to change the error to a warning instead if you think that is better.

acceptance/launcher_test.go
natalieparellano: Ah, this makes sense. I am partial to having the values be different, seems more durable than a comment.

launch/launcher.go
natalieparellano: I see. In this case an error might be better, as a warning could be missed.

launch/launcher.go
ekcasey: Creating name functions as you suggested turned out to be more difficult than I anticipated because many of these functions manipulated the receiver. I ended up doing a slightly different refactor https://github.com/buildpacks/lifecycle/pull/473/commits/4147c88b565d80820557452e47eb044ae9f6e1cf

Let me know if you think this is more readable. I can back it out if it isn't.

launch/launcher.go
```diff
@@ -124,6 +146,85 @@ func (l *Launcher) env(process Process) error {
...
+}
+
+func supportsExecD(bp Buildpack) bool {
+	if bp.API == "" {
+		return false
+	}
+	return api.MustParse(bp.API).Compare(api.MustParse("0.5")) >= 0
+}
+
+func (l *Launcher) buildpackFiles(process Process, dirName string, filters ...func(Buildpack) bool) ([]string, error) {
```
yaelharel: It looks like we're using only `process.Type` in this function, so maybe it's better that `buildpackFiles` will get only it instead of `Process`.

launch/launcher.go
yaelharel: Do you want to get back to line 215? Meaning, go to the next buildpack in `l.Buildpacks`?
If yes, isn't it better to call `break` here?

launch/launcher.go
```diff
@@ -124,6 +146,85 @@ func (l *Launcher) env(process Process) error {
 	})
 }
 
+func (l *Launcher) execD(process Process) error {
```
yaelharel: What do you think about changing the function name to `launchWithExecutable` in order to be consistent with `launchWithShell` (from `shell.go`)?

launch/launcher.go
yaelharel:
```suggestion
func (l *Launcher) getBuildpackFiles(process Process, dirName string, filters ...func(Buildpack) bool) ([]string, error) {
```

launch/shell.go
```diff
@@ -45,44 +44,8 @@ func (l *Launcher) launchWithShell(self string, process Process) error {
 }
 
 func (l *Launcher) profiles(process Process) ([]string, error) {
```
yaelharel, line 46: What do you think about renaming this function to `getProfileDFiles` (or something similar starting with `get`)?

launch/launcher.go
```diff
@@ -75,76 +92,119 @@ func (l *Launcher) launchDirect(process Process) error {
...
+}
+
+type action func(path string) error
+type bpPredicate func(bp Buildpack) bool
+
+func (l *Launcher) eachBuildpack(fn action, predicates ...bpPredicate) error {
+	for _, bp := range l.Buildpacks {
+		var skip bool
+		for _, pred := range predicates {
+			skip = skip || !pred(bp)
```
natalieparellano, line 127:
```suggestion
			skip = skip || !pred(bp) // skip unless all predicates return true
```

launch/launcher.go
```diff
@@ -75,76 +92,119 @@ func (l *Launcher) launchDirect(process Process) error {
...
+		}
+		return nil
+	})
+}
 
-	return Process{}, false
+func (l *Launcher) doExecD(process Process) error {
+	return l.eachBuildpack(func(path string) error {
+		return eachLayer(path, l.doLayerExecD(process))
+	}, supportsExecD)
```
natalieparellano, line 110:
```suggestion
	return l.eachBuildpack(func(bpDir string) error {
		return eachLayer(bpDir, l.doLayerExecD(process))
	}, supportsExecD)
```

launch/shell.go
```diff
@@ -46,44 +45,10 @@ func (l *Launcher) launchWithShell(self string, process Process) error {
...
-		return eachDir(absPath, func(path string) error {
-			if err := appendFilesInDir(filepath.Join(path, "profile.d")); err != nil {
-				return err
-			}
-			if process.Type != "" {
-				return appendFilesInDir(filepath.Join(path, "profile.d", process.Type))
-			}
-			return nil
-		})
+		return eachLayer(path, l.populateLayerProfiles(process, &profiles))
```
natalieparellano, line 49: Having trouble leaving a code suggestion, but path -> bpDir might be clearer here as well.

launch/shell.go
```diff
@@ -92,11 +57,31 @@ func (l *Launcher) profiles(process Process) ([]string, error) {
...
+		profiles = append(profiles, appProfile)
+	}
 
 	return profiles, nil
 }
 
+func (l *Launcher) populateLayerProfiles(proc Process, profiles *[]string) action {
+	return func(path string) error {
+		if err := eachFile(filepath.Join(path, "profile.d"), func(path string) error {
+			*profiles = append(*profiles, path)
```
natalieparellano, line 70:
```suggestion
		if err := eachFile(filepath.Join(path, "profile.d"), func(profilePath string) error {
			*profiles = append(*profiles, profilePath)
```

launch/shell.go
```diff
@@ -92,11 +57,31 @@ func (l *Launcher) profiles(process Process) ([]string, error) {
...
+			return nil
+		}); err != nil {
+			return err
+		}
+		if proc.Type == "" {
+			return nil
+		}
+		return eachFile(filepath.Join(path, "profile.d", proc.Type), func(path string) error {
+			*profiles = append(*profiles, path)
```
natalieparellano, line 79:
```suggestion
		return eachFile(filepath.Join(path, "profile.d", proc.Type), func(profilePath string) error {
			*profiles = append(*profiles, profilePath)
```

launch/launcher.go
natalieparellano: Left a couple of suggestions to improve the readability, but overall I think this makes sense!

launch/launcher.go
ekcasey: This is out of date after 4147c88 but I think it is meaningfully different from `launchWithShell` because `launchWithShell` actually launches the process whereas `execD` (now `doExecD` only executes the `exec.d` binaries).

launch/shell.go
ekcasey: I went with `getProfiles` because the returned values include `<app-dir>/.profile` if it exists which isn't in a `profile.d` directory.

launch/launcher.go
ekcasey: I kind of like like passing around the process because functions and methods can have symmetrical signatures regardless of the field(s) they read or eventually will read from the process.

acceptance/testdata/launcher/posix/exec.d/main.go
```diff
@@ -0,0 +1,33 @@
...
+
+func main() {
+	fmt.Printf("%s was executed\n", os.Args[0])
+	wd, err := os.Getwd()
+	if err != nil {
+		fmt.Println("ERROR: failed to get working dir:", err)
+		os.Exit(1)
+	}
+	fmt.Println("Exec.d Working Dir:", wd)
+	f := os.NewFile(3, "/dev/fd/3")
```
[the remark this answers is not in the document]
ekcasey: changed to `fd3`

launch/shell.go
ekcasey: I renamed the other use of path to `layerDir` which I think should make this use of `path` clearer.

launch/launcher.go
yaelharel: Can you please elaborate?
If this function doesn't use anything other than `process.Type`, why do we need to pass `Process`?
If things will change in the future, we can always change the function signature.
Am I missing something?

launch/launcher.go
```diff
@@ -166,77 +117,94 @@ func supportsExecD(bp Buildpack) bool {
...
+	})
+}
+
+func eachFile(dir string, action action) error {
+	return eachInDir(dir, action, func(fi os.FileInfo) bool {
+		return !fi.IsDir()
+	})
+}
+
+func eachInDir(dir string, action action, pred func(fi os.FileInfo) bool) error {
```
yaelharel, line 195:
```suggestion
func eachInDir(dir string, action action, predicate func(fi os.FileInfo) bool) error {
```
To be consistent with line 126.

launch/launcher.go
```diff
@@ -166,77 +117,94 @@ func supportsExecD(bp Buildpack) bool {
...
+		if err := l.Env.AddEnvDir(filepath.Join(path, "env.launch")); err != nil {
+			return err
+		}
+		if proc.Type == "" {
 			return nil
-		})
-	}, filters...); err != nil {
-		return nil, err
+		}
+		return l.Env.AddEnvDir(filepath.Join(path, "env.launch", proc.Type))
```
yaelharel, line 163: I see that you call `AddEnvDir` once with `proc.Type` and once without it (line 157).
I guess it's because of the files that exist in the path.
Can you please elaborate?

launch/launcher.go
```diff
@@ -166,77 +117,94 @@ func supportsExecD(bp Buildpack) bool {
...
+	return func(path string) error {
+		if err := eachFile(filepath.Join(path, "exec.d"), func(path string) error {
+			return l.ExecD.ExecD(path, l.Env)
+		}); err != nil {
 			return err
 		}
+		if proc.Type == "" {
+			return nil
+		}
+		return eachFile(filepath.Join(path, "exec.d", proc.Type), func(path string) error {
```
yaelharel, line 177: Same question as the one in `doLayerEnvFiles`

launch/process.go
```diff
@@ -57,6 +57,16 @@ func (l *Launcher) processForLegacy(cmd []string) (Process, error) {
 	return l.userProvidedProcess(cmd)
 }
 
+func (l *Launcher) findProcessType(kind string) (Process, bool) {
```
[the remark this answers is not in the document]
ekcasey: yeah I agree.  I actually just moved this function between files, but while we are looking param names we should clean this up. Unfortunately `type` is a reserved keyword, but we can go with `pType` (we use that name elsewhere).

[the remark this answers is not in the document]
yaelharel: Yeah, `type` is a reserved keyword. Maybe we can also go with `desiredType`/`requestedType` or something else similar.

launch/launcher.go
ekcasey: We are processing the env files in two different directories. The first `<layer>/env` contains files that apply to all process types. The second `<layer>/env/<process-type>` contains files that are specific to the given process-type.

launch/launcher.go
ekcasey: same answer :) https://github.com/buildpacks/spec/blob/main/buildpack.md#process-4

launch/process.go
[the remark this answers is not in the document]
ekcasey: I like to keep the names short when there is no ambiguity. Given the name of the function is `findProcessType` I think `pType` is clear. Given how short the method is, I think one _could_ make a pretty compelling argument for naming it `t` but i'll spare us 😉. 

https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#variable-names

launch/shell.go
```diff
@@ -62,19 +64,37 @@ func (l *Launcher) profiles(process Process) ([]string, error) {
 	return profiles, nil
 }
 
-func (l *Launcher) isScript(process Process) (bool, error) {
+func (l *Launcher) populateLayerProfiles(proc Process, profiles *[]string) action {
```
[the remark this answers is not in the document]
ekcasey: done

launch/launcher.go
yaelharel: Thanks!

yaelharel (approved):

PR #476, opened 2020-11-20
Mention creator in README

### LINKED ISSUE #311: Add docs for Creator
Given that we can now use the Creator to run all phases, it would be helpful to mention that in the README, perhaps even with some more rigorous documentation.

ekcasey:
* Resolves #311

Mentions the existence of `creator` in the README and tweaks phase descriptions slightly.

This PR doesn't even attempt to fix most of our README problems but it fixes one large omission.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,19 +23,23 @@ Lifecycle Version | Platform APIs              | Buildpack APIs |
 
 ### Build
 
-* `detector` - chooses buildpacks (via `/bin/detect`)
-* `analyzer` - restores launch layer metadata from the previous build
-* `restorer` - restores cache
-* `builder` -  executes buildpacks (via `/bin/build`)
-* `exporter` - creates image and stores cache
+Either:
+* `detector` - Chooses buildpacks (via `/bin/detect`) and produces a build plan.
+* `analyzer` - Restores layer metadata from the previous image and from the cache.
+* `restorer` - Restores cached layers.
+* `builder` -  Executes buildpacks (via `/bin/build`).
+* `exporter` - Creates an image and caches layers.
+
+Or:
+* `creator` - Runs the five phases listed above in order.
 
 ### Run
 
-* `launcher` - invokes choice of process
+* `launcher` - Invokes a chosen process.
 
 ### Rebase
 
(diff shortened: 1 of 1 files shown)


jabrown85 (approved):
👏

README.md
```diff
@@ -23,19 +23,23 @@ Lifecycle Version | Platform APIs              | Buildpack APIs |
...
 ### Build
 
-* `detector` - chooses buildpacks (via `/bin/detect`)
-* `analyzer` - restores launch layer metadata from the previous build
-* `restorer` - restores cache
-* `builder` -  executes buildpacks (via `/bin/build`)
-* `exporter` - creates image and stores cache
+Either:
+* `detector` - Chooses buildpacks (via `/bin/detect`) and produces a build plan.
+* `analyzer` - Restores layer metadata from the previous image and from the cache.
```
yaelharel, line 28: Is it `and` or `or`?
Can it happen that there will be both previous image and a cache and the `analyzer` will restore metadata from both of them?
Can we have conflicts in such case?

README.md
```diff
@@ -23,19 +23,23 @@ Lifecycle Version | Platform APIs              | Buildpack APIs |
...
+* `restorer` - Restores cached layers.
+* `builder` -  Executes buildpacks (via `/bin/build`).
+* `exporter` - Creates an image and caches layers.
+
+Or:
+* `creator` - Runs the five phases listed above in order.
```
yaelharel, line 34: Do we want to add that they'll run in the same container or is it too much information for here?

ekcasey: Given that the lifecycle doesn't orchestrate containers but instead runs inside a container I don't think we need to mention that here. We _should_ provide more clarity about how to use these binaries (probably replacing parts of our readme with a link to the docs instead of duplicating soon-to-be-improved docs here. For now I just want to get this one quick fix in.

README.md
ekcasey: It restores metadata for launch layers from the previous image and metadata for cached layers from the cache.

https://github.com/buildpacks/spec/blob/main/platform.md#layer-analysis

yaelharel: Thanks!

Issue #453, opened 2020-10-23
Exporter always tries to delete cache image


jabrown85:
### Summary
When using the `-cache-image` flag, the `exporter` always issues a delete to the registry.

---

##### Steps

1. Run a `registry:2` service ephemerally to ensure no images exist via `docker run -it -p 5000:5000 registry:2`
2. Run `/cnb/lifecycle/creator repo/img` with `-cache-image=localhost:5000/repo/img:build-cache` using the local registry as the target registry.
3. Observe that lifecycle sends a `DELETE` for the "previous" cache image, when there could not have been one.

##### Current behavior

Lifecycle logs
```
Adding 1/1 app layer(s)
Adding layer 'launcher'
Adding layer 'config'
Adding label 'io.buildpacks.lifecycle.metadata'
Adding label 'io.buildpacks.build.metadata'
Adding label 'io.buildpacks.project.metadata'
*** Images (sha256:62a0edeb85a97bd975367e9e9b57d75c2ae8b1af3b7c477080b4f4b48ac489f7):
      192.168.86.34:5000/frozen-woodland-11119/function
Adding cache layer 'heroku/go:shim'
Unable to delete previous cache image: DELETE http://192.168.86.34:5000/v2/frozen-woodland-11119/function/manifests/sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10: UNSUPPORTED: The operation is unsupported.complete
```
Registry logs
```
ERRO[0261] response completed with error                 err.code=unsupported err.message="The operation is unsupported." go.version=go1.11.2 http.request.host="192.168.86.34:5000" http.request.id=f030c486-e117-4125-8af3-1542da211d04 http.request.method=DELETE http.request.remoteaddr="172.17.0.1:43940" http.request.uri="/v2/frozen-woodland-11119/function/manifests/sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10" http.request.useragent=go-containerregistry http.response.contenttype="application/json; charset=utf-8" http.response.duration="926.028µs" http.response.status=405 http.response.written=78 vars.name="frozen-woodland-11119/function" vars.reference="sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10"
[...]

PR #477, opened 2020-11-20
fix: Do not delete cache-img if the cache-img did not exist

jabrown85:
When using image cache, the commit of new cache deletes the old cache. There is a check for if there is a previous cache image, but that check was happening after the new cache was written to that tag.

https://github.com/buildpacks/lifecycle/issues/453

### DIFF
--- a/cache/image_cache.go
+++ b/cache/image_cache.go
@@ -95,12 +95,15 @@ func (c *ImageCache) Commit() error {
 		return errCacheCommitted
 	}
 
+	// Check if the cache image exists prior to saving the new cache at that same location
+	origImgExists := c.origImage.Found()
+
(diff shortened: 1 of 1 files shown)


jabrown85:
I was going to add a test, but this failure case only produces a line output. I wasn't sure if there was a great way to test this, I didn't immediately see one but I'll be glad to add a test if there is a pattern to follow.

ekcasey (approved):

ekcasey:
> I was going to add a test, but this failure case only produces a line output. I wasn't sure if there was a great way to test this, I didn't immediately see one but I'll be glad to add a test if there is a pattern to follow.

Yeah that makes sense for now.

PR #483, opened 2020-12-03
Changes default env file action to override

ekcasey:
* Applies to buildpacks with API 0.5+

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/env"
 	"github.com/buildpacks/lifecycle/launch"
 	"github.com/buildpacks/lifecycle/layers"
 )
@@ -31,7 +32,7 @@ type Builder struct {
 
 type BuildEnv interface {
 	AddRootDir(baseDir string) error
-	AddEnvDir(envDir string) error
+	AddEnvDir(envDir string, defaultAction env.ActionType) error
 	WithPlatform(platformDir string) ([]string, error)
 	List() []string
 }
@@ -131,7 +132,7 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 		if err := cmd.Run(); err != nil {
 			return nil, NewLifecycleError(err, ErrTypeBuildpack)
 		}
-		if err := setupEnv(b.Env, bpLayersDir); err != nil {
+		if err := b.setupEnv(api.MustParse(bp.API), bpLayersDir); err != nil {
 			return nil, err
 		}
 		var bpPlanOut BuildpackPlan
@@ -248,12 +249,12 @@ func (bom *BOMEntry) convertVersionToMetadata() error {
 	return nil
 }
 
-func setupEnv(env BuildEnv, layersDir string) error {
+func (b *Builder) setupEnv(bpAPI *api.Version, layersDir string) error {
 	if err := eachDir(layersDir, func(path string) error {
 		if !isBuild(path + ".toml") {
 			return nil
 		}
-		return env.AddRootDir(path)
+		return b.Env.AddRootDir(path)
 	}); err != nil {
 		return err
 	}
@@ -262,10 +263,10 @@ func setupEnv(env BuildEnv, layersDir string) error {
 		if !isBuild(path + ".toml") {
 			return nil
 		}
-		if err := env.AddEnvDir(filepath.Join(path, "env")); err != nil {
+		if err := b.Env.AddEnvDir(filepath.Join(path, "env"), env.DefaultActionType(bpAPI)); err != nil {
 			return err
 		}
-		return env.AddEnvDir(filepath.Join(path, "env.build"))
+		return b.Env.AddEnvDir(filepath.Join(path, "env.build"), env.DefaultActionType(bpAPI))
 	})
 }
 
--- a/env/build.go
+++ b/env/build.go
@@ -19,7 +19,7 @@ var BuildEnvIncludelist = []string{
 
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
-// NewLaunchEnv returns an build-time Env from the given environ.
+// NewBuildEnv returns an build-time Env from the given environ.
 //
 // Only keys in the BuildEnvIncludelist will be added to the Environment.
 func NewBuildEnv(environ []string) *Env {
--- a/env/build.go
+++ b/env/build.go
@@ -19,7 +19,7 @@ var BuildEnvIncludelist = []string{
 
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
-// NewBuildEnv returns an build-time Env from the given environ.
+// NewBuildEnv returns an build-time Env from the given environment.
 //
 // Only keys in the BuildEnvIncludelist will be added to the Environment.
 func NewBuildEnv(environ []string) *Env {
--- a/env/build.go
+++ b/env/build.go
@@ -25,7 +25,7 @@ var ignoreEnvVarCase = runtime.GOOS == "windows"
 func NewBuildEnv(environ []string) *Env {
 	return &Env{
 		RootDirMap: POSIXBuildEnv,
-		Vars:       varsFromEnviron(environ, ignoreEnvVarCase, isNotIncluded),
+		Vars:       varsFromEnv(environ, ignoreEnvVarCase, isNotIncluded),
 	}
 }
 
--- a/env/build.go
+++ b/env/build.go
@@ -19,6 +19,9 @@ var BuildEnvIncludelist = []string{
 
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
+// NewLaunchEnv returns an build-time Env from the given environ.
+//
+// Only keys in the BuildEnvIncludelist will be added to the Environment.
 func NewBuildEnv(environ []string) *Env {
 	return &Env{
 		RootDirMap: POSIXBuildEnv,
--- a/env/env.go
+++ b/env/env.go
@@ -27,14 +27,14 @@ func (p *Env) AddRootDir(dir string) error {
 		return err
 	}
 	for dir, vars := range p.RootDirMap {
-		newDir := filepath.Join(absDir, dir)
-		if _, err := os.Stat(newDir); os.IsNotExist(err) {
+		childDir := filepath.Join(absDir, dir)
+		if _, err := os.Stat(childDir); os.IsNotExist(err) {
 			continue
 		} else if err != nil {
 			return err
 		}
 		for _, key := range vars {
-			p.Vars.Set(key, newDir+prefix(p.Vars.Get(key), os.PathListSeparator))
+			p.Vars.Set(key, childDir+prefix(p.Vars.Get(key), os.PathListSeparator))
 		}
 	}
 	return nil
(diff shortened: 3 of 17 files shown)


env/build.go
```diff
@@ -19,6 +19,9 @@ var BuildEnvIncludelist = []string{
 
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
+// NewLaunchEnv returns an build-time Env from the given environ.
```
jabrown85, line 22:
```suggestion
// NewBuildEnv returns an build-time Env from the given environ.
```

env/env_test.go
```diff
@@ -102,61 +102,111 @@ func testEnv(t *testing.T, when spec.G, it spec.S) {
...
+					}
+					if s := cmp.Diff(out, expected); s != "" {
+						t.Fatalf("Unexpected env:\n%s\n", s)
+					}
+				})
+			})
+
+			when("default action matches a suffix", func() {
+				it("should preform that action", func() {
+					it("should prepend with pathlist separator or given delimiter", func() {
```
jabrown85, line 190:
```suggestion
					it("should override the value with the new value", func() {
```

env/env_test.go
```diff
@@ -102,61 +102,111 @@ func testEnv(t *testing.T, when spec.G, it spec.S) {
...
+						"VAR_NORMAL_DELIM_NEW=value-normal-delim",
+						"VAR_NORMAL_NEW=value-normal",
+					}
+					if s := cmp.Diff(out, expected); s != "" {
+						t.Fatalf("Unexpected env:\n%s\n", s)
+					}
+				})
+			})
+
+			when("default action matches a suffix", func() {
```
jabrown85, line 188:
```suggestion
			when("default action is Override", func() {
```

launch/launcher_test.go
```diff
@@ -499,25 +505,30 @@ func testLauncher(t *testing.T, when spec.G, it spec.S) {
 						process.Type = "some-process-type"
 					})
 
-					it("should apply type-specific env modifications", func() {
+					it.Focus("should apply type-specific env modifications", func() {
```
jabrown85, line 508:
```suggestion
					it("should apply type-specific env modifications", func() {
```

ekcasey: oops!

jabrown85 (approved):

env/build.go
```diff
@@ -19,6 +19,9 @@ var BuildEnvIncludelist = []string{
 
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
+// NewBuildEnv returns an build-time Env from the given environ.
```
yaelharel, line 22:
```suggestion
// NewBuildEnv returns an build-time Env from the given environment.
```

env/vars.go
```diff
@@ -7,6 +7,21 @@ type Vars struct {
 	ignoreCase bool
 }
 
+func varsFromEnviron(environ []string, ignoreCase bool, removeKey func(string) bool) *Vars {
```
yaelharel, line 10: What do you think about `varsFromEnvionrnment` or `varsFromEnv`?

env/vars.go
```diff
@@ -7,6 +7,21 @@ type Vars struct {
...
+	vars := NewVars(nil, ignoreCase)
+	for _, kv := range environ {
```
yaelharel, line 12: Same comment about `environ`.
I think if we're trying to make it a short name, it's better to use `env`, but personally I'm also fine with `environment`

env/env_test.go
```diff
@@ -102,61 +102,111 @@ func testEnv(t *testing.T, when spec.G, it spec.S) {
...
+						"VAR_NORMAL_NEW=value-normal",
+					}
+					if s := cmp.Diff(out, expected); s != "" {
+						t.Fatalf("Unexpected env:\n%s\n", s)
+					}
+				})
+			})
+
+			when("default action is Override", func() {
+				it("should preform that action", func() {
```
natalieparellano, line 189: Do we need the outer `it`?

natalieparellano (commented):
I made a few suggestions regarding comments and variable names but overall this looks great!

builder.go
```diff
@@ -248,12 +249,12 @@ func (bom *BOMEntry) convertVersionToMetadata() error {
 	return nil
 }
 
-func setupEnv(env BuildEnv, layersDir string) error {
+func (b *Builder) setupEnv(bpAPI *api.Version, layersDir string) error {
```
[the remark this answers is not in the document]
yaelharel: I don't know what's the reason that Emily made this change, but I think it should be like that because this function relates to the `Builder`.

env/env.go
```diff
@@ -48,27 +40,65 @@ func (p *Env) AddRootDir(baseDir string) error {
...
+
+// DefaultActionType returns the default action to preform for an unsuffixed env file as specified for the given
+// buildpack API
+func DefaultActionType(bpAPI *api.Version) ActionType {
+	if bpAPI == nil {
+		return ActionTypeOverride
+	}
+	if bpAPI.Compare(api.MustParse("0.5")) < 0 {
+		return ActionTypePrependPath
+	}
```
yaelharel, line 72:
```suggestion
	if bpAPI != nil &&  bpAPI.Compare(api.MustParse("0.5")) < 0 {
		return ActionTypePrependPath
	}
```

env/env.go
```diff
@@ -48,27 +40,65 @@ func (p *Env) AddRootDir(baseDir string) error {
...
+	return false
+}
+
+type ActionType string
+
+const (
+	ActionTypePrepend     ActionType = "prepend"
+	ActionTypeAppend      ActionType = "append"
+	ActionTypeOverride    ActionType = "override"
```
yaelharel, line 59: Maybe I missed it, but where do we use these three constants?
I see that we're calling `DefaultActionType` in many places, but it returns either `ActionTypeOverride` or `ActionTypePrependPath`.

launch/shell.go
```diff
@@ -90,23 +90,35 @@ func (l *Launcher) isScript(proc Process) (bool, error) {
...
+			}
+			return api, nil
 		}
-		if isLegacyProcess(bpAPI) {
-			return true, nil
-		}
-		return false, nil
 	}
-	return false, fmt.Errorf("process type '%s' provided by unknown buildpack '%s'", proc.Type, proc.BuildpackID)
+	return nil, fmt.Errorf("process type '%s' provided by unknown buildpack '%s'", proc.Type, proc.BuildpackID)
```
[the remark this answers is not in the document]
ekcasey: It really shouldn't happen. We would only end up here if there was a bug in export or somebody copied in an invalid `metadata.toml` at launch.

builder.go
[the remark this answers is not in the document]
ekcasey: Two reasons:
1. What @yaelharel said. This is specific to the build environment so it makes semantic sense to me to put it Builder rather than having it at the package level.
2. Trying to keep the numbers of parameters manageable. If `*Builder` is the receiver we don't need to pass in the `BuildEnv`

If we have to undo this shortly to merge it with your refactor in #461 that is understandable.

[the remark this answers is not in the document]
natalieparellano: Ah yeah, `setupEnv` ended up in `BuildpackTOML.Build()` in my refactor so that prompted the question. But it won't be a big deal to change it later.

env/vars.go
ekcasey: I went with `environ` as a nod to [os.Environ](https://golang.org/pkg/os/#Environ) which returns the environment in the format expected here. I originally avoid `env` because it matches the package name, but on second thought there is no practical problem with that and the linter doesn't seem to mind so maybe it's kosher.

env/env.go
ekcasey: We use them here: https://github.com/buildpacks/lifecycle/pull/483/commits/d58ab15f71c49a6d063dc3d44975c2df73aedb71#diff-1e4c1fcda16b2f94c80249aa8e58bff941c8621fe28458d767fd3502c10c4f31R90-R101

Although that is a within package use and therefore they don't _need_ to be exported, it made sense to me to export them from a package design perspective, if we are exporting the other two. I.E if this package was a standalone entity what would make the most sense? Happy to hear the counter argument though.

I originally considered just passing the buildpack API into `AddEnvDir` and not exporting these at all, but decided on the current design because it is very explicit and maintains a separation of concerns. I could see `DefaultActionType` moving into a `buildpack` package that handles all buildpack API dependent logic in the future.

env/vars.go
yaelharel: ohhh...I wasn't aware of `os.Environ`.
I'm just more used to `env` but if the convention is `environ`, I'm totally ok with that.

env/env.go
yaelharel: I don't know why I ignored the usage in `AddEnvDir`. Probably it was too late :-D
I think the current implementation is great.
Thanks!

env/vars.go
yaelharel: Just saw that you changed it to `env`. Thanks!

env/vars.go
yaelharel: Thanks for making this change!

yaelharel (approved):

env/env.go
```diff
@@ -7,68 +7,95 @@ import (
...
+	ActionTypeDefault     ActionType = "default"
+	ActionTypePrependPath ActionType = ""
+)
+
+// DefaultActionType returns the default action to preform for an unsuffixed env file as specified for the given
```
natalieparellano, line 64:
```suggestion
// DefaultActionType returns the default action to perform for an unsuffixed env file as specified for the given
```

env/env.go
```diff
@@ -7,68 +7,95 @@ import (
...
+func DefaultActionType(bpAPI *api.Version) ActionType {
+	if bpAPI != nil && bpAPI.Compare(api.MustParse("0.5")) < 0 {
+		return ActionTypePrependPath
+	}
+	return ActionTypeOverride
+}
+
+// AddEnvDir modified the Env given a directory containing env files. For each file in the envDir, if the file has
+// a period delimited suffix, the action matching the given suffix will be preformed. If the file has no suffix,
+// the default action will be preformed. If the suffix does not match a known type, AddEnvDir will ignore the file.
```
natalieparellano, line 75:
```suggestion
// a period delimited suffix, the action matching the given suffix will be performed. If the file has no suffix,
// the default action will be performed. If the suffix does not match a known type, AddEnvDir will ignore the file.
```

env/env_test.go
```diff
@@ -102,61 +102,112 @@ func testEnv(t *testing.T, when spec.G, it spec.S) {
...
-			mkfile(t, "[]", filepath.Join(tmpDir, "VAR_PREPEND_DELIM.delim"), filepath.Join(tmpDir, "VAR_PREPEND_DELIM_NEW.delim"))
+				mkfile(t, "value-default", filepath.Join(tmpDir, "VAR_DEFAULT.default"), filepath.Join(tmpDir, "VAR_DEFAULT_NEW.default"))
+				mkfile(t, "value-override", filepath.Join(tmpDir, "VAR_OVERRIDE.override"), filepath.Join(tmpDir, "VAR_OVERRIDE_NEW.override"))
+				mkfile(t, "value-ignore", filepath.Join(tmpDir, "VAR_IGNORE.ignore"))
+			})
 
-			mkfile(t, "value-default", filepath.Join(tmpDir, "VAR_DEFAULT.default"), filepath.Join(tmpDir, "VAR_DEFAULT_NEW.default"))
-			mkfile(t, "value-override", filepath.Join(tmpDir, "VAR_OVERRIDE.override"), filepath.Join(tmpDir, "VAR_OVERRIDE_NEW.override"))
-			mkfile(t, "value-ignore", filepath.Join(tmpDir, "VAR_IGNORE.ignore"))
+			it("preform the matching action", func() {
```
natalieparellano, line 121:
```suggestion
			it("performs the matching action", func() {
```

natalieparellano (approved):

PR #485, opened 2020-12-10
Bumps imgutil to fix daemon image bugs

ekcasey:
Falls back to slower behavior if the daemon does not accept empty layers. This allows the lifecycle to be compatible with podman which requires all layers to be preset during a docker load: buildpacks/imgutil#80

Fixes "manifest.json had unexpected number of entries"  errors in analyze, export, and rebase resulting from a bug in our daemon save logic: buildpacks/imgutil#63

### DIFF
--- a/cmd/lifecycle/exporter.go
+++ b/cmd/lifecycle/exporter.go
@@ -259,6 +259,7 @@ func initDaemonImage(imagName string, runImageRef string, analyzedMD lifecycle.A
 		opts = append(opts, local.WithPreviousImage(analyzedMD.Image.Reference))
 	}
 
+	var appImage imgutil.Image
 	appImage, err := local.NewImage(
 		imagName,
 		docker,
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Microsoft/go-winio v0.4.15-0.20190919025122-fc70bd9a86b5 // indirect
 	github.com/apex/log v1.9.0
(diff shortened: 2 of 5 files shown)


cmd/lifecycle/exporter.go
```diff
@@ -259,6 +259,7 @@ func initDaemonImage(imagName string, runImageRef string, analyzedMD lifecycle.A
 		opts = append(opts, local.WithPreviousImage(analyzedMD.Image.Reference))
 	}
 
+	var appImage imgutil.Image
```
jabrown85, line 262: Is this needed?

jabrown85 (approved):

cmd/lifecycle/exporter.go
ekcasey: Yes because `local.NewImage` now returns a `*local.Image` instead of an `imgutil.Image` type.  Without this, https://github.com/buildpacks/lifecycle/pull/485/files#diff-92c6ed289675df1e2edf598694dfc99f22a591496c130d10a43e86830cbd3686R282, fails because `cache.NewCachingImage` returns an `imgutil.Image` type and you can change the type of the variable without redeclaring it. However, since `*local.Image` implements `imgutil.Image` if you just declare the variable with the interface type everything works fine.

yaelharel: This is very confusing but it looks like this is the way Golang works (because an interface is a pointer to itself).
Thanks for the clarification @ekcasey!

yaelharel (approved):

natalieparellano (approved):

PR #494, opened 2020-12-28
Prepare analyzer to be run before detect by removing the group option

jkutner:
This prepares the `analyzer` to run before the `detector`, which requires removing the `-group` flag. The buildpack group is determined by the detector. Instead, buildpack metadata is read from the previous image's label.


natalieparellano:
@jkutner could we close this one in favor of https://github.com/buildpacks/lifecycle/pull/505?

jkutner:
@natalieparellano yes

PR #498, opened 2021-01-12
Manifest size

### LINKED ISSUE #490: [RFC 0067] Add Image Manifest Size to report.toml
[RFC 0067](https://github.com/buildpacks/rfcs/blob/main/text/0067-report-toml-image-manifest-size.md)
https://github.com/buildpacks/rfcs/pull/121
https://github.com/buildpacks/spec/issues/169

**Given** platform API `0.6` or greater
**When** `creator` `exporter` or `rebaser` succeeds in writing an image to a registry
**Then** `report.toml` should include `image.manifest-size` which reports the size of the manifest in bytes

```
[image]
tags = ["<tag reference>"]
digest = "<image digest>"
image-id = "<imageID>"
manifest-size = <manifest size in bytes>
```

yaelharel:
Add the manifest size to report.toml in the exporter, rebaser and creator

Fixes #490


yaelharel:
This PR depends on https://github.com/buildpacks/imgutil/pull/86.

cmd/lifecycle/exporter.go
```diff
@@ -259,7 +259,9 @@ func initDaemonImage(imagName string, runImageRef string, analyzedMD lifecycle.A
 		opts = append(opts, local.WithPreviousImage(analyzedMD.Image.Reference))
 	}
 
-	appImage, err := local.NewImage(
+	var appImage imgutil.Image
+	var err error
+	appImage, err = local.NewImage(
```
natalieparellano, line 264: I'm a little confused by this diff - this func is now called `initDaemonAppImage` on main: https://github.com/buildpacks/lifecycle/blob/9bb33743a01758f15d6135005adf395dca7fe13b/cmd/lifecycle/exporter.go#L263

yaelharel: This diff is due to some changes in `imgutil` that don't relate to my changes. I just had to bump `imgutil` locally in order to test my code and I got an error on this assignment and it was solved by line 262.
Anyway I'll go ahead and rebase my branch.

yaelharel: I've pushed updated code after rebasing my branch.
I'm going to resolve this conversation.
Please reopen it if something is still unclear.

save.go
```diff
@@ -46,6 +47,15 @@ func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (Im
...
 	}
 
+	manifestSize, sizeErr := image.ManifestSize()
+	if sizeErr != nil {
+		return ImageReport{}, sizeErr
+	}
+	if manifestSize != 0 {
+		imageReport.ManifestSize = strconv.FormatInt(manifestSize, 10)
+		logger.Debugf("\n*** Manifest Size: %d\n", manifestSize)
+	}
```
natalieparellano, line 57: I don't see that we have different logic depending on whether the platform API is 0.6+. Rather than pass api-related logic into `saveImage`, maybe we could have the exporter / rebaser do some post-processing to strip this info out of the report for lower platform APIs. What do you think?

save.go
```diff
@@ -46,6 +47,15 @@ func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (Im
 	default:
 	}
 
+	manifestSize, sizeErr := image.ManifestSize()
+	if sizeErr != nil {
+		return ImageReport{}, sizeErr
```
natalieparellano, line 52: See [comment](https://github.com/buildpacks/lifecycle/pull/498/files?file-filters%5B%5D=.go&file-filters%5B%5D=.mod&file-filters%5B%5D=.sum#r557731001), maybe it makes sense to just return the report that we have instead of failing, so we avoid breaking builds for lower platform APIs in case we fail to get the manifest size for some reason.

save.go
natalieparellano: Similarly to the above, we should strip manifest size from the report for daemon builds.

natalieparellano (changes requested):
See comments, I believe we'll need to add platform API ~~and daemon related logic~~.

save.go
yaelharel: I don't think that lower platform APIs are supposed to fail in this case because the function that `ManifestSize` is using internally isn't new (see https://github.com/buildpacks/imgutil/pull/86/files).
But anyway, I'm fine that instead of failing, we can just print a message and not include the value in the report.

exporter.go
```diff
@@ -72,9 +72,10 @@ type BuildReport struct {
 }
 
 type ImageReport struct {
-	Tags    []string `toml:"tags"`
-	ImageID string   `toml:"image-id,omitempty"`
-	Digest  string   `toml:"digest,omitempty"`
+	Tags         []string `toml:"tags"`
+	ImageID      string   `toml:"image-id,omitempty"`
+	Digest       string   `toml:"digest,omitempty"`
+	ManifestSize string   `toml:",manifest-size,omitempty"`
```
ekcasey, line 78:
```suggestion
	ManifestSize int64   `toml:"manifest-size,omitempty"`
```
Why not just make this an `int64` instead of converting to a string?

save.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/fbf519cd2b6d74906571808c3207936d56eb4757

save.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/fbf519cd2b6d74906571808c3207936d56eb4757

exporter.go
yaelharel: You're totally right :) I'll make this change. Thanks!

natalieparellano: If this is `int64`, will we be able to "unset" it for older platform APIs?

yaelharel: I think that if it's set to `0`, it omitted from `report.toml` thanks to the `omitempty` flag.

exporter_test.go
```diff
@@ -639,6 +639,72 @@ version = "4.5.6"
 				})
 			})
 
+			when("checking the image manifest", func() {
```
natalieparellano, line 642: I wonder if we could pull this test out of the `previous image exists` contexts and put it at the top level, as the functionality being tested doesn't depend on the previous image existing or not.

exporter_test.go
```diff
@@ -639,6 +639,72 @@ version = "4.5.6"
...
+							h.AssertEq(t, report.Image.ManifestSize, int64(0))
+						})
+					})
+				})
+
+				when("platform API is >= 0.6", func() {
+					it.Before(func() {
+						exporter.PlatformAPI = api.MustParse("0.6")
+						exporter.Buildpacks = []lifecycle.GroupBuildpack{
+							// when PlatformAPI is >= 0.5, we need to set Builpacks API
```
natalieparellano, line 665: I think I know why this is true (https://github.com/buildpacks/lifecycle/blob/d9c7de5395c1ba66682b3e548f8955a11b265c17/exporter.go#L442) but maybe it would be better to just set the API when `exporter.Buildpacks` is instantiated, so that we don't have to remember to do it for every future test?

yaelharel: Yeah, the line that you posted is the reason. Didn't want to indicate the specific function.
But I agree that it's better to move it to `it.Before` of these unit tests.

exporter_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/e32d468109f145d40f80ec070f7300d3f807ed80

exporter_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/e32d468109f145d40f80ec070f7300d3f807ed80

exporter.go
yaelharel: I added a print of `report.toml` after it's written and you were right - `manifest-size` is printed even if it's `0` and `PlatformAPI=0.4`:
```
Setting CNB_PLATFORM_API=0.4
...
content: [build]
[image]
  tags = ["yaelharel/test-app:my-tag"]
  digest = "sha256:b5a675383bcec612389b81bd5f70772adb452706d014965c19685e94480c77e4"
  manifest-size = 0
```

This is weird because the following code outputs `{}`:
```
package main
import (
	"encoding/json"
	"fmt"
)
type MyStr struct {
	Num int64 `json:",omitempty"`
}
func main() {
	x := MyStr{
		Num : 0,
	}
	j, _ := json.Marshal(x)
	fmt.Println(string(j))
}
```

Maybe `Marshal` works differently than `lifecycle.WriteTOML`.

yaelharel: So first I worked with `json` instead of `toml`, but the way to solve it is with `omitzero` instead of `omitempty`.

yaelharel: With `omitzero`:
```
Setting CNB_PLATFORM_API=0.4
...
*** Digest: sha256:c0885c0619709bfc53731570359dd553fb6e36d2493d757d51f304082ba80616

*** Manifest Size: 3185
content: [build]

[image]
  tags = ["yaelharel/test-app2:my-tag2"]
  digest = "sha256:c0885c0619709bfc53731570359dd553fb6e36d2493d757d51f304082ba80616"
```

In this example `manifest-size` doesn't exist in `report.toml` since `PlatformAPI` is `0.4`.

yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/069147f1c476abbe01ee8c1b9a044d34ac8cafa1

save.go
```diff
@@ -46,6 +46,15 @@ func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (Im
 	default:
 	}
 
+	manifestSize, sizeErr := image.ManifestSize()
+	if sizeErr != nil {
+		// ignore the manifest size if it's unavailable
+		logger.Infof("*** Manifest size is unavailable (%s):\n", sizeErr.Error())
```
natalieparellano, line 52:
```suggestion
		logger.Infof("*** Manifest size is unavailable: %s\n", sizeErr.Error())
```

rebaser_test.go
```diff
@@ -54,7 +55,8 @@ func testRebaser(t *testing.T, when spec.G, it spec.S) {
 		additionalNames = []string{"some-repo/app-image:foo", "some-repo/app-image:bar"}
 
 		rebaser = &lifecycle.Rebaser{
-			Logger: &log.Logger{Handler: &discard.Handler{}},
+			Logger:      &log.Logger{Handler: &discard.Handler{}},
+			PlatformAPI: api.MustParse("0.4"),
```
natalieparellano, line 59: Could we use something like `api.Platform.Latest()` here? The tests that require the API version to be lower than some boundary could then set it explicitly, otherwise it would just be latest. This is a pattern that we started here: https://github.com/buildpacks/lifecycle/blob/d9c7de5395c1ba66682b3e548f8955a11b265c17/builder_test.go#L67 and should be helpful as we support ever more api versions.

tools/go.mod
```diff
@@ -3,16 +3,24 @@ module github.com/buildpacks/lifecycle/tools
 go 1.15
 
 require (
+	github.com/Azure/azure-sdk-for-go v42.3.0+incompatible // indirect
+	github.com/Azure/go-autorest/autorest v0.10.2 // indirect
+	github.com/Azure/go-autorest/autorest/validation v0.2.0 // indirect
```
natalieparellano, line 8: I saw these changes too when updating the go.mod and I'm not sure what to make of them. Eventually I was able to get the go.mod to have only the diff that I was expecting (imgutil + ggcr) but I don't understand what happened. I feel like, maybe we should investigate, or maybe it doesn't matter - could someone who understands go.mod a bit better opine here? cc @ekcasey

exporter_test.go
```diff
@@ -134,8 +134,8 @@ func testExporter(t *testing.T, when spec.G, it spec.S) {
 
 		exporter = &lifecycle.Exporter{
 			Buildpacks: []lifecycle.GroupBuildpack{
-				{ID: "buildpack.id", Version: "1.2.3"},
-				{ID: "other.buildpack.id", Version: "4.5.6", Optional: false},
+				{ID: "buildpack.id", Version: "1.2.3", API: "0.1"},
+				{ID: "other.buildpack.id", Version: "4.5.6", API: "0.1", Optional: false},
```
natalieparellano, line 138: I still feel like we should set these values here: https://github.com/buildpacks/lifecycle/blob/d9c7de5395c1ba66682b3e548f8955a11b265c17/exporter_test.go#L137-L138, so we won't have to remember to set the api for every test case.

exporter_test.go
```diff
@@ -1116,6 +1074,116 @@ version = "4.5.6"
 			})
 		})
 
+		when("checking the image manifest", func() {
```
natalieparellano, line 1077: Would it make sense to wrap these in a `when("report.toml", ...` like here: https://github.com/buildpacks/lifecycle/blob/d9c7de5395c1ba66682b3e548f8955a11b265c17/exporter_test.go#L1119?

exporter_test.go
```diff
@@ -613,48 +613,6 @@ version = "4.5.6"
 				h.AssertEq(t, fakeAppImage.IsSaved(), true)
 			})
 
-			when("image has a digest identifier", func() {
```
natalieparellano, line 616: Pulled these out of the `when("previous image exists", ...` context since the assertions aren't dependent on a previous image.

rebaser_test.go
yaelharel: Good idea.
I'll make this change both here and in `exporter_test.go`.

exporter_test.go
natalieparellano, line 138: Could we try using the latest buildpack api here too? `0.1` feels a little strange, because it's not a supported version. I think in practice the version doesn't actually matter, except for this test: https://github.com/buildpacks/lifecycle/blob/d9c7de5395c1ba66682b3e548f8955a11b265c17/exporter_test.go#L1124-L1125.

go.mod
```diff
@@ -4,11 +4,12 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Microsoft/go-winio v0.4.15-0.20190919025122-fc70bd9a86b5 // indirect
 	github.com/apex/log v1.9.0
-	github.com/buildpacks/imgutil v0.0.0-20201211223552-8581300fe2b2
+	github.com/buildpacks/imgutil v0.0.0-20210115182929-e2b7b1a5467a
 	github.com/containerd/containerd v1.3.3 // indirect
 	github.com/docker/cli v0.0.0-20200312141509-ef2f64abbd37 // indirect
 	github.com/docker/docker v1.4.2-0.20190924003213-a8608b5b67c7
 	github.com/golang/mock v1.4.4
+	github.com/golang/protobuf v1.4.2 // indirect
```
natalieparellano, line 12: Same comment as https://github.com/buildpacks/lifecycle/pull/498/files#r561087744.

natalieparellano:
This is looking good! I added a couple nits and a couple questions.

go.mod
yaelharel: Going to remove it.

exporter_test.go
yaelharel: Exactly. It doesn't matter. We just have to set it in order to not fail [here](https://github.com/buildpacks/lifecycle/blob/d9c7de5395c1ba66682b3e548f8955a11b265c17/exporter.go#L442).
I thought about setting it to `0.1` like the version is set to `1.2.3` or `4.5.6` although these are just random versions.
I can change it to something else.
@natalieparellano, what do you think?

tools/go.mod
yaelharel: I was able to clean `go.mod` as well.

save.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/721cf13f78654cd7978864642557e02ce6a9931d

rebaser_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/721cf13f78654cd7978864642557e02ce6a9931d

exporter_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/721cf13f78654cd7978864642557e02ce6a9931d

go.mod
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/721cf13f78654cd7978864642557e02ce6a9931d

exporter_test.go
natalieparellano: To me `0.1` reads like an old version of the api rather than a random value, which is why I'd lean towards just having it be the latest. Then we could remove the lines that set it to `0.5` in the test that I linked to.

yaelharel: I don't have strong opinion on either approach so I'll make this change.

yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498/commits/99f51baecde6a658cb9f5ee0651d6870b5cdeb5b

jabrown85 (approved):
LGTM - hopefully we can rebase onto https://github.com/buildpacks/lifecycle/pull/506 to remove all the go mod churn.

yaelharel:
@jabrown85, sure. Maybe we can merge #506 first and then I'll rebase onto main.
cc @natalieparellano

natalieparellano:
#506 is merged now cc @yaelharel

natalieparellano (approved):

PR #505, opened 2021-01-19
Run analyzer before detector

jabrown85:
- Run `analyzer` before `detector`
- Move `analyzer` logic that depends on `group.toml` to `restorer`

Outstanding Questions:
- `creator` has a `SkipRestore` the entirely skips `restorer`. Should this continue to do that? Or should that flag perhaps mean we only skip the pre-0.6 `restorer` behavior?

[Related RFC](https://github.com/buildpacks/rfcs/pull/135)
[Draft Spec PR](https://github.com/buildpacks/spec/pull/197)


acceptance/analyzer_test.go
```diff
@@ -112,6 +112,62 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 
+	when("Platform API = 0.6", func() {
```
natalieparellano, line 115:
```suggestion
	when("Platform API > 0.5", func() {
```
(for consistency with other context)

analyzer_test.go
```diff
@@ -53,9 +54,10 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		h.AssertNil(t, err)
 
 		analyzer = &lifecycle.Analyzer{
-			Buildpacks: []lifecycle.GroupBuildpack{{ID: "metadata.buildpack"}, {ID: "no.cache.buildpack"}, {ID: "no.metadata.buildpack"}},
-			LayersDir:  layerDir,
-			Logger:     &log.Logger{Handler: &discard.Handler{}},
+			Buildpacks:  []lifecycle.GroupBuildpack{{ID: "metadata.buildpack"}, {ID: "no.cache.buildpack"}, {ID: "no.metadata.buildpack"}},
+			LayersDir:   layerDir,
+			Logger:      &log.Logger{Handler: &discard.Handler{}},
+			PlatformAPI: api.MustParse("0.5"),
```
natalieparellano, line 60: A pattern we've been trying to follow in tests of other phases (build?) is to have the "default" platform API version be the latest one, and have separate contexts for the earlier platform APIs. In theory, this would make it easier to drop support for earlier APIs if/when the time comes.

restorer.go
```diff
@@ -87,3 +114,85 @@ func (r *Restorer) restoreLayer(cache Cache, sha string) error {
 
 	return layers.Extract(rc, "")
 }
+
+func (r *Restorer) analyzeLayers(appMeta LayersMetadata, meta CacheMetadata) error {
```
natalieparellano, line 118: It seems this is duplicating a fair amount of logic in analyzer.go. Is there a way to extract it into a function that could be shared by both phases? (That said, maybe there's value in keeping the code separate, so that it could more easily diverge?)

restorer.go
```diff
@@ -1,34 +1,50 @@
...
+	Buildpacks  []GroupBuildpack
+	Logger      Logger
+	SkipLayers  bool
+	PlatformAPI *api.Version
 }
 
 // Restore attempts to restore layer data for cache=true layers, removing the layer when unsuccessful.
 // If a usable cache is not provided, Restore will remove all cache=true layer metadata.
-func (r *Restorer) Restore(cache Cache) error {
+func (r *Restorer) Restore(img imgutil.Image, cache Cache) error {
```
natalieparellano, line 27: Where does the image come from for earlier platform APIs?

restorer_test.go
```diff
@@ -320,11 +331,12 @@ func testRestorer(t *testing.T, when spec.G, it spec.S) {
 			when("there is a cache=true escaped layer", func() {
 				var meta string
 				it.Before(func() {
+					restorer.PlatformAPI = api.MustParse("0.5")
```
natalieparellano, line 334: This might be more readable if wrapped in a context like "platform API < 0.6"

restorer_test.go
```diff
@@ -372,14 +384,15 @@ func testRestorer(t *testing.T, when spec.G, it spec.S) {
 
 			when("there are multiple cache=true layers", func() {
 				it.Before(func() {
+					restorer.PlatformAPI = api.MustParse("0.5")
```
natalieparellano, line 387: This might be more readable if wrapped in a context like "platform API < 0.6"

cmd/lifecycle/restorer.go
```diff
@@ -32,12 +47,27 @@ func (r *restoreCmd) DefineFlags() {
...
+	if api.MustParse(r.platformAPI).Compare(api.MustParse("0.6")) >= 0 { // platform API >= 0.6
+		cmd.FlagUseDaemon(&r.useDaemon)
+		cmd.FlagSkipLayers(&r.skipLayers)
+	}
 }
 
 func (r *restoreCmd) Args(nargs int, args []string) error {
-	if nargs > 0 {
-		return cmd.FailErrCode(errors.New("received unexpected Args"), cmd.CodeInvalidArgs, "parse arguments")
+	if api.MustParse(r.platformAPI).Compare(api.MustParse("0.6")) < 0 { // platform API < 0.6
```
natalieparellano, line 57: Not to throw a wrench in this feature, but given the complexity involved I wonder if it's worth trying to add a few high-level acceptance tests around restorer.

natalieparellano (commented):
This is exciting work! I added a few comments to try to understand it better.

restorer.go
jabrown85: This logic actually _moved_ from analyzer to restorer. Analyzer is currently using this same functions

restorer.go
jabrown85: it would be `nil` and unused in the earlier platform API code path. I don't like it and would love to get feedback on a better path forward. You can see in `creator` a similar ugly pattern emerging with `analyze(lifecycle.BuildpackGroup{}, nil)` in one code path of the `analyzer` execution.

cmd/lifecycle/restorer.go
jabrown85: Yeah - I was sort of afraid someone might say that 😬 

I don't mind throwing some time at some acceptance tests. I may attempt to do it on a fresh branch off of main and then bring it in to this branch once we are happy. Just in case it fails, I don't want it clogging up this work.

restorer.go
natalieparellano: I do wonder if there's a way this could be architected differently to make it a bit cleaner. Thinking...

cmd/lifecycle/restorer.go
natalieparellano: So happy to see https://github.com/buildpacks/lifecycle/pull/512 :)

jabrown85:
I rebased this on `main` 😅  - I think we are ready for a real review. We probably should also discuss with platform about implications for `pack`/`tekton` and how we can communicate with other platforms so they can be ready for the switch.

jkutner:
@jabrown85 is it possible that platforms like tekton could run analyzer before _and_ after detector in order to support multiple compat versions? Are there any consequences of analyzer running twice (or analyzer and restorer running one after the other)?

jkutner (approved):

jabrown85:
> @jabrown85 is it possible that platforms like tekton could run analyzer before _and_ after detector in order to support multiple compat versions? Are there any consequences of analyzer running twice (or analyzer and restorer running one after the other)?

I have not tested that but I'm pretty sure it would NOT work. I'm not sure we need it though. The `CNB_PLATFORM_API` is set at the platform level, not at the builder level. I don't think a single tekton template to support multiple platform versions is needed today.

The templates today are pinned to `0.3`
https://github.com/buildpacks/tekton-integration/blob/main/task/buildpacks/0.3/buildpacks.yaml#L63-L64
https://github.com/buildpacks/tekton-integration/blob/main/task/buildpacks-phases/0.2/buildpacks-phases.yaml#L56-L57

I guess the question of should we expect platforms to be able to support multiple platform versions simultaneously should be answered though.

jkutner:
@jabrown85 I don't think we should _expect_ platform support multiple versions. And there's always a path that's forking (i.e. it runs a completely different pipeline for the lifecycle).

restorer_test.go
```diff
@@ -238,28 +248,34 @@ func testRestorer(t *testing.T, when spec.G, it spec.S) {
...
 
-			when("there is a cache=true layer", func() {
-				var meta string
+			when("platform API < 0.6", func() {
 				it.Before(func() {
-					meta = `cache=true
+					restorer.PlatformAPI = api.MustParse("0.5")
+				})
+
+				when("there is a cache=true layer", func() {
```
natalieparellano, line 256: Does it make sense to also have an assertion for platform api 0.6+?

restorer_test.go
```diff
@@ -238,28 +248,34 @@ func testRestorer(t *testing.T, when spec.G, it spec.S) {
 				h.AssertNil(t, os.RemoveAll(tarTempDir))
 			})
 
-			when("there is a cache=true layer", func() {
-				var meta string
+			when("platform API < 0.6", func() {
```
natalieparellano, line 251: Does it make sense to group `platform API < 0.6` assertions together?

analyzer.go
```diff
@@ -33,88 +31,32 @@ func (a *Analyzer) Analyze(image imgutil.Image, cache Cache) (platform.AnalyzedM
 		appMeta = platform.LayersMetadata{}
 	}
 
-	for _, bp := range a.Buildpacks {
-		if store := appMeta.MetadataForBuildpack(bp.ID).Store; store != nil {
-			if err := WriteTOML(filepath.Join(a.LayersDir, launch.EscapeID(bp.ID), "store.toml"), store); err != nil {
-				return platform.AnalyzedMetadata{}, err
-			}
+	if a.PlatformAPI.Compare(api.MustParse("0.6")) < 0 { // platform API < 0.6
```
natalieparellano, line 34:
```suggestion
	if a.analyzeLayers() // returns true or false depending on the platform api
```

cmd/lifecycle/restorer.go
```diff
@@ -33,12 +48,27 @@ func (r *restoreCmd) DefineFlags() {
 	cmd.FlagLayersDir(&r.layersDir)
 	cmd.FlagUID(&r.uid)
 	cmd.FlagGID(&r.gid)
+	if api.MustParse(r.platformAPI).Compare(api.MustParse("0.6")) >= 0 { // platform API >= 0.6
```
natalieparellano, line 51:
```suggestion
	if r.analyzeLayers() {
```

cmd/lifecycle/restorer.go
```diff
@@ -33,12 +48,27 @@ func (r *restoreCmd) DefineFlags() {
...
+		cmd.FlagUseDaemon(&r.useDaemon)
+		cmd.FlagSkipLayers(&r.skipLayers)
+	}
 }
 
 func (r *restoreCmd) Args(nargs int, args []string) error {
-	if nargs > 0 {
-		return cmd.FailErrCode(errors.New("received unexpected Args"), cmd.CodeInvalidArgs, "parse arguments")
+	if api.MustParse(r.platformAPI).Compare(api.MustParse("0.6")) < 0 { // platform API < 0.6
```
natalieparellano, line 58:
```suggestion
	if !r.analyzeLayers() {
```

cmd/lifecycle/restorer.go
```diff
@@ -57,6 +87,16 @@ func (r *restoreCmd) Privileges() error {
 		return cmd.FailErr(err, "resolve keychain")
 	}
 
+	if api.MustParse(r.platformAPI).Compare(api.MustParse("0.6")) >= 0 { // platform API >= 0.6
```
natalieparellano, line 90:
```suggestion
	if r.analyzeLayers() {
```

restorer.go
```diff
@@ -1,36 +1,53 @@
...
-			r.Logger.Info("Layer cache not found")
+	meta, err := r.retrieveMetadataFrom(cache)
+	if err != nil {
+		return err
+	}
+
+	if r.PlatformAPI.Compare(api.MustParse("0.6")) >= 0 { // platform API >= 0.6
+		var appMeta platform.LayersMetadata
+		// continue even if the label cannot be decoded
+		if err := DecodeLabel(img, platform.LayerMetadataLabel, &appMeta); err != nil {
```
natalieparellano, line 40: Don't we already do this in analyzer?

restorer.go
natalieparellano: Would there be value in creating a `lifecycle.LayerAnalyzer`?

natalieparellano (commented):
@jabrown85 I have to say that I am still getting my head around this, but leaving some comments in the meantime!

restorer.go
jabrown85: That is a good point. I guess we should read this from `analyzed.toml` here yeah?

analyzer.go
```diff
@@ -1,22 +1,20 @@
...
 
 type Analyzer struct {
-	Buildpacks []buildpack.GroupBuildpack
-	LayersDir  string
-	Logger     Logger
-	SkipLayers bool
+	Buildpacks  []buildpack.GroupBuildpack
+	LayersDir   string
+	Logger      Logger
+	SkipLayers  bool
```
ekcasey, line 16: Maybe we should move this into a separate grouping and add a doc comment that makes it clear, this is only for older APIs?

analyzer.go
```diff
@@ -1,22 +1,20 @@
...
+	PlatformAPI *api.Version
 }
 
 // Analyze restores metadata for launch and cache layers into the layers directory.
```
ekcasey, line 20: We should update this comment to describe the new platform API dependency behavior.

ekcasey (commented):
I haven't gone through all the details yet but I have two high level concerns:

1. This is complicated enough that it would be really helpful to at least have a draft spec PR that lists the new expected flags for analyzer. I have some questions about the high level plan. E.g:
    * To enable us to validate cache compatibility in the future and in the interest of checking registry access as early as possible, should we be writing the digest reference and cache metadata from a cache image to a file during analyze for use by the restorer later?
    * While we are making this changes, should we check that we have write access to the destination tag(s). Even if we don't want to do that _now_ it would change decisions we make about the usage (we may want to accept the previous image using an optional --previous-image flag like `creator` does. We may not want to require a destination image at all for the time being but later we could default the previous image to the destination image)

1. All the conditional make the code a bit hard to read. Also, this interfaces which accommodate both APIs seem a bit weird. If I was using the lifecycle as a library it would be difficult (I need to know to pass a nil cache to analyzer, etc.). It's been a long time coming but it seems like the differences between APIs is finally large enough to require a rearchitecture?

analyzer.go
```diff
@@ -33,18 +32,15 @@ func (a *Analyzer) Analyze(image imgutil.Image, cache Cache) (platform.AnalyzedM
 		appMeta = platform.LayersMetadata{}
 	}
 
-	for _, bp := range a.Buildpacks {
-		if store := appMeta.MetadataForBuildpack(bp.ID).Store; store != nil {
-			if err := WriteTOML(filepath.Join(a.LayersDir, launch.EscapeID(bp.ID), "store.toml"), store); err != nil {
-				return platform.AnalyzedMetadata{}, err
-			}
+	if a.analyzeLayers() {
+		meta, err := a.LayerAnalyzer.RetrieveMetadataFrom(cache)
```
natalieparellano, line 36:
```suggestion
		cacheMeta, err := a.LayerAnalyzer.RetrieveMetadataFrom(cache)
```

analyzer.go
```diff
@@ -33,18 +32,15 @@ func (a *Analyzer) Analyze(image imgutil.Image, cache Cache) (platform.AnalyzedM
...
+		if err != nil {
+			return platform.AnalyzedMetadata{}, err
 		}
-	}
 
-	if a.SkipLayers {
-		a.Logger.Infof("Skipping buildpack layer analysis")
-	} else if err := a.analyzeLayers(appMeta, cache); err != nil {
-		return platform.AnalyzedMetadata{}, err
+		if err := a.LayerAnalyzer.Analyze(a.Buildpacks, a.SkipLayers, appMeta, meta); err != nil {
```
natalieparellano, line 41:
```suggestion
		if err := a.LayerAnalyzer.Analyze(a.Buildpacks, a.SkipLayers, appMeta, cacheMeta); err != nil {
```

analyzer.go
natalieparellano: Could `LayerAnalyzer.Analyze` return the cache metadata, and analyzer just ignores it? I feel like this would read better than having `RetrieveMetadataFrom` as a separate call, unless I'm missing something.

restorer.go
```diff
@@ -4,33 +4,43 @@ import (
...
 // If a usable cache is not provided, Restore will remove all cache=true layer metadata.
-func (r *Restorer) Restore(cache Cache) error {
+func (r *Restorer) Restore(img imgutil.Image, cache Cache) error {
 	// Create empty cache metadata in case a usable cache is not provided.
-	var meta platform.CacheMetadata
-	if cache != nil {
-		var err error
-		if !cache.Exists() {
-			r.Logger.Info("Layer cache not found")
+	meta, err := r.LayerAnalyzer.RetrieveMetadataFrom(cache)
```
natalieparellano, line 29: This feels a bit weird to me in that if `!r.analyzesLayers()` I'd expect the `LayerAnalyzer` not to be used... could it be broken out as a separate responsibility?

detector_test.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 //go:generate mockgen -package testmock -destination testmock/resolver.go github.com/buildpacks/lifecycle Resolver
+//go:generate mockgen -package testmock -destination testmock/layer_analyzer.go github.com/buildpacks/lifecycle LayerAnalyzer
```
natalieparellano, line 25: Does this belong here?

analyzer_test.go
```diff
@@ -198,263 +122,111 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
+							h.AssertEq(t, md.Image.Reference, "s0m3D1g3sT")
+							h.AssertEq(t, md.Metadata, appImageMetadata)
+						})
+					})
 				})
 			})
 		})
 
-		when("image is not found", func() {
+		when("platform API < 0.6", func() {
```
natalieparellano, line 157: I propose moving all `when`s concerning earlier platform apis to the bottom of the file, it would make it easier to understand what is specific to which api.

analyzer_test.go
```diff
@@ -94,78 +98,11 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
-					{"metadata.buildpack/launch-build-cache.sha", "launch-build-cache-sha"},
-					{"metadata.buildpack/launch-cache.sha", "launch-cache-sha"},
-					{"no.cache.buildpack/some-layer.sha", "some-layer-sha"},
-				} {
-					got := h.MustReadFile(t, filepath.Join(layerDir, data.name))
-					h.AssertStringContains(t, string(got), data.want)
-				}
-			})
-
-			it("does not restore launch=false layer metadata", func() {
```
natalieparellano, line 132: Where did this assertion move to?

analyzer_test.go
```diff
@@ -94,78 +98,11 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
-				_, err := analyzer.Analyze(image, testCache)
-				h.AssertNil(t, err)
-
-				h.AssertPathDoesNotExist(t, filepath.Join(layerDir, "metadata.buildpack", "launch-false.toml"))
-				h.AssertPathDoesNotExist(t, filepath.Join(layerDir, "metadata.buildpack", "launch-false.sha"))
-			})
-
-			it("does not restore build=true, cache=false layer metadata", func() {
```
natalieparellano, line 140: Same question as https://github.com/buildpacks/lifecycle/pull/505/files#r582333920

analyzer_test.go
```diff
@@ -94,78 +98,11 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 
 		when("image exists", func() {
 			it.Before(func() {
-				metadata := h.MustReadFile(t, filepath.Join("testdata", "analyzer", "app_metadata.json"))
+				metadata := h.MustReadFile(t, filepath.Join("testdata", "restorer", "app_metadata.json"))
 				h.AssertNil(t, image.SetLabel("io.buildpacks.lifecycle.metadata", string(metadata)))
 				h.AssertNil(t, json.Unmarshal(metadata, &appImageMetadata))
 			})
 
-			it("restores layer metadata", func() {
```
natalieparellano, line 102: Same question as https://github.com/buildpacks/lifecycle/pull/505/files#r582333920

analyzer_test.go
```diff
@@ -94,78 +98,11 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
-					{"metadata.buildpack/launch-build-cache.toml", "[metadata]\n  launch-build-cache-key = \"launch-build-cache-value\""},
-					{"metadata.buildpack/launch-cache.toml", "[metadata]\n  launch-cache-key = \"launch-cache-value\""},
-					{"no.cache.buildpack/some-layer.toml", "[metadata]\n  some-layer-key = \"some-layer-value\""},
-				} {
-					got := h.MustReadFile(t, filepath.Join(layerDir, data.name))
-					h.AssertStringContains(t, string(got), data.want)
-				}
-			})
-
-			it("restores layer sha files", func() {
```
natalieparellano, line 117: Same question as https://github.com/buildpacks/lifecycle/pull/505/files#r582333920

analyzer_test.go
```diff
@@ -198,263 +122,111 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
-					metadata := h.MustReadFile(t, filepath.Join("testdata", "analyzer", "cache_metadata.json"))
-					var cacheMetadata platform.CacheMetadata
-					h.AssertNil(t, json.Unmarshal(metadata, &cacheMetadata))
-					h.AssertNil(t, testCache.SetMetadata(cacheMetadata))
-					h.AssertNil(t, testCache.Commit())
-
-					analyzer.Buildpacks = append(analyzer.Buildpacks, buildpack.GroupBuildpack{ID: "escaped/buildpack/id"})
-				})
-
-				it("restores cache=true layer metadata", func() {
```
natalieparellano, line 390: Where did this (and the following two) move to?

analyzer_test.go
jabrown85: > restores app and cache layer sha files, prefers app sha
I think that covers this. It was asserting the same files and contents

analyzer_test.go
jabrown85: > restores app and cache layer metadata
I _think_ that covers the same situation.

layer_analyzer_test.go
```diff
@@ -0,0 +1,303 @@
...
+				h.AssertNil(t, err)
+
+				path := filepath.Join(layerDir, "escaped_buildpack_id", "escaped-bp-layer.toml")
+				got := h.MustReadFile(t, path)
+				want := "[metadata]\n  escaped-bp-layer-key = \"escaped-bp-layer-value\""
+
+				h.AssertStringContains(t, string(got), want)
+			})
+
+			it("restores each store metadata", func() {
```
natalieparellano, line 214: Is this a new assertion?

analyzer_test.go
jabrown85: > restores app and cache layer metadata
That test covers this file already

detector_test.go
jabrown85: I'll be honest, I only put it here because the other one was there. Where _should_ this go? I haven't used `mockgen` before this project. Should I pop this on `restorer_test.go` instead?

metadata_retriever.go
```diff
@@ -0,0 +1,40 @@
...
+		var err error
+		if !cache.Exists() {
+			mr.Logger.Info("Layer cache not found")
+		}
+		cacheMeta, err = cache.RetrieveMetadata()
+		if err != nil {
+			return cacheMeta, errors.Wrap(err, "retrieving cache metadata")
+		}
+	} else {
+		mr.Logger.Debug("Usable cache not provided, using empty cache metadata.")
```
natalieparellano, line 36:
```suggestion
		mr.Logger.Debug("Usable cache not provided, using empty cache metadata")
```

layer_analyzer_test.go
jabrown85: Yeah, sort of. It's really a continuation of existing tests, but there wasn't one for exactly this situation. It might've been covered by other tests, but I figured the explicit test would be an improvement.

metadata_retriever.go
natalieparellano: (for consistency)

jabrown85: It did have the period before. I don't mind removing it tho!

wburningham:
Just checking in on this PR. What is the status or next steps?

natalieparellano:
@wburningham I believe we're waiting for this spec PR to work its way through: https://github.com/buildpacks/spec/pull/197

analyzer_test.go
```diff
@@ -387,78 +232,278 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
 
-					files, err := ioutil.ReadDir(layerDir)
-					h.AssertNil(t, err)
-					h.AssertEq(t, len(files), 0)
+		when("platform API < 0.6", func() {
+			it.Before(func() {
+				analyzer.PlatformAPI = api.MustParse("0.5")
+			})
+
+			when("image is not found", func() {
```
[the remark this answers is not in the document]
natalieparellano: IIUC this will go in https://github.com/buildpacks/spec/issues/206

natalieparellano:
Verified with Tekton that when `CNB_PLATFORM_API=0.6` a dev lifecycle from this branch will run analyze, then detect (for both creator & 5 phases) and that the second build will restore metadata from app & cache, data from cache, and will reuse app layers. Environment is still around if there are other flows we want to test or points to verify.

acceptance/analyzer_test.go
```diff
@@ -217,369 +220,92 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
-						h.AssertMatch(t, output, "2222 3333 .+ staging")
-					})
+					h.AssertPathExists(t, filepath.Join(copyDir, "layers", "some-other-buildpack-id"))
+					h.AssertPathDoesNotExist(t, filepath.Join(copyDir, "layers", "some-buildpack-id"))
 				})
 			})
-		})
-	})
-
-	when("registry case", func() {
```
natalieparellano, line 478: Where did these move to?

cmd/lifecycle/analyzer.go
```diff
@@ -45,25 +48,36 @@ type analyzeArgs struct {
 
 func (a *analyzeCmd) DefineFlags() {
 	cmd.FlagAnalyzedPath(&a.analyzedPath)
-	cmd.FlagCacheDir(&a.cacheDir)
 	cmd.FlagCacheImage(&a.cacheImageTag)
-	cmd.FlagGroupPath(&a.groupPath)
+	if a.analyzeLayers() {
+		cmd.FlagCacheDir(&a.cacheDir)
```
natalieparellano, line 53: Should `cmd.FlagCacheImage(&a.cacheImageTag)` go here? The spec PR removes `-cache-image` as a flag.

acceptance/analyzer_test.go
```diff
@@ -70,7 +71,10 @@ func TestAnalyzer(t *testing.T) {
 	h.DockerBuild(t,
 		analyzeImage,
 		analyzeDockerContext,
-		h.WithFlags("-f", filepath.Join(analyzeDockerContext, variables.DockerfileName)),
+		h.WithFlags(
+			"--build-arg", fmt.Sprintf("cnb_platform_api=%s", api.Platform.Latest()),
```
natalieparellano, line 75: Now that we have acceptance tests for the restorer, I think it makes sense to test the latest platform api there was well. IIUC the restorer acceptance tests are exercising platform api 0.3 (since we don't set `CNB_PLATFORM_API` in the image). We'll probably want to update them so that they test 0.5 and "latest".

cmd/lifecycle/analyzer.go
jabrown85: Yes, you are correct. I'll make that change

yaelharel:
HI @jabrown85, I saw that you created a [new PR](https://github.com/buildpacks/lifecycle/pull/582) for this issue.
Can we draft/close this PR?

jabrown85:
Sure @yaelharel! I'll draft it until we pull the code for the tests, which haven't yet been migrated.

jabrown85:
Close in favor of #582

Issue #504, opened 2021-01-19
Pin to older golang.org/x/sys


natalieparellano:
Note, we'll need to do this for imgutil also.

### Background

After upgrading to ggcr 0.4.0, the lifecycle and imgutil are failing to compile with the following error:
`errors in package: [/Users/pivotal/workspace/imgutil/local/local.go:21:2: could not import github.com/docker/docker/pkg/jsonmessage (/Users/pivotal/go/pkg/mod/github.com/docker/docker@v1.4.2-0.20190924003213-a8608b5b67c7/pkg/jsonmessage/jsonmessage.go:10:2: could not import github.com/docker/docker/pkg/term (/Users/pivotal/go/pkg/mod/github.com/docker/docker@v1.4.2-0.20190924003213-a8608b5b67c7/pkg/term/tc.go:13:33: SYS_IOCTL not declared by package unix))]`

We believe this is because ggcr 0.4.0 does not reference `golang.org/x/sys` as `// indirect` (as it did previously): https://github.com/google/go-containerregistry/pull/904/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L38

Thus with ggcr 0.4.0 the version of `golang.org/x/sys` is newer: `golang.org/x/sys v0.0.0-20200930185726-fdedc70b468f` as opposed to `golang.org/x/sys v0.0.0-20200523222454-059865788121`. It appears that in moving from go 1.11 to go 1.13 the newer `golang.org/x/sys` removed the `SYS_IOCTL` declaration for darwin. Related go issue: https://go-review.googlesource.com/c/sys/+/265857/

### Fix

Propose adding `replace (
    golang.org/x/sys => golang.org/x/sys v0.0.0-20200523222454-059865788121
)` to our go.mod to pin to a specific version for now.

### Longer term

It looks like the `SYS_IOCTL` removal is in the process of being reverted.

The `pkg/term` codepath that is currently failing is deprecated in the latest version of docker - perhaps upgrading to a newer docker version would resolve this issue. However we wish to stay pinned to the same version of docker that ggcr is using. We should inquire if ggcr intends to upgrade docker.

PR #506, opened 2021-01-19
Pin to older version of golang.org/x/sys

natalieparellano:
Fixes https://github.com/buildpacks/lifecycle/issues/504


jabrown85 (approved):

natalieparellano:
I'm confused... I don't see my go.mod change. Also this should probably include the bump to imgutil that has the same fix.

yaelharel (approved):

tools/go.mod
```diff
@@ -4,12 +4,12 @@ go 1.15
 
 require (
 	github.com/BurntSushi/toml v0.3.1
-	github.com/buildpacks/imgutil v0.0.0-20201211223552-8581300fe2b2
+	github.com/buildpacks/imgutil v0.0.0-20210119193758-3436c122b313
```
jabrown85, line 7: I don't quite understand how this `tools/` became pinned to this version. I would have expected we may need a `replace golang.org/x/sys => golang.org/x/sys v0.0.0-20200523222454-059865788121` in this `go.mod` as well. 😕

jabrown85 (approved):
Part of me rather revert to ggcr `0.3.0` until this is fixed there, but I don't think the risk of this `replace` is high enough to block.

tools/go.mod
natalieparellano: I believe since imgutil is already pinned, everything should work as expected: https://github.com/buildpacks/imgutil/commit/3436c122b31395e516adc4cc566823b2eb6d13fc

`tools` is used by our Makefile, so it everything is compiling with this version of imgutil.

ekcasey (approved):

PR #507, opened 2021-01-20
feat: set working directory on built image

josegonzalez:
Refs buildpacks/rfcs#134

Signed-off-by: Jose Diaz-Gonzalez <jose@seatgeek.com>

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -127,6 +127,14 @@ func (e *Exporter) Export(opts ExportOptions) (ExportReport, error) {
 		return ExportReport{}, err
 	}
 
+	// platform API > 0.5
+	if e.PlatformAPI.Compare(api.MustParse("0.5")) > 0 {
+		e.Logger.Debugf("Setting WORKDIR: '%s'", opts.AppDir)
+		if err := e.setWorkingDir(opts); err != nil {
+			return ExportReport{}, errors.Wrap(err, "setting workdir")
+		}
+	}
+
 	entrypoint, err := e.entrypoint(buildMD.toLaunchMD(), opts.DefaultProcessType)
 	if err != nil {
 		return ExportReport{}, errors.Wrap(err, "determining entrypoint")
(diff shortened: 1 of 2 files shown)


jabrown85 (approved):

exporter.go
```diff
@@ -362,6 +366,10 @@ func (e *Exporter) setEnv(opts ExportOptions, launchMD launch.Metadata) error {
 	return nil
 }
 
+func (e *Exporter) setWorkingDir(opts ExportOptions) error {
+	return opts.WorkingImage.SetWorkingDir(opts.AppDir)
```
natalieparellano, line 370: It might be nice to add a debug statement, similar to here: https://github.com/buildpacks/lifecycle/blob/ee2fd58e6e5c16de1b450f4ddcf819849c6a6d5c/exporter.go#L135

natalieparellano:
@josegonzalez thanks for the PR! After some discussion, we are wondering if it makes sense to tie this to a platform API change. It seems that technically the working directory for the app image could be inherited from the run image, and therefore there's a slight risk of clobbering what might already be set. What do you think?

josegonzalez:
Yeah I think that makes sense @natalieparellano. What would you all need me to do to... get the ball rolling on that?

natalieparellano:
Thanks @josegonzalez - could you add a guard on the platform API similar to what we do [here](https://github.com/buildpacks/lifecycle/blob/ee2fd58e6e5c16de1b450f4ddcf819849c6a6d5c/exporter.go#L446) or [here](https://github.com/buildpacks/lifecycle/blob/ee2fd58e6e5c16de1b450f4ddcf819849c6a6d5c/exporter.go#L153)? The spec changes will get merged into platform/0.6 anyway so we should be good on that front.

josegonzalez:
@natalieparellano i think i did it correctly but happy to move that check (and the log call) elsewhere.

natalieparellano:
@josegonzalez this is great! Thank you for the contribution!

natalieparellano (approved):

PR #508, opened 2021-01-21
Add <layers>/order.toml resolution to detector

jabrown85:
This adds an optional path to the resolution of `order.toml`. The new path is `<layers>/order.toml` and will be selected prior to `/cnb/order.toml` if there is a file present. This will allow platforms to create an `order.toml` in the `<layers>` volume in another step and have that accepted automatically by the lifecycle when. that occurs.

Blocked on:

- [x] Waiting on [Spec](https://github.com/buildpacks/spec/pull/182) approval and spec version designation
- [x] Need tests

### DIFF
--- a/acceptance/detector_test.go
+++ b/acceptance/detector_test.go
@@ -3,6 +3,7 @@
 package acceptance
 
 import (
+	"fmt"
 	"io/ioutil"
 	"math/rand"
 	"os"
@@ -16,6 +17,7 @@ import (
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
+	"github.com/buildpacks/lifecycle/api"
 	"github.com/buildpacks/lifecycle/buildpack"
 	"github.com/buildpacks/lifecycle/cmd"
 	"github.com/buildpacks/lifecycle/platform"
@@ -35,7 +37,11 @@ func TestDetector(t *testing.T) {
 	rand.Seed(time.Now().UTC().UnixNano())
 
 	h.MakeAndCopyLifecycle(t, "linux", detectorBinaryDir)
-	h.DockerBuild(t, detectImage, detectDockerContext)
+	h.DockerBuild(t,
+		detectImage,
+		detectDockerContext,
+		h.WithArgs("--build-arg", fmt.Sprintf("cnb_platform_api=%s", api.Platform.Latest())),
+	)
 	defer h.DockerImageRemove(t, detectImage)
 
 	spec.Run(t, "acceptance-detector", testDetector, spec.Parallel(), spec.Report(report.Terminal{}))
@@ -64,7 +70,7 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
 
 	when("read buildpack order file failed", func() {
 		it("errors", func() {
-			// no order.toml file in the default directory
+			// no order.toml file in the default search locations
 			command := exec.Command("docker", "run", "--rm", detectImage)
 			output, err := command.CombinedOutput()
 			h.AssertNotNil(t, err)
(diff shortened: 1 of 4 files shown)


acceptance/testdata/detector/Dockerfile
```diff
@@ -3,6 +3,8 @@ FROM ubuntu:bionic
 ARG cnb_uid=1234
 ARG cnb_gid=1000
 
+ENV CNB_PLATFORM_API=0.6
```
jabrown85, line 6: I wasn't sure how the acceptance test are supposed to be targeting the latest platform API. I did this for now, but would love to know if there is a more automatic way so we don't have to bump these dockerfiles that I might've missed.

tools/go.sum
```diff
@@ -71,6 +71,7 @@ github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMn
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
 github.com/containerd/containerd v1.3.0/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtMXFTttgp+kVtyUA=
+github.com/containerd/containerd v1.3.3 h1:LoIzb5y9x5l8VKAlyrbusNPXqBY0+kviRloxFUMFwKc=
```
jabrown85, line 74: This file changed as a consequence of `go mod tidy` in the `tools/` folder.

acceptance/testdata/detector/Dockerfile
natalieparellano: I don't think we have an established pattern for this. Maybe we could make the api version an ARG - and then when we `h.DockerBuild` the image, we could pass `h.WithArgs("--build-arg", fmt.Sprintf("platformAPI=%s", api.Platform.Latest()))`. Individual tests could override the platform api version by passing `--env` when we run the image.

cmd/flags.go
```diff
@@ -120,7 +121,23 @@ func FlagNoColor(skip *bool) {
...
 
 func FlagOrderPath(orderPath *string) {
-	flagSet.StringVar(orderPath, "order", EnvOrDefault(EnvOrderPath, DefaultOrderPath), "path to order.toml")
+	flagSet.StringVar(orderPath, "order", EnvOrDefault(EnvOrderPath, PlaceholderOrderPath), "path to order.toml")
+}
+
+func DefaultOrderPath(platformAPI, layersDir string) string {
+	cnbOrderPath := filepath.Join(rootDir, "cnb", "order.toml")
+
+	// prior to Platform API 0.6, the default as /cnb/order.toml
```
natalieparellano, line 130:
```suggestion
	// prior to Platform API 0.6, the default is /cnb/order.toml
```

acceptance/detector_test.go
```diff
@@ -193,4 +193,125 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
+
+				// check group.toml
+				tempGroupToml := filepath.Join(copyDir, "layers", "group.toml")
+				var buildpackGroup lifecycle.BuildpackGroup
+				_, err := toml.DecodeFile(tempGroupToml, &buildpackGroup)
+				h.AssertNil(t, err)
+				h.AssertEq(t, buildpackGroup.Group[0].ID, "simple_buildpack")
+				h.AssertEq(t, buildpackGroup.Group[0].Version, "simple_buildpack_version")
+			})
+		})
```
natalieparellano, line 315: Is this context duplicating lines 245-265?

It might be more readable to pull the `when("platform api < 0.6"` context to be just under `when("/cnb/order.toml is present"`.

natalieparellano: Ohh I see - this one doesn't have `"--volume", otherOrderTOMLPath+":"+"/cnb/order.toml"`. TBH I'm not sure we need this context, but maybe I'm wrong...

jabrown85: I went back and forth on this. I think I like it better where it is at, but if you feel strongly I can try and move things around.

natalieparellano: I'm okay leaving it where it is!

acceptance/detector_test.go
```diff
@@ -193,4 +199,125 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
 			h.AssertStringContains(t, logs, expectedAppDir)
 		})
 	})
+
+	when("-order is not provided directly", func() {
```
yaelharel, line 203: What do you think about the following order under this "when":

when("platform api < 0.6")
{
    when ("/cnb/order.toml and /layers/order.toml are present") {...}

    when(“only /layers/order.toml is present”) // to add - there will be an error, right? {...}
}

when("/cnb/order.toml and /layers/order.toml are present") {...}

when(“only /cnb/order.toml” is present”) {...}

when(“only /layers/order.toml is present”) {...}

I would also add a test where `-order` is provided.

natalieparellano:
@jabrown85 now that https://github.com/buildpacks/spec/pull/182 is merged into platform/0.6, can this be undrafted?

jabrown85:
@natalieparellano thanks for the ping! I rebased to resolve the conflict. I think this is ready for review!

natalieparellano:
I think we have an import problem in the acceptance tests...

natalieparellano (approved):

cmd/flags.go
```diff
@@ -120,7 +121,23 @@ func FlagNoColor(skip *bool) {
 }
 
 func FlagOrderPath(orderPath *string) {
-	flagSet.StringVar(orderPath, "order", EnvOrDefault(EnvOrderPath, DefaultOrderPath), "path to order.toml")
+	flagSet.StringVar(orderPath, "order", EnvOrDefault(EnvOrderPath, PlaceholderOrderPath), "path to order.toml")
+}
+
+func DefaultOrderPath(platformAPI, layersDir string) string {
```
yaelharel, line 127: Why didn't we use `defaultPath` as we're doing with:
```
DefaultAnalyzedFile
DefaultGroupFile
DefaultPlanFile
DefaultProjectMetadataFile
DefaultReportFile
```
?

PR #511, opened 2021-01-22
Fix goroutine race in detector

### LINKED ISSUE #510: Lifecycle detect go routine is no longer safe
### Summary
It appears that with the latest version lifecycle a environment variable (`CNB_BUILDPACK_DIR` [#325](https://github.com/buildpacks/lifecycle/pull/325)) is set to a different path at times when multiple processes (of detect) are running.

This is caused by the [go routine](https://github.com/buildpacks/lifecycle/blob/main/detector.go#L354) in the detect stage that changes the pathing/envs within the DetectConfig.

---

##### Steps

1. Have multiple source directories
2. Run the detect binary against them in parallel
3. Note the path in DetectConfig will be different 

##### Current behavior
When running detect in parallel the buildpack detection files will mismatch as the go routine runs and overwrites packs that are taking longer than others. This causes paths within DetectConfig to be different than the contents of the detection.

For example, running detect against a large php app and nodejs app might end up with the path within the nodejs app as the one used for the php app. 

##### Expected
The detector should be able to be run multiple times in parallel without effecting other detections.

We may want to mutex the entire DetectConfig file in the `detector.go$detect(...)` function though this might make the go routine useless. 

---

##### lifecycle version
v0.10.1

##### platform version(s)
v0.3

##### anything else?
I'm very new to this project but if will contribute a PR if we can track down a solid solution :rocket:

kamaln7:
Fixes #510 

@jonfriesen and I had been trying to debug a weird regression we started seeing after we upgraded from a `0.8.x` version of lifecycle to `0.10.1`. Eventually we narrowed it down to random Buildpacks intermittently having their `CNB_BUILDPACK_DIR` env var set to that of a different Buildpack like Jon described in #510.

The issue seems to be stemming from the fact that the goroutine runs `info.Detect(c)` which does the following:

```go
	cmd.Env = c.FullEnv
	if b.Buildpack.ClearEnv {
		cmd.Env = c.ClearEnv
	}
	cmd.Env = append(cmd.Env, EnvBuildpackDir+"="+b.Dir)
```

Copying the slice `c.FullEnv`/`c.ClearEnv` as-is and then appending to it inside the goroutine is unsafe. `cmd.Env` and `c.FullEnv` share the same underlying array. When it is appended to it affects the underlying array on `c.FullEnv`/`c.ClearEnv` which is also being used by the other goroutines.

Copying the slice using `newSlice = append([]string{}, oldSlice...)` ensures that the items are copied to a brand new slice and  makes it safe to modify inside the goroutine.

I also pass the `key` and `info` variables to the goroutine as function arguments although that didn't seem to be the cause of the issue we were seeing.

May I also suggest adding the `-race` flag to the test suite?

### DIFF
--- a/detector.go
+++ b/detector.go
@@ -271,9 +271,9 @@ func (b *BuildpackTOML) Detect(c *DetectConfig) DetectRun {
 	cmd.Dir = appDir
 	cmd.Stdout = out
 	cmd.Stderr = out
-	cmd.Env = c.FullEnv
+	cmd.Env = append([]string{}, c.FullEnv...)
 	if b.Buildpack.ClearEnv {
-		cmd.Env = c.ClearEnv
+		cmd.Env = append([]string{}, c.ClearEnv...)
 	}
 	cmd.Env = append(cmd.Env, EnvBuildpackDir+"="+b.Dir)
 
@@ -351,12 +351,12 @@ func (bg BuildpackGroup) detect(done []GroupBuildpack, wg *sync.WaitGroup, c *De
 		}
 		done = append(done, bp)
 		wg.Add(1)
-		go func() {
+		go func(key string, info *BuildpackTOML) {
 			if _, ok := c.runs.Load(key); !ok {
 				c.runs.Store(key, info.Detect(c))
 			}
 			wg.Done()
-		}()
+		}(key, info)
 	}
 
 	wg.Wait()


natalieparellano:
@kamaln7 @jonfriesen thanks for the issue and PR! Taking a look...

jabrown85 (approved):

natalieparellano:
@kamaln7 could you rebase your branch on top of `release/0.10.2`? We can cut a patch soon.

Also, is it possible you could explain why the arguments to the go func would be necessary? After some discussion we believe that because `key` and `info` are initialized inside the loop, there isn't a risk of a race.

kamaln7:
Thanks for reviewing @natalieparellano! Yes, I'll rebase the branch now.

As for why pass `key` and `info` as arguments to the goroutine closure, it's mostly a precaution. Everything seems to work fine even without it so I don't think it is causing any issues in its current state (the race detector did not complain either). So it's really just to avoid issues popping up in the future in case changes are introduced that area of the codebase.

I'll defer to you on whether to remove it or not - happy to do it either way 🙂

natalieparellano:
Ahh that makes sense - that falls in line with our thinking also. I did trying adding an unsafe use of `bp` within the func, and `go vet` complained with `loopclosure: loop variable bp captured by func literal` so I'm inclined to say that it's not necessary, but it's not doing any harm either. I'm happy to merge the PR as-is.

natalieparellano (approved):

kamaln7:
Great, thank you!

PR #512, opened 2021-01-22
Add restorer acceptance tests

jabrown85:
Following the existing acceptance test patterns, the new acceptance test will use the test data mounted into an image to prove restorer is properly restoring data as expected.

The contents of the test data was created in-container and copied out of the container using the buildpacks in the testdata folders.


acceptance/testdata/restorer/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:bionic
+
+ARG cnb_uid=1234
+ARG cnb_gid=1000
+
+COPY ./container/ /
+
+# this is required because git on windows does not support our naming of sha256:sha.tar :(
+COPY ./container/cache/committed/29acd01e6fba0963e796f63eb3b517c2b2c6fb2ebcca5180f2bdda82c8b922d0.tar /cache/committed/sha256:29acd01e6fba0963e796f63eb3b517c2b2c6fb2ebcca5180f2bdda82c8b922d0.tar
```
jabrown85, line 9: I'm totally open to any suggestions on making this less gross. Windows can't check out a file named `sha256:<sha>`, so I am currently just copying it over with the correct name.

jabrown85: [Thread](https://buildpacks.slack.com/archives/CSFE49VGR/p1611336072013100)

acceptance/restorer_test.go
```diff
@@ -0,0 +1,130 @@
...
+		it("errors", func() {
+			command := exec.Command("docker", "run", "--rm", restorerImage, "some-arg")
+			output, err := command.CombinedOutput()
+			h.AssertNotNil(t, err)
+			expected := "failed to parse arguments: received unexpected Args"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("called without any cache arguments", func() {
```
natalieparellano, line 50:
```suggestion
	when("called without any cache flag", func() {
```

acceptance/testdata/restorer/Dockerfile
natalieparellano: I can't think of how to make this less gross :(

I wonder though if instead of building the cache into the container we could just mount it in - then it would work more like `pack`. Also if we wanted to test cache image or cache directory not available we'd already have a restore image that works for that.

acceptance/restorer_test.go
```diff
@@ -0,0 +1,130 @@
...
+				os.RemoveAll(copyDir)
+			})
+
+			it("restores cached layer data", func() {
+				h.DockerRunAndCopy(t,
+					containerName,
+					copyDir,
+					restorerImage,
+					"/layers",
+					h.WithFlags("--user", "0"),
```
natalieparellano, line 84: Why is this necessary?

natalieparellano (commented):
@jabrown85 it is awesome to see this. I'm sorry for being slow in reviewing this. I left a couple of questions and suggestions.

acceptance/testdata/restorer/container/layers/cacher_buildpack/uncached-layer.sha
```diff
@@ -0,0 +1 @@
+sha256:45e752b0b440c89a362b7c24a8ac4eb7245a26abda2942e124d95964ce57db41
```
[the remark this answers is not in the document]
jabrown85: This file came from a `docker cp` from a container to be exactly as they are from the previous step. I kind of like leaving it as-is so we don't make assumptions around it having a newline when processing it.

acceptance/testdata/restorer/container/layers/cacher_buildpack/cached-layer.sha
```diff
@@ -0,0 +1 @@
+sha256:29acd01e6fba0963e796f63eb3b517c2b2c6fb2ebcca5180f2bdda82c8b922d0
```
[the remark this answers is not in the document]
jabrown85: This file came from a `docker cp` from a container to be exactly as they are from the previous step. I kind of like leaving it as-is so we don't make assumptions around it having a newline when processing it.

acceptance/testdata/restorer/container/layers/analyzed.toml
```diff
@@ -0,0 +1,34 @@
+[image]
+  reference = "192.168.86.20:5000/builtimg@sha256:f5006340c3a8613a4200cd965db29386069a2c3b427a2f0086c22d74c565610a"
```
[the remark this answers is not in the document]
jabrown85: It isn't used by restorer. It came over when I `docker cp`'d the folders. I like that it sort of represents the state of the world when restorer will be executed but am happy to remove it.

acceptance/testdata/restorer/container/cache/committed/io.buildpacks.lifecycle.cache.metadata
```diff
@@ -0,0 +1 @@
+{"buildpacks":[{"key":"cacher_buildpack","version":"cacher_v1","layers":{"cached-layer":{"sha":"sha256:29acd01e6fba0963e796f63eb3b517c2b2c6fb2ebcca5180f2bdda82c8b922d0","data":null,"build":false,"launch":false,"cache":true}}},{"key":"unused_buildpack","version":"v1","layers":{"cached-layer":{"sha":"sha256:4974a59c1c25cacec0d8e2623104618edb3d3e892a2a140dd35044bc1151e185","data":null,"build":false,"launch":false,"cache":true}}}]}
```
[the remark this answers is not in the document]
jabrown85: Yeah, I have mixed feelings about that. Generating the tars on the fly would mean that we aren't testing that lifecycle can read a tar written by lifecycle _during an actual export_. Maybe that isn't an important part of the test and isn't worth keeping the binary data. I sort of worry that the creation of the tar will become most of this test and have divergent behavior from reality. I'm not opposed to making this suggested change though.

acceptance/restorer_test.go
jabrown85: The `restorer` does a `chown` to the `uid`/`gid` on the `layers` and `cache` dirs before dropping privs.

acceptance/testdata/restorer/container/layers/cacher_buildpack/uncached-layer.sha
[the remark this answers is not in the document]
natalieparellano: Ohhh that's interesting! So that is how the analyzer is writing this file...

acceptance/restorer_test.go
natalieparellano: I see. But since we do that in the Dockerfile too, we could run this as `--user 1234` if we wanted right? I guess this line is sort of implying that restorer **must** be run as root, when in fact we only NEED to do that if using the docker daemon.

acceptance/testdata/restorer/Dockerfile
jabrown85: I did actually try that. I didn't dig too deep into why it fails though, but it doesn't like the `sha256:blah` naming when using `docker`'s `--volume` support. There is a chance the `--mount` syntax may work better though...

```
» docker run -it --volume "/Users/jesse.brown/dev/buildpacks/lifecycle/acceptance/testdata/restorer/container/cache/committed/29acd01e6fba0963e796f63eb3b517c2b2c6fb2ebcca5180f2bdda82c8b922d0.tar:/cache/committed/sha256:test.tar" heroku/buildpacks:20 bash
docker: Error response from daemon: invalid mode: test.tar.
See 'docker run --help'.
```

acceptance/restorer_test.go
jabrown85: Using `--user 1234` does _not_ work. If we expect that to work while using docker daemon, it does not seem to. It still attempts to `chown` the `/layers` directory even though it should already be owned by `1234`.

```
[docker run --user 1234 --name test-container-hzcuhdnfou lifecycle/acceptance/restorer -cache-dir /cache], exit status 1, , ERROR: failed to chown volumes: chown /layers: operation not permitted
```

Is this expected because we are using docker volumes?

acceptance/testdata/restorer/Dockerfile
natalieparellano: Could you mount the entire cache directory to avoid having to specify the tar name?

acceptance/testdata/restorer/container/cache/committed/io.buildpacks.lifecycle.cache.metadata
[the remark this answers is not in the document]
natalieparellano: Hmm... if we want to test operating on the result of an actual export I'd be inclined to just run a build with the `creator` in the test setup. I think having a sort of magic file makes it hard to understand what the test is validating and how.

acceptance/testdata/restorer/Dockerfile
jabrown85: The original problem this solves is that we can't have the `sha256:abc.tar` names in our git repo at all, because of windows. So it has to be done in the CLI or the Dockerfile, something that could support that naming scheme.

natalieparellano: If we generate the tars on the fly though, this problem goes away : )

acceptance/testdata/restorer/container/cache/committed/io.buildpacks.lifecycle.cache.metadata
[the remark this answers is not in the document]
jabrown85: I could take a stab at generating some of the test data on the fly and see how it starts to look. It would get rid of the binary + the filename problem on windows.

acceptance/restorer_test.go
natalieparellano: This surprises me. Maybe we also need to pass the group id?

jabrown85: The `Dockerfile` ends with `USER 1234:1000`, so omitting the `--user` should have worked but doesn't. I did try explicitly codifying it into the test to make sure and it fails as well.
```
--- FAIL: TestRestorer/acceptance-restorer/using_cache-dir/there_is_cache_present_from_a_previous_build/restores_cached_layer_data (1.65s)
            docker.go:56: Expected nil: failed to execute command: [docker run --user 1234:1000 --name test-container-qjlelhxuou lifecycle/acceptance/restorer -cache-dir /cache], exit status 1, , ERROR: failed to chown volumes: chown /layers: operation not permitted
```

acceptance/testdata/restorer/container/layers/analyzed.toml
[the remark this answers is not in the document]
natalieparellano: I think we could take it out, it may add noise and lead to further questions as to "why is this here"?

acceptance/restorer_test.go
natalieparellano: This is interesting. So I think the disconnect here is that the restorer (and probably other phases too) when not provided a `-uid` or `-gid` flag sets those values to 0 / root instead of the user & group that the lifecycle itself is running as.

Thus when we're passing `--user 1234` (run restorer as `1234`) the lifecycle is still using `uid 0` when checking `canWrite`: https://github.com/buildpacks/lifecycle/blob/77735c391b3f7b8376616a04cad8013c46b16e59/priv/user_linux.go#L45 and thus tries to chown `/layers` even though it's already owned by `1234`. Since you can't chown as non-root it all fails...

I tried passing `-uid 1234` as an argument to the restorer and it doesn't have a problem with `/layers` but then tries to chown `/cache` :(

I think this is all tangential to what we're trying to accomplish here - maybe we could throw a comment on the `--user 0` and get around to tackling this more carefully when filling out the rest of the restorer tests. The analyzer tests currently have some unpleasant bash workarounds to get the cache directory to be readable when mounted in as a volume: https://github.com/buildpacks/lifecycle/blob/ee2fd58e6e5c16de1b450f4ddcf819849c6a6d5c/acceptance/analyzer_test.go#L464-L466

acceptance/testdata/restorer/Dockerfile
natalieparellano: Example of how we mount the cache directory for analyzer acceptance: https://github.com/buildpacks/lifecycle/blob/ee2fd58e6e5c16de1b450f4ddcf819849c6a6d5c/acceptance/analyzer_test.go#L455-L462

acceptance/restorer_test.go
ekcasey: Because `CNB_USER_ID` and `CNB_GROUP_ID` are required to be set in the build image by the spec, we can assume we always have a good value for `uid` and `gid` whether they come from flags or the environments.

acceptance/testdata/restorer/container/cache/committed/io.buildpacks.lifecycle.cache.metadata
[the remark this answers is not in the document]
jabrown85: I took a path of doing the `tar` of the expected cache _inside_ the `Dockerfile`. This, I think, will be useful as we could introduce `Dockerfile.windows` in the future that tars it up and names the file the way windows containers expect. We could explore using a volume, but this seemed like a good compromise with less go code to generate tarballs in tmp directories and whatnot. Check it out and let me know what you think!

acceptance/restorer_test.go
jabrown85: I removed all the `--user "0"` instances. Since I'm doing the tar work without volumes, I can safely do the chown in the `Dockerfile` and know that `restorer` won't have to. 😃

acceptance/testdata/restorer/container/cache/committed/io.buildpacks.lifecycle.cache.metadata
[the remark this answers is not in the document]
natalieparellano: Hmm... I definitely like this better than hard coding the tar files :)

Did you take a look at `layers.Factory.DirLayer`? That seems pretty easy to hook into and might give exactly what we want (and be cross platform too).

acceptance/testdata/restorer/container/cnb/buildpacks/cacher_buildpack/cacher_v1/bin/build
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
natalieparellano, line 1: If we're generating the tar on the fly we could get rid of everything in `acceptance/testdata/restorer/container/cnb/buildpacks` as it's not required for the tests to run successfully.

jabrown85: I removed all the unused files now

acceptance/testdata/restorer/container/cache/committed/io.buildpacks.lifecycle.cache.metadata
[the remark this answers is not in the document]
jabrown85: I did briefly look at that helper. It could likely be leveraged, but not without some changes. The helper was building tars with paths like `/Users/myusername/some/path...`, which would not restore in the expected `/layers` location once inside the container and being restored by `restorer`.

acceptance/testdata/restorer/container/cnb/buildpacks/cacher_buildpack/cacher_v1/bin/build
natalieparellano: It looks like the tests all still pass if the following files are removed:

* acceptance/testdata/restorer/container/cnb/orders/cache_creator_order.toml
* acceptance/testdata/restorer/container/cnb/orders/cache_restorer_order.toml
* acceptance/testdata/restorer/container/layers/cacher_buildpack/uncached-layer.sha
* acceptance/testdata/restorer/container/layers/cacher_buildpack/uncached-layer.toml

acceptance/testdata/restorer/container/cache/committed/io.buildpacks.lifecycle.cache.metadata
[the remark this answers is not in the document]
natalieparellano: I see... in this case let's go with the Dockerfile approach, we can always modify or extend it later.

acceptance/testdata/restorer/container/cnb/buildpacks/cacher_buildpack/cacher_v1/bin/build
jabrown85: Yeah, I removed the orders. Good spot.

natalieparellano (approved):
Looks good to me! @ekcasey did you want to review before we merge?

acceptance/restorer_test.go
```diff
@@ -0,0 +1,127 @@
...
+					"/layers",
+					h.WithArgs("-cache-dir", "/cache"),
+				)
+
+				// check no content is not present from unused buildpack
+				unusedBpLayer := filepath.Join(copyDir, "layers", "unused_buildpack")
+				h.AssertPathDoesNotExist(t, unusedBpLayer)
+			})
+		})
+
```
yaelharel, line 125:
```suggestion
		})
```

acceptance/testdata/restorer/container/to_cache/unused_buildpack/layers/unused_buildpack/cached-layer/data
```diff
@@ -0,0 +1 @@
+cached-data
```
yaelharel, line 1: Maybe it's better to change this a bit to something different than the content of `acceptance/testdata/restorer/container/to_cache/cacher_buildpack/layers/cacher_buildpack/cached-layer/data` so the code will be more readable.
What do you think?

acceptance/testdata/restorer/container/layers/cacher_buildpack/uncached-layer.toml
```diff
@@ -0,0 +1,3 @@
+build = false
+launch = true
```
yaelharel, line 2: Why is `launch` `true` here and `false` in `acceptance/testdata/restorer/container/layers/cacher_buildpack/cached-layer.toml`?

acceptance/testdata/restorer/container/layers/cacher_buildpack/uncached-layer.sha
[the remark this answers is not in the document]
yaelharel: Can we write a comment on how this file was generated?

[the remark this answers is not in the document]
jabrown85: I removed this file entirely. It is needed to prove out the test this was originally built for.

acceptance/testdata/restorer/container/layers/cacher_buildpack/uncached-layer.toml
jabrown85: It was from my test data, I went ahead and made the 2 layers identical except for name and the fact that one exists in the cache being restored and one doesn't. I hope the intent is now more clear :)

yaelharel (approved):
The code looks great! Thank you for adding these tests!

PR #513, opened 2021-01-22
Run mod tidy

natalieparellano:
Run `go mod tidy` at the root of the repo and in `tools`. Otherwise, after the tests are run our repo may have local changes and the lifecycle image will be tagged with `<sha>-dirty`. This may cause our post-release automation that re-tags the image to fail.


natalieparellano:
Probably we should also update our automation to never add the `-dirty`, since we'll never want that as part of the tag in non-local development.

ekcasey (approved):

PR #519, opened 2021-01-29
Fix tools & add make tidy

natalieparellano:
Tools needs to be pinned to older x/sys as well.


natalieparellano:
Should we automatically run `make tidy` when we test / format / lint?

Makefile
```diff
@@ -203,6 +203,10 @@ format: install-goimports
 	@echo "> Formating code..."
 	$(if $(shell goimports -l -w -local github.com/buildpacks/lifecycle .), @echo Fixed formatting errors. Re-run && exit 1)
 
+tidy:
```
yaelharel, line 206: I think it's worth adding this to `make unit/all/acceptance`.
Or maybe we can create another target that is running `lint format tidy` and use it when running `make unit/all/acceptance`.

natalieparellano: Added to test targets

yaelharel (approved):

jabrown85 (approved):
This LGTM - I could see omitting `tidy` from `unit` and `acceptance` for local dev experience speed reasons, but this aligns with `format` and `lint` so 👍

PR #520, opened 2021-02-01
Default process

### LINKED ISSUE #457: [RFC 0064] Platform assumes web is default process for older buildpacks
[RFC 0064](https://github.com/buildpacks/rfcs/blob/main/text/0064-buildpacks-contribute-default-process-type.md)
https://github.com/buildpacks/rfcs/pull/110
https://github.com/buildpacks/spec/issues/160

Also, if none is specified, don't set a default process-type, even if there is only one process-type.
https://github.com/buildpacks/spec/pull/137

### LINKED ISSUE #458: [RFC 0064] Buildpacks contribute default process type
[RFC 0064](https://github.com/buildpacks/rfcs/blob/main/text/0064-buildpacks-contribute-default-process-type.md)
https://github.com/buildpacks/rfcs/pull/110
https://github.com/buildpacks/spec/issues/159

yaelharel:
- Add support for Builpack API 0.6
- Add support of default process type to buildpacks

Fixes #457
Fixes #458 

cc @dwillist


buildpacktoml.go
```diff
@@ -254,6 +260,33 @@ func (b *BuildpackTOML) readOutputFiles(bpLayersDir, bpPlanPath string, bpPlanIn
...
+	if api.MustParse(platformAPI).Compare(api.MustParse("0.6")) < 0 || api.MustParse(bpAPI).Compare(api.MustParse("0.6")) >= 0 {
+		return
+	}
+
+	for i := range processes {
+		if processes[i].Type == "web" {
+			processes[i].Default = true
+		}
+	}
+}
```
natalieparellano, line 274: Could we move this logic into builder.go? Having `BuildpackTOML` care only about buildpack api things and `Builder` care (mostly) about platform api things gives a nice separation and is easier to test all the permutations. [Here](https://github.com/buildpacks/lifecycle/blob/fae1fe27c0acdfabe9538e3607e72af61c62ede9/builder.go#L104-L108) is an example of some post-processing that the `Builder` does depending on the platform api version.

builder_test.go
```diff
@@ -310,26 +310,187 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								Direct:      true,
+								BuildpackID: "A",
+							},
+							{
+								Type:        "some-other-type",
+								Command:     "some-other-command",
+								Args:        []string{"some-other-arg"},
+								Direct:      true,
+								BuildpackID: "B",
+							},
```
natalieparellano, line 326: I'm confused as to why the expectation changed when the setup stayed the same - it looks like the processes are in reverse order now - was that intentional?

builder_test.go
```diff
@@ -411,6 +572,7 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 		when("platform api < 0.4", func() {
 			it.Before(func() {
 				builder.PlatformAPI = api.MustParse("0.3")
+				config.PlatformAPI = "0.3"
```
natalieparellano, line 575: If we follow the suggestion in https://github.com/buildpacks/lifecycle/pull/520/files#r568171284 we won't need to modify the config that gets passed to each buildpack.

buildpacktoml_test.go
```diff
@@ -286,32 +288,107 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
+									{Type: "web", Command: "other-cmd", BuildpackID: "A", Default: false},
+								},
+								Slices: []layers.Slice{},
+							}); s != "" {
+								t.Fatalf("Unexpected metadata:\n%s\n", s)
+							}
+						})
+					})
+
+					it("should include processes and use the default value that is set", func() {
```
natalieparellano, line 327: Is this a repeat of the previous context?

buildpacktoml_test.go
```diff
@@ -286,32 +288,107 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
+						it("should include processes and use the default value that is set", func() {
+							h.Mkfile(t,
+								`[[processes]]`+"\n"+
+									`type = "some-type"`+"\n"+
+									`command = "some-cmd"`+"\n"+
+									`default = true`+"\n"+
+									`[[processes]]`+"\n"+
+									`type = "web"`+"\n"+
+									`command = "other-cmd"`+"\n",
+								// default is false and therefore doesn't appear
```
natalieparellano, line 305: Having `default= = false` in the TOML would also be fine, if we want to do it that way because it's clearer.

buildpacktoml_test.go
```diff
@@ -286,32 +288,107 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
+									`command = "some-cmd"`+"\n"+
+									`default = true`+"\n"+
+									`[[processes]]`+"\n"+
+									`type = "other-type"`+"\n"+
+									`command = "other-cmd"`+"\n"+
+									`[[processes]]`+"\n"+
+									`type = "some-type"`+"\n"+
+									`command = "some-other-cmd"`+"\n"+
+									`default = true`+"\n",
+								// default is false and therefore doesn't appear
```
natalieparellano, line 370: Copy / paste error I think

buildpacktoml_test.go
```diff
@@ -523,6 +600,31 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
+					h.Mkfile(t,
+						`[[processes]]`+"\n"+
+							`type = "some-type"`+"\n"+
+							`command = "some-cmd"`+"\n"+
+							`default = true`+"\n"+
+							`[[processes]]`+"\n"+
+							`type = "other-type"`+"\n"+
+							`command = "other-cmd"`+"\n"+
+							`default = true`+"\n",
+						// default is false and therefore doesn't appear
```
natalieparellano, line 619: Copy / paste error I think

buildpacktoml_test.go
```diff
@@ -767,6 +869,78 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
 			})
 		})
+
+		when("buildpack api < 0.6", func() {
+			it.Before(func() {
+				bpTOML.API = "0.5"
+				mockEnv.EXPECT().WithPlatform(platformDir).Return(append(os.Environ(), "TEST_ENV=Av1"), nil)
+			})
+
+			when("platform API < 0.6", func() {
```
natalieparellano, line 879: https://github.com/buildpacks/lifecycle/pull/520/files#r568171284 would help avoid the combinatorial explosion of buildpack & platform apis.

builder_test.go
```diff
@@ -310,26 +310,187 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+							},
+						}); s != "" {
+							t.Fatalf("Unexpected:\n%s\n", s)
+						}
+
+						expected := "Warning: redefining the following default process types with processes not marked as default: [override-type]"
+						h.AssertStringContains(t, stdout.String(), expected)
+					})
+
+					it("should preserve ordering of processes", func() {
```
natalieparellano, line 413: I don't think this is necessary to be spec-compliant, though maybe it's good to have it?

builder.go
```diff
@@ -175,25 +192,54 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
 	return false
 }
 
-type processMap map[string]launch.Process
+// orderedProcesses is a mapping from process types to Processes, it will preserve ordering.
+// processList is the ordered list
+// wwe keep typeToPtr map in order to delete elements when others are overriding them
```
natalieparellano, line 197:
```suggestion
// we keep typeToProcess map in order to delete elements when others are overriding them
```

natalieparellano: Following on https://github.com/buildpacks/lifecycle/pull/520/files#r568185952, this seems to be adding a bit of complexity that is maybe not necessary? Or am I missing something?

natalieparellano: Ohh I see - it is needed so that we can do `launchMD.FindLastDefaultProcessType()`...

exporter.go
```diff
@@ -371,16 +371,22 @@ func (e *Exporter) entrypoint(launchMD launch.Metadata, defaultProcessType strin
 	if !e.supportsMulticallLauncher() {
 		return launch.LauncherPath, nil
 	}
-	if defaultProcessType == "" {
-		if len(launchMD.Processes) == 1 {
```
natalieparellano, line 375: I believe for older platform api, we still want to preserve the behavior of setting the only process as the default.

exporter.go
```diff
@@ -371,16 +371,22 @@ func (e *Exporter) entrypoint(launchMD launch.Metadata, defaultProcessType strin
...
-			return launch.ProcessPath(launchMD.Processes[0].Type), nil
+
+	if defaultProcessType != "" {
+		defaultProcess, ok := launchMD.FindProcessType(defaultProcessType)
+		if !ok {
+			if e.PlatformAPI.Compare(api.MustParse("0.6")) < 0 {
+				e.Logger.Warn(processTypeWarning(launchMD, defaultProcessType))
+				return launch.LauncherPath, nil
+			}
+			return "", fmt.Errorf("tried to set %s to default but it doesn't exist", defaultProcessType)
```
natalieparellano, line 382: Elsewhere in this file we're using a pattern of `errors.Wrap(err, "do a thing")`

exporter_test.go
```diff
@@ -1299,6 +1307,17 @@ version = "4.5.6"
 	})
 }
 
+func checkEntrypoint(t *testing.T, image *fakes.Image, expectedDir string, expectedFile string) {
```
natalieparellano, line 1310: It might be nice to pass the expected entrypoint into the function, then this could read like `assertHasEntrypoint(t, image, expectedEntrypoint)`. There'd be slightly more duplication in passing the correct expectation for Linux vs. Windows, but it might read easier.

exporter_test.go
```diff
@@ -910,156 +911,163 @@ version = "4.5.6"
...
-						} else {
-							h.AssertEq(t, ep[0], `/cnb/process/some-process-type`)
-						}
+						assertLogEntry(t, logHandler, "no default process type")
+						checkEntrypoint(t, fakeAppImage, "lifecycle", "launcher")
 					})
+				})
 
-					it("does not set CNB_PROCESS_TYPE", func() {
+				when("there is a default=true process in metadata.toml", func() {
```
natalieparellano, line 980: This context might also benefit from a `it("doesn't set CNB_PROCESS_TYPE"` expectation.

natalieparellano (changes requested):
@yaelharel @dwillist it's great to see this, nice work wrangling all of the complexity of this feature :)

I think we want to preserve the existing "choose the only process as default" behavior for platform api 0.4 and 0.5. 

Beyond that I think the refactor I suggested (move platform api logic into builder.go) should make this easier to reason about and test - what do you think?

buildpacktoml_test.go
yaelharel: It was changed as part of commit https://github.com/buildpacks/lifecycle/pull/520/commits/845f01c70137cd8ddf8c399a1b6a93f3199d6b72.

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/845f01c70137cd8ddf8c399a1b6a93f3199d6b72.

buildpacktoml.go
yaelharel: Good idea. Please see https://github.com/buildpacks/lifecycle/pull/520/commits/845f01c70137cd8ddf8c399a1b6a93f3199d6b72.

buildpacktoml_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/845f01c70137cd8ddf8c399a1b6a93f3199d6b72.

builder.go
yaelharel: Exactly. We need this struct because we care about the order of the processes.

exporter.go
yaelharel: There are also some places in this file that we're using `fmt.Errorf` (for example: https://github.com/buildpacks/lifecycle/blob/fae1fe27c0acdfabe9538e3607e72af61c62ede9/exporter.go#L431).
I think it's better to use `fmt.Errorf` in this case since we don't have an error to wrap.
What do you think?

builder_test.go
yaelharel: I see that you marked this conversation as resolved. Is it clear?

buildpacktoml_test.go
yaelharel: We thought that it's fine to not write `default` to the TOML when its value is `false`, but I don't have strong opinion on that.
If you think it'll be clearer to write it anyway, we can make this change.

builder_test.go
yaelharel: I see that you marked this conversation as resolved.
Do you still want to discuss it?

exporter.go
yaelharel: Makes sense. I'll add it back.

buildpacktoml_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/d7d5d23a49855cf3585cff952586e1e79213dc4a.

buildpacktoml_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/d7d5d23a49855cf3585cff952586e1e79213dc4a.

exporter_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/d7d5d23a49855cf3585cff952586e1e79213dc4a.

exporter_test.go
yaelharel: I don't mind to rename the function to `assertHasEntrypoint` but I think it's better not to pass the 2 entrypoints (one for each OS).
I think it's better to hide the differences between the entrypoints of linux and windows because it prevents code duplication and therefore typos while writing new tests.
In order to make the code clearer, maybe we can pass only one entrypoint and do string manipulations on it in the function in order to check the other expected entrypoint, although it might be confusing.
What do you think?

yaelharel:
@natalieparellano, thank you for your feedback!
Please see the changes I've made and the comments I left to some of your ideas.

exporter.go
natalieparellano: Ohh I missed that!

builder.go
```diff
@@ -25,6 +26,7 @@ type BuildpackStore interface {
 
 type Buildpack interface {
 	Build(bpPlan BuildpackPlan, config BuildConfig) (BuildResult, error)
+	BuilpackAPI() string
```
natalieparellano, line 29:
```suggestion
	BuildpackAPI() string
```

builder.go
```diff
@@ -94,10 +96,22 @@ func (b *Builder) Build() (*BuildMetadata, error) {
 			return nil, err
 		}
 
+		updateDefaultProcesses(br.Processes, api.MustParse(bpTOML.BuilpackAPI()), b.PlatformAPI)
```
natalieparellano, line 99: You could also call `bp.API` here to avoid having to modify the interface.

builder_test.go
```diff
@@ -121,8 +121,8 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 						},
 					},
 				}
-				bpA := testmock.NewMockBuildpack(mockCtrl)
-				bpB := testmock.NewMockBuildpack(mockCtrl)
+				bpA := testmock.NewMockBuildpack(mockCtrl, builder.Group.Group[0].API)
```
natalieparellano, line 124: To force `bpA` to return a specific value for `BuildpackAPI` you can do `bpA. EXPECT().BuildpackAPI().Return(my-value)`

builder_test.go
```diff
@@ -310,36 +310,286 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+						}
+					})
+
+					it("should warn when overriding default process type, with non-default process type", func() {
+						bpA := testmock.NewMockBuildpack(mockCtrl, builder.Group.Group[0].API)
+						buildpackStore.EXPECT().Lookup("A", "v1").Return(bpA, nil)
+						bpA.EXPECT().Build(gomock.Any(), config).Return(lifecycle.BuildResult{
+							Processes: []launch.Process{
+								{
+									Type:        "override-type",
```
natalieparellano, line 345: What about `override-with-default` and `override-with-not-default` for these process names? That might make the test a bit easier to read and understand.

builder_test.go
```diff
@@ -310,36 +310,286 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								BuildpackID: "B",
+								Default:     true,
 							},
 						}); s != "" {
 							t.Fatalf("Unexpected:\n%s\n", s)
 						}
 					})
+
+					when("builpack API < 0.6", func() {
+						it("set web processes as default", func() {
```
natalieparellano, line 501:
```suggestion
						it("sets web processes as default", func() {
```

builder.go
```diff
@@ -175,25 +206,54 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
...
+// we keep typeToProcess map in order to delete elements when others are overriding them.
+type orderedProcesses struct {
+	typeToProcess map[string]*list.Element
+	processList   *list.List
+}
 
-func (m processMap) add(l []launch.Process) {
-	for _, proc := range l {
-		m[proc.Type] = proc
+func newProcessStruct() orderedProcesses {
```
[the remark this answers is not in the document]
jabrown85: nit: Personally, I would prefer pretty much anything to the `Struct` literal in the naming of the methods/vars.

builder.go
```diff
@@ -175,25 +206,54 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
...
-func (m processMap) list() []launch.Process {
-	var keys []string
-	for key := range m {
-		keys = append(keys, key)
+func (m orderedProcesses) add(listToAdd []launch.Process) ([]string, error) {
+	result := []string{}
+	for _, proc := range listToAdd {
+		// when we replace a default process type with a non-default process type, add to result
+		if p, ok := m.typeToProcess[proc.Type]; ok {
+			cast, success := (p.Value).(launch.Process)
```
jabrown85, line 229:
```suggestion
			existingProc, success := (p.Value).(launch.Process)
```

or `foundProc` maybe

buildpacktoml.go
```diff
@@ -254,6 +262,20 @@ func (b *BuildpackTOML) readOutputFiles(bpLayersDir, bpPlanPath string, bpPlanIn
 	return br, nil
 }
 
+// check that there are not multiple default processes in a buildpack with different process types
+func validateProcesses(processes []launch.Process) error {
+	defaultType := ""
```
jabrown85, line 267: Does this validation block need to be in a version guard for either platform or bp API?

buildpacktoml_test.go
```diff
@@ -286,32 +286,71 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
+							br, err := bpTOML.Build(lifecycle.BuildpackPlan{}, config)
+							if err != nil {
+								t.Fatalf("Unexpected error:\n%s\n", err)
+							}
+							if s := cmp.Diff(br, lifecycle.BuildResult{
+								BOM:         nil,
+								Labels:      []lifecycle.Label{},
+								MetRequires: nil,
+								Processes: []launch.Process{
+									{Type: "some-type", Command: "some-cmd", BuildpackID: "A", Default: true},
```
jabrown85, line 344: Is it expected the process list will have two with the same type as `Default: true`? I kind of figured the last-in-wins would mean the resulting process collection would only ever show a single entry with `Default: true`.

builder_test.go
natalieparellano: But I think if we follow https://github.com/buildpacks/lifecycle/pull/520/files#r568666640 we won't need it...

builder_test.go
```diff
@@ -310,36 +310,286 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+							bpA := testmock.NewMockBuildpack(mockCtrl, builder.Group.Group[0].API)
+							buildpackStore.EXPECT().Lookup("A", "v1").Return(bpA, nil)
+							bpA.EXPECT().Build(gomock.Any(), config).Return(lifecycle.BuildResult{
+								Processes: []launch.Process{
+									{
+										Type:        "some-type",
```
natalieparellano, line 507: Probably we also want a `web` here (bpA is the newer api, right?) to test that we do NOT set it as default if the buildpack doesn't set it.

exporter.go
```diff
@@ -371,16 +371,28 @@ func (e *Exporter) entrypoint(launchMD launch.Metadata, defaultProcessType strin
...
-			return launch.ProcessPath(launchMD.Processes[0].Type), nil
+
+	if defaultProcessType == "" && e.PlatformAPI.Compare(api.MustParse("0.6")) < 0 && len(launchMD.Processes) == 1 {
+		// if there is only one process, we set it to the default for platform API < 0.6
+		e.Logger.Infof("Setting default process type '%s'", launchMD.Processes[0].Type)
+		return launch.ProcessPath(launchMD.Processes[0].Type), nil
+	}
+
+	if defaultProcessType != "" {
+		defaultProcess, ok := launchMD.FindProcessType(defaultProcessType)
```
jabrown85, line 382: The `launchMD.FindProcessType` finds the first matching process declared in the metadata with the matching type. Is the order of `launchMD.Processes` going to shift at all to make this unexpected for users on `< 0.6`?

I don't _think_ there is anything to worry about here, but I wanted to ask just in case. It looks innocent but if we are storing multiple processes of the same type and have changed that order we may write the incorrect process.

builder_test.go
```diff
@@ -310,36 +310,286 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								Processes: []launch.Process{
+									{
+										Type:        "web",
+										Command:     "web-command",
+										Args:        []string{"web-arg"},
+										Direct:      false,
+										BuildpackID: "B",
+									},
+									{
+										Type:        "other",
```
natalieparellano, line 536: How about `web` and `not-web` for the process names? It might make it easier to understand what the test is doing.

buildpacktoml.go
```diff
@@ -58,6 +58,10 @@ func (b *BuildpackTOML) String() string {
 	return b.Buildpack.Name + " " + b.Buildpack.Version
 }
 
+func (b *BuildpackTOML) BuilpackAPI() string {
```
natalieparellano, line 61:
```suggestion
func (b *BuildpackTOML) BuildpackAPI() string {
```

buildpacktoml_test.go
```diff
@@ -286,32 +286,71 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
+							Processes: []launch.Process{
+								{Type: "some-type", Command: "some-cmd", BuildpackID: "A", Default: true},
+								{Type: "web", Command: "other-cmd", BuildpackID: "A", Default: false},
+							},
+							Slices: []layers.Slice{},
+						}); s != "" {
+							t.Fatalf("Unexpected metadata:\n%s\n", s)
+						}
+					})
+					when("when there are multiple default=true processes but they are all of the same type", func() {
```
natalieparellano, line 319: Hmm... I don't think this is expected behavior. A buildpack shouldn't be able to declare multiple processes of the same type.

testmock/buildpack.go
```diff
@@ -16,6 +16,7 @@ import (
 type MockBuildpack struct {
 	ctrl     *gomock.Controller
 	recorder *MockBuildpackMockRecorder
+	api      string
```
natalieparellano, line 19: This is generated code :)

exporter_test.go
natalieparellano: Hmm... what about passing e.g., `filepath.Join(rootDir, "cnb", "process", "some-process-type"+execExt)` with some os-based constants like here: https://github.com/buildpacks/lifecycle/blob/fae1fe27c0acdfabe9538e3607e72af61c62ede9/cmd/flags.go#L17

natalieparellano: Here's an example of tests where we have os-based duplication: https://github.com/buildpacks/lifecycle/blob/main/archive/reader_test.go

I'd argue that the tests are nevertheless more readable this way...

builder.go
```diff
@@ -175,25 +206,54 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
...
+			if !success {
+				return []string{}, fmt.Errorf("can't cast an element from the list to a process")
+			}
+			if cast.Default && !proc.Default {
+				result = append(result, proc.Type)
+			}
+			m.processList.Remove(p)
+		}
+		m.processList.PushBack(proc)
+		m.typeToProcess[proc.Type] = m.processList.Back()
```
natalieparellano, line 239: I have to confess that this code feels a bit overly complex for me, but maybe that's just a limit of my understanding. Could `processMap` somehow also hold on to the index of the buildpack in `range b.Group.Group`, and then when we have the complete map, we sort it by that index?

buildpacktoml_test.go
```diff
@@ -286,32 +286,71 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
...
+							MetRequires: nil,
+							Processes: []launch.Process{
+								{Type: "some-type", Command: "some-cmd", BuildpackID: "A", Default: true},
+								{Type: "web", Command: "other-cmd", BuildpackID: "A", Default: false},
+							},
+							Slices: []layers.Slice{},
+						}); s != "" {
+							t.Fatalf("Unexpected metadata:\n%s\n", s)
+						}
+					})
```
natalieparellano, line 318:
```suggestion
					})
					
```

buildpacktoml_test.go
natalieparellano, line 319:
```suggestion
					when("there are multiple default=true processes but they are all of the same type", func() {
```

buildpacktoml_test.go
```diff
@@ -523,6 +562,30 @@ func testBuildpackTOML(t *testing.T, when spec.G, it spec.S) {
 					})
 				})
 			})
+
+			when("when there is more than one default=true process", func() {
```
natalieparellano, line 566:
```suggestion
			when("there is more than one default=true process", func() {
```

exporter.go
natalieparellano: I think the previous use of `procMap` ensured that we only ever had one process of a given type.

builder.go
natalieparellano: This is just a suggestion, maybe the current implementation is the best way of doing it...

buildpacktoml_test.go
natalieparellano: I'm okay leaving it this way!

buildpacktoml.go
natalieparellano: I had the same question myself and I'm not sure...

builder.go
yaelharel: You're so right. This is what happens when making changes too late. Thanks Natalie, I'll make this change.

builder.go
yaelharel: mmmm...this is an interesting suggestion. The problem that I see is about overriding. I mean, there could be some numbers in the range that won't exist and then it won't be very easy to sort this map.
If you think I'm wrong and you know how to implement it, I would love to discuss it!

buildpacktoml_test.go
yaelharel: You're right that a builpdpack shouldn't be able to declare multiple processes of the same type but do we check it anywhere?
I don't think that we check it in `builpacktoml_test.go` before introducing the new functionality.
What is the expected behavior?
We thought this is a good practice but maybe we just want to fail.

builder_test.go
yaelharel: I'll create another test for it. We can't do it in the same test because they'll override one another.

buildpacktoml_test.go
natalieparellano: I think it was always possible for an individual buildpack to declare multiple processes of the same type, but because we were putting those processes into a map keyed by type, in the end the list only has one process per type. Maybe we should update the spec to be more clear on this point before introducing a new failure... but I think we could remove this test to avoid suggesting that this is something buildpacks _should_ be doing. What do you think?

buildpacktoml.go
yaelharel: mmmm...this validation is relevant only for `buildpackAPI >= 0.6`, but since for older versions this validation will never fail (because we introduced the `default` field only in `buildpackAPI 0.6`), we thought it's ok to not check it.
What do you think?

exporter.go
yaelharel: That's right. We used `procMap` in the `builder` to keep only one process for each type, and now we're doing it with the new struct `orderedProcesses`.
This is why we need the `map` in this struct.

buildpacktoml_test.go
yaelharel: I totally agree with everything you wrote. I'm just wondering about deleting this test.
I think that I prefer to keep it so the behavior will be clear for the reader. I can write a comment that explains that it's not something that buildpacks should do.
But I definitely think that we should make a spec change and then fail whenever this thing happens.
What do you think?

buildpacktoml_test.go
yaelharel: This is interesting. We apply the "last wins" algorithm only in the `exporter` so the builder doesn't care about it.
Correct me if I'm wrong but you're saying that the `builder` should just write one process per buildpack if there is a default process in it, right?
My biggest concern about it is that I'm not sure if it's the `builder` responsibility to do that but I'm not 100% sure.

builder.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

builder.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

buildpacktoml.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

builder.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

builder.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

testmock/buildpack.go
yaelharel: I deleted it. Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

buildpacktoml_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

buildpacktoml_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

buildpacktoml_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

dwillist:
Just throwing up some manual acceptance resources: 
Steps:
1) build a lifecycle archive off of this branch:
  `LIFECYCLE_VERSION=1.2.3 make build-image-linux`
2) compile a pack version based off of [this PR](https://github.com/buildpacks/pack/pull/1051) (so we can actually test default processes & over-rides)
  `make build (from pack repository root)`
3) using the above `pack` binary use the following `builder.toml` file to make a builder:
```
# Buildpacks to include in builder
[lifecycle]
  uri = "<path/to/lifecycle/tarball/created/in/step/1.tar.gz>"

# Stack that will be used by the builder
[stack]
id = "io.buildpacks.samples.stacks.bionic"
run-image = "cnbs/sample-stack-run:bionic"
build-image = "cnbs/sample-stack-build:bionic"
```

and trust this builder running `pack trust-builder <builder-name>`

4) Use the attached bulidpacks to run a build against the [ruby-bundler samples app](https://github.com/buildpacks/samples/tree/main/apps/ruby-bundler) 
- [ruby-bundler.zip](https://github.com/buildpacks/lifecycle/files/5914529/ruby-bundler.zip)
- [snake-oil.zip](https://github.com/buildpacks/lifecycle/files/5914530/snake-oil.zip)

Unzip the buildpacks and change the process types and buildpack API's in their `buildpack.toml`s 
`pack build test-app --buildpack <path/to/unzipped/ruby-bunder> --buildpack <path/to/unzipped/snake-oil> --builder <builder created in step 3>`

buildpacktoml.go
jabrown85: That sounds good to me!

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/63f4e9a470ff92f133caa5ffff56d047d6fa3295.

buildpacktoml_test.go
jabrown85: Hmm. That is interesting. My expectations were just from reading the proposed change and trying to align that with the test output. 

I kind of feel like it is partly the builder's responsibility, but I think the exporter would still need to do similar validation. Kind of like duplicating client/server validation. I wouldn't be against it existing in both, but as long as it exists in the exporter I'm 👍

natalieparellano: I think if a single buildpack tries to write `default = true` for more than one process it should be a failure even if the processes have the same type. We could deal with the edge case of a single buildpack writing duplicate processes separate from this PR, since it existed before.

buildpacktoml_test.go
natalieparellano: Responded here: https://github.com/buildpacks/lifecycle/pull/520#discussion_r568975384

exporter_test.go
natalieparellano: In general I agree with you that duplication is bad but I think we should also prioritize readability. It's nice when scanning through the tests if you can immediately understand what they are asserting and don't need to navigate elsewhere in the code to see what a helper function is doing. Though I'm already familiar with this code, I found it a bit confusing so I feel that others might as well...

I feel that https://github.com/buildpacks/lifecycle/pull/520#discussion_r568744924 could go a long way toward reducing the os-based duplication though. We'd have duplication of `filepath.Join` but all other logic could remain in the helper.

buildpacktoml.go
natalieparellano: Related [Slack conversation](https://buildpacks.slack.com/archives/CSFE49VGR/p1612291496030300)

builder_test.go
natalieparellano: I was confused as to why this was there before I saw `launch. FindLastDefaultProcessType`

It looks like the current implementation is always returning the processes sorted alphabetically: https://github.com/buildpacks/lifecycle/blob/fae1fe27c0acdfabe9538e3607e72af61c62ede9/builder.go#L186-L197

I am not sure, but I think the exporter _might_ be relying on this ordering when processing the layers. We should probably understand this better.

buildpacktoml_test.go
yaelharel: @natalieparellano, I like your idea that whenever a buildpack is trying to write `default = true` for more than one process, the builder should fail (no matter if the processes have the same or different type).
@jabrown85, are you good with this decision?

buildpacktoml_test.go
yaelharel: I'm marking this conversation as resolved because I think [this](https://github.com/buildpacks/lifecycle/pull/520#discussion_r568685339) conversation is dealing with the same concern..

buildpacktoml.go
yaelharel: It also relates to [this comment](https://github.com/buildpacks/lifecycle/pull/520#discussion_r569081128).

exporter_test.go
yaelharel: I changed this code [here](https://github.com/buildpacks/lifecycle/pull/520/commits/972cf6de0cba7f026367a6dd49532bbf94ded7bf).

builder_test.go
yaelharel: I need to check the `sort.Strings` call but, as you wrote, the reason we introduced the test that is making sure that we're preserving the order, is since starting from `platform API 0.6` with `buildpack API 0.6`, we care about the order of the processes that sets the `default` field. For older APIs, we didn't need that because the user was able to set a default process only with passing the `-process-type` flag.

natalieparellano: It seems the current alphabetical ordering is required for reproducibility ([Slack convo](https://buildpacks.slack.com/archives/CSFE49VGR/p1612364432036100)). We should probably add a test for this, so that we don't accidentally change it in the future.

To communicate "last default process" to the exporter, we could add a new field in metadata.toml.

buildpacktoml.go
natalieparellano: To summarize the various conversations, I think when a buildpack declaring an older buildpack api tries to implement a newer buildpack api feature, we should warn and ignore the configuration. So in this case if a buildpack on 0.5 writes `default = true` to launch.toml, we should warn and set `default = false`.

buildpacktoml_test.go
jabrown85: I'm 👍 on adding additional validation here.

builder.go
natalieparellano: If we add a field to metadata.toml (per https://github.com/buildpacks/lifecycle/pull/520#discussion_r569489888) this code should go away... the builder could just keep track of a single "default process" and update it for each buildpack when processing the build result.

buildpacktoml.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/8ef28ee8fc57b4ca50e83bd20938479fbcaf60e9.

buildpacktoml_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/badd029b1d958fe32cb82734998d22b65e454cb0.

builder.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/48b609971e977b9ae0976f961226955d0e4dbf3f.

builder.go
```diff
@@ -175,25 +203,56 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
...
+	}
+}
 
-func (m processMap) add(l []launch.Process) {
-	for _, proc := range l {
-		m[proc.Type] = proc
+// This function adds the processes from listToAdd to processMap
+// it sets m.defaultType to the last default process
+// if a non-default process overrides a default process, it returns its type and unset m.defaultType
+func (m *processMap) add(listToAdd []launch.Process) string {
```
yaelharel, line 221: This function is a bit complex because we want to warn only if the last default process type is being override by a non default process type.

Example A:
default X --> default Y --> non-default Y
There is no default
We'll warn on Y

Example B:
default X --> default Y --> non-default Y --> default Z
Z is the default
We won't warn on Y

Example C:
default Y --> default X --> non-default Y
X is the default
We won't warn on Y

yaelharel:
@jabrown85 @natalieparellano @ekcasey, we're going to run some manual tests tomorrow, but we've made all of the changes that we talked about so I think it's good time for another review.
Thank you!

cc @dwillist

exporter_test.go
```diff
@@ -1307,6 +1323,13 @@ version = "4.5.6"
 	})
 }
 
+func checkEntrypoint(t *testing.T, image *fakes.Image, entrypointPath string) {
```
natalieparellano, line 1326: I was thinking about our discussion earlier, and I think you're right, there's no need to put the `filepath.Join` in the argument. I think if the call were to read e.g., `assertHasEntrypoint("cnb", "lifecycle", "launcher")` that would be clear enough. It was really the omission of the leading `cnb` that was making it a bit hard to follow before.

builder.go
natalieparellano: I think in example B it's okay to warn when non-default Y replaces default Y, because at that point Y **is** the default. Does that make it any simpler?

launch/launch.go
```diff
@@ -33,6 +34,18 @@ func (m Metadata) FindProcessType(pType string) (Process, bool) {
...
+	defaultFound := false
+	var lastDefaultProcess Process
+	for _, p := range m.Processes {
+		if p.Default {
+			lastDefaultProcess = p
+			defaultFound = true
+		}
+	}
+	return lastDefaultProcess, defaultFound
+}
```
natalieparellano, line 47: Is this function needed anymore, now that we have `Metadata.Default`?

builder.go
```diff
@@ -175,25 +203,57 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
...
+	for proc := range m.typeToProcess {
+		keys = append(keys, proc)
 	}
 	sort.Strings(keys)
-	procs := []launch.Process{}
+	result := []launch.Process{}
 	for _, key := range keys {
-		procs = append(procs, m[key])
+		processWithNoDefault := m.typeToProcess[key]
+		processWithNoDefault.Default = false // we set the default to false so it won't be part of metadata.toml
```
natalieparellano, line 253: Maybe we could add a `NoDefault()` function on `Process`, like we have with `bp.noAPI()` or `bp.noHomepage()`.

builder.go
```diff
@@ -175,25 +203,57 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
...
+// if a non-default process overrides a default process, it returns its type and unset m.defaultType
+func (m *processMap) add(listToAdd []launch.Process) string {
+	result := ""
+	for _, procToAdd := range listToAdd {
+		if procToAdd.Default {
+			m.defaultType = procToAdd.Type
+			result = ""
+		} else {
+			existingProc, ok := m.typeToProcess[procToAdd.Type]
+			if ok && existingProc.Type == m.defaultType { // existingProc.Default = true
```
natalieparellano, line 229:
```suggestion
			procToAdd.Type == m.defaultType { // non-default process overrides a default process
```
I **think** this will achieve the same result

natalieparellano: This function might read a bit easier if you returned the entire warning instead of just the type that was replaced.

builder.go
```diff
@@ -94,10 +94,23 @@ func (b *Builder) Build() (*BuildMetadata, error) {
...
+
 		bom = append(bom, br.BOM...)
 		labels = append(labels, br.Labels...)
 		plan = plan.filter(br.MetRequires)
-		procMap.add(br.Processes)
+
+		replacedDefault := processMap.add(br.Processes)
+		if err != nil {
+			return nil, err
+		}
```
natalieparellano, line 106: This could be removed I believe

builder.go
```diff
@@ -175,25 +203,57 @@ func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
...
+		m.typeToProcess[procToAdd.Type] = procToAdd
 	}
+
+	return result
 }
 
+// list returns a sorted array of processes.
+// The array is sorted based on the process types.
+// The list is sorted for reproducibility.
```
natalieparellano, line 243: Love this additional context in the comment ❤️

metadata.go
```diff
@@ -16,12 +16,13 @@ const (
...
-	Launcher   LauncherMetadata `toml:"-" json:"launcher"`
-	Processes  []launch.Process `toml:"processes" json:"processes"`
-	Slices     []layers.Slice   `toml:"slices" json:"-"`
+	BOM                         []BOMEntry       `toml:"bom" json:"bom"`
+	Buildpacks                  []GroupBuildpack `toml:"buildpacks" json:"buildpacks"`
+	Labels                      []Label          `toml:"labels" json:"-"`
+	Launcher                    LauncherMetadata `toml:"-" json:"launcher"`
+	Processes                   []launch.Process `toml:"processes" json:"processes"`
+	Slices                      []layers.Slice   `toml:"slices" json:"-"`
+	BuildpackDefaultProcessType string           `toml:"buildpack-default-process-type, omitempty" json:"buildpack-default-process-type"`
```
natalieparellano, line 25: The `json` may need an omitempty as well

builder_test.go
```diff
@@ -334,6 +334,273 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+							}
+
+							expected := "Warning: redefining the following default process type with a process not marked as default: override-type"
+							h.AssertStringContains(t, stdout.String(), expected)
+
+							h.AssertEq(t, metadata.BuildpackDefaultProcessType, "")
+						})
+					})
+
+					when("there is a web process", func() {
```
natalieparellano, line 504: We might want to add a test that when platform api < 0.6, we don't set default = true for web. But I'm not sure if it's really necessary.

builder_test.go
```diff
@@ -334,6 +334,273 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+										},
+									},
+								}, nil)
+
+								metadata, err := builder.Build()
+								if err != nil {
+									t.Fatalf("Unexpected error:\n%s\n", err)
+								}
+
+								if s := cmp.Diff(metadata.Processes, []launch.Process{
```
natalieparellano, line 583: Maybe it's worth adding a comment here about why `Default = false` in the metadata.

exporter_test.go
```diff
@@ -918,156 +921,169 @@ version = "4.5.6"
...
+
+			it("saves the image for all provided AdditionalNames", func() {
+				_, err := exporter.Export(opts)
+				h.AssertNil(t, err)
+				h.AssertContains(t, fakeAppImage.SavedNames(), append(opts.AdditionalNames, fakeAppImage.Name())...)
+			})
+		})
 
-				when("platform API is < 0.4", func() {
+		when("default process", func() {
```
natalieparellano, line 940: I'm so happy we pulled this into the top-level context!

natalieparellano:
@yaelharel @dwillist this is looking great! I left couple suggestions, mostly cosmetic. I think we'll want to add the `omitempty` to `json:"buildpack-default-process-type"` to avoid changing the `io.buildpacks.build.metadata` label, but overall this seems to be just about ready.

launch/launch.go
yaelharel: No, we need to remove it.
Thanks!

builder.go
yaelharel: I like this idea. We'll do it.

exporter_test.go
yaelharel: We changed the function name and kept the `filepath.Join` as is.

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/files#diff-5dce70063c4b1b07e611b3fcd32da0445a0b95970028c8992beb9c955e38320fR253.

exporter_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/7b24df17bb72f57cfba08d780795cbaba5aaddca.

launch/launch.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/7b24df17bb72f57cfba08d780795cbaba5aaddca.

builder.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/7b24df17bb72f57cfba08d780795cbaba5aaddca.

metadata.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/7b24df17bb72f57cfba08d780795cbaba5aaddca.

builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/520/commits/7b24df17bb72f57cfba08d780795cbaba5aaddca.

natalieparellano (approved):

natalieparellano:
@jabrown85 did you want to have another look before we merge?

yaelharel:
We did a few more changes thanks to @natalieparellano feedback!

We also ran the following tests manually (see https://github.com/buildpacks/lifecycle/pull/520#issuecomment-771996608 for the exact commands):

- Build a lifecycle image based on our code
- Build a pack with the changes from https://github.com/buildpacks/pack/pull/1051
- Create a builder with the new lifecycle image
- Run some scenarios while passing the `--buildpack` flag in the right order and test the expected results:
    - Print of the default process
    - Warning in case a buildpack overrides the default process
    - Metadata content by getting into the app image and checking the `config/metadata.toml` file (`default=false` and `buildpack-default-process-type=false` isn't printed to the file)

The scenarios that we tested:
- Old buildpack with a web process
⇒ default = web
- New buildpack with a default worker process
⇒ default = worker
- Old buildpack with no web process
⇒ no default
- New buildpack with a default worker process → new buildpack with a default another-worker process
⇒ default = another-worker
- New buildpack with a default worker process → old builpdack with a web process
⇒ default = web
- New buildpack with a default worker process → new buildpack with a non-default worker process
⇒ no default + override warning
- New buildpack with a default Y process → new buildpack with a default X process → new buildpack with a non-default Y process
⇒ default = X (no warning)

builder.go
yaelharel: First, after the changes [here](https://github.com/buildpacks/lifecycle/pull/520/commits/7b24df17bb72f57cfba08d780795cbaba5aaddca), I think that the function is less complicated.

What would you want to do in the following example:
```
default X --> default Y --> non-default Y --> default Z --> default A --> non-default A --> default B
```
Do you think we should warn on both Y and A?
I don't think so but I'm not sure.
If we want to warn on both of them, the function will look like:
```
func (m *processMap) add(listToAdd []launch.Process) string {
	overriddenProcesses := []string
	for _, procToAdd := range listToAdd {
		if procToAdd.Default {
			m.defaultType = procToAdd.Type
		} else if procToAdd.Type == m.defaultType {
			// non-default process overrides a default process
			m.defaultType = ""
                        overriddenProcesses = append(overriddenProcesses, proceToAdd.Type)
		}
		m.typeToProcess[procToAdd.Type] = procToAdd
	}
        warning := fmt.Sprintf("Warning: redefining the following default process type with a process not marked as default: %s\n", strings.Join(overridenProcesses, ", "))
	return warning
}
```
What do you think?

jabrown85 (approved):

PR #527, opened 2021-02-12
Follow symlinks when building env from files.

### LINKED ISSUE #526: Directories are being picked up in `/platform/env` as files
### Summary
<!--- Please provide a general summary of the issue. -->
Detector is failing to load the `/platform/env` due to a symlink to a directory.

### Reproduction

##### Steps
<!--- What steps should be taken to reproduce the issue? -->

1. Create `..data` symlink to `..a_dir` directory in `/platform/env`
2. Run detector
3. See failure referencing `..data` 

##### Current behavior
<!--- What happened? Logs, etc. could go here. -->
It looks as though lifecycle is reading `..data` as a file and not ignoring it as a directory. By default, directories are ignored by the platform env loading seen [here](https://github.com/buildpacks/lifecycle/blob/ca47d3defd627d7ebb16e5e3b98bb9cc9b2d7ca4/env/env.go#L164).

```
===> DETECTING                                                                                                                                                                                                         │
│ ERROR: failed to read full env: read /platform/env/..data: is a directory
```

##### Expected
<!--- What did you expect to happen? -->
I expect detector to succeed in loading the platform env, ignoring the `..data` symlink to the directory.

---

### Context

##### lifecycle version
<!--- If you can find this, it helps us pin down the issue. For example, run `pack inspect-builder BUILDER` which should report the lifecycle version in question. -->
`0.10.1+36da170`

##### anything else?
<!--- Tekton task version, kpack version, etc. -->
This is in kubernetes and I found it by mounting an entire secret without using subpaths. When you do not use subpaths when mounting a configmap or secret, you get symlinks so that kubernetes can keep the file up to date. If you use subpaths, the symlink will not be present and it works as expected. See [here](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod) and [here](https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically)

jabrown85:
Symlinks to directories should be evaluated so they can be skipped when building env.

Fixes #526

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -165,11 +165,7 @@ func eachEnvFile(dir string, fn func(k, v string) error) error {
 			continue
 		}
 		if f.Mode()&os.ModeSymlink != 0 {
-			lnPath, err := filepath.EvalSymlinks(filepath.Join(dir, f.Name()))
-			if err != nil {
-				return err
-			}
-			lnFile, err := os.Stat(lnPath)
+			lnFile, err := os.Stat(filepath.Join(dir, f.Name()))
 			if err != nil {
 				return err
 			}
--- a/env/env.go
+++ b/env/env.go
@@ -164,6 +164,19 @@ func eachEnvFile(dir string, fn func(k, v string) error) error {
 		if f.IsDir() {
 			continue
 		}
+		if f.Mode()&os.ModeSymlink != 0 {
+			lnPath, err := filepath.EvalSymlinks(filepath.Join(dir, f.Name()))
+			if err != nil {
+				return err
+			}
+			lnFile, err := os.Stat(lnPath)
+			if err != nil {
+				return err
+			}
+			if lnFile.IsDir() {
+				continue
+			}
+		}
 		value, err := ioutil.ReadFile(filepath.Join(dir, f.Name()))
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


jabrown85:
TODO: Rebase once https://github.com/buildpacks/lifecycle/pull/528 or similar is merged

natalieparellano (approved):

jabrown85:
Rebased on `main` to fix CI failures

env/env.go
```diff
@@ -164,6 +164,19 @@ func eachEnvFile(dir string, fn func(k, v string) error) error {
...
 		}
+		if f.Mode()&os.ModeSymlink != 0 {
+			lnPath, err := filepath.EvalSymlinks(filepath.Join(dir, f.Name()))
+			if err != nil {
+				return err
+			}
+			lnFile, err := os.Stat(lnPath)
+			if err != nil {
+				return err
+			}
```
ekcasey, line 175:
```suggestion
			lnFile, err := os.Stat(filepath.Join(dir, f.Name()))
			if err != nil {
				return err
			}
```

`os.Stat` already stats the link target (`os.Lstat` would stat the link) so there is no need to evalulate the links first.

ekcasey (approved):

PR #531, opened 2021-02-18
Pass file handle to exec.d on Windows

natalieparellano:


natalieparellano:
This isn't actually ready for feedback

PR #534, opened 2021-02-24
Pass file handle to exec.d on Windows

### LINKED ISSUE #496: Pass file handle to execd on windows
buildpacks/spec#164

natalieparellano:
Resolves #496 


acceptance/variables_windows.go
```diff
@@ -0,0 +1,10 @@
+package acceptance
+
+const (
+	containerBaseImage = "" // Not used in Windows tests
```
jabrown85, line 4: Maybe something like
```
type RuntimeEnv interface {
  BinaryExtension() string
  ContainerBaseImage() string
}

// in test
test := "fromImage=" + runtime.ContainerBaseImage()
```
The windows implementation of that interface could throw `NotImplemented`, which would be much more clear when someone runs into a test that interfaces with a helper that isn't yet supported.

acceptance/launcher_test.go
```diff
@@ -44,21 +47,28 @@ func TestLauncher(t *testing.T) {
...
+
+				expected := fmt.Sprintf("%s was executed\n", execDHelper)
+				expected += fmt.Sprintf("Exec.d Working Dir: %s\n", workDir)
+				expected += fmt.Sprintf("%s was executed\n", execDCheckerHelper)
+				expected += fmt.Sprintf("Exec.d Working Dir: %s\n", workDir)
 				expected += "sourced bp profile\n"
 				expected += "sourced app profile\n"
-				expected += "VAR_FROM_EXEC_D: orig-val:val-from-exec.d:val-from-exec.d-for-process-type-exec.d-checker"
+				if runtime.GOOS == "windows" {
+					expected += `"VAR_FROM_EXEC_D: orig-val:val-from-exec.d:val-from-exec.d-for-process-type-exec.d-checker"`
```
jabrown85, line 67: What causes this difference? Would it make sense to lessen the assertion to a contains to remove this OS conditional you think?

acceptance/launcher_test.go
```diff
@@ -353,3 +363,14 @@ func assertOutput(t *testing.T, cmd *exec.Cmd, expected ...string) {
 		}
 	}
 }
+
+func copyPosixFixture(t *testing.T) {
```
jabrown85, line 367: Is this copying things from `posix/` because the `Dockerfile` is re-building the go binaries? I wonder if building a known named image to use would be more discoverable. Like `buidpacksio/exechelper` and both `Dockerfile` would change to `FROM buidpacksio/exechelper as builder`. This obviously assumes the dockerfile has been built.

ekcasey (commented):
Looks good overall!

acceptance/launcher_test.go
```diff
@@ -353,3 +363,14 @@ func assertOutput(t *testing.T, cmd *exec.Cmd, expected ...string) {
...
+	windowsContext := filepath.Join("testdata", "launcher", "windows")
+	h.AssertNil(t, os.MkdirAll(filepath.Join(windowsContext, "exec.d"), 0755))
+	h.RecursiveCopy(
+		t,
+		filepath.Join("testdata", "launcher", "posix", "exec.d"),
+		filepath.Join(windowsContext, "exec.d"),
+	)
+	os.RemoveAll(filepath.Join(windowsContext, "container", "layers", "0.5_buildpack", "some_layer", "exec.d", "exec.d-checker", ".gitkeep"))
```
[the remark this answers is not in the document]
natalieparellano: I'm not totally sure that it's necessary, but I saw it here: https://github.com/buildpacks/lifecycle/blob/7f1c2db042cfef1a5adbe052a89d30a15a63aa80/acceptance/testdata/launcher/posix/Dockerfile#L11 and when trying to replicate with `del /F` in the Windows Dockerfile I got a permissions issue. I'll experiment to see if it's required...

acceptance/variables_windows.go
natalieparellano: Ah, you are right! This is an outgrowth of the time we tried to make the analyzer acceptance tests work on Windows. I can clean it up...

acceptance/launcher_test.go
natalieparellano: Sounds good... I think this is a difference in how `printf` on Linux and `echo` on Windows process args.

acceptance/launcher_test.go
natalieparellano: Related discussion: https://github.com/buildpacks/lifecycle/pull/534#discussion_r582321165

acceptance/launcher_test.go
```diff
@@ -28,6 +30,7 @@ func TestLauncher(t *testing.T) {
 	if daemonOS == "windows" {
 		launchDockerContext = filepath.Join("testdata", "launcher", "windows")
 		launcherBinaryDir = filepath.Join("testdata", "launcher", "windows", "container", "cnb", "lifecycle")
+		copyPosixFixture(t)
```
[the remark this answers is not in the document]
natalieparellano: I like moving it and modifying the build context!

launch/exec_d_windows.go
```diff
@@ -0,0 +1,22 @@
...
+	"os/exec"
+	"strconv"
+	"syscall"
+)
+
+const EnvExecDHandle = "CNB_EXEC_D_HANDLE"
+
+func setHandle(cmd *exec.Cmd, pw *os.File) error {
+	handle := pw.Fd()
+	if err := syscall.SetHandleInformation(syscall.Handle(handle), syscall.HANDLE_FLAG_INHERIT, 1); err != nil {
```
micahyoung, line 15: I've been using the `golang.org/x/sys/windows` package for all my syscalls these days. They're a mostly just more user-friendly wrappers, though sometimes have better error checking and more defined constants than `syscall`.

Also, it appears that third argument is the correct value but should probably use the same constant instead ([MSFT docs](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-sethandleinformation#:~:text=A%20mask%20that%20specifies)):
> ...
> `dwMask`
> A mask that specifies the bit flags to be changed. **Use the same constants shown in the description of dwFlags**.
> 
> `dwFlags`
> Set of bit flags that specifies properties of the object handle. This parameter can be 0 or one or more of the following values.
> | Value | Meaning |
> |---|---|
> | HANDLE_FLAG_INHERIT 0x00000001 | If this flag is set, a child process created with the bInheritHandles parameter of CreateProcess set to TRUE will inherit the object handle.
> |...| |
 
So the call might look like:
```
windows.SetHandleInformation(windows.Handle(handle), windows.HANDLE_FLAG_INHERIT, windows.HANDLE_FLAG_INHERIT)
```

micahyoung (changes requested):
Looks great! Just a couple tiny changes I think maybe around the syscalls.

acceptance/variables_windows.go
natalieparellano: Removed unneeded var

acceptance/launcher_test.go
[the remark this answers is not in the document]
natalieparellano: It is necessary... but I couldn't delete it inside the docker build. Trying to create the directory in the build instead...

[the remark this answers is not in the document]
natalieparellano: Because I couldn't run `del /F <path to .gitkeep>` during the build I moved the .gitkeep up two directories and created empty directories in the Dockerfile: https://github.com/buildpacks/lifecycle/pull/534/files#diff-b45234da4232089107bcd78584060e81019e23a33c8fe5418e81b2c3cefacf69R8. IDK if there's a better way to do it...

natalieparellano:
I think this is ready for another round!

acceptance/testdata/analyzer/app-image/Dockerfile
```diff
@@ -1,6 +1,4 @@
-ARG fromImage
-
-FROM $fromImage
+FROM ubuntu:bionic
```
ekcasey, line 1: Not necessary for this PR but for some (not all) of these test fixtures I wonder of building them directly with `imgutil` would be easier than managing `Dockerfile` test fixture (and eventually parameterizing for windows etc).

acceptance/testdata/launcher/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN go build -o helper ./src/exec.d
 
 FROM ubuntu:bionic
 
-COPY container /
+COPY posix/container /
```
ekcasey, line 9: Also not necessary for this PR but I probably should have named this directory `linux` when I originally introduced the launcher acceptance tests. It would be significantly less confusing.

launch/exec_d_unix.go
```diff
@@ -0,0 +1,13 @@
...
+
+package launch
+
+import (
+	"os"
+	"os/exec"
+)
+
+func setHandle(cmd *exec.Cmd, pw *os.File) error {
+	cmd.ExtraFiles = []*os.File{pw}
```
ekcasey, line 11:
```suggestion
func setHandle(cmd *exec.Cmd, f *os.File) error {
	cmd.ExtraFiles = []*os.File{f}
```
This function doesn't need to know that the input file is a pipe writer.

launch/exec_d_windows.go
```diff
@@ -0,0 +1,21 @@
...
+	"os"
+	"os/exec"
+
+	"golang.org/x/sys/windows"
+)
+
+const EnvExecDHandle = "CNB_EXEC_D_HANDLE"
+
+func setHandle(cmd *exec.Cmd, pw *os.File) error {
+	handle := pw.Fd()
```
ekcasey, line 14:
```suggestion
func setHandle(cmd *exec.Cmd, f *os.File) error {
	handle := f.Fd()
```

ekcasey (approved):

micahyoung (approved):

acceptance/testdata/launcher/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN go build -o helper ./src/exec.d
```
yaelharel, line 6: Does it make sense to rename this file to `Dockerfile.linux` to keep the convention with `Dockerfile.windows`?
Same comment applies to other files that were renamed.

acceptance/testdata/launcher/exec.d/fd_unix.go
```diff
@@ -0,0 +1,9 @@
+//+build linux darwin
```
yaelharel, line 1: Should we call this file `fd_linux` instead of `fd_unix`? I'm not sure what is more accurate.

acceptance/testdata/launcher/windows/container/layers/config/metadata.toml
```diff
@@ -1,3 +1,8 @@
+[[buildpacks]]
+  id = "0.5/buildpack"
+  version = "0.0.1"
+  api = "0.5"
```
yaelharel, line 4: Don't we want to add it to buildpack api 0.6?
And if yes, is it possible to just use the latest api here?

launch/exec_d_test.go
```diff
@@ -45,7 +45,13 @@ func testExecD(t *testing.T, when spec.G, it spec.S) {
 			h.AssertNil(t, err)
 			wd, err := os.Getwd()
 			h.AssertNil(t, err)
-			path = filepath.Join(tmpDir, "execd")
+
+			exe := ""
```
yaelharel, line 49:
```suggestion
			ext := ""
```

acceptance/testdata/launcher/Dockerfile
natalieparellano: I took this pattern from the `tools/` directory.

acceptance/testdata/launcher/exec.d/fd_unix.go
natalieparellano: In other places where we have _windows.go files we have a _unix.go counterpart. Since it will also compile on Mac I think _unix.go is more accurate.

acceptance/testdata/launcher/windows/container/layers/config/metadata.toml
natalieparellano: Ohh... that's interesting. This is a weird one because technically exec.d support was introduced in buildpack api 0.5 (according to the spec) but for Windows it's different. I am happy to change the fixture to avoid any confusion.

launch/exec_d_test.go
natalieparellano: I chose this var name for parity with https://github.com/buildpacks/lifecycle/blob/4bfb55d7d0a489fb3a3b7f3c7c083ff40fe85499/launch/launcher.go#L18. I don't really care what the var is called but I feel that it would be better for it to be consistent.

acceptance/testdata/launcher/Dockerfile
natalieparellano: (not sure if Dockerfile.linux would be better but just wanting to keep a single convention where possible!)

launch/exec_d_test.go
yaelharel: I don't really care either, so I'm fine with be consistent with what already exists.

yaelharel (approved):

jabrown85 (approved):

Issue #522, opened 2021-02-03
[RFC 0074] Move layer types to new table in layer.toml


ekcasey:
[RFC 0074](https://github.com/buildpacks/rfcs/blob/main/text/0074-layer-table.md)
https://github.com/buildpacks/rfcs/pull/132

https://github.com/buildpacks/spec/issues/185

natalieparellano:
Goes hand-in-hand with https://github.com/buildpacks/lifecycle/issues/395

yaelharel:
#537 was merged. Closing this issue.

PR #537, opened 2021-03-02
Opt in layer caching

### LINKED ISSUE #395: [RFC 0052] Opt-in Layer Caching
[RFC 0052](https://github.com/buildpacks/rfcs/blob/main/text/0052-opt-in-layer-caching.md)
https://github.com/buildpacks/rfcs/pull/99
https://github.com/buildpacks/spec/issues/132

For buildpacks with API >= 0.5 require buildpacks to explicitly indicate intent to reuse layers as describe in the RFC.

For buildpack with API < 0.5, keep the existing behavior.

yaelharel:
- Set the launch, cache and build flags to false when restoring layer metadata

- Move the build, launch and cache flags into a types table in the toml file for buildpack API >= 0.6
Fail if the flags are at the top level

- For buildpack API < 0.6, the flags should be at the top level of the toml file
Warn and ignore if the flags are in a types table

- The exporter should ignore the launch, build and cache flags if they're not in the right format

- Rename <layers>/<layer> to <layers>/<layer>.ignore for buildpack API >= 0.6 if all of the types flags are set to false

Fixes #395 and #522


yaelharel:
A small comment about the tests that are testing old buildpack API (buildpack API < 0.6).
Some of them appear in the middle of the unit tests and some of them are at the end, separate from the tests that are testing the new buildpack API.
I think that the later is better because it's clearer and if we decide to stop supporting old buildpack API at some day, it's will be easier to delete the irrelevant tests.
Although I didn't put all of the new tests at the end of each unit test file because it meant to change the structure of some of the tests and I didn't want to make big changes that are irrelevant to the issues that this PR is trying to solve.
I think that we should create a refactoring PR to deal with these changes.

buildpack/build_test.go
```diff
@@ -460,7 +478,7 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 				})
 			})
 
-			it("should error when launch.toml is not writable", func() {
+			it.Pend("should error when launch.toml is not writable", func() { // TODO: ask Natalie - do we need this test?
```
yaelharel, line 481: @natalieparellano

layers.go
```diff
@@ -106,41 +106,35 @@ func (bd *bpLayersDir) findLayers(f func(layer bpLayer) bool) []bpLayer {
...
-	if _, err := toml.DecodeFile(tomlPath, &data); err != nil {
+	layerMetadataFile, err := buildpack.DecodeLayerMetadataFile(tomlPath, bp.API)
+	if err != nil {
 		return platform.BuildpackLayerMetadata{}, err
 	}
 	sha, err := ioutil.ReadFile(bp.path + ".sha")
 	if err != nil {
 		if os.IsNotExist(err) {
-			return data, nil
+			return platform.BuildpackLayerMetadata{LayerMetadata: platform.LayerMetadata{SHA: ""}, LayerMetadataFile: layerMetadataFile}, nil
```
jabrown85, line 133: Won't `SHA` already be `""`?

yaelharel: Yes it is, I just thought to make it clearer because a few lines after, we set it to `string(sha)`.
I'll remove it if you don't think it's necessary.
I'm good either way.

jabrown85: I'm fine either way, just making sure I understood. If it is for clarity, I think that is fine

buildpack/files.go
```diff
@@ -198,3 +202,111 @@ type LayerMetadataFile struct {
...
+		return LayerMetadataFile{Data: lmtf.Data, Build: lmtf.Types.Build, Launch: lmtf.Types.Launch, Cache: lmtf.Types.Cache}, nil
+	}
+	var lmf LayerMetadataFile
+	if _, err := toml.DecodeFile(path, &lmf); err != nil {
+		return LayerMetadataFile{}, err
+	}
+	return lmf, nil
+}
+
+func typesInTopLevel(path string) (bool, error) {
```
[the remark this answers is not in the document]
yaelharel: Great idea!

buildpack/build_test.go
```diff
@@ -572,6 +590,25 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+						filepath.Join(appDir, "layers-A-v1", "layer"),
+					)
+					h.Mkfile(t,
+						"build=true\ncache=true\nlaunch=true",
+						filepath.Join(appDir, "layers-A-v1", "layer.toml"),
+					)
+
+					_, err := bpTOML.Build(buildpack.Plan{}, config)
+					h.AssertNotNil(t, err)
+					expected := "the launch, cache and build flags should be in the types table"
```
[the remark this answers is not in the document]
yaelharel: I don't really care but I'm just curious to know - why do you think an Oxford comma is necessary here?

buildpack/build_test.go
```diff
@@ -853,6 +890,24 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 				expected := "Warning: default processes aren't supported in this buildpack api version. Overriding the default value to false for the following processes: [type-with-default]"
 				h.AssertStringContains(t, stdout.String(), expected)
 			})
+
+			when("the launch, cache and build flags are in the types table", func() {
```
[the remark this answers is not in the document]
yaelharel: Same question as [here](https://github.com/buildpacks/lifecycle/pull/537#discussion_r586738135).

buildpack/build_test.go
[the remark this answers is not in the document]
jabrown85: It is not necessary at all. Merely a suggestion. I prefer them, but I am not an english major. It just popped out aesthetically to me 😃

analyzer.go
```diff
@@ -89,6 +89,7 @@ func (a *Analyzer) analyzeLayers(appMeta platform.LayersMetadata, cache Cache) e
 				continue
 			}
 			a.Logger.Infof("Restoring metadata for %q from app image", identifier)
+			layer.UnsetFlags()
```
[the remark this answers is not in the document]
yaelharel: As far as I understand, we would like to set all of the flags to false just before writing them (calling `a.writeLayerMetadata`).
In your example, when `launch = false`, we won't get to `writeLayerMetadata`, meaning we won't write this layer's metadata to a toml file.
So I'm not sure how to add it to `MetadataForBuildpack` and whether it's the right thing to do.

buildpack/files.go
```diff
@@ -198,3 +202,111 @@ type LayerMetadataFile struct {
...
+	defer fh.Close()
+
+	var lmtf layerMetadataTomlFile
+	if _, err = toml.DecodeFile(path, &lmtf); err != nil {
+		return false, err
+	}
+	return lmtf.Types.Build || lmtf.Types.Launch || lmtf.Types.Cache, nil
+}
+
+func typesInRightFormat(path, buildpackAPI string) bool {
```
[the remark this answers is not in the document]
yaelharel: I've made the separation between `typesInRightFormat` and `checkTypesFormat` because it felt more clear for me but I'm open to make changes.
Talking about your suggestion, I'm not sure how to actually do it while keeping it simple because for `buildpackAPI < 0.6`, we would like to send a warning if needed, and for `buildpackAPI >= 0.6`, we would like to return an error if needed.
What do you think?

buildpack/build.go
```diff
@@ -81,13 +81,47 @@ func (b *Descriptor) Build(bpPlan Plan, config BuildConfig) (BuildResult, error)
...
 	if err := b.setupEnv(config.Env, bpLayersDir); err != nil {
 		return BuildResult{}, err
 	}
 
 	return b.readOutputFiles(bpLayersDir, bpPlanPath, bpPlan, config.Out)
 }
 
+func renameLayerDirIfNeeded(layerDir, buildpackAPI string) error {
+	// rename <layers>/<layer> to <layers>/<layer>.ignore if buildpack API >= 0.6 and all of the types flags are set to false
+	if api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) >= 0 && areTypesFalse(layerDir+".toml", buildpackAPI) {
```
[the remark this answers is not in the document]
yaelharel: I agree that it's pretty ugly that we're parsing the `buildpackAPI` over and over again.
I like the idea of doing something like:
```
func (b *Descriptor) checkTypesFormat(layersDir string, out io.Writer) error {
  if api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) < 0 {
    return eachDir...
  }
  return eachDir...
}
```
Did you mean to do this or to do something else?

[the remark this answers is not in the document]
jabrown85: Yeah, that is what I had in mind.

buildpack/files.go
```diff
@@ -198,3 +202,111 @@ type LayerMetadataFile struct {
...
+
+	if supportsTypesTable(buildpackAPI) {
+		var lmtf layerMetadataTomlFile
+		if _, err = toml.DecodeFile(path, &lmtf); err != nil {
+			return LayerMetadataFile{}, err
+		}
+		return LayerMetadataFile{Data: lmtf.Data, Build: lmtf.Types.Build, Launch: lmtf.Types.Launch, Cache: lmtf.Types.Cache}, nil
+	}
+	var lmf LayerMetadataFile
+	if _, err := toml.DecodeFile(path, &lmf); err != nil {
```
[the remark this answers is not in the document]
yaelharel: Following your idea [here](https://github.com/buildpacks/lifecycle/pull/537#discussion_r585936785), there are 3 `toml.DecodeFile` in this file.
2 in the `DecodeLayerMetadataFile` function, each one to decode to the right format.
1 in the `typesInRightFormat` function including checking the return value of `toml.MetaData`.
Do you think there is a way to reduce it while keeping the code clear and simple?

buildpack/files.go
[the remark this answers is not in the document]
jabrown85: We could probably make the signature `validateLayerTypes(path, buildpackAPI string, out io.Writer
) error`. In the `< 0.6` it would never return an error but possibly write a warning.

analyzer.go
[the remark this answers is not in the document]
jabrown85: That makes sense. What about `a.writeLayerMetadata` taking in a parameter to ignore the layer values? I'm wondering what the value of modifying the layers returned by `appMeta.MetadataForBuildpack` is. Asked another way, If code outside of this loop starts interacting with the `appLayers` collection, do they expect the flags to be unset or what they were in the metdata file?

buildpack/files.go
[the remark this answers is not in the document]
yaelharel: What do you think about the following idea:
Having one function `DecodeLayerMetadataFile` that returning `LayerMetadataFile, error, bool`.
The boolean will be `true` if the toml file was in the expected format depending on the `buildpack API`, and `false` otherwise.
Also, when calling `toml.DecodeFile`, it'll be into the following `struct`:
type layerMetadataTomlFile struct {
	Data  interface{} `toml:"metadata"`
	Types typesTable  `toml:"types"`
         Build  bool        `json:"build" toml:"build"`
	Launch bool        `json:"launch" toml:"launch"`
	Cache  bool        `json:"cache" toml:"cache"`
}
In this way we won't lose data and we'll be able to call `toml.DecodeFile` only once.

analyzer.go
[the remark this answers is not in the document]
yaelharel: Interesting question.
I don't think that we're planning on interacting with the `appLayers` outside of this loop but if we would, I would guess we would like these flags to be false.
We can pass a flag to `writeLayerMetadata` but I'm not sure whether it'll be clearer.
What do you think?

buildpack/build.go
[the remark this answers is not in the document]
yaelharel: Please see my idea [here](https://github.com/buildpacks/lifecycle/pull/537#discussion_r586784095).

buildpack/files.go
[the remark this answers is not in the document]
jabrown85: I was just thinking about a small utility function to at least reduce the boilerplate around `toml.Decode` and the pointer
```
func decodeLayerTOMLFile(path string) (layerMetadataTomlFile, error)
{
    var lmtf layerMetadataTomlFile
    if _, err = toml.DecodeFile(path, &lmtf); err != nil {
        return LayerMetadataFile{}, err
    }
    return lmtf, nil
}
```

Calling code is just a bit shorter
```
if supportsTypesTable(buildpackAPI) {
     lmtf, err := decodeLayerTOMLFile(path); 
     if err != nil {
	return LayerMetadataFile{}, err
     }
     return LayerMetadataFile{Data: lmtf.Data, Build: lmtf.Types.Build, Launch: lmtf.Types.Launch, Cache: lmtf.Types.Cache}, nil

}
return decodeLayerTOMLFile(path);
```
It may not be worth it, but that is what I had in mind when I wrote the comment!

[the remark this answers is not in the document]
yaelharel: mmmmm but I think that `decodeLayerTOMLFile` cannot be called as is for old buildpack API and for new buildpack API.
Right?

analyzer.go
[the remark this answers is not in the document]
jabrown85: The flag to `writeLayerMetadata` only makes sense if we expect `appLayers` to be remain original values. Right now `appLayers` is a mix. The layers that are being restored are all updated to `false`. The layers that are not getting restored are their original values. That is what prompted my original comment.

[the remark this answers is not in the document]
yaelharel: Interesting. Although I'm not sure it'll be clearer, you convinced me that this is the right thing to do.
Thanks!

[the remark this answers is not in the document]
yaelharel: Actually, thinking about it again, `writeLayerMetadata` doesn't even need to get a flag.
It should always ignore the values of the flags.

buildpack/files.go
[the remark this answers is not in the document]
jabrown85: Your idea seems like a good path to explore! 

We could also explore two distinct structs and them both implementing a `LayerTOML` interface `type LayerTOMLFile interface { Build() bool, Launch() bool, Warnings() []string etc...}`. A single parse driven by the Buildpack API and the translation to the interface would be implemented by the struct itself.

The metadata from the parsing of the toml could be stored in the structs and interrogated by a `Warnings() []string` method when needed. If you want to explore something like that, I can try and pair with you sometime. It might be bigger than we are ready to tackle for this PR tho 😛 

```
func (l layerTOMLv2) Build() bool {
 return l.Types.Build
}

func (l layerTOMLv1) Warnings() []string {
 if l.parseMetadata.IsDefined("types") { ...  }
}
```

analyzer.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/de2f8e688570dc8f2c17f94b92b944c5652cc6e0.

buildpack/build.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/de2f8e688570dc8f2c17f94b92b944c5652cc6e0

buildpack/files.go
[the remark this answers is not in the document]
yaelharel: Interesting!
I saw your comment after pushing the commit https://github.com/buildpacks/lifecycle/pull/537/commits/de2f8e688570dc8f2c17f94b92b944c5652cc6e0.
Please review it and let's discuss your idea vs my commit tomorrow.

buildpack/files.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/de2f8e688570dc8f2c17f94b92b944c5652cc6e0.

buildpack/build.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/de2f8e688570dc8f2c17f94b92b944c5652cc6e0.

buildpack/files.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/de2f8e688570dc8f2c17f94b92b944c5652cc6e0.

yaelharel:
Hi @jabrown85, thanks a lot for your feedback!
I pushed a new PR that address most of your comments: https://github.com/buildpacks/lifecycle/pull/537/commits/de2f8e688570dc8f2c17f94b92b944c5652cc6e0.
As I wrote https://github.com/buildpacks/lifecycle/pull/537#discussion_r587054086, I saw your comment only after pushing my commit and your idea is interesting.
Let's review the two directions together tomorrow and decide what is the best approach for this.

buildpack/files.go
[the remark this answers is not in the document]
jabrown85: Nice improvements! Another struct makes the intent clear and the mapping is explicit.

jabrown85 (approved):
This LGTM!

buildpack/build.go
```diff
@@ -147,17 +192,17 @@ func (b *Descriptor) runBuildCmd(bpLayersDir, bpPlanPath string, config BuildCon
 }
 
 func (b *Descriptor) setupEnv(buildEnv BuildEnv, layersDir string) error {
```
yaelharel, line 194: Right now, we decode each toml file at least 2 times - once as part of `checkTypesFormat`, and at least once in this function (`setupEnv`).
I'm wondering if we can merge this function with `checkTypesFormat` so we decode each toml file only once. Then we won't need that `isBuild` will decode the file, we'll just check the `Build` flag in the struct.
@ekcasey, any thoughts?

buildpack/files.go
```diff
@@ -192,9 +197,57 @@ func containsName(unmet []Unmet, name string) bool {
...
+
+	for _, encoder := range encoders {
+		if encoder.IsSupported(buildpackAPI) {
+			return encoder.Encode(fh, lmf)
+		}
+	}
+	return errors.New("couldn't find an encoder")
+}
+
+func DecodeLayerMetadataFile(path, buildpackAPI string) (layertypes.LayerMetadataFile, string /*warning/error*/, error) {
```
jabrown85, line 228: This isn't something that should change in this PR, but I wonder if we should introduce a pattern for raising errors that could be sometimes emitted as logs/warnings. The caller code might look like:

```go
decodedTOML, err := decoder.Decode(path, buildpackAPI); 
if err != nil {
  if errors.IsTerminal(err) { 
    return err
  }
  out.Write([]byte(err.Error()))
}
... processing logic with decodedTOML...
```

buildpack/files.go
```diff
@@ -192,9 +197,57 @@ func containsName(unmet []Unmet, name string) bool {
...
+func EncodeFalseFlags(lmf layertypes.LayerMetadataFile, path, buildpackAPI string) error {
+	fh, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	defer fh.Close()
+
+	lmf.UnsetFlags()
+
+	encoders := []EncoderDecoder{
```
jabrown85, line 215: nit: something like `DefaultEncoders()` would allow us to add/remove a version encoder in one spot in the future

buildpack/v06/encoderdecoder.go
```diff
@@ -0,0 +1,56 @@
...
+	Launch bool `toml:"launch"`
+	Cache  bool `toml:"cache"`
+}
+
+type layerMetadataTomlFile struct {
+	Data  interface{} `toml:"metadata"`
+	Types typesTable  `toml:"types"`
+}
+
+type EncoderDecoder06 struct {
```
jabrown85, line 24: Can this struct be private? `encoderDecoder06`

buildpack/v05/encoderdecoder.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"os"
+
+	"github.com/BurntSushi/toml"
+
+	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/buildpack/layertypes"
+)
+
+type EncoderDecoder05 struct {
```
jabrown85, line 12: Can this struct be private? `encoderDecoder05`

buildpack/v05/encoderdecoder.go
```diff
@@ -0,0 +1,42 @@
...
+func (d *EncoderDecoder05) IsSupported(buildpackAPI string) bool {
+	return api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) < 0
+}
+
+func (d *EncoderDecoder05) Encode(file *os.File, lmf layertypes.LayerMetadataFile) error {
+	return toml.NewEncoder(file).Encode(lmf)
+}
+
+func (d *EncoderDecoder05) Decode(path string) (layertypes.LayerMetadataFile, string, error) {
+	var lmf layertypes.LayerMetadataFile
```
jabrown85, line 28: I think I would rather see a private struct here to decode into and map to `layertypes.LayerMetadataFile` like we do in the `06` encoder. It's not a big deal, but this decode path going directly to the resulting struct means we could introduce new fields and the `05` users would be able to populate those fields.

jabrown85 (approved):
I love the improvements! 💜  I added a few comments/nits, but I am very +1 to this!

buildpack/v06/encoderdecoder.go
yaelharel: Unless I'm missing something, I don't think it's possible.
We're returning this struct in `NewEncoderDecoder` and calling to all of the functions on it from `buildpack/files.go`.

buildpack/v05/encoderdecoder.go
yaelharel: The same as https://github.com/buildpacks/lifecycle/pull/537#discussion_r587935792. Am I missing something?

buildpack/v05/encoderdecoder.go
yaelharel: I totally agree with what you're suggesting but I'm wondering whether we should introduce this struct now or whether we should wait for the time when we need to add new fields.
I lean a bit to the later since if it takes time until we introduce new fields, someone who isn't familiar with the code might wonder why we add this redundant struct.
What do you think?

buildpack/v06/encoderdecoder.go
jabrown85: As long as you aren't constructing structs directly via `decoder := EncoderDecoder06{}`, `NewEncoderDecoder` and the methods you want to expose is enough. The struct itself can remain private and not exposed to devs importing the package. This would restrict them to use `NewEncoderDecoder` to create a `dncoderDecoder06` as well.

buildpack/v05/encoderdecoder.go
jabrown85: Replied in the other thread.

buildpack/v05/encoderdecoder.go
jabrown85: I don't have a strong opinion on this, but I personally lean toward creating explicit structs for each version and treating the `layertypes.LayerMetadataFile` as the internal interface we code against outside of parsing/decoding/encoding. My main concern is around not remembering that `EncoderDecoder05` is _directly_ decoding to the internal interface.  But again, no strong preference.

buildpack/v06/encoderdecoder.go
yaelharel: I'm sorry, my fault. I got an error last night from the linter and I was too tired to actually read it :)
Anyway, you're right. Please see https://github.com/buildpacks/lifecycle/pull/537/commits/2373122d4321c49d25e9d495e98b6755013b4095.

buildpack/v05/encoderdecoder.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/2373122d4321c49d25e9d495e98b6755013b4095

buildpack/v05/encoderdecoder.go
yaelharel: Following https://github.com/buildpacks/lifecycle/pull/537/commits/2373122d4321c49d25e9d495e98b6755013b4095 and a discussion I had with @jromero, I'm wondering whether we want to keep `buildpack/files.go` as is or split it so each go file will be related to a different file.
If we do it, in order to prevent circular imports, we'll need to redefine the `EncoderDecoder` interface in `v05/encoderdecorder.go` and `v06/encoderdecorder.go`, and we'll need to define `LayerMetadataFile` in `v05/encoderdecorder.go`.
@jabrown85 @natalieparellano, what do you think?
@natalieparellano, during the refactoring you did recently, I remember that there was a reason that you didn't split `buildpack/files.go` but I don't remember what it was. I guess it can be related to our decision.

jabrown85: I think @natalieparellano recently had a change in opinion on not splitting them up in the first place. As far as the `EncoderDecoder` interface, it should be in a different package than the version specific implementations. The interfaces can live under `lifecycle` in a single place perhaps. Sort of like docker does [here](https://github.com/moby/moby/blob/master/client/interface.go) with the `client` package. All the interfaces living in `lifecycle` will allow the `test` and `vXX` packages to reference them. It also makes it very obvious if a `vXX` package is relying on another `vXX` package instead of the defined interface.

buildpack/build_test.go
natalieparellano: I think so? This test was there before I did the refactor. Is there a reason to remove it?

yaelharel: Yes, the way we fake that `launch.toml` is not writable is by making it a directory, and it makes us troubles with the new functionality.
@ekcasey and I were wondering whether we can delete this test or should we think about a different way to fake a not-writable file.
What do you think?

cache_test.go
```diff
@@ -280,6 +268,61 @@ func testCache(t *testing.T, when spec.G, it spec.S) {
...
+							initializeCache(t, exporter, &testCache, cacheDir, layersDir, metadataTemplate)
+						})
+						it("sets cache metadata", func() {
+							err := exporter.Cache(layersDir, testCache)
+							h.AssertNil(t, err)
+
+							metadata, err := testCache.RetrieveMetadata()
+							h.AssertNil(t, err)
+
+							t.Log("adds layer shas to metadata")
```
natalieparellano, line 313: I noticed we don't have a `SHA` assertion here - is that intended?

exporter_test.go
```diff
@@ -273,6 +273,25 @@ func testExporter(t *testing.T, when spec.G, it spec.S) {
...
+					exporter.Buildpacks = []buildpack.GroupBuildpack{{ID: "other.buildpack.id", API: api.Buildpack.Latest().String()}}
+					fakeAppImage.AddPreviousLayer("local-non-reusable-layer-digest", "")
+					opts.OrigMetadata = platform.LayersMetadata{
+						Buildpacks: []platform.BuildpackLayersMetadata{{
+							ID:     "other.buildpack.id",
+							Layers: map[string]platform.BuildpackLayerMetadata{"local-non-reusable-layer": {LayerMetadata: platform.LayerMetadata{SHA: "local-non-reusable-layer-digest"}}},
+						}},
+					}
+				})
+				it("ignores the flag and doesn't reuse cached launch layers even if the local sha matches the sha in the metadata", func() {
```
natalieparellano, line 287: I believe this was originally proposed as a failure case - did it change?

analyzer_test.go
```diff
@@ -99,10 +104,11 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				h.AssertNil(t, json.Unmarshal(metadata, &appImageMetadata))
 			})
 
-			it("restores layer metadata", func() {
+			it("restores layer metadata and unset the launch, build and cache flags", func() {
```
natalieparellano, line 107:
```suggestion
			it("restores layer metadata and unsets the launch, build and cache flags", func() {
```

buildpack/build_test.go
```diff
@@ -340,6 +341,23 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 					}
 				})
 			})
+
+			when("the launch, cache and build flags are false", func() {
+				it("rename <layers>/<layer> to <layers>/<layer>.ignore", func() {
```
natalieparellano, line 346:
```suggestion
				it("renames <layers>/<layer> to <layers>/<layer>.ignore", func() {
```

buildpack/files.go
natalieparellano: Is the `/*warning/error*/` intended?

analyzer_test.go
```diff
@@ -384,19 +420,20 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					h.AssertNil(t, testCache.SetMetadata(cacheMetadata))
 					h.AssertNil(t, testCache.Commit())
 
-					analyzer.Buildpacks = append(analyzer.Buildpacks, buildpack.GroupBuildpack{ID: "escaped/buildpack/id"})
+					analyzer.Buildpacks = append(analyzer.Buildpacks, buildpack.GroupBuildpack{ID: "escaped/buildpack/id", API: api.Buildpack.Latest().String()})
 				})
 
-				it("restores cache=true layer metadata", func() {
+				it("restores cache=true layer metadata and unset the launch, build and cache flags", func() {
```
natalieparellano, line 426:
```suggestion
				it("restores cache=true layer metadata and unsets the launch, build and cache flags", func() {
```

analyzer_test.go
```diff
@@ -111,9 +117,35 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
 			})
 
+			when("buildpack API < 0.6", func() {
+				it.Before(func() {
+					analyzer.Buildpacks = []buildpack.GroupBuildpack{
+						{ID: "metadata.buildpack", API: "0.5"},
+						{ID: "no.cache.buildpack", API: "0.5"},
+					}
+				})
+				it("restores layer metadata and unset the launch, build and cache flags in top level", func() {
```
natalieparellano, line 131:
```suggestion
				it("restores layer metadata and unsets the launch, build and cache flags in top level", func() {
```

buildpack/build.go
```diff
@@ -81,13 +82,57 @@ func (b *Descriptor) Build(bpPlan Plan, config BuildConfig) (BuildResult, error)
...
+	// rename <layers>/<layer> to <layers>/<layer>.ignore if buildpack API >= 0.6 and all of the types flags are set to false
+	if !layerMetadataFile.Launch && !layerMetadataFile.Cache && !layerMetadataFile.Build {
+		if err := os.Rename(layerDir, layerDir+".ignore"); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (b *Descriptor) checkTypesFormat(layersDir string, out io.Writer) error {
```
natalieparellano, line 106: It seems this function is doing a bit more than just checking the types, perhaps the name could be more descriptive?

analyzer_test.go
```diff
@@ -195,13 +227,14 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					h.AssertNil(t, testCache.SetMetadata(cacheMetadata))
 					h.AssertNil(t, testCache.Commit())
 
-					analyzer.Buildpacks = append(analyzer.Buildpacks, buildpack.GroupBuildpack{ID: "escaped/buildpack/id"})
+					analyzer.Buildpacks = append(analyzer.Buildpacks, buildpack.GroupBuildpack{ID: "escaped/buildpack/id", API: api.Buildpack.Latest().String()})
 				})
 
-				it("restores app and cache layer metadata", func() {
+				it("restores app and cache layer metadata and unset the launch, build and cache flags", func() {
```
natalieparellano, line 233:
```suggestion
				it("restores app and cache layer metadata and unsets the launch, build and cache flags", func() {
```

analyzer_test.go
```diff
@@ -288,20 +322,22 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					h.AssertPathDoesNotExist(t, filepath.Join(layerDir, "metadata.buildpack", "cache-false.sha"))
 				})
 
-				it("restores escaped buildpack layer metadata", func() {
+				it("restores escaped buildpack layer metadata and unset the launch, build and cache flags", func() {
```
natalieparellano, line 325:
```suggestion
				it("restores escaped buildpack layer metadata and unsets the launch, build and cache flags", func() {
```

natalieparellano:
This is looking really good! I left a few comments. It's awesome to see how clean the implementation turned out.

buildpack/files.go
yaelharel: Yes. One of the return values is a string that will be shown as an error or a warning depending on the buildpack API version.
If you have a better comment, I would love to change the current one :)

cache_test.go
yaelharel: Wrong copy-paste. Thanks!

buildpack/files.go
natalieparellano: Hmm... why couldn't `Decoder.Decode` return the message in an error when it is supposed to be an error?

exporter_test.go
yaelharel: I don't think it should be a failure case. The exporter ignores flags that aren't in the "right place" (at the types table for buildpack API < 0.6, and at the top level for buildpack API >= 0.6).
Am I missing something?

natalieparellano: I think in practice we wouldn't get here, because it would fail in the build phase: https://github.com/buildpacks/lifecycle/pull/537/files#diff-0fd47ba02691a569b10e75c699543569da61397717979463429aa1a4a5c5b287R594

Probably it's better to be consistent, just for ease of comprehension.

buildpack/build.go
yaelharel: I agree but I'm not sure what is the right name.
The function:
- Checks the types format and error/warn if the format is wrong
- Rename the layer directory to `.ignore` if the buildpack API >= 0.6 and all the flags are set to false.

What do you think about `doTypesManupulations`?
Any better ideas?

natalieparellano: Hmm... `processLayerTypes`? IDK!

analyzer_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/1f8bae1526197856d67ae94b129c8e87e77bb576.

analyzer_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/1f8bae1526197856d67ae94b129c8e87e77bb576.

analyzer_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/1f8bae1526197856d67ae94b129c8e87e77bb576.

analyzer_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/1f8bae1526197856d67ae94b129c8e87e77bb576.

buildpack/build_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/1f8bae1526197856d67ae94b129c8e87e77bb576.

analyzer_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/1f8bae1526197856d67ae94b129c8e87e77bb576.

cache_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/1f8bae1526197856d67ae94b129c8e87e77bb576.

exporter_test.go
yaelharel: Ohh, of course that we would fail in the build phase, but I wanted to be clear that the exporter doesn't check it again and will just ignore it.
I can delete this test but I'm not sure it's the right thing to do.
I can add a comment about it.
What do you think?

buildpack/files.go
yaelharel: Because for buildpack API < 0.6, we return a warning in this case: https://github.com/buildpacks/lifecycle/pull/537/files#diff-e348e4b0ed5d889979d2fb076558d56656b133c84892137fdd110002e8948ec8R113-R116

analyzer_test.go
```diff
@@ -111,9 +117,33 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				} {
 					got := h.MustReadFile(t, filepath.Join(layerDir, data.name))
 					h.AssertStringContains(t, string(got), data.want)
+					h.AssertStringContains(t, string(got), unsetFlags)
 				}
 			})
 
+			when("buildpack API < 0.6", func() {
```
natalieparellano, line 124: I think we might have an edge case where the previous <layer>.toml was written by a buildpack with the same id but implementing a different api. In this case I think we should ensure that we restore the file in the format that the current buildpack is expecting to find it. This might involve some changes to ensure that we're using the right encoder & decoder as they could be different.

yaelharel: I don't think that we have any problem here because we didn't change the `.json` format. So in your example, the `analyzer` will read the `.json` file of the old buildpack and will write the `.toml` file in the new format.

buildpack/v05/encoderdecoder.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/d6aa6e72ae980771d30efd95383086f0e1995f67.

buildpack/build.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/d6aa6e72ae980771d30efd95383086f0e1995f67.

exporter_test.go
natalieparellano: Hmm... I'd almost argue that the exporter shouldn't ignore it, and should have the same behavior WRT inconsistent data as the builder. I think we did something similar when deprecating version from the BOM - builder failed out with data that it shouldn't get because detector would have failed. IDK!

buildpack/build_test.go
ekcasey: I think we can just delete it.

buildpack/files.go
```diff
@@ -3,8 +3,13 @@
 package buildpack
 
 import (
+	"errors"
 	"fmt"
+	"os"
 
+	"github.com/buildpacks/lifecycle/buildpack/layertypes"
+	v05 "github.com/buildpacks/lifecycle/buildpack/v05"
+	v06 "github.com/buildpacks/lifecycle/buildpack/v06"
```
ekcasey, line 12:
```suggestion
	"github.com/buildpacks/lifecycle/buildpack/v05"
	"github.com/buildpacks/lifecycle/buildpack/v06"
```

ekcasey: Also maybe `api05` and `api06` would be better names?

buildpack/layertypes/types.go
```diff
@@ -0,0 +1,8 @@
+package layertypes
```
ekcasey, line 1: I see why this might have been convenient (to avoid import cycles), but I don't love having a separate package just for this type. I wonder if we could make an interface in the buildpack package
```
type Layer interface {
    Metadata() interface{}
    Types() LayerTypes
}

type LayerTypes {
	Build  bool
	Launch bool
	Cache  bool 
}
```
And have the `v0x` packages return something that implements the interface. Then we could have different structs representing the different file schemas, and which would be quite clear. `Encode` and `Decode` could accept a Layer interface type.

buildpack/v05/encoderdecoder.go
```diff
@@ -0,0 +1,42 @@
...
+type EncoderDecoder05 struct {
+}
+
+func NewEncoderDecoder() *EncoderDecoder05 {
+	return &EncoderDecoder05{}
+}
+
+func (d *EncoderDecoder05) IsSupported(buildpackAPI string) bool {
+	return api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) < 0
+}
```
ekcasey, line 21: Is it weird to call it `EncoderDecorder05` but have it work for APIs `0.2-0.5` maybe we want to create a package for each supported api? It would be a lot of duplication but it would be explicit and afterwards we could start pushing other API dependent logic into this packages?

We don't need to do this now but I wonder if our lives would be easier in the future if we just pushed everything into an api package and started a new API by copy paste the package and making the desired changes. The top level `buildpack` package could just just the buildpack API and delegate, so higher level workflows like `builder` wouldn't need to know about the api packages.

cache_test.go
```diff
@@ -280,6 +268,62 @@ func testCache(t *testing.T, when spec.G, it spec.S) {
...
+					       "cache": true,
+					       "sha": "%s",
+					       "data": {"old":"data"}
+					     }
+					   }
+					 }
+					]
+					}`
+							initializeCache(t, exporter, &testCache, cacheDir, layersDir, metadataTemplate)
+						})
```
ekcasey, line 305:
```suggestion
						})
						
```

cache_test.go
```diff
@@ -280,6 +268,62 @@ func testCache(t *testing.T, when spec.G, it spec.S) {
...
+					   "layers": {
+					     "cache-true-layer": {
+					       "cache": true,
+					       "sha": "%s",
+					       "data": {"old":"data"}
+					     }
+					   }
+					 }
+					]
+					}`
```
ekcasey, line 303: I think we should either remove the deep indentation here (so it's realistic) or indent it further (so it looks nice).

exporter_test.go
```diff
@@ -273,6 +273,25 @@ func testExporter(t *testing.T, when spec.G, it spec.S) {
...
+				it.Before(func() {
+					exporter.Buildpacks = []buildpack.GroupBuildpack{{ID: "other.buildpack.id", API: api.Buildpack.Latest().String()}}
+					fakeAppImage.AddPreviousLayer("local-non-reusable-layer-digest", "")
+					opts.OrigMetadata = platform.LayersMetadata{
+						Buildpacks: []platform.BuildpackLayersMetadata{{
+							ID:     "other.buildpack.id",
+							Layers: map[string]platform.BuildpackLayerMetadata{"local-non-reusable-layer": {LayerMetadata: platform.LayerMetadata{SHA: "local-non-reusable-layer-digest"}}},
+						}},
+					}
+				})
```
ekcasey, line 286:
```suggestion
				})
				
```

exporter_test.go
ekcasey: I agree with @natalieparellano that we should redo the check. We are unlikely to get into this situation but I like the consistency.

layers.go
```diff
@@ -106,41 +107,35 @@ func (bd *bpLayersDir) findLayers(f func(layer bpLayer) bool) []bpLayer {
 	return selectedLayers
 }
 
-func (bd *bpLayersDir) newBPLayer(name string) *bpLayer {
+func (bd *bpLayersDir) newBPLayer(name, buildpackAPI string) *bpLayer {
```
ekcasey, line 110: Seems like we should be eventually pushing this down into the `buildpack` package with the other buildpack API dependent logic.

testdata/exporter/app-slices/layers/buildpack.id/layer2.toml
```diff
@@ -1,4 +1,5 @@
-cache = true
-launch = true
+[types]
+  cache = true
+  launch = true
 [metadata]
  layer2key = "layer2val"
```
ekcasey, line 5:
```suggestion
 layer2key = "layer2val"
 
```

testdata/exporter/app-slices/layers/other.buildpack.id/new-launch-layer.toml
```diff
@@ -1 +1,2 @@
-launch = true
\ No newline at end of file
+[types]
+  launch = true
```
ekcasey, line 2:
```suggestion
  launch = true

```

testdata/exporter/escaped-bpid/layers/some_escaped_bp_id/some-layer.toml
```diff
@@ -1 +1,2 @@
-launch = true
\ No newline at end of file
+[types]
+  launch = true
```
ekcasey, line 2:
```suggestion
  launch = true

```

testdata/exporter/previous-image-exists/layers/buildpack.id/new-launch-layer.toml
```diff
@@ -1,3 +1,4 @@
-launch = true
+[types]
+  launch = true
 [metadata]
   somekey = "someval"
```
ekcasey, line 4:
```suggestion
  somekey = "someval"

```

testdata/exporter/app-slices/layers/other.buildpack.id/new-launch-layer.toml
ekcasey: I know it was already like this, but we really should have a new line at the end of all these data files.

testdata/exporter/previous-image-exists/layers/old.buildpack.id/launch-layer.toml
```diff
@@ -0,0 +1,2 @@
+[types]
+  launch = true
```
ekcasey, line 2:
```suggestion
  launch = true

```

testdata/exporter/previous-image-exists/layers/old.buildpack.id/local-non-reusable-layer/layer5contents
```diff
@@ -0,0 +1 @@
+this contents match the metadata (by sha)
```
ekcasey, line 1:
```suggestion
this contents match the metadata (by sha)

```

testdata/exporter/previous-image-exists/layers/other.buildpack.id/local-non-reusable-layer/layer5contents
```diff
@@ -0,0 +1 @@
+this contents match the metadata (by sha)
```
ekcasey, line 1:
```suggestion
this contents match the metadata (by sha)

```

testdata/exporter/previous-image-exists/layers/other.buildpack.id/new-launch-layer.toml
```diff
@@ -1 +1,2 @@
-launch = true
\ No newline at end of file
+[types]
+  launch = true
```
ekcasey, line 2:
```suggestion
  launch = true

```

testdata/exporter/previous-image-not-exist/layers/buildpack.id/layer2.toml
```diff
@@ -1,4 +1,5 @@
-cache = true
-launch = true
+[types]
+  cache = true
+  launch = true
 [metadata]
  layer2key = "layer2val"
```
ekcasey, line 5:
```suggestion
  layer2key = "layer2val"

```

ekcasey (commented):
I left some architectural suggestion that I see as 'nice-to-have's rather than 'must-have's.

However, I do think using a real logger for messages printed by the builder is important, so we get consistent log level configurability and formatting.

buildpack/build_test.go
yaelharel: Thanks!
Please see https://github.com/buildpacks/lifecycle/pull/537/commits/9bd6a05415c17939ad6fa3105eae95823aac568f.

buildpack/build.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/9bd6a05415c17939ad6fa3105eae95823aac568f.

analyzer.go
```diff
@@ -109,7 +109,7 @@ func (a *Analyzer) analyzeLayers(appMeta platform.LayersMetadata, cache Cache) e
 				continue
 			}
 			a.Logger.Infof("Restoring metadata for %q from cache", identifier)
-			if err := a.writeLayerMetadata(buildpackDir, name, layer); err != nil {
+			if err := a.writeLayerMetadataFlags(buildpackDir, name, layer); err != nil {
```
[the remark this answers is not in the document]
yaelharel: I'm not sure why I renamed it from `writeLayerMetadata` to `writeLayerMetadataFlags`. Probably there was a reason at some step of the refactoring but I agree that `writeLayerMetadata` makes more sense here. I'll rename it.
Thanks!

analyzer_test.go
```diff
@@ -288,20 +320,22 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					h.AssertPathDoesNotExist(t, filepath.Join(layerDir, "metadata.buildpack", "cache-false.sha"))
 				})
 
-				it("restores escaped buildpack layer metadata", func() {
+				it("restores escaped buildpack layer metadata and unsets the launch, build and cache flags", func() {
```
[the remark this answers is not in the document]
yaelharel: I'm ok with removing it completely.
What do you mean by "Once per top-level when seems like enough."?
Where would you put it?
I guess you're not talking about adding it with "cache exists", right?

testdata/exporter/previous-image-not-exist/layers/buildpack.id/layer2.toml
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/previous-image-exists/layers/other.buildpack.id/new-launch-layer.toml
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/previous-image-exists/layers/other.buildpack.id/local-non-reusable-layer/layer5contents
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/previous-image-exists/layers/old.buildpack.id/local-non-reusable-layer/layer5contents
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/previous-image-exists/layers/old.buildpack.id/launch-layer.toml
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/previous-image-exists/layers/buildpack.id/new-launch-layer.toml
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/escaped-bpid/layers/some_escaped_bp_id/some-layer.toml
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/app-slices/layers/other.buildpack.id/new-launch-layer.toml
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

testdata/exporter/app-slices/layers/buildpack.id/layer2.toml
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

exporter_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

cache_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

buildpack/files.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

analyzer.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/53c486316a8f616bd11b4c5ecb4a2a304c6b9a93.

layers.go
yaelharel: In order to do it, we need to make a big change since we need this information in the following flow (and in some other similar flows):
```
# exporter.go
addBuildpackLayers
    bpDir, err := readBuildpackLayersDir --> calls to newBPLayer
    bpDir.findLayers(forLaunch) --> calls to read --> calls to buildpack.DecodeLayerMetadataFile (needs the buildpack API version)
```

I'll keep it as is for now and we can make this change in a different PR.

buildpack/build.go
```diff
@@ -81,13 +82,57 @@ func (b *Descriptor) Build(bpPlan Plan, config BuildConfig) (BuildResult, error)
...
+
+func (b *Descriptor) processLayerTypes(layersDir string, out io.Writer) error {
+	if api.MustParse(b.API).Compare(api.MustParse("0.6")) < 0 {
+		return eachDir(layersDir, b.API, func(path, buildpackAPI string) error {
+			_, msg, err := DecodeLayerMetadataFile(path+".toml", buildpackAPI)
+			if err != nil {
+				return err
+			}
+			if msg != "" {
+				if _, err = out.Write([]byte(msg)); err != nil {
```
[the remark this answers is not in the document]
natalieparellano: Now that you pointed it out, I think we are also doing something similar here: https://github.com/buildpacks/lifecycle/blob/11ca8ef99c1626fa34ea087e74993d8746c7cf61/buildpack/build.go#L291 and here: https://github.com/buildpacks/lifecycle/blob/11ca8ef99c1626fa34ea087e74993d8746c7cf61/builder.go#L79

We'll probably want to pass a logger to the builder (like we do with analyzer). We could probably also pass this logger to the encoder/decoder to avoid the weirdness mentioned here: https://github.com/buildpacks/lifecycle/pull/537#discussion_r587842939 and here: https://github.com/buildpacks/lifecycle/pull/537#discussion_r593278527

buildpack/v05/encoderdecoder.go
yaelharel: I agree that calling it `EncoderDecoder05` while it's relevant for buildpack APIs `0.2-0.5` is very weird!
But I'm also not a fan of having a lot of duplicated code.
I think it really depends on what happens in the next versions. If there are many changes every version so probably it's worth having a package per version, but if there won't be any changes in the next few versions, maybe it's not worth having so much code duplicated for no reason.
I'll keep it as is for now and will add a comment that we need to work on it.

buildpack/build.go
```diff
@@ -81,13 +82,57 @@ func (b *Descriptor) Build(bpPlan Plan, config BuildConfig) (BuildResult, error)
...
+		})
+	}
+	return eachDir(layersDir, b.API, func(path, buildpackAPI string) error {
+		layerMetadataFile, msg, err := DecodeLayerMetadataFile(path+".toml", buildpackAPI)
+		if err != nil {
+			return err
+		}
+		if msg != "" {
+			return errors.New(msg)
+		}
```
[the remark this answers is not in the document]
yaelharel: We had this discussion before and I'm not sure.
If anyone has a strong opinion on this, I can make this change, otherwise I'll keep it as is for now.

analyzer_test.go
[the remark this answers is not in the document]
yaelharel: Following our discussion, I'll delete this part only from this specific test.

exporter_test.go
yaelharel: Just for the record - what really convinced me to redo the check is that someone can theoretically replace only our build phase but keep our export phase.

yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/10e7898473c8b663ac444985d8b6fe429f57b8ba.

analyzer_test.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/10e7898473c8b663ac444985d8b6fe429f57b8ba.

buildpack/v05/encoderdecoder.go
yaelharel: Added a comment: https://github.com/buildpacks/lifecycle/pull/537/commits/10e7898473c8b663ac444985d8b6fe429f57b8ba

cache_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/10e7898473c8b663ac444985d8b6fe429f57b8ba.

layers.go
yaelharel: In order to prevent a big refactoring, I also had to add `Logger` to `newBPLayer` so it's now part of the `bpLayer` struct (see https://github.com/buildpacks/lifecycle/pull/537/commits/10e7898473c8b663ac444985d8b6fe429f57b8ba).
I added  a `TODO` comment so we'll change it in a different PR.

buildpack/build.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/537/commits/f80c0ccd77f7f3c0564ce125570dd551fe44b19f.
I didn't pass the `Logger` to the `Decoder` for now since the interface is defined in the `buildpack` package and in order to prevent circular imports, we can't import it in the `v05` and `v06` packages.
I'm planning to refactor these files a bit in order to get rid of the `layertypes` package, so I might do it in one of the next commits.

buildpack/build.go
```diff
@@ -147,17 +181,17 @@ func (b *Descriptor) runBuildCmd(bpLayersDir, bpPlanPath string, config BuildCon
 }
 
 func (b *Descriptor) setupEnv(buildEnv BuildEnv, layersDir string) error {
-	if err := eachDir(layersDir, func(path string) error {
-		if !isBuild(path + ".toml") {
+	if err := eachDir(layersDir, b.API, func(path, buildpackAPI string) error {
+		if !isBuild(path+".toml", buildpackAPI) {
```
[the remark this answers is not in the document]
yaelharel: As you wrote, we'll leave this change to a different PR :)

buildpack/layertypes/types.go
yaelharel: Unless I miss understood your suggestion, I don't think it's possible.
Tried to make a simple example to show the problem: https://play.golang.org/p/AxPN3bKNx8U
Please let me know if I'm missing something.
In the meantime, I'll leave the code as is.

ekcasey: Yeah I screwed up this suggestion a bit. You couldn't have `Layer.Types()` return a `LayerTypes` without defining `LayerTypes` in a third package still. You would need something like:
```
type Layer interface {
    Metadata() interface{}
    IsBuild() bool
    IsLaunch() bool
    IsCache() bool
}
```

ekcasey: Alright, after looking at this more closely I realize my suggestion doesn't actually work in this case, it would require some rethinking. I'll resolve this.

buildpack/files.go
yaelharel: I added a TODO comment in the following commit: https://github.com/buildpacks/lifecycle/pull/537/commits/3ca28cd072aa3714b4528c4cf7062f7ef0d094c4.
I thought about adding the `Logger` to the `DecodeLayerMetadataFile` function and then for `buildpack API < 0.6`, we'll use the logger to print a warning if needed (in case the flags are in a types table), and for `buildpack API >= 0.6`, we'll just return an error if needed (in case the flags are at the top level).
The only "problem" is that the `Logger` interface is defined in the `buildpack` package and we can't import it to the `v05` and `v06` packages because we'll have circular imports.
We can redefine the `Logger` interface in each of these packages, but I think that instead we should use the `layertypes` package and define it once there. We should probably rename this package and we can also decide to make it an "interfaces package".
I'll leave it for a different PR.

buildpack/build.go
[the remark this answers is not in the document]
yaelharel: Just for the record - the commit https://github.com/buildpacks/lifecycle/pull/537/commits/f3807d339969c123835663f8f65132b17cd2a8f6 is also relevant.

natalieparellano (approved):

yaelharel:
@natalieparellano  and I ran the following manual tests (each scenario is independent so all of the changes need to be reverted between the scenarios).

### Set the environment
Build a lifecycle binary
Update `~/workspace/samples/builders/bionic/builder.toml` with this binary
```pack builder create my-builder --config ../samples/builders/bionic/builder.toml```

### Scenarios for buildpack API < 0.6
```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/buildpack.toml
# set the buildpack api to 0.5
```

- Set the `launch` flag at the top level → re-build → the flags values don’t change

```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# add the following lines:
# testlayer="$layers_dir"/test
# mkdir -p "$testlayer"
# echo -e “launch = true” > "$layers_dir/test.toml"

pack build test-old --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder

vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# delete the lines that you added before and add the following line:
# cat "$layers_dir/test.toml"

pack build test-old --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder
### expected output:
# build = false
# launch = true
# cache = false

docker inspect test-old | jq
### expected output - the metadata.json contains the right values:
# "io.buildpacks.lifecycle.metadata": … "build\":false,\"launch\":true,\"cache\":false
```

- Set the `cache` flag under a types table + set the `launch` flag at the top level → get a warning + successful build → re-build → check that the `cache` flag isn’t set to true

```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# add the following lines:
# testlayer="$layers_dir"/test
# mkdir -p "$testlayer"
# echo -e “launch = true\n[types]\n  cache = true” > "$layers_dir/test.toml"

pack build test-old-warning --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder
### expected output:
# Warning: Types table isn't supported in this buildpack api version. The launch, build and cache flags should be in the top level. Ignoring the values in the types table.
# Successfully built image test-old-warning

vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# delete the lines that you added before and add the following line:
# cat "$layers_dir/test.toml"

pack build test-old-warning --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder
### expected output:
# build = false
# launch = true
# cache = false
```

### Scenarios for buildpack API >= 0.6
```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/buildpack.toml
# set the buildpack api to 0.6
```

- Set the `launch` flag at the top level → get an error

```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# add the following lines:
# testlayer="$layers_dir"/test
# mkdir -p "$testlayer"
# echo -e “launch = true” > "$layers_dir/test.toml"

pack build test-new-error --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder
### expected output:
# ERROR: failed to build: the launch, cache and build flags should be in the types table of /layers/samples_bash-script/test.toml
```

- Set the `launch` flag under a types table → re-build → at the beginning of the build script, the flags are already set to false

```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# add the following lines:
# testlayer="$layers_dir"/test
# mkdir -p "$testlayer"
# echo -e “[types]\n  launch = true” > "$layers_dir/test.toml"

pack build test-new --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder

vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# delete the lines that you added before and add the following line:
# cat "$layers_dir/test.toml"

pack build test-new --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder
### expected output:
# [types]
#   build = false
#   launch = false
#   cache = false

docker inspect test-new | jq
### expected output - the metadata.json contains the right values:
# "io.buildpacks.lifecycle.metadata": … "build\":false,\"launch\":true,\"cache\":false
```

### Scenarios for renaming the layers directory to `.ignore` if buildpack API >= 0.6 and all of the flags are set to false
```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/buildpack.toml
# set the buildpack api to 0.6

cp ~/workspace/samples/apps/bash-script/bash-script-buildpack ~/workspace/samples/apps/bash-script/bash-script-buildpack-1

vim ~/workspace/samples/apps/bash-script/project.toml
# add the new buildpack so the file will look like this:
# [project]
# id = "io.buildpacks.bash-script"
# name = "Bash Script"
# version = "1.0.0"
# 
# [build]
# exclude = [
#     "README.md",
#     "bash-script-buildpack",
#     "bash-script-buildpack-1"
# ]
# 
# [[build.buildpacks]]
# uri = "bash-script-buildpack/"
# 
# [[build.buildpacks]]
# uri = "bash-script-buildpack-1/"
```

- `bash-script-buildpack` implements buildpack API 0.6 + creates a layer where all of the flags are set to false → the layer is renamed to `.ignore` before `bash-script-buildpack-1` is running

```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# add the following lines:
# testlayer="$layers_dir"/test
# mkdir -p "$testlayer"
# echo -e “[types]\n  launch = false” > "$layers_dir/test.toml"

vim ~/workspace/samples/apps/bash-script/bash-script-buildpack-1/bin/build
# add the following line:
# ls "/layers/samples_bash-script"

pack build test-new-ignore --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder
### expected output:
# launch.toml
# test.ignore
# test.toml
```

- `bash-script-buildpack` implements buildpack API 0.5 + creates a layer where all of the flags are set to false → the layer isn’t renamed to `.ignore` before `bash-script-buildpack-1` is running

```
vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/buildpack.toml
# set the buildpack api to 0.5

vim ~/workspace/samples/apps/bash-script/bash-script-buildpack/bin/build
# add the following lines:
# testlayer="$layers_dir"/test
# mkdir -p "$testlayer"
# echo -e “launch = false” > "$layers_dir/test.toml"

vim ~/workspace/samples/apps/bash-script/bash-script-buildpack-1/bin/build
# add the following line:
# ls "/layers/samples_bash-script"

pack build test-old-ignore --path ~/workspace/samples/apps/bash-script --builder my-builder --trust-builder
### expected output:
# launch.toml
# test
# test.toml
```

PR #541, opened 2021-03-05
Fixes analyzer acceptance tests for Windows

micahyoung:
This enables almost all the analyzer acceptance tests for Windows. This does not yet fix other acceptance tests for Windows, as this is intended to introduce the patterns those tests follow in later commits.

In general, uses tests helpers and variables for OS differences. Existing helpers workaround the following limitations on windows:
* Windows can't seed volumes from stopped containers so it copies to seeds to container paths instead
* Windows can't use "host" network so it access the registry on the default network, using any IP with a "/32" in the daemon's "insecure-registry" (from imgutil Registry helper)


acceptance/variables/variables.go
```diff
@@ -1,14 +1,53 @@
 package variables
 
-const (
-	ContainerBaseImage = "ubuntu:bionic"
-	DockerfileName     = "Dockerfile"
-	VolumeHelperImage  = "busybox"
-)
+type VariableHelper struct {
```
natalieparellano, line 3: In the exec.d PR I decided to move this os-specific logic into the `acceptance` package: https://github.com/buildpacks/lifecycle/blob/11ca8ef99c1626fa34ea087e74993d8746c7cf61/acceptance/variables_unix.go

I don't know if it was the right call. I'm happy to embrace this pattern instead...

testhelpers/docker.go
```diff
@@ -49,15 +49,34 @@ func DockerRun(t *testing.T, image string, ops ...DockerCmdOp) string {
 	return Run(t, exec.Command("docker", append([]string{"run", "--rm"}, args...)...))
 }
 
-func DockerRunAndCopy(t *testing.T, containerName, copyDir, image, path string, ops ...DockerCmdOp) string {
+//DockerRunAndCopy runs a container and once stopped, outputCtrPath is copied to outputDir
+func DockerRunAndCopy(t *testing.T, containerName, outputDir, outputCtrPath, image string, ops ...DockerCmdOp) string {
```
natalieparellano, line 53: This signature does read better :)

acceptance/analyzer_test.go
```diff
@@ -364,32 +372,37 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					})
 
 					when("no auth registry", func() {
+						var noAuthRegCacheImage string
+
 						it.Before(func() {
 							metadata := minifyMetadata(t, filepath.Join("testdata", "analyzer", "cache_image_metadata.json"), platform.CacheMetadata{})
-							cacheImage, cacheAuthConfig = buildNoAuthRegistryImage(
+							cacheImageName := "some-cache-image-" + h.RandString(10)
+							authRegCacheImage, _ = buildRegistryImage(
```
natalieparellano, line 380: Maybe we could rename this helper `buildAuthRegistryImage`

acceptance/analyzer_test.go
```diff
@@ -364,32 +372,37 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
...
 								t,
-								"some-cache-image-"+h.RandString(10),
+								cacheImageName,
 								filepath.Join("testdata", "analyzer", "cache-image"),
-								"--build-arg", "fromImage="+variables.ContainerBaseImage,
+								"--build-arg", "fromImage="+vh.ContainerBaseImage(),
 								"--build-arg", "metadata="+metadata,
 							)
+
+							noAuthRegCacheImage = noAuthRegistry.RepoName(cacheImageName)
```
natalieparellano, line 388: I found myself scratching my head wondering "when do we push this image" but now I understand (I think) this works because we're using the shared storage volume. I feel like we might want to document this in a comment somewhere...

testhelpers/docker.go
```diff
@@ -49,15 +49,34 @@ func DockerRun(t *testing.T, image string, ops ...DockerCmdOp) string {
...
 	args := formatArgs([]string{image}, ops...)
 
 	output := Run(t, exec.Command("docker", append([]string{"run"}, args...)...))
-	Run(t, exec.Command("docker", "cp", containerName+":"+path, copyDir))
+	Run(t, exec.Command("docker", "cp", containerName+":"+outputCtrPath, outputDir))
 	return output
 }
 
+//DockerSeedRunAndCopy copies srcDir to container's srcCtrPath before container is started. Once stopped, outputCtrPath is copied to outputDir
+//Only works for WCOW when seeding to container directory (not a mounted volume)
```
natalieparellano, line 63:
```suggestion
//On WCOW, only works when seeding to container directory (not a mounted volume)
```

natalieparellano: At first I assumed this function only works for WCOW

natalieparellano:
@micahyoung it's awesome to see this! I added a couple nits / comments on clarity but overall this is looking great!

acceptance/variables/variables.go
micahyoung: Good call. I replaced `VariableHelper` with `const`s in `variables.go` (and a `ctrPath()` function for some simple path conversion)

micahyoung:
Trying to get this to green again after rebasing... Should be ready shortly

micahyoung:
Looks like this [Docker engine issue](https://docs.docker.com/engine/release-notes/#:~:text=docker%20start%20--attach) was biting us (usage https://github.com/buildpacks/lifecycle/pull/541/files#diff-4315aabae04f875bdf06b568ddd38191e3f377f95cf1899e55fa2f2741b14bf7R70).

I'll have to check again but I'm hoping runner's docker CLI got bumped (or will shortly) and this should get back to green.

.github/workflows/build.yml
```diff
@@ -42,6 +42,34 @@ jobs:
...
+      - name: BUGFIX - Downgrade Docker CLI if 20.10.0
+        shell: powershell
+        run: |
+          $dockerVersion=(docker version -f '{{.Client.Version}}')
+          if ($dockerVersion -ne "20.10.0") {
+            echo Newer docker version has been released: $dockerVersion
+            echo Remove this step if greater than: 20.10.5
+          }
+
+          curl.exe -o "$env:USERPROFILE\go\bin\docker.exe" -L https://github.com/StefanScherer/docker-cli-builder/releases/download/19.03.14/docker.exe
```
micahyoung, line 54: This step downgrades the docker CLI from 20.10.0 to 19.03.14 to workaround this [bug](https://docs.docker.com/engine/release-notes/#:~:text=docker%20start%20--attach) that get's fixed in 20.10.5: 
> Revert docker/cli#2960 to fix hanging in docker start --attach
See the `DockerSeedRunAndCopy` comments for more details

testhelpers/docker.go
```diff
@@ -54,15 +49,34 @@ func DockerRun(t *testing.T, image string, ops ...DockerCmdOp) string {
...
+func DockerSeedRunAndCopy(t *testing.T, containerName, srcDir, srcCtrPath, outputDir, outputCtrPath, image string, ops ...DockerCmdOp) string {
+	ops = append(ops, WithFlags("--name", containerName))
+	args := formatArgs([]string{image}, ops...)
+
+	output := Run(t, exec.Command("docker", append([]string{"create"}, args...)...))
+	output += Run(t, exec.Command("docker", "cp", srcDir, containerName+":"+srcCtrPath))
+	output += Run(t, exec.Command("docker", "start", "--attach", containerName))
```
micahyoung, line 70: This is the line that does not work with docker CLI 20.10.0 but works in every other version of Docker, hence the downgrade in the GHA workflow. I attempted to rewrite this to avoid this syntax but couldn't find a simple, cross-platform implementation.

.github/workflows/build.yml
micahyoung: FWIW, a 20.10.5 CLI could be built from source via `go get` but this would slow down every run

testhelpers/docker.go
```diff
@@ -96,17 +110,19 @@ func PushImage(dockerCli dockercli.CommonAPIClient, ref string, auth string) err
 	return nil
 }
 
+//SeedDockerVolume only works with Linux daemons as Windows only mounts volumes for started containers
 func SeedDockerVolume(t *testing.T, srcPath string) string {
 	volumeName := "test-volume-" + RandString(10)
 	containerName := "test-volume-helper-" + RandString(10)
+	volumeHelperImage := "alpine"
```
micahyoung, line 117: Since this helper can only be used on Linux, I made the image name local. Eventually, when Windows implements this functionality, this function will also be ported.

testhelpers/docker.go
micahyoung: Worked around this by using 20.10.5 in GHA

natalieparellano (approved):
This is awesome! Thank you @micahyoung :)

PR #543, opened 2021-03-09
Add output before saving image during export to indicate progress

jabrown85:
Resolves: https://github.com/buildpacks/lifecycle/issues/500

### DIFF
--- a/save.go
+++ b/save.go
@@ -15,7 +15,7 @@ import (
 func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (platform.ImageReport, error) {
 	var saveErr error
 	imageReport := platform.ImageReport{}
-	logger.Infof("Publishing %s\n", image.Name())
+	logger.Infof("Saving %s\n", image.Name())
 	if err := image.Save(additionalNames...); err != nil {
 		var ok bool
 		if saveErr, ok = err.(imgutil.SaveError); !ok {
--- a/save.go
+++ b/save.go
@@ -15,6 +15,7 @@ import (
 func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (platform.ImageReport, error) {
 	var saveErr error
 	imageReport := platform.ImageReport{}
+	logger.Infof("Publishing %s\n", image.Name())
 	if err := image.Save(additionalNames...); err != nil {
 		var ok bool
 		if saveErr, ok = err.(imgutil.SaveError); !ok {
--- a/save.go
+++ b/save.go
@@ -15,7 +15,7 @@ import (
 func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (platform.ImageReport, error) {
 	var saveErr error
 	imageReport := platform.ImageReport{}
-	logger.Infof("Saving %s\n", image.Name())
+	logger.Infof("Saving %s...\n", image.Name())
 	if err := image.Save(additionalNames...); err != nil {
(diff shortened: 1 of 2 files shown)


save.go
```diff
@@ -15,6 +15,7 @@ import (
 func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (platform.ImageReport, error) {
 	var saveErr error
 	imageReport := platform.ImageReport{}
+	logger.Infof("Publishing %s\n", image.Name())
```
jabrown85, line 18: I thought about listing all the image names like we do in the `*** Images` section. I also thought about a more generic message like `Pushing image(s)...`. Let me know what you think!

yaelharel: Maybe it's better that we'll print the same value as we print in the following line: https://github.com/buildpacks/lifecycle/blob/203221553c114ecbad7378e16ff02385dd86d23d/save.go#L34
I think that printing `image.Name()` is clearer but maybe there is a reason that we print `shortID(id)` in the other place.

jabrown85: Will the `id` be available via `image.Identifier()` _before_ the push? I thought that was likely invalid to use until a push to a registry had occurred.

yaelharel: ohhhh interesting. I'm not sure :\
If it's not, maybe you can just add `image.Name` to the print after the push (in addition to `shortID(id)`).
I think that in this way it'll be easier to the reader to see the progress.

jabrown85: The output already has the name it looks like
```
Publishing 192.168.86.20:5005/myrepo/img **new line
*** Images (sha256:cb66dc1599dd7be223b5dc3e88dc9f2eef1aec0aaf609c56aef2cb3ee394ff84):
      192.168.86.20:5005/myrepo/img
```

jabrown85: All the tags are output for the image that was pushed, so I think it lines up ok.

yaelharel: You're right. I missed that we're printing more information after https://github.com/buildpacks/lifecycle/blob/203221553c114ecbad7378e16ff02385dd86d23d/save.go#L34
Thanks for the clarification!

yaelharel (approved):

ekcasey (approved):

save.go
```diff
@@ -15,6 +15,7 @@ import (
 func saveImage(image imgutil.Image, additionalNames []string, logger Logger) (platform.ImageReport, error) {
 	var saveErr error
 	imageReport := platform.ImageReport{}
+	logger.Infof("Saving %s\n", image.Name())
```
natalieparellano, line 18:
```suggestion
	logger.Infof("Saving %s...\n", image.Name())
```

natalieparellano: IDK if this is really needed it is just a preference. Having tried a build with this branch out, I must say the message when there's a pause is psychologically satisfying :)

natalieparellano (approved):

save.go
natalieparellano: Actually there's no `...` on the other "Doing XXX" log messages so what we have now is probably more consistent.

natalieparellano: @jabrown85 I'm sorry... I made this comment and then tried to take it back :/ I am totally fine with whatever we decide to do :)

jabrown85: I liked the `...` suggestion. But if someone feels strongly that we shouldn't introduce that, I'll back it out.

PR #545, opened 2021-03-15
Condense exit codes

natalieparellano:
Pending spec PR: https://github.com/buildpacks/spec/pull/196

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -101,6 +101,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		it("errors", func() {
 			cmd := exec.Command(
 				"docker", "run", "--rm",
+				"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
 				analyzeImage,
 				analyzerPath,
 			)
@@ -116,6 +117,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		it("warns", func() {
 			output := h.DockerRun(t,
 				analyzeImage,
+				h.WithFlags("--env", "CNB_PLATFORM_API="+latestPlatformAPI),
 				h.WithArgs(analyzerPath, "some-image"),
 			)
 
@@ -130,6 +132,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 
 			output := h.DockerRun(t,
 				analyzeImage,
+				h.WithFlags("--env", "CNB_PLATFORM_API="+latestPlatformAPI),
 				h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s some-image; ls -al /layers", analyzerPath)),
 			)
 
@@ -149,6 +152,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				"/layers",
 				h.WithFlags(
 					"--volume", cacheVolume+":"+"/cache", // use a cache so that we can observe the effect of other-group.toml on /layers
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
 				),
 				h.WithArgs(
 					analyzerPath,
@@ -170,6 +174,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				copyDir,
 				analyzeImage,
 				"/some-dir/some-analyzed.toml",
+				h.WithFlags("--env", "CNB_PLATFORM_API="+latestPlatformAPI),
 				h.WithArgs(
 					analyzerPath,
 					"-analyzed", "/some-dir/some-analyzed.toml",
@@ -188,7 +193,10 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 				copyDir,
 				analyzeImage,
 				"/layers/analyzed.toml",
-				h.WithFlags(dockerSocketMount...),
+				h.WithFlags(append(
+					dockerSocketMount,
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+				)...),
 				h.WithArgs(analyzerPath, "-daemon", "some-image"),
 			)
 
@@ -222,7 +230,10 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 					copyDir,
 					analyzeImage,
 					"/layers",
-					h.WithFlags(dockerSocketMount...),
+					h.WithFlags(append(
+						dockerSocketMount,
+						"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					)...),
 					h.WithArgs(analyzerPath, "-daemon", appImage),
 				)
 
@@ -236,7 +247,10 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 						copyDir,
 						analyzeImage,
 						"/layers",
-						h.WithFlags(dockerSocketMount...),
+						h.WithFlags(append(
+							dockerSocketMount,
+							"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+						)...),
 						h.WithArgs(
 							analyzerPath,
 							"-daemon",
@@ -280,7 +294,10 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 							copyDir,
 							analyzeImage,
 							"/layers",
-							h.WithFlags(dockerSocketMount...),
+							h.WithFlags(append(
+								dockerSocketMount,
+								"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+							)...),
 							h.WithArgs(
 								analyzerPath,
 								"-daemon",
@@ -321,8 +338,9 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 									"/layers",
 									h.WithFlags(append(
 										dockerSocketMount,
-										"--env", "CNB_REGISTRY_AUTH="+cacheAuthConfig,
 										"--network", "host",
+										"--env", "CNB_REGISTRY_AUTH="+cacheAuthConfig,
+										"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
 									)...),
 									h.WithArgs(
 										analyzerPath,
@@ -344,8 +362,9 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 									analyzeImage,
 									"/layers",
 									h.WithFlags(
-										"--env", "DOCKER_CONFIG=/docker-config",
 										"--network", "host",
+										"--env", "DOCKER_CONFIG=/docker-config",
+										"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
 									),
 									h.WithArgs(
 										analyzerPath,
@@ -387,7 +406,10 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 								copyDir,
 								analyzeImage,
 								"/layers",
-								h.WithFlags("--network", "host"),
+								h.WithFlags(
+									"--network", "host",
+									"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+								),
 								h.WithArgs(
 									analyzerPath,
 									"-cache-image",
@@ -414,6 +436,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 						h.WithFlags(append(
 							dockerSocketMount,
 							"--volume", cacheVolume+":"+"/cache",
+							"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
 						)...),
 						h.WithArgs(
 							analyzerPath,
@@ -438,6 +461,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 							h.WithFlags(append(
 								dockerSocketMount,
 								"--volume", cacheVolume+":/cache",
+								"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
 							)...),
 							h.WithBash(
 								fmt.Sprintf("chown -R 9999:9999 /cache; chmod -R 775 /cache; %s -daemon -cache-dir /cache some-image; ls -alR /cache", analyzerPath),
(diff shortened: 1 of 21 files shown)


natalieparellano:
Undrafting as the associated spec PR is ready to be merged

jabrown85:
This is a spec change, should we returning the previous exit codes when `API < 0.6`?

natalieparellano:
> This is a spec change, should we returning the previous exit codes when `API < 0.6`?

🤦  I forgot about this...

jabrown85:
Maybe we can trying changing them from `const CodeFailedDetect = 20` to `var CodeFailedDetect = 20` and perhaps an `init()` in `exit.go` could read `CNB_PLATFORM_API` to change the value to the old value?

I don't know how well that would work, but if we _only_ ever accept the api version via env var - maybe we could get away with `init()` doing some of the work so we don't have to have the version passed around everywhere.

cmd/flags.go
```diff
@@ -245,11 +245,17 @@ func BoolEnv(k string) bool {
 	return b
 }
 
-func EnvOrDefault(key string, defaultVal string) string {
-	if envVal := os.Getenv(key); envVal != "" {
-		return envVal
+func EnvOrDefault(key string, defaultVal string, predicates ...func(string) bool) string {
```
jromero, line 248: ❤️ Love the predicate addition. fn programing 🤌

jabrown85:
I had a go at what an `init` style might look like for my own learning purposes. My thought was that perhaps we could always have lifecycle's `platform` be the latest. This could help remove some of the APIVersion being passed around as well.

https://play.golang.org/p/6apKcDLrqQ9

The calling code, in this case the lifecycle/cmds or tests, could `platform.OverridePlatform(v0X.NewPlatform())`  to change versions. If not overridden, however, the platform will essentially be an alias to the current latest supported version.

I'm not glued to the idea or anything - I wanted to throw out the pattern in case we think it could remove some mapping/translation code. I think it could be a useful pattern for things that aren't _truly_ dynamic.

cmd/flags.go
natalieparellano: Unfortunately this code is no longer needed, but I share your sentiment!

natalieparellano:
I realized I made this PR harder than it needed to be by requiring `platform` to determine codes that didn't actually change between platforms (and aren't likely to in the future). Without this, `platform` isn't passed around as much - and where it is, it's likely to be used for other things besides determining the code. I'm much happier with it now - looking forward to hearing what others thing :)

platform/v05/exit.go
```diff
@@ -0,0 +1,31 @@
+package v05
+
+var exitCodes = map[string]int{
```
yaelharel, line 3: Do we really need an `exit.go` file?
Can't we just have one file that implements the Platform interface for version 0.5 (and the same for version 0.6)?
Don't have a strong opinion on this. Just curios.

platform/v05/exit.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// rebase phase errors: 600-699
+	"RebaseError": 602, // RebaseError indicates generic rebase error
+
+	// launch phase errors: 700-799
+	"LaunchError": 702, // LaunchError indicates generic launch error
+}
+
+func (p *Platform) CodeFor(errType string) int {
+	return exitCodes[errType]
```
yaelharel, line 30: What will happen when we want to introduce new exit codes?
In the current implementation, we should probably need to add them to both api < 0.6 and to api >=0.5.
Of course that theoretically, we shouldn't get to this function with new codes but I think that with the current implementation, it's better to do it so we won't get a bad error for searching a key that doesn't exist in the map.
Since I prefer not to add new codes to old apis, I think we should change this function to check first if the key exists in the map and if not, return a general error.
Moreover, I don't like so much the fact that `errType` is a string. I feel that contributors might make mistake when writing them.
I'm not sure what is the best way to make this change but what do you think about having a common file for all apis that will contain a list of constants and we'll use it everywhere?

go.sum
```diff
@@ -290,6 +290,7 @@ github.com/konsorten/go-windows-terminal-sequences v1.0.3 h1:CE8S1cTafDpPvMhIxNJ
 github.com/konsorten/go-windows-terminal-sequences v1.0.3/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
 github.com/kr/logfmt v0.0.0-20140226030751-b84e30acd515/go.mod h1:+0opPa2QZZtGFBFZlji/RkVcI2GknAs/DXo4wKdlNEc=
 github.com/kr/pretty v0.1.0/go.mod h1:dAy3ld7l9f0ibDNOQOHHMYYIIbhfbHSm3C4ZsoJORNo=
+github.com/kr/pretty v0.2.0 h1:s5hAObm+yFO5uHYt5dYjxi2rXrsnmRpJx4OYvIWUaQs=
```
yaelharel, line 293: Just wondering - do you understand why these two lines have been added to this file?

cmd/lifecycle/main.go
```diff
@@ -19,21 +20,22 @@ func main() {
 		cmd.Exit(err)
 	}
 
+	var platform cmd.Platform = lplatform.NewPlatform(platformAPI)
```
yaelharel, line 23:
```suggestion
	var newPlatform cmd.Platform = lplatform.NewPlatform(platformAPI)
```
It is just a suggestion because I just don't like seeing in the code lines like `platform: platform`.

In addition I have a small question - do we need `cmd.Platform` because otherwise it'll be considered as `*Platform`?

yaelharel (commented):
Hi @natalieparellano, overall it looks great!
I left some questions and suggestions.
In addition, not for this PR, I think we should add some acceptance tests for all phases to check the error codes. I think it won't be very easy to do so but I think it's important.

platform/v05/exit.go
natalieparellano: +1 to checking that the code exists.

I agree that keying on a string feels weird. What would the file of constants look like?

go.sum
natalieparellano: Nope...

platform/v05/exit.go
yaelharel: I thought about using enums but after searching in Google, the implementation of enums in golang is a bit different.
Here are some links that I found and might be useful:
https://www.educative.io/edpresso/what-is-an-enum-in-golang
https://stackoverflow.com/questions/14426366/what-is-an-idiomatic-way-of-representing-enums-in-go

cmd/launcher/main.go
```diff
@@ -51,7 +54,7 @@ func runLaunch() error {
 	}
 
 	if err := launcher.Launch(os.Args[0], os.Args[1:]); err != nil {
-		return cmd.FailErrCode(err, cmd.CodeLaunchError, "launch")
+		return cmd.FailErrCode(err, platform.CodeFor("LaunchError"), "launch")
```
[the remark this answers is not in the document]
natalieparellano: I love this idea. @yaelharel had the same suggestion. Fixed in 8faa92e

platform/v05/exit.go
natalieparellano: Thank you for the suggestion! Fixed in 8faa92e and bf07a2a.

jabrown85: I haven't checked our lint rules, but I know there are configurable lint rules for ensuring exhaustive switch statements with enums. Maybe removing the map in favor of a switch in each version's `exit.go` would work and alert us to handling new errors as they are introduced?

cmd/lifecycle/main.go
natalieparellano: `platform := lplatform.NewPlatform(platformAPI)` works too, I'll make this change. 

I'm not sure `newPlatform` adds any useful information vs. `platform` and I like the current terseness so propose keeping it as is for now. Though I could be persuaded if there is another variable name that is more descriptive... we do have a bunch of `thing: thing` in our code so extending this pattern doesn't bother me.

platform/v05/exit.go
natalieparellano: Summarizing other discussion, we plan to extend the platform interface to cover other platform related logic so having a separate file probably makes sense.

acceptance/detector_test.go
```diff
@@ -81,14 +101,21 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
+				detectImage,
+			)
 			output, err := command.CombinedOutput()
 			h.AssertNotNil(t, err)
 			failErr, ok := err.(*exec.ExitError)
 			if !ok {
 				t.Fatalf("expected an error of type exec.ExitError")
 			}
-			h.AssertEq(t, failErr.ExitCode(), cmd.CodeFailedDetect)
+			h.AssertEq(t, failErr.ExitCode(), 20)
```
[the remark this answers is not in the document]
natalieparellano: Hmm... I thought about this, but the alternative is `lplatform.NewPlatform(latestPlatformAPI).CodeFor(cmd.FailedDetect)` which is just very long (though we could put it in a variable to make it more readable) and I don't love using application code in the test expectation. I also kind of like the "code as documentation" result here... but if you feel strongly about it I can make this change.

acceptance/analyzer_test.go
```diff
@@ -101,6 +101,7 @@ func testAnalyzer(t *testing.T, when spec.G, it spec.S) {
 		it("errors", func() {
 			cmd := exec.Command(
 				"docker", "run", "--rm",
+				"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
```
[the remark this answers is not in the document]
natalieparellano: I kind of like the explicitness of the current implementation - even the current `h.WithFlags()` and `h.WithArgs()` helpers are hard for me to read :/

acceptance/detector_test.go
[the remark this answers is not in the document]
jabrown85: I don't feel strong - maybe a comment for `// cmd.FailedDetect ` would be sufficient

natalieparellano:
> Hi @natalieparellano, overall it looks great!
> I left some questions and suggestions.
> In addition, not for this PR, I think we should add some acceptance tests for all phases to check the error codes. I think it won't be very easy to do so but I think it's important.

@yaelharel I agree with you. I updated the various issues that we have re: acceptance tests to include something like "should include a test of the exit code in case of error (see Detector as an example)". I tried to see how we might add such a test for Analyzer but all of the code paths that might lead to a `cmd.AnalyzeError` would be pretty hard to force in a test (such as a failure in `os.Create()` so maybe we leave that one out. I suspect we might discover that some of the other phases also have this problem :/

acceptance/analyzer_test.go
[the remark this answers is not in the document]
jabrown85: It's fine - I hope we get to a point where the only time you see `CNB_PLATFORM_API` is when it is NOT the latest platform API. I think we have some tests that are using docker args to build the image with the latest api version, so I think we are trending in that direction.

cmd/launcher/main.go
```diff
@@ -34,13 +37,13 @@ func runLaunch() error {
...
 
-	defaultProcessType := defaultProcessType(api.MustParse(platformAPI), md)
+	defaultProcessType := defaultProcessType(api.MustParse(platform.API()), md)
 
 	launcher := &launch.Launcher{
 		DefaultProcessType: defaultProcessType,
 		LayersDir:          cmd.EnvOrDefault(cmd.EnvLayersDir, cmd.DefaultLayersDir),
 		AppDir:             cmd.EnvOrDefault(cmd.EnvAppDir, cmd.DefaultAppDir),
-		PlatformAPI:        api.MustParse(platformAPI),
+		PlatformAPI:        api.MustParse(platform.API()),
```
jabrown85, line 46: Do you think we'll be able to remove `api.MustParse` in the future? Can the new `platform.API()` return `api.Version`? I see the constructor already has `api.MustParse(apiStr)`.

cmd/launcher/main.go
```diff
@@ -51,7 +54,7 @@ func runLaunch() error {
 	}
 
 	if err := launcher.Launch(os.Args[0], os.Args[1:]); err != nil {
-		return cmd.FailErrCode(err, cmd.CodeLaunchError, "launch")
+		return cmd.FailErrCode(err, platform.CodeFor(cmd.LaunchError), "launch")
```
jabrown85, line 57: Now that we have a `LifecycleExitError`, we could refactor this to prevent callers from passing an unsupported. Something to think about in the future.
```
return cmd.FailErrCode(err, platform, cmd.LaunchError, "launch")
```

jabrown85 (approved):

acceptance/detector_test.go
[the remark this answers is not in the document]
yaelharel: I also thought about it but I think it's fine and even a good idea for the test.
But I also like the idea of adding a comment.

cmd/launcher/main.go
yaelharel: I think it's a good idea but I'm not sure if we should do it as part of this PR.
We're using a string that is representing an api all over the code. Maybe we should dedicate a PR to change it.

platform/platform.go
```diff
@@ -0,0 +1,24 @@
...
+
+func NewPlatform(apiStr string) Platform {
+	var platform Platform
+	platformAPI := api.MustParse(apiStr)
+	if platformAPI.Compare(api.MustParse("0.5")) > 0 { // platform API > 0.5
+		platform = v06.NewPlatform(apiStr)
+	} else {
+		platform = v05.NewPlatform(apiStr)
+	}
+	return platform
```
yaelharel, line 23:
```suggestion
	platformAPI := api.MustParse(apiStr)
	if platformAPI.Compare(api.MustParse("0.6")) < 0 { // platform API < 0.6
		return v05.NewPlatform(apiStr)
	}
	return v06.NewPlatform(apiStr)
```

I would change the order of the condition because if at some point we stop supporting api < 0.6, it'll be a bit easier to change the code.

yaelharel (approved):
This looks great!
Thank you for making the changes :)
I added a small suggestion but other than that, I think it can be merged.

cmd/launcher/main.go
natalieparellano: I like the idea of doing this in a separate PR.

platform/platform.go
natalieparellano: Nice!

acceptance/detector_test.go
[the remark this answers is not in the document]
natalieparellano: Adding a comment!

cmd/launcher/main.go
natalieparellano: Right now it's very convenient that `cmd` doesn't depend on platform - because then we can have `cmd.LaunchError` without creating a circular dependency (cmd -> platform -> v05/v06 -> cmd). But if we decide to do this I'm sure we can find another place for those constants.

PR #548, opened 2021-03-18
Add gosec

natalieparellano:
Adds gosec to our linter


buildpack/detect.go
```diff
@@ -54,12 +54,12 @@ func (b *Descriptor) Detect(config *DetectConfig) DetectRun {
 	defer os.RemoveAll(planDir)
 
 	planPath := filepath.Join(planDir, "plan.toml")
-	if err := ioutil.WriteFile(planPath, nil, 0777); err != nil {
+	if err := ioutil.WriteFile(planPath, nil, 0600); err != nil {
```
natalieparellano, line 57: Not 100% sure on this change, but the acceptance tests are passing, so I think it's fine?

layers.go
```diff
@@ -176,7 +176,7 @@ func (bp *bpLayer) hasLocalContents() bool {
 }
 
 func (bp *bpLayer) writeSha(sha string) error {
-	if err := ioutil.WriteFile(bp.path+".sha", []byte(sha), 0777); err != nil {
+	if err := ioutil.WriteFile(bp.path+".sha", []byte(sha), 0777); err != nil { //nolint:gosec
```
natalieparellano, line 179: I am pretty sure we still need `0777` as we can't be sure of the user who is going to read this file - but I am happy to be corrected.

testhelpers/testhelpers.go
```diff
@@ -27,7 +27,7 @@ import (
 func RandString(n int) string {
 	b := make([]byte, n)
 	for i := range b {
-		b[i] = 'a' + byte(rand.Intn(26))
+		b[i] = 'a' + byte(rand.Intn(26)) //nolint:gosec
```
natalieparellano, line 30: We could change this function to use crypto/rand, but it's really gross.

buildpack/detect.go
jabrown85: I don't think we need to execute `plan.toml` - so this feels right

layers.go
jabrown85: I think we could get away with `666` - `777` is only needed if we are going to need execute permissions on the `.sha` file, which I don't believe we need.

jabrown85 (commented):
Had one question that pertains to a lot of the `//nolint` lines to gain context, but looks pretty good to me.

acceptance/acceptance_test.go
```diff
@@ -149,5 +149,5 @@ func testVersion(t *testing.T, when spec.G, it spec.S) {
 }
 
 func lifecycleCmd(phase string, args ...string) *exec.Cmd {
-	return exec.Command(filepath.Join(buildDir, runtime.GOOS, "lifecycle", phase), args...)
+	return exec.Command(filepath.Join(buildDir, runtime.GOOS, "lifecycle", phase), args...) //nolint:gosec
```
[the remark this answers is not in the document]
natalieparellano: I think it's the fact that `args` is an input to the function and gets passed to `exec`. If I take out the `//nolint` the error message is `G204: Subprocess launched with function call as argument or cmd arguments`.

layers.go
natalieparellano: Haha `0666` still requires `//nolint` but that's a good point! I think we could make this change as it's a step in the right direction.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
jabrown85: Ah, ok. I see [here](https://github.com/securego/gosec/blob/master/rules/subproc.go#L42-L61). Maybe we want to do `// #nosec G204` instead? Just so we don't miss other `gosec` errors/warnings.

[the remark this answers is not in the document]
natalieparellano: That's a great idea!

jabrown85 (approved):

PR #549, opened 2021-03-24
Bump imgutil

natalieparellano:
Picks up the change to retry image fetch on EOF error.


jabrown85 (approved):
I am excited to move to go 1.16 so we can potentially remove the `/tools` folder 🥳

PR #553, opened 2021-03-25
buildpack API >= 0.6 - omit the types table when encoding the layer.toml file

yaelharel:
The launch, build and cache flags are set to false.

Make this change to allow the buildpack author to append new values to the flags instead of changing the existing ones
(according to the toml's spec, there shouldn't be duplicate values).

### DIFF
--- a/buildpack/v06/encoderdecoder.go
+++ b/buildpack/v06/encoderdecoder.go
@@ -10,17 +10,6 @@ import (
 	"github.com/buildpacks/lifecycle/buildpack/layertypes"
 )
 
-type typesTable struct {
-	Build  bool `toml:"build"`
-	Launch bool `toml:"launch"`
-	Cache  bool `toml:"cache"`
-}
-
-type layerMetadataTomlFile struct {
-	Data  interface{} `toml:"metadata"`
-	Types typesTable  `toml:"types"`
-}
-
 type EncoderDecoder06 struct {
 }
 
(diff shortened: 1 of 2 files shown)


yaelharel:
A small addition to #537.

yaelharel:
Thank you @ekcasey for thinking about this scenario!!!

buildpack/v06/encoderdecoder.go
```diff
@@ -32,20 +21,26 @@ func (d *EncoderDecoder06) IsSupported(buildpackAPI string) bool {
...
+	return toml.NewEncoder(file).Encode(dtf)
 }
 
 func (d *EncoderDecoder06) Decode(path string) (layertypes.LayerMetadataFile, string, error) {
+	type typesTable struct {
+		Build  bool `toml:"build"`
+		Launch bool `toml:"launch"`
+		Cache  bool `toml:"cache"`
+	}
+	type layerMetadataTomlFile struct {
```
ekcasey, line 39: FWIW if you wanted to keep the single `layerMetadataTomlFile` struct you could change it to
```
type layerMetadataTomlFile struct {
	Data  interface{} `toml:"metadata"`
	Types *typesTable  `toml:"types,omitempty"`
}
```
and leave `Types` set to `nil` when encoding

ekcasey (approved):

buildpack/v06/encoderdecoder.go
yaelharel: Yeah, you're right. I thought about it but since we're using this struct only in this file, I felt that being explicit is better.
If at some point there will be more functions in this file that use the `layerMetadataTomlFile` struct, then it'll make sense to me to define it once in this file and delete `dataTomlFile`.
What do you think?

natalieparellano (approved):

yaelharel:
Ran some manual tests similar to what described [here](https://github.com/buildpacks/lifecycle/pull/537#issuecomment-802246439).

PR #554, opened 2021-03-26
Add a unit test

yaelharel:
rename `<layers>/<layer>` to `<layers>/<layer>.ignore` if the `launch`, `build` and `cache` flags aren't specified

### DIFF
--- a/buildpack/build_test.go
+++ b/buildpack/build_test.go
@@ -372,6 +372,7 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 							filepath.Join(layersDir, "A", "layer"),
 						)
 						h.Mkfile(t,
+							"",
 							filepath.Join(layersDir, "A", "layer.toml"),
 						)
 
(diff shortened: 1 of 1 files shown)


jabrown85 (approved):

buildpack/build_test.go
```diff
@@ -349,19 +349,37 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+					})
+				})
+
+				when("the flags aren't specified in <layer>.toml", func() {
+					it("renames <layers>/<layer> to <layers>/<layer>.ignore", func() {
+						h.Mkdir(t,
+							filepath.Join(layersDir, "A", "layer"),
+						)
+						h.Mkfile(t,
+							filepath.Join(layersDir, "A", "layer.toml"),
```
natalieparellano, line 375: I think calling `h.Mkfile()` with one argument is a no-op, as it ranges over `paths`. That is interesting to know that the test is still passing but I think it would be more accurate to create the file with blank data.

yaelharel: Thanks! Please see https://github.com/buildpacks/lifecycle/pull/554/commits/6d5219299b503d794427d6ae41f23a35ac0c6d99.

natalieparellano (approved):

PR #559, opened 2021-03-31
Add print statements

natalieparellano:
Not a real PR, using this to create a lifecycle image that we can use for testing.


natalieparellano:
There are better ways to do this. We have a make target `build-image-linux`.

PR #560, opened 2021-03-31
Remove k8schain

jabrown85:
The k8schain gcp provider is causing lifecycle to hang indefinitely. We believe it is similar to https://github.com/kubernetes/kubernetes/issues/86245. Reverting out this feature until we can reliably bring it back in.


jabrown85:
Confirmed that running a workload like this fails:

```
» cat product_name
Google%
» docker run -it -v /some/workspace:/workspace -v $(pwd)/product_name:/sys/class/dmi/id/product_name paketobuildpacks/builder:base bash

cnb@e61093db08ec:/layers$ /cnb/lifecycle/detector
4 of 6 buildpacks participating
paketo-buildpacks/go-dist       0.3.1
paketo-buildpacks/go-mod-vendor 0.2.0
paketo-buildpacks/go-build      0.3.0
paketo-buildpacks/procfile      4.0.0
cnb@e61093db08ec:/layers$ /cnb/lifecycle/analyzer testimg

cnb@e61093db08ec:/layers$ /cnb/lifecycle/analyzer testimg
************* defining flags
************* parsing flags
************* setting log level
************* parsing args
Warning: Not restoring cached layer metadata, no cache flag specified.
************* privileges
************* creating k8schain

...waits for eternity...
```

PR #563, opened 2021-04-07
Add codecov

natalieparellano:
Tested on fork: https://github.com/natalieparellano/lifecycle/actions/runs/724102070

I am not sure about the thresholds. We'll probably want to play around with them to see what makes sense for lifecycle.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,16 @@ jobs:
           chmod +x deps/bin/jq
           echo "${PWD}/deps/bin" >> $GITHUB_PATH
       - name: Test
+        env:
+          TEST_COVERAGE: 1
         run: make test
+      - name: Upload Coverage
+        uses: codecov/codecov-action@v1
+        with:
+          file: ./out/tests/coverage-unit.txt
+          flags: unit,os_linux
+          fail_ci_if_error: true
+          verbose: true
   test-windows:
     runs-on: windows-latest
     steps:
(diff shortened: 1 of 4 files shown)


codecov.yml
```diff
@@ -0,0 +1,18 @@
+codecov:
+  notify:
+    after_n_builds: 4
```
natalieparellano, line 3: Took this straight from `pack` but I'm actually not sure why it's 4.

natalieparellano: Looking at https://docs.codecov.io/docs/notifications#section-preventing-notifications-until-after-n-builds I think actually we want this number to be 2.

.github/workflows/build.yml
```diff
@@ -26,7 +26,16 @@ jobs:
           chmod +x deps/bin/jq
           echo "${PWD}/deps/bin" >> $GITHUB_PATH
       - name: Test
+        env:
+          TEST_COVERAGE: 1
         run: make test
+      - name: Upload Coverage
```
jabrown85, line 32: Is this the step that will fail the build if the code coverage is not high enough? or is that configured in codecov and added as a different GitHub check?

natalieparellano: I **think** so... checking...

natalieparellano: So it won't actually fail `test-linux` or `test-windows` or `build` - it will show up as a separate "job" that's failing: https://github.com/natalieparellano/lifecycle/pull/2

jabrown85 (approved):

PR #583, opened 2021-04-19
Move cache metadata

natalieparellano:
@jabrown85 this is what I had in mind :)


jabrown85 (approved):
It looks good to me!

analyzer.go
```diff
@@ -43,8 +44,13 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 		appMeta = platform.LayersMetadata{}
 	}
 
+	cacheMeta, err = retrieveCacheMetadata(a.Cache, a.Logger)
+	if err != nil {
+		return platform.AnalyzedMetadata{}, err
+	}
```
yaelharel, line 50: Don't we want to put it under the if condition in line 52?

natalieparellano: Good point! I'll make this change.

analyzer.go
```diff
@@ -55,6 +61,25 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 	}, nil
 }
 
+func retrieveCacheMetadata(cache Cache, logger Logger) (platform.CacheMetadata, error) {
```
yaelharel, line 64: I don't think this is the best place for this function.
What do you think about moving it to a common file (since both the analyzer and the restorer are using it)?

layer_metadata_restorer_test.go
```diff
@@ -68,7 +59,7 @@ func testLayerMetadataRestorer(t *testing.T, when spec.G, it spec.S) {
 
 		when("app and cache metadata are not present", func() {
 			it("does not restore any metadata", func() {
-				_, err := layerAnalyzer.Restore(buildpacks, layersMetadata, nil)
+				err := layerAnalyzer.Restore(buildpacks, layersMetadata, cacheMetadata)
```
yaelharel, line 62: Maybe it's better to rename `cacheMetadata` to `emptyCacheMetadata` for readability?

natalieparellano: Maybe... but that's a bit outside the scope of this PR.

yaelharel: You're right. `cacheMetadata` is sometimes empty but sometimes is not.
I'm ok with leaving it as is.

analyzer.go
natalieparellano: I'll move it into layer_metadata_restorer.go since both phases are using that struct. It feels too specific for utils.go

yaelharel (approved):
Looks great!

PR #585, opened 2021-04-20
Remove cleanup of images stored in test registry

natalieparellano:

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -968,6 +968,11 @@ func testAnalyzerBuilder(apiString string) func(t *testing.T, when spec.G, it sp
 
 		when("layers path is provided", func() {
 			it("uses the group path at the working directory and writes analyzed.toml at the working directory", func() {
+				h.SkipIf(t,
+					api.MustParse(apiString).Compare(api.MustParse("0.5")) >= 0,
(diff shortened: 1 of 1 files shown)


jabrown85 (approved):

acceptance/analyzer_test.go
```diff
@@ -474,9 +474,8 @@ func testAnalyzerBuilder(apiString string) func(t *testing.T, when spec.G, it sp
 								)
 							})
 
-							it.After(func() {
-								h.DockerImageRemove(t, authRegCacheImage)
-							})
+							// Don't attempt to remove the image, as it's stored in the test registry, which is ephemeral.
```
yaelharel, line 477: Does it worth adding to all of these comments a reference to `defer registry.Stop(t)` for readability purpose?
Also, not related to this PR, but I think that the name of the `Stop` function isn't accurate since `ContainerKill` and `ContainerRemove` are being called there.

yaelharel (approved):
Other than my small comment, it looks good to me!

acceptance/analyzer_test.go
yaelharel: I didn't mean that we should mention `defer registry.Stop` in the comment since we might change the function name.
I don't have a specific idea so let's leave the comment as is for now and we can always change it in a future PR if we think about a better comment.

Issue #587, opened 2021-04-20
[Analyze & Restore] Unit tests should mock LayerMetadataRestorer 


natalieparellano:
### Description
https://github.com/buildpacks/lifecycle/pull/582 introduces a `LayerMetadataRestorer` that analyze and/or restore can call when they are running with a platform API that supports restoring layer metadata (writing `<layer>.toml` files) as part of that phase.

https://github.com/buildpacks/lifecycle/pull/582 does not mock the `LayerMetadataRestorer` in tests because we are trying to change as little as possible to ensure that we keep the existing test coverage.

### Proposed solution
Unit tests should mock `LayerMetadataRestorer`. We can then (potentially) delete some tests from the analyze & restore units as long as the behavior is covered in the tests for `LayerMetadataRestorer`.

natalieparellano:
Blocked on #582

natalieparellano:
Resolved by #605

PR #588, opened 2021-04-21
Make analyzer units into table test

natalieparellano:
Also remove platform from layer metadata restorer because it's not used.


analyzer_test.go
```diff
@@ -1,13 +1,650 @@
...
 
 func TestAnalyzer(t *testing.T) {
-	spec.Run(t, "Analyzer", testAnalyzer06, spec.Report(report.Terminal{}))
-	spec.Run(t, "Analyzer", testAnalyzer07, spec.Report(report.Terminal{}))
+	for _, api := range api.Platform.Supported {
+		spec.Run(t, "unit-analyzer/"+api.String(), testAnalyzerBuilder(api.String()), spec.Parallel(), spec.Report(report.Terminal{}))
+	}
+}
+
+func testAnalyzerBuilder(apiString string) func(t *testing.T, when spec.G, it spec.S) {
```
natalieparellano, line 35: To get a better idea of the changes, it might be best to compare this with main.

natalieparellano:
Pending un-draft on answering the TODO

layer_metadata_restorer_test.go
```diff
@@ -40,8 +40,7 @@ func testLayerMetadataRestorer(t *testing.T, when spec.G, it spec.S) {
 		layerDir, err = ioutil.TempDir("", "lifecycle-layer-dir")
 		h.AssertNil(t, err)
 		logger := &log.Logger{Handler: &discard.Handler{}}
-		platform := platform.NewPlatform("0.7")
-		layerAnalyzer = lifecycle.NewLayerMetadataRestorer(logger, layerDir, platform, skipLayers)
+		layerAnalyzer = lifecycle.NewLayerMetadataRestorer(logger, layerDir, skipLayers)
```
yaelharel, line 43: Just to make sure I'm not missing something: we're planning to delete `restorer_platform06_test.go` and `restorer_platform07_test.go`, right?

natalieparellano: Yeah that'd be next.

natalieparellano: There's also this issue which is related: https://github.com/buildpacks/lifecycle/issues/587

analyzer_test.go
```diff
@@ -1,13 +1,649 @@
...
+	return func(t *testing.T, when spec.G, it spec.S) {
+		var (
+			analyzer         *lifecycle.Analyzer
+			cacheDir         string
```
yaelharel, line 39: It's not really important but I know that usually we're trying to put all of the variables of the same type one after the other.

analyzer_test.go
```diff
@@ -1,13 +1,649 @@
...
+	"github.com/buildpacks/lifecycle/cmd"
+	"github.com/buildpacks/lifecycle/platform"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+	"github.com/buildpacks/lifecycle/testmock"
 )
 
 func TestAnalyzer(t *testing.T) {
-	spec.Run(t, "Analyzer", testAnalyzer06, spec.Report(report.Terminal{}))
-	spec.Run(t, "Analyzer", testAnalyzer07, spec.Report(report.Terminal{}))
+	for _, api := range api.Platform.Supported {
```
yaelharel, line 30:
```suggestion
	for _, platformApi := range api.Platform.Supported {
```

analyzer_test.go
```diff
@@ -1,13 +1,649 @@
...
+		spec.Run(t, "unit-analyzer/"+api.String(), testAnalyzerBuilder(api.String()), spec.Parallel(), spec.Report(report.Terminal{}))
+	}
+}
+
+func testAnalyzerBuilder(apiString string) func(t *testing.T, when spec.G, it spec.S) {
```
yaelharel, line 35:
```suggestion
func testAnalyzerBuilder(platformApiString string) func(t *testing.T, when spec.G, it spec.S) {
```

analyzer_test.go
```diff
@@ -1,13 +1,649 @@
...
+				when("buildpack API < 0.6", func() {
+					it.Before(func() {
+						analyzer.Buildpacks = []buildpack.GroupBuildpack{
+							{ID: "metadata.buildpack", API: "0.5"},
+							{ID: "no.cache.buildpack", API: "0.5"},
+						}
+					})
+
+					it("restores layer metadata and doesn't change the values of the launch, build and cache flags in top level", func() {
+						h.SkipIf(t, api.MustParse(apiString).Compare(api.MustParse("0.7")) >= 0, "Platform API >= 0.7 does not restore layer metadata")
```
yaelharel, line 149: Can we move this line to the `it.Before` on line 141?

analyzer_test.go
```diff
@@ -1,13 +1,649 @@
...
+
+						path := filepath.Join(layersDir, "no.cache.buildpack", "some-layer.toml")
+						got := h.MustReadFile(t, path)
+						want := "[metadata]\n  some-layer-key = \"some-layer-value\""
+
+						h.AssertStringContains(t, string(got), want)
+					})
+
+					it("does not restore layers for undetected buildpacks", func() {
+						h.SkipIf(t, api.MustParse(apiString).Compare(api.MustParse("0.7")) >= 0, "Platform API >= 0.7 does not restore layer metadata")
```
yaelharel, line 221: Please see https://github.com/buildpacks/lifecycle/pull/588/files#r618456024

analyzer_test.go
jabrown85: IIRC `SkipIf` has to be inside an `it` block.

analyzer_test.go
```diff
@@ -1,13 +1,649 @@
...
+						{"metadata.buildpack/store.toml", "[metadata]\n  [metadata.metadata-buildpack-store-data]\n    store-key = \"store-val\""},
+						{"no.cache.buildpack/store.toml", "[metadata]\n  [metadata.no-cache-buildpack-store-data]\n    store-key = \"store-val\""},
+					} {
+						got := h.MustReadFile(t, filepath.Join(layersDir, data.name))
+						h.AssertStringContains(t, string(got), data.want)
+					}
+				})
+
+				when("cache exists", func() {
+					it.Before(func() {
```
yaelharel, line 254: I think you can put the `h.SkipIf` here instead of all of the nested `it` and `when`.

yaelharel (approved):
The PR looks good!
I put some comments on it.

Regarding my comments about moving the `h.SkipIf` to the `it.Before`:
I'm not sure if we should move it. My initial thought was to reduce the number of `SkipIf` calls and to not run the `it.Before` steps if they're not needed. But it might be clearer to have it written down next to each `it` for better readability and in case we'll add more tests under `when`.

analyzer_test.go
yaelharel: Interesting. I was able to put it under `it.Before`.

PR #591, opened 2021-04-26
Asset packages

dwillist:
Small changes to the build environment logic that do the following: 
- Add the `CNB_ASSETS` environment variable to the `build` environement if it is present and the platform API supports assets (`>= 0.7`)
- Split `detect` and `build` environements so `CNB_ASSETS` is never present during `detect`

In summary this will allow the `CNB_ASSETS` environment variable to be set in the `ENV` label on a builder and actually end up being visible to buildpacks if the platform api >= 0.7.

In general this is needed to enable the following support matrix between `pack` and the `lifecycle`

| pack used to build supports platform api >= 0.7 | pack used to create builder supports platform api >= 0.7 | lifecycle supports platform api >= 0.7 | conclusion                                                                                                                                                                                                                                      |
|-------------------------------------------------|----------------------------------------------------------|----------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| T                                               | T                                                        | T                                      | CNB_ASSETS can be set in build env! and buildpacks can use it                                                                                                                                                                                   |
| T                                               | T                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API Do not allow assets to be added to builders like this.                                                                                                                              |
| T                                               | F                                                        | T                                      | CNB_ASSETS will not be set on the builder :( and assets were not allowed on the  builder when it was created. But we should support adding assets using: "pack build ... --asset-package" at build time by adding CNB_ASSETS to the environment |
| T                                               | F                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API...                                                                                                                                                                                  |
| F                                               | T                                                        | T                                      | Cannot use assets as pack used to build does not support platform api 0.8                                                                                                                                                                       |
| F                                               | T                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API...                                                                                                                                                                                  |
| F                                               | F                                                        | T                                      | Cannot use assets as pack used to build does not support platform api 0.8                                                                                                                                                                       |
| F                                               | F                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API...                                                                                                                                                                                  |


cmd/lifecycle/detector.go
```diff
@@ -88,7 +88,7 @@ func (da detectArgs) detect() (buildpack.Group, platform.BuildPlan, error) {
 		return buildpack.Group{}, platform.BuildPlan{}, err
 	}
 
-	envv := env.NewBuildEnv(os.Environ())
+	envv := env.NewDetectEnv(os.Environ())
```
natalieparellano, line 91: It would be awesome if we could add an acceptance test for this - like have a buildpack that prints its env as part of detect and verify that we don't see `CNB_ASSETS`.

env/build.go
```diff
@@ -37,6 +56,26 @@ func matches(k1, k2 string) bool {
 	return k1 == k2
 }
 
+func isNotIncludedForAPI(platformAPI *api.Version) func(k string) bool {
```
natalieparellano, line 59:
```suggestion
func isNotSupported(platformAPI *api.Version) func(k string) bool {
```
Just a suggestion, feel free to ignore (it made more sense to me this way)

cmd/lifecycle/builder.go
```diff
@@ -83,12 +83,13 @@ func (ba buildArgs) build(group buildpack.Group, plan platform.BuildPlan) error
...
 
+	platformAPI := api.MustParse(ba.platform.API())
 	builder := &lifecycle.Builder{
 		AppDir:         ba.appDir,
 		LayersDir:      ba.layersDir,
 		PlatformDir:    ba.platformDir,
-		PlatformAPI:    api.MustParse(ba.platform.API()),
-		Env:            env.NewBuildEnv(os.Environ()),
+		PlatformAPI:    platformAPI,
+		Env:            env.NewBuildEnv(os.Environ(), platformAPI),
```
natalieparellano, line 92: Our use of `platform` is still evolving, but what about passing the entire `platform` to `NewBuildEnv` and on to `isNotIncludedForAPI`? Then we could do something like `if !platform.SupportsAssetPackages { /* return true */ }`

natalieparellano: I think our long-term dream is for all instances of `platformAPI.Compare(api.MustParse("0.8"))` to end up in the platform package, and the platform interface returns something meaningful instead.

env/build.go
```diff
@@ -17,12 +19,29 @@ var BuildEnvIncludelist = []string{
 	"no_proxy",
 }
 
+// List of environment variables to be included for
+// particular platform API versions
```
natalieparellano, line 23:
```suggestion
// asset packages
```

I assume if we need to update the build env for future platform api versions we'd create a new array and call it something other than `BuildEnvIncludeAssets`.

env/build.go
```diff
@@ -17,12 +19,29 @@ var BuildEnvIncludelist = []string{
...
+}
+
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
 // NewBuildEnv returns an build-time Env from the given environment.
 //
-// Only keys in the BuildEnvIncludelist will be added to the Environment.
-func NewBuildEnv(environ []string) *Env {
+// Keys in the BuildEnvIncludelist will be added to the Environment.
+// if platformAPI >= 0.8 keys from BuildEnvIncludeAssets will be added to the environment
```
natalieparellano, line 33:
```suggestion
// if platform supports asset packages, keys from BuildEnvIncludeAssets will be added to the environment
```

env/build.go
```diff
@@ -17,12 +19,29 @@ var BuildEnvIncludelist = []string{
...
+func NewBuildEnv(environ []string, platformAPI *api.Version) *Env {
+	return &Env{
+		RootDirMap: POSIXBuildEnv,
+		Vars:       varsFromEnv(environ, ignoreEnvVarCase, isNotIncludedForAPI(platformAPI)),
+	}
+}
+
+// NewBuildEnv returns an detect-time Env from the given environment.
```
natalieparellano, line 41:
```suggestion
// NewDetectEnv returns an detect-time Env from the given environment.
```

natalieparellano (commented):
Still going through it, but adding some comments so that I can reference in other comments :)

env/build.go
```diff
@@ -37,6 +56,26 @@ func matches(k1, k2 string) bool {
...
+		return isNotIncluded
+	}
+
+	return func(k string) bool {
+		return isNotIncluded(k) && isNotIncludedForAssets(k)
+	}
+}
+
+func isNotIncludedForAssets(k string) bool {
+	for _, wk := range BuildEnvIncludeAssets {
```
natalieparellano, line 70: If we do the suggestion here https://github.com/buildpacks/lifecycle/pull/591/files#r625318099 this is where we could check the platform support and return true.

env/build.go
```diff
@@ -37,6 +56,26 @@ func matches(k1, k2 string) bool {
 	return k1 == k2
 }
 
+func isNotIncludedForAPI(platformAPI *api.Version) func(k string) bool {
+	if platformAPI.Compare(api.MustParse("0.8")) < 0 {
+		return isNotIncluded
+	}
```
natalieparellano, line 62: This could go away if we do https://github.com/buildpacks/lifecycle/pull/591/files#r625322067

env/env.go
```diff
@@ -14,8 +14,9 @@ import (
 // Env is used to modify and return environment variables
 type Env struct {
 	// RootDirMap maps directories in a posix root filesystem to a slice of environment variables that
-	RootDirMap map[string][]string
-	Vars       *Vars
+	RootDirMap  map[string][]string
+	Vars        *Vars
+	PlatformAPI *api.Version
```
natalieparellano, line 19: Where is this used?

env/build_test.go
```diff
@@ -19,51 +21,113 @@ func TestBuildEnv(t *testing.T) {
...
-				"NOT_INCLUDED=not-included",
-				"PATH=some-path",
-				"LD_LIBRARY_PATH=some-ld-library-path",
-				"LIBRARY_PATH=some-library-path",
-				"CPATH=some-cpath",
-				"PKG_CONFIG_PATH=some-pkg-config-path",
+		when("platform API >= 0.8", func() {
+			var platformAPI *api.Version
+			it.Before(func() {
+				platformAPI = api.MustParse("0.8")
```
natalieparellano, line 27:
```suggestion
				platformAPI = api.Platform.Latest()
```

env/build_test.go
```diff
@@ -19,51 +21,113 @@ func TestBuildEnv(t *testing.T) {
...
+						"https_proxy=some-https-proxy",
+						"no_proxy=some-no-proxy",
+					)
+				}
+				if s := cmp.Diff(out, expectedVars); s != "" {
+					t.Fatalf("Unexpected env\n%s\n", s)
+				}
+			})
+		})
+		when("platform API <0.8", func() {
```
natalieparellano, line 78:
```suggestion
		when("platform API < 0.7", func() {
```

env/build_test.go
```diff
@@ -19,51 +21,113 @@ func TestBuildEnv(t *testing.T) {
...
+			var platformAPI *api.Version
+			it.Before(func() {
+				platformAPI = api.MustParse("0.7")
```
natalieparellano, line 81:
```suggestion
				platformAPI = api.MustParse("0.6")
```

env/build_test.go
```diff
@@ -76,7 +140,7 @@ func testBuildEnv(t *testing.T, when spec.G, it spec.S) {
 			it("ignores case when initializing", func() {
 				benv := env.NewBuildEnv([]string{
 					"Path=some-path",
-				})
+				}, api.MustParse("0.0"))
```
natalieparellano, line 143:
```suggestion
				}, api.Platform.Latest())
```

env/build_test.go
```diff
@@ -86,7 +150,7 @@ func testBuildEnv(t *testing.T, when spec.G, it spec.S) {
 		it("allows keys with '='", func() {
 			benv := env.NewBuildEnv([]string{
 				"CNB_STACK_ID=included=true",
-			})
+			}, api.MustParse("0.0"))
```
natalieparellano, line 153:
```suggestion
			}, api.Platform.Latest())
```

env/build_test.go
```diff
@@ -95,7 +159,7 @@ func testBuildEnv(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("assign the build time root dir map", func() {
-			benv := env.NewBuildEnv([]string{})
+			benv := env.NewBuildEnv([]string{}, api.MustParse("0.0"))
```
natalieparellano, line 162:
```suggestion
			benv := env.NewBuildEnv([]string{}, api.Platform.Latest())
```

env/build_test.go
```diff
@@ -19,51 +21,113 @@ func TestBuildEnv(t *testing.T) {
...
-				"LIBRARY_PATH=some-library-path",
-				"NO_PROXY=some-no-proxy",
-				"PATH=some-path",
-				"PKG_CONFIG_PATH=some-pkg-config-path",
-			}
-			// Environment variables in Windows are case insensitive, and are added by the lifecycle in uppercase.
-			if runtime.GOOS != "windows" {
-				expectedVars = append(
-					expectedVars,
+			it("can include CNB_ASSETS env var", func() {
```
natalieparellano, line 29: This test might be easier to understand if you kept the old `it("includes expected vars",` assertion and added a new assertion that just checks for `CNB_ASSETS`.

env/build_test.go
```diff
@@ -19,51 +21,113 @@ func TestBuildEnv(t *testing.T) {
...
+					t.Fatalf("Unexpected env\n%s\n", s)
+				}
+			})
+		})
+		when("platform API <0.8", func() {
+			var platformAPI *api.Version
+			it.Before(func() {
+				platformAPI = api.MustParse("0.7")
+			})
+			it("includes expected vars", func() {
```
natalieparellano, line 83:
```suggestion
			it("does not include CNB_ASSETS env var", func() {
```
(with same comment as https://github.com/buildpacks/lifecycle/pull/591/files?diff=split&w=1#r625334442 )

env/build.go
```diff
@@ -17,12 +19,29 @@ var BuildEnvIncludelist = []string{
...
+	return &Env{
+		RootDirMap: POSIXBuildEnv,
+		Vars:       varsFromEnv(environ, ignoreEnvVarCase, isNotIncludedForAPI(platformAPI)),
+	}
+}
+
+// NewBuildEnv returns an detect-time Env from the given environment.
+//
+// only keys in the BuildEnvIncludelist will be added to the Environment.
+func NewDetectEnv(environ []string) *Env {
```
natalieparellano, line 44: Could we add a test for this as well?

cmd/lifecycle/detector.go
natalieparellano: (since we already have acceptance tests for the detector)

env/build.go
```diff
@@ -19,32 +17,46 @@ var BuildEnvIncludelist = []string{
...
 }
 
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
 // NewBuildEnv returns an build-time Env from the given environment.
 //
 // Keys in the BuildEnvIncludelist will be added to the Environment.
-// if platformAPI >= 0.8 keys from BuildEnvIncludeAssets will be added to the environment
-func NewBuildEnv(environ []string, platformAPI *api.Version) *Env {
+// if the platform supoprts asset packages keys from AssetsEnvVars will be added to the environment
```
natalieparellano, line 39:
```suggestion
// If the platform supports asset packages, keys from AssetsEnvVars will be added to the environment.
```

env/vars.go
```diff
@@ -10,6 +12,9 @@ type Vars struct {
 func varsFromEnv(env []string, ignoreCase bool, removeKey func(string) bool) *Vars {
 	vars := NewVars(nil, ignoreCase)
 	for _, kv := range env {
+		//if strings.Contains(kv, "CNB_ASSETS") {
+		//	panic(fmt.Sprintf("key found: %s\n", kv))
+		//}
```
natalieparellano, line 17: :)

env/launch_test.go
```diff
@@ -80,9 +78,9 @@ func testLaunchEnv(t *testing.T, when spec.G, it spec.S) {
 			})
 
 			it("ignores case when initializing", func() {
-				benv := env.NewBuildEnv([]string{
+				benv := env.NewLaunchEnv([]string{
```
natalieparellano, line 81: Nice catch!

acceptance/testdata/detector/container/cnb/buildpacks/env_print_buildpack/env_print_buildpack_version/bin/detect
```diff
@@ -0,0 +1,17 @@
...
+
+[[requires]]
+name = "some_requirement"
+[requires.metadata]
+version = "some_version"     # Optional
+EOL
+
+cat >> "${plan_path}" <<EOL
+env = """$(env)"""
+EOL
```
natalieparellano, line 17: nit: missing newline

acceptance/detector_test.go
```diff
@@ -121,6 +122,61 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
+		})
+		it.After(func() {
+			if h.DockerContainerExists(t, containerName) {
+				h.Run(t, exec.Command("docker", "rm", containerName))
+			}
+			os.RemoveAll(copyDir)
+		})
+
+		it.Focus("does not appear in the detection environment", func() {
+			h.DockerRunAndCopy(t,
```
natalieparellano, line 141: You could just use `h.DockerRun` if you're okay with https://github.com/buildpacks/lifecycle/pull/591/files/2ab4678bd230c0aeb6d0c081991f992b0d3953fc..5ff86130c9a0a0f5176c617849564ecbe62ffb4e?file-filters%5B%5D=No+extension#r630437959

env/build_test.go
```diff
@@ -20,13 +22,71 @@ func TestBuildEnv(t *testing.T) {
...
+			}
+			// Environment variables in Windows are case insensitive, and are added by the lifecycle in uppercase.
+			if runtime.GOOS != "windows" {
+				expectedVars = append(
+					expectedVars,
+					"http_proxy=some-http-proxy",
+					"https_proxy=some-https-proxy",
+					"no_proxy=some-no-proxy",
+				)
+			}
```
natalieparellano: Same comment as https://github.com/buildpacks/lifecycle/pull/591/files/2ab4678bd230c0aeb6d0c081991f992b0d3953fc..5ff86130c9a0a0f5176c617849564ecbe62ffb4e?file-filters%5B%5D=.go&file-filters%5B%5D=.toml#r630444456

env/build.go
```diff
@@ -19,32 +17,46 @@ var BuildEnvIncludelist = []string{
...
+func NewBuildEnv(environ []string, platform Platform) *Env {
+	envFilter := composeFilters(inMatchList(BuildEnvIncludelist), inMatchList(flattenMap(POSIXBuildEnv)))
+	if platform.SupportsAssetPackages() {
+		envFilter = composeFilters(envFilter, inMatchList(AssetsEnvVars))
+	}
 	return &Env{
 		RootDirMap: POSIXBuildEnv,
-		Vars:       varsFromEnv(environ, ignoreEnvVarCase, isNotIncludedForAPI(platformAPI)),
+		Vars:       varsFromEnv(environ, ignoreEnvVarCase, envFilter),
```
natalieparellano, line 47: Thank you! All the logical inversions were making my brain hurt.

natalieparellano:
Left a few more comments!

env/build.go
```diff
@@ -107,3 +83,41 @@ var POSIXBuildEnv = map[string][]string{
...
+		return true
+	}
+}
+
+func inMatchList(list []string) removeFilter {
+	return func(key string) bool {
+		for _, wk := range list {
+			if matches(wk, key) {
+				// keep in env
+				return false
```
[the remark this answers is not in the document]
dwillist: I think this is a bit awkward no mater what as this is passed into to a `removeKey` (here)[https://github.com/buildpacks/lifecycle/blob/9ceda0d9dce039948ec800b3d92b8ec5846fda31/env/vars.go#L10]. Should we then not remove things that return true?

env/build.go
```diff
@@ -88,27 +87,16 @@ var POSIXBuildEnv = map[string][]string{
 // if removeFilter(key) returns true, then the env var will be removed from the respective environment
 type removeFilter func(key string) bool
 
-func composeFilters(filters ...removeFilter) removeFilter {
+func inMatchList(lists ...[]string) removeFilter {
```
natalieparellano, line 90:
```suggestion
func isNotMember(lists ...[]string) removeFilter {
```

natalieparellano: (Otherwise I would expect the function to return `true` when it matches)

env/build_test.go
```diff
@@ -5,16 +5,16 @@ import (
...
-	"github.com/buildpacks/lifecycle/env/testmock"
-
 	"github.com/google/go-cmp/cmp"
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
 	"github.com/buildpacks/lifecycle/env"
 	h "github.com/buildpacks/lifecycle/testhelpers"
+
+	"github.com/golang/mock/gomock"
```
natalieparellano, line 15: Before line 8

env/build_test.go
```diff
@@ -5,16 +5,16 @@ import (
...
+
+	"github.com/buildpacks/lifecycle/env/testmock"
```
natalieparellano, line 17: After line 12

natalieparellano: (this is annoying, apologies... I wish there was a way to do it automatically)

env/build_test.go
```diff
@@ -178,6 +143,8 @@ func testBuildEnv(t *testing.T, when spec.G, it spec.S) {
 				}
 			})
 			it("ignores case when initializing", func() {
+				platform.EXPECT().SupportsAssetPackages().Return(false)
```
natalieparellano, line 146: This is just a nit, but could we move it to the `Before` and change it to `true` (to be consistent with the "latest" platform). I just don't want to give the impression that somehow support (or not support) for asset packages is related to Windows.

platform/v06/platform.go
```diff
@@ -1,17 +1,13 @@
...
-}
-
-func (p *Platform) API() string {
-	return p.api
-}
-
-func (p *Platform) SupportsAssetPackages() bool {
-	return false
+	return &Platform{
+		Platform: v05.NewPlatform(apiStr),
```
natalieparellano, line 11: I am pretty sure that I wrote the original code, but I find it weird that the constructor in the `v05` package even takes a string as an argument. Shouldn't it always be `"0.5"`?

Separately though, I think we might start to get confused if we had a `v05.Platform` that returned `"0.6"` for its API... thinking...

natalieparellano: Here is what I was thinking of: https://play.golang.org/p/adQu0qhSHej

natalieparellano: I feel like... it works, but maybe it's too confusing and we're better off copying all of the methods to each platform. Curious to know what others think!

natalieparellano: Or maybe instead of silently delegating to the previous platform, we have a `previousPlatform` field and a method that explicitly calls the method on the previous platform, like:
```
func (p *Platform) CodeFor(errType cmd.LifecycleExitError) int {
	return p.previousPlatform.CodeFor(errType)
}
```

platform/platform.go
```diff
@@ -1,21 +1,34 @@
 package platform
 
 import (
+	"fmt"
+
 	"github.com/buildpacks/lifecycle/api"
-	"github.com/buildpacks/lifecycle/cmd"
-	v05 "github.com/buildpacks/lifecycle/platform/v05"
+	"github.com/buildpacks/lifecycle/platform/common"
+	pre06 "github.com/buildpacks/lifecycle/platform/pre06"
```
natalieparellano, line 8:
```suggestion
	"github.com/buildpacks/lifecycle/platform/pre06"
```
(redundant alias)

platform/platform.go
```diff
@@ -1,21 +1,34 @@
...
-type Platform interface {
-	API() string
-	CodeFor(errType cmd.LifecycleExitError) int
+var platform03 = pre06.NewPlatform("0.3")
+var platform04 = pre06.NewPlatform("0.4")
+var platform05 = pre06.NewPlatform("0.5")
+var platform06 = v06.NewPlatform(platform05)
+var platform07 = v07.NewPlatform(platform06)
+
+var platformMap = map[string]common.Platform{
```
natalieparellano, line 19:
```suggestion
var supportedPlatforms = map[string]common.Platform{
```

natalieparellano: (more descriptive)

platform/platform.go
```diff
@@ -1,21 +1,34 @@
...
+	"0.6": platform06,
+	"0.7": platform07,
 }
 
-func NewPlatform(apiStr string) Platform {
+func NewPlatform(apiStr string) (common.Platform, error) {
 	platformAPI := api.MustParse(apiStr)
-	if platformAPI.Compare(api.MustParse("0.6")) < 0 { // platform API < 0.6
-		return v05.NewPlatform(apiStr)
+	p, ok := platformMap[platformAPI.String()]
```
natalieparellano, line 29:
```suggestion
	p, ok := platformMap[apiStr]
```
This would allow you to delete the line above I believe.

platform/pre06/exit.go
```diff
@@ -1,4 +1,4 @@
-package v05
+package pre06
```
natalieparellano, line 1: This is a nice change :)

platform/platform.go
natalieparellano: I'm not sure why the linter doesn't also think the aliases for v06 and v07 are redundant...

platform/v06/platform.go
```diff
@@ -1,13 +1,22 @@
...
 
+import (
+	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/platform/common"
+)
+
 type Platform struct {
-	api string
+	api *api.Version
+	common.Platform
```
natalieparellano, line 10:
```suggestion
	previousPlatform common.Platform
```

natalieparellano: I still somewhat dislike the embedded struct. It's very convenient to embed it, but I feel that it might be better to be more explicit. With the suggestion above, for every method in the interface we'd need to write something like
```
func (p *Platform) SupportsAssetPackages() bool {
	return previousPlatform.SupportsAssetPackages()
}
```
or a new implementation. It is slightly annoying but it might help us not forget to override the previous logic when we need to. And it might make it more obvious to someone new to the code where the implementation is coming from.

That said, I'd be happy to hear (and learn from) other opinions!

env/build_test.go
```diff
@@ -18,87 +20,135 @@ func TestBuildEnv(t *testing.T) {
...
+			})
+			it("does not include CNB_ASSETS env var", func() {
+				benv := env.NewBuildEnv([]string{
+					"CNB_ASSETS=some-assets-path",
+				}, platform)
+				out := benv.List()
+				sort.Strings(out)
+				var expectedVars []string
+				// Environment variables in Windows are case insensitive, and are added by the lifecycle in uppercase.
+				if s := cmp.Diff(out, expectedVars); s != "" {
```
natalieparellano, line 131:
```suggestion
				if s := cmp.Diff(benv.List(), []string{}); s != "" {
```

env/build_test.go
```diff
@@ -18,87 +20,135 @@ func TestBuildEnv(t *testing.T) {
...
-				"HOSTNAME=some-hostname",
-				"HTTPS_PROXY=some-https-proxy",
-				"HTTP_PROXY=some-http-proxy",
-				"LD_LIBRARY_PATH=some-ld-library-path",
-				"LIBRARY_PATH=some-library-path",
-				"NO_PROXY=some-no-proxy",
-				"PATH=some-path",
-				"PKG_CONFIG_PATH=some-pkg-config-path",
-			}
+			var expectedVars []string
```
natalieparellano, line 39: This could be made terser as suggested here: https://github.com/buildpacks/lifecycle/pull/591/files#r635441542

env/build.go
```diff
@@ -17,15 +17,45 @@ var BuildEnvIncludelist = []string{
...
-		Vars:       varsFromEnv(environ, ignoreEnvVarCase, isNotIncluded),
+		Vars:       varsFromEnv(environ, ignoreEnvVarCase, envFilter),
+	}
+}
+
+// NewDetectEnv returns an detect-time Env from the given environment.
+//
+// only keys in the BuildEnvIncludelist will be added to the Environment.
+func NewDetectEnv(environ []string) *Env {
+	envFilter := isNotMember(append(BuildEnvIncludelist, flattenMap(POSIXBuildEnv)...))
```
natalieparellano, line 54:
```suggestion
	envFilter := isNotMember(BuildEnvIncludelist, flattenMap(POSIXBuildEnv)
```

natalieparellano: I think this will also work...

env/build.go
```diff
@@ -68,3 +82,30 @@ var POSIXBuildEnv = map[string][]string{
 		"PKG_CONFIG_PATH",
 	},
 }
+
+// removeFilter is generic function used to filter env vars out of the environment
+// if removeFilter(key) returns true, then the env var will be removed from the respective environment
+type removeFilter func(key string) bool
```
natalieparellano, line 88: Since this is only used by `isNotMember` now I wonder if we need it. Alternatively we could change the signature of `varsFromEnv` to accept this type.

env/build.go
```diff
@@ -17,15 +17,45 @@ var BuildEnvIncludelist = []string{
...
+}
+
 var ignoreEnvVarCase = runtime.GOOS == "windows"
 
 // NewBuildEnv returns an build-time Env from the given environment.
 //
-// Only keys in the BuildEnvIncludelist will be added to the Environment.
-func NewBuildEnv(environ []string) *Env {
+// Keys in the BuildEnvIncludelist will be added to the Environment.
+// if the platform supports asset packages, keys from AssetsEnvVars will be added to the environment.
```
natalieparellano, line 38:
```suggestion
// If the platform supports asset packages, keys from AssetsEnvVars will be added to the environment.
```

env/build.go
```diff
@@ -17,15 +17,45 @@ var BuildEnvIncludelist = []string{
...
 	return &Env{
 		RootDirMap: POSIXBuildEnv,
-		Vars:       varsFromEnv(environ, ignoreEnvVarCase, isNotIncluded),
+		Vars:       varsFromEnv(environ, ignoreEnvVarCase, envFilter),
+	}
+}
+
+// NewDetectEnv returns an detect-time Env from the given environment.
+//
+// only keys in the BuildEnvIncludelist will be added to the Environment.
```
natalieparellano, line 52:
```suggestion
// Only keys in the BuildEnvIncludelist will be added to the Environment.
```

env/build_test.go
dwillist: Whoops sorry missed that

cmd/lifecycle/builder.go
```diff
@@ -83,12 +83,13 @@ func (ba buildArgs) build(group buildpack.Group, plan platform.BuildPlan) error
 		return cmd.FailErrCode(err, ba.platform.CodeFor(cmd.BuildError), "build")
 	}
 
+	platformAPI := api.MustParse(ba.platform.API())
```
natalieparellano, line 86: Do we need this variable?

cmd/exit.go
```diff
@@ -38,6 +38,7 @@ const (
 type Platform interface {
 	API() string
 	CodeFor(errType LifecycleExitError) int
+	SupportsAssetPackages() bool
```
natalieparellano, line 41: There is some interesting entanglement happening between cmd and platform packages - I am not sure what to do about it, maybe it's fine as it is. @jabrown85 @yaelharel any thoughts here?

platform/v06/platform.go
natalieparellano: See other discussion: https://github.com/buildpacks/lifecycle/pull/591#discussion_r635433290

env/build_test.go
natalieparellano: Sorry I should have said "as also suggested" ...I wasn't able to make a new suggestion because some of the lines didn't change...

natalieparellano:
This is looking great! I left a few nits. I'd be curious to hear opinions about https://github.com/buildpacks/lifecycle/pull/591#discussion_r635433290 (`previousPlatform` delegation) and https://github.com/buildpacks/lifecycle/pull/591#discussion_r635461363 (relationship between `cmd` and `platform` packages) but these aren't blockers on my side.

platform/v06/platform.go
jabrown85: I don't have a strong opinion here, I lean a bit toward the embedding as it is. Mostly just because of duplication avoided.

cmd/exit.go
jabrown85: I don't love the entanglement happening. Why do we need `SupportsAssetPackages` in the cmd interface? Is it because we have end up sending `cmd.Platform` into a `lifecycle.Platform` function somewhere?

natalieparellano: It's [here](https://github.com/dwillist/lifecycle/blob/3ffc711f63cece48178021066f97b3ea9e636ac6/cmd/lifecycle/builder.go#L92) in cmd/lifecycle/builder.go: `env.NewBuildEnv(os.Environ(), ba.platform)`

env/build.go
natalieparellano: Removing for now.

platform/v06/platform.go
natalieparellano: Hmm... normally I'm not a fan of duplication but I like that you could read the code for any of the platform/vX packages and it would essentially describe the platform spec (granted, we haven't moved all platform-related logic into this package, but eventually...). Having the code mirror the spec seems like a win from an organizational perspective.

One other reason not to embed the struct - take the definition of `API()`:
```
func (p *Platform) API() string {
	return p.api.String()
}
```

If we forgot to add this method to a new platform package, the code would compile but the new platform would return the previous platform's api version. With the `previousPlatform` field, if we forget to add a method the code would not compile. This seems preferable.

Since I inherited this PR I added a7cdb8e with my preferred implementation 😛  but I'm open to changing it back depending on opinions.

jabrown85 (approved):
This looks pretty good to me! I am not sure how I'll feel about the platform interface changes in practice, but nothing worth blocking over.

natalieparellano:
I realized that this PR is missing this little bit of logic that came up in discussion recently: https://github.com/buildpacks/spec/pull/228#discussion_r648629762

I'll rework it a bit to add that in...

env/build_test.go
```diff
@@ -42,89 +49,120 @@ func testBuildEnv(t *testing.T, when spec.G, it spec.S) {
...
-			})
-			it("includes expected vars", func() {
-				benv := env.NewBuildEnv([]string{
-					"CNB_STACK_ID=some-stack-id",
-					"HOSTNAME=some-hostname",
-					"HOME=some-home",
-					"HTTPS_PROXY=some-https-proxy",
-					"https_proxy=some-https-proxy",
-					"HTTP_PROXY=some-http-proxy",
+		it("includes expected vars", func() {
```
natalieparellano, line 52: I changed the nesting of tests in order to not repeat this assertion a bunch of times for the various combos of platform & buildpack. It's probably easiest to view this diff with main.

env/build_test.go
```diff
@@ -66,27 +99,13 @@ func testBuildEnv(t *testing.T, when spec.G, it spec.S) {
 			}
 		})
 
-		when("building in Windows", func() {
```
natalieparellano, line 69: Moved to the end of the file

natalieparellano:
The small change in logic unfortunately did not result in a small change in code :/

However, I now believe this to be complete.

natalieparellano:
This sadly still has a bug... looking...

natalieparellano:
Hopefully fixed now (pending CI). This is one that we should definitely do manual acceptance on, since there's no acceptance tests for the builder. (cc @yaelharel )

jabrown85 (approved):

natalieparellano:
Confirmed that for older buildpacks, `CNB_ASSETS` is not included.

Using this buildpack:
```
$  cat /tmp/validate/asset_packages/old/some-buildpack/1.0.0/bin/build 
#!/usr/bin/env bash
set -eo pipefail

layers_dir="$1"
env_dir="$2/env"
plan_path="$3"

env

exit 0
```

And this group:
```
$  cat /tmp/validate/asset_packages/old/group.toml 
[[group]]
id = "some-buildpack"
version = "1.0.0"
api = "0.4"
```

When not supported by platform, it is not printed:
```
$  CNB_PLATFORM_API="0.6" CNB_ASSETS="some-assets-path" docker run -v $PWD:/workspace -v /tmp/validate/asset_packages/old:/cnb/buildpacks -v /tmp/validate/asset_packages/old:/layers my-builder /cnb/lifecycle/builder
HOSTNAME=2c55f8e9c4ab
CNB_STACK_ID=io.buildpacks.stacks.bionic
PWD=/workspace
HOME=/home/cnb
CNB_BUILDPACK_DIR=/cnb/buildpacks/some-buildpack/1.0.0
SHLVL=1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env
```

What supported by platform, it is not printed:
```
$  CNB_PLATFORM_API="0.7" CNB_ASSETS="some-assets-path" docker run -v $PWD:/workspace -v /tmp/validate/asset_packages/old:/cnb/buildpacks -v /tmp/validate/asset_packages/old:/layers my-builder /cnb/lifecycle/builder
HOSTNAME=a1df262e2242
CNB_STACK_ID=io.buildpacks.stacks.bionic
PWD=/workspace
HOME=/home/cnb
CNB_BUILDPACK_DIR=/cnb/buildpacks/some-buildpack/1.0.0
SHLVL=1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env
```

natalieparellano:
Confirmed that for newer buildpacks, CNB_ASSETS is included when supported by the platform.

Using this buildpack:
```
$  cat /tmp/validate/asset_packages/new/some-buildpack/1.0.0/bin/build
#!/usr/bin/env bash
set -eo pipefail

layers_dir="$1"
env_dir="$2/env"
plan_path="$3"

env

exit 0
```

And this group:
```
$  cat /tmp/validate/asset_packages/new/group.toml 
[[group]]
id = "some-buildpack"
version = "1.0.0"
api = "0.7"
```

When not supported by platform, it is not printed:
```
$  docker run --env CNB_PLATFORM_API="0.6" --env CNB_ASSETS="some-assets-path" -v $PWD:/workspace -v /tmp/validate/asset_packages/new:/cnb/buildpacks -v /tmp/validate/asset_packages/new:/layers my-builder /cnb/lifecycle/builder
HOSTNAME=be1949d1d214
CNB_STACK_ID=io.buildpacks.stacks.bionic
PWD=/workspace
HOME=/home/cnb
CNB_BUILDPACK_DIR=/cnb/buildpacks/some-buildpack/1.0.0
SHLVL=1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env
```

What supported by platform, it is printed:
```
$  docker run --env CNB_PLATFORM_API="0.7" --env CNB_ASSETS="some-assets-path" -v $PWD:/workspace -v /tmp/validate/asset_packages/new:/cnb/buildpacks -v /tmp/validate/asset_packages/new:/layers my-builder /cnb/lifecycle/builder
HOSTNAME=8e3172b2572d
CNB_STACK_ID=io.buildpacks.stacks.bionic
PWD=/workspace
HOME=/home/cnb
CNB_BUILDPACK_DIR=/cnb/buildpacks/some-buildpack/1.0.0
CNB_ASSETS=some-assets-path
SHLVL=1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env
```

PR #609, opened 2021-05-06
(bugsbash) change template

asquare14:
Signed-off-by: aSquare14 <atibhi.a@gmail.com>
opt.semgrep.import-text-template
'text/template' does not escape HTML content. If you need to escape HTML content, use 'html/template' instead.

### DIFF
--- a/tools/packager/main.go
+++ b/tools/packager/main.go
@@ -6,10 +6,10 @@ import (
 	"compress/gzip"
 	"flag"
 	"fmt"
+	"html/template"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


natalieparellano:
@asquare14 thanks for the PR! Sadly I think this is another one where we'll likely want to ignore the warning from muse-dev instead of making the change. With `html/template` the version printed in lifecycle.toml looks like `version = "0.11.2-45&#43;d690a9a"` as opposed to `version = "0.11.2-55+ac7046c"` which is what we want.

Issue #471, opened 2020-11-16
[Analyze] Validate stack IDs


natalieparellano:
The buildpacks [spec](https://github.com/buildpacks/spec/blob/main/buildpack.md#process) states that

> The lifecycle MUST run `/bin/detect` for all buildpacks in a group in a container using common stack with a common set of mixins. The lifecycle MUST fail detection if any of those buildpacks does not list that stack in `buildpack.toml`. The lifecycle MUST fail detection if any of those buildpacks specifies a mixin associated with that stack in `buildpack.toml` that is unavailable in the container.

However, the lifecycle does not currently validate stack IDs (leaving this up to the platform). 

The lifecycle should validate that the provided buildpacks and the provided stack can work together by validating that all buildpacks list the provided stack (or `*`, if on a buildpack API that contains https://github.com/buildpacks/rfcs/blob/main/text/0056-any-stack-buildpacks.md). The lifecycle can examine `CNB_STACK_ID` to know what stack was provided.

ekcasey:
In addition, we should check that the following all match
1. The previous image stack ID
2. The build environment stack ID
3. The run image stack ID

We should also consider checking whether the layers in the cache created for the correct stack ID. If we decided to do this:
1. We will need to store stack ID in the cache metadata.
2. For backwards compatibility, in cases where the cache metadata is missing the stack ID we should be optimistic and assume the layers were created for the correct stack.

natalieparellano:
@jabrown85 do we want to move this to 0.12.0? I think it'll go nicely with https://github.com/buildpacks/spec/issues/206

jabrown85:
That sounds good to me @natalieparellano

natalieparellano:
> We should also consider checking whether the layers in the cache were created for the correct stack ID. If we decided to do this:

>We will need to store stack ID in the cache metadata.
For backwards compatibility, in cases where the cache metadata is missing the stack ID we should be optimistic and assume the layers were created for the correct stack.

I think this probably makes sense to be a separate issue.

natalieparellano:
Q: If stack IDs are the same, can we assume that the images have the same platform & architecture?

natalieparellano:
Blocked on https://github.com/buildpacks/lifecycle/issues/530

natalieparellano:
Removing this from 0.12.0 milestone for the same reason as https://github.com/buildpacks/lifecycle/issues/573#issuecomment-867809036

PR #617, opened 2021-05-12
Validate stack during analyze phase

jabrown85:
During analyze, Platform 0.7 and above will validate the build and stack image if the data is available to do so.

Addresses Part of: #471

[Spec](https://github.com/buildpacks/spec/blob/platform/0.7/platform.md#analyzer) being implemented:
> The lifecycle MUST fail if the stack ID of the `<run-image>` does not match `<stack-id>`

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -73,6 +73,36 @@ func TestAnalyzer(t *testing.T) {
 	h.AssertNil(t, os.RemoveAll(filepath.Join(targetDockerConfig, "config.json")))
 	h.RecursiveCopy(t, authRegistry.DockerDirectory, targetDockerConfig)
 
+	// build run-images into test registry
+	runImageContext := filepath.Join("testdata", "analyzer", "run-image")
+	buildAuthRegistryImage(
+		t,
+		"company/stack:bionic",
+		runImageContext,
+		"-f", filepath.Join(runImageContext, dockerfileName),
+		"--build-arg", "stackid=io.buildpacks.stacks.bionic",
+	)
+	buildAuthRegistryImage(
+		t,
+		"company/stack:centos",
+		runImageContext,
+		"-f", filepath.Join(runImageContext, dockerfileName),
+		"--build-arg", "stackid=io.company.centos",
+	)
+
+	// build run-image into daemon
+	h.DockerBuild(
+		t,
+		"localcompany/stack:bionic",
+		runImageContext,
+		h.WithArgs(
+			"-f", filepath.Join(runImageContext, dockerfileName),
+			"--build-arg", "stackid=io.buildpacks.stacks.bionic",
+		),
+	)
+
+	defer h.DockerImageRemove(t, "localcompany/stack:bionic")
+
 	// Setup test container
 
 	h.MakeAndCopyLifecycle(t, daemonOS, analyzerBinaryDir)
@@ -81,6 +111,7 @@ func TestAnalyzer(t *testing.T) {
 		analyzeDockerContext,
 		h.WithFlags(
 			"-f", filepath.Join(analyzeDockerContext, dockerfileName),
+			"--build-arg", "registry="+noAuthRegistry.Host+":"+noAuthRegistry.Port,
 		),
 	)
 	defer h.DockerImageRemove(t, analyzeImage)
@@ -253,7 +284,10 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					copyDir,
 					ctrPath("/some-dir/some-analyzed.toml"),
 					analyzeImage,
-					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI),
+					h.WithFlags(
+						"--network", registryNetwork,
+						"--env", "CNB_PLATFORM_API="+platformAPI,
+					),
 					h.WithArgs(execArgs...),
 				)
 
@@ -277,6 +311,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					h.WithFlags(append(
 						dockerSocketMount,
 						"--env", "CNB_PLATFORM_API="+platformAPI,
+						"--env", "CNB_STACK_PATH=/cnb/local-bionic-stack.toml", // /cnb/local-bionic-stack.toml has `io.buildpacks.stacks.bionic` and points to run image `localcompany/stack:bionic` with same stack id
 					)...),
 					h.WithArgs(execArgs...),
 				)
@@ -316,6 +351,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						h.WithFlags(append(
 							dockerSocketMount,
 							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--env", "CNB_STACK_PATH=/cnb/local-bionic-stack.toml", // /cnb/local-bionic-stack.toml has `io.buildpacks.stacks.bionic` and points to run image `localcompany/stack:bionic` with same stack id
 						)...),
 						h.WithArgs(
 							ctrPath(analyzerPath),
@@ -624,6 +660,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					ctrPath("/layers/analyzed.toml"),
 					analyzeImage,
 					h.WithFlags(
+						"--network", registryNetwork,
 						"--env", "CNB_PLATFORM_API="+platformAPI,
 					),
 					h.WithArgs(ctrPath(analyzerPath), "some-image"),
@@ -1063,6 +1100,166 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				h.AssertPathExists(t, filepath.Join(copyDir, "some-other-layers", "another-buildpack-id")) // another-buildpack-id is found in the provided -layers directory: /some-other-layers/group.toml
 
 				assertAnalyzedMetadata(t, filepath.Join(copyDir, "some-other-layers", "analyzed.toml")) // analyzed.toml is written at the provided -layers directory: /some-other-layers
+			})
+		})
+
+		when("validating stack", func() {
+			it.Before(func() {
+				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not validate stack")
+			})
+
+			when("stack metadata is present", func() {
+				when("stacks match", func() {
+					it("passes validation", func() {
+						execArgs := []string{ctrPath(analyzerPath), "some-image"}
+						h.DockerRun(t,
+							analyzeImage, // /cnb/stack.toml has `io.buildpacks.stacks.bionic` and points to run image `company/stack:bionic` with same stack id
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("CNB_RUN_IMAGE is present", func() {
+					it("uses CNB_RUN_IMAGE for validation", func() {
+						execArgs := []string{ctrPath(analyzerPath), "some-image"}
+
+						h.DockerRun(t,
+							analyzeImage,
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--env", "CNB_STACK_PATH=/cnb/mismatch-stack.toml", // /cnb/mismatch-stack.toml points to run image `company/stack:centos`
+								"--env", "CNB_RUN_IMAGE="+noAuthRegistry.RepoName("company/stack:bionic"),
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("stack metadata file is invalid", func() {
+					it("fails validation", func() {
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--network", registryNetwork,
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--env", "CNB_STACK_PATH=/cnb/bad-stack.toml",
+							analyzeImage,
+							ctrPath(analyzerPath),
+							"some-image",
+						) // #nosec G204
+						output, err := cmd.CombinedOutput()
+
+						h.AssertNotNil(t, err)
+						expected := "get stack metadata"
+						h.AssertStringContains(t, string(output), expected)
+					})
+				})
+
+				when("run image inaccessible", func() {
+					it("fails validation", func() {
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--network", registryNetwork,
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--env", "CNB_RUN_IMAGE=fake.example.com/company/example:20",
+							analyzeImage,
+							ctrPath(analyzerPath),
+							"some-image",
+						) // #nosec G204
+						output, err := cmd.CombinedOutput()
+
+						h.AssertNotNil(t, err)
+						expected := "failed to resolve run image"
+						h.AssertStringContains(t, string(output), expected)
+					})
+				})
+
+				when("run image has mirrors", func() {
+					it("uses expected mirror for run-image", func() {
+						execArgs := []string{ctrPath(analyzerPath), noAuthRegistry.RepoName("apprepo/myapp")} // image located on same registry as mirror
+
+						h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							ctrPath("/layers/analyzed.toml"),
+							analyzeImage,
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--env", "CNB_STACK_PATH=/cnb/run-mirror-stack.toml", // /cnb/run-mirror-stack.toml points to run image on gcr.io and mirror on test registry
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("daemon case", func() {
+					when("stacks match", func() {
+						it("passes validation", func() {
+							execArgs := []string{ctrPath(analyzerPath), "-daemon", "some-image"}
+
+							h.DockerRunAndCopy(t,
+								containerName,
+								copyDir,
+								ctrPath("/layers/analyzed.toml"),
+								analyzeImage,
+								h.WithFlags(append(
+									dockerSocketMount,
+									"--network", registryNetwork,
+									"--env", "CNB_PLATFORM_API="+platformAPI,
+									"--env", "CNB_STACK_PATH=/cnb/local-bionic-stack.toml", // /cnb/local-bionic-stack.toml has `io.buildpacks.stacks.bionic` and points to run image `localcompany/stack:bionic` with same stack id
+								)...),
+								h.WithArgs(execArgs...),
+							)
+						})
+					})
+				})
+			})
+
+			when("stack metadata is not present", func() {
+				when("CNB_RUN_IMAGE and CNB_STACK_ID are set", func() {
+					it("passes validation", func() {
+						execArgs := []string{ctrPath(analyzerPath), "some-image"}
+
+						h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							ctrPath("/layers/analyzed.toml"),
+							analyzeImage,
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--env", "CNB_STACK_PATH=/cnb/file-does-not-exist.toml",
+								"--env", "CNB_RUN_IMAGE="+noAuthRegistry.RepoName("company/stack:bionic"),
+								"--env", "CNB_STACK_ID=io.buildpacks.stacks.bionic",
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("run image and stack id are not provided as arguments or in the environment", func() {
+					it("fails validation", func() {
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--network", registryNetwork,
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--env", "CNB_STACK_PATH=/cnb/file-does-not-exist.toml",
+							analyzeImage,
+							ctrPath(analyzerPath),
+							"some-image",
+						) // #nosec G204
+						output, err := cmd.CombinedOutput()
+
+						h.AssertNotNil(t, err)
+						expected := "a run image must be specified when there is no stack metadata available"
+						h.AssertStringContains(t, string(output), expected)
+					})
+				})
 			})
 		})
 	}
--- a/acceptance/testdata/analyzer/analyze-image/Dockerfile
+++ b/acceptance/testdata/analyzer/analyze-image/Dockerfile
@@ -19,3 +19,44 @@ RUN chown -R $CNB_USER_ID:$CNB_GROUP_ID /layers
 
 # ensure docker config directory is root owned and NOT world readable
 RUN chown -R root /docker-config; chmod -R 700 /docker-config
+
+ARG registry
+
+# write some stack.toml files to use in tests
+RUN echo "\
+[run-image]\n\
+ image = \"${registry}/company/stack:bionic\"\n\
+ mirrors = []\n\
+[build-image]\n\
+ stack-id = \"io.buildpacks.stacks.bionic\"\n\
+ mixins = []\n\
+" > /cnb/stack.toml
+
+RUN echo "\
+[run-image]\n\
+ image = \"${registry}/company/stack:centos\"\n\
+ mirrors = []\n\
+[build-image]\n\
+ stack-id = \"io.buildpacks.stacks.bionic\"\n\
+ mixins = []\n\
+" > /cnb/mismatch-stack.toml
+
+RUN echo "\
+[run-image]\n\
+ image = \"gcr.io/paketobuildpacks/invalidimg:20\"\n\
+ mirrors = [\"${registry}/company/stack:bionic\"]\n\
+[build-image]\n\
+ stack-id = \"io.buildpacks.stacks.bionic\"\n\
+ mixins = []\n\
+" > /cnb/run-mirror-stack.toml
+
+RUN echo "\
+[run-image]\n\
+ image = \"localcompany/stack:bionic\"\n\
+ mirrors = []\n\
+[build-image]\n\
+ stack-id = \"io.buildpacks.stacks.bionic\"\n\
+ mixins = []\n\
+" > /cnb/local-bionic-stack.toml
+
+RUN echo "[run-images" > /cnb/bad-stack.toml
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -161,10 +161,6 @@ func (a *analyzeCmd) Exec() error {
 		}
 	}
 
-	if err := a.validateStack(); err != nil {
-		return cmd.FailErr(err, "validate stack")
-	}
-
 	analyzedMD, err := a.analyze()
 	if err != nil {
 		return err
@@ -177,7 +173,29 @@ func (a *analyzeCmd) Exec() error {
 	return nil
 }
 
+func (aa analyzeArgs) validateStack() error {
+	if !aa.supportsStackValidation() {
+		return nil
+	}
+
+	var stackMD platform.StackMetadata
+	if _, err := toml.DecodeFile(aa.stackPath, &stackMD); err != nil && !os.IsNotExist(err) {
+		return cmd.FailErr(err, "get stack metadata")
+	}
+
+	runImage, err := aa.getRunImage(stackMD)
+	if err != nil {
+		return cmd.FailErr(err, "resolve run image")
+	}
+
+	return lifecycle.ValidateStack(stackMD, runImage)
+}
+
 func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
+	if err := aa.validateStack(); err != nil {
+		return platform.AnalyzedMetadata{}, cmd.FailErr(err, "validate stack")
+	}
+
 	var (
 		img imgutil.Image
 		err error
@@ -213,46 +231,28 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 	return analyzedMD, nil
 }
 
-func (a *analyzeCmd) validateStack() error {
-	if !a.supportsStackValidation() {
-		return nil
-	}
-
-	var stackMD platform.StackMetadata
-	if _, err := toml.DecodeFile(a.stackPath, &stackMD); err != nil && !os.IsNotExist(err) {
-		return cmd.FailErr(err, "get stack metadata")
-	}
-
-	runImage, err := a.getRunImage(stackMD)
-	if err != nil {
-		return cmd.FailErr(err, "resolve run image")
-	}
-
-	return lifecycle.ValidateStack(stackMD, runImage)
-}
-
-func (a *analyzeCmd) getRunImage(stackMD platform.StackMetadata) (imgutil.Image, error) {
-	if a.runImageRef == "" {
-		runImageRef, err := lifecycle.ResolveRunImage(stackMD, a.imageName)
+func (aa *analyzeArgs) getRunImage(stackMD platform.StackMetadata) (imgutil.Image, error) {
+	if aa.runImageRef == "" {
+		runImageRef, err := lifecycle.ResolveRunImage(stackMD, aa.imageName)
 		if err != nil {
 			return nil, err
 		}
-		a.runImageRef = runImageRef
+		aa.runImageRef = runImageRef
 	}
 
 	var runImage imgutil.Image
 	var err error
-	if a.useDaemon {
+	if aa.useDaemon {
 		runImage, err = local.NewImage(
-			a.runImageRef,
-			a.docker,
-			local.FromBaseImage(a.runImageRef),
+			aa.runImageRef,
+			aa.docker,
+			local.FromBaseImage(aa.runImageRef),
 		)
 	} else {
 		runImage, err = remote.NewImage(
-			a.runImageRef,
-			a.keychain,
-			remote.FromBaseImage(a.runImageRef),
+			aa.runImageRef,
+			aa.keychain,
+			remote.FromBaseImage(aa.runImageRef),
 		)
 	}
 	return runImage, err
@@ -273,8 +273,8 @@ func (a *analyzeCmd) restoresLayerMetadata() bool {
 	return !a.platformAPIVersionGreaterThan06()
 }
 
-func (a *analyzeCmd) supportsStackValidation() bool {
-	return a.platformAPIVersionGreaterThan06()
+func (aa *analyzeArgs) supportsStackValidation() bool {
+	return api.MustParse(aa.platform.API()).Compare(api.MustParse("0.7")) >= 0
 }
 
 func (a *analyzeCmd) platformAPIVersionGreaterThan06() bool {
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"fmt"
+	"os"
 
+	"github.com/BurntSushi/toml"
 	"github.com/buildpacks/imgutil"
 	"github.com/buildpacks/imgutil/local"
 	"github.com/buildpacks/imgutil/remote"
@@ -159,6 +161,10 @@ func (a *analyzeCmd) Exec() error {
 		}
 	}
 
+	if err := a.validateStack(); err != nil {
+		return cmd.FailErr(err, "validate stack")
+	}
+
 	analyzedMD, err := a.analyze()
 	if err != nil {
 		return err
@@ -207,6 +213,51 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 	return analyzedMD, nil
 }
 
+func (a *analyzeCmd) validateStack() error {
+	if !a.supportsStackValidation() {
+		return nil
+	}
+
+	var stackMD platform.StackMetadata
+	if _, err := toml.DecodeFile(a.stackPath, &stackMD); err != nil && !os.IsNotExist(err) {
+		return cmd.FailErr(err, "get stack metadata")
+	}
+
+	runImage, err := a.getRunImage(stackMD)
+	if err != nil {
+		return cmd.FailErr(err, "resolve run image")
+	}
+
+	return lifecycle.ValidateStack(stackMD, runImage)
+}
+
+func (a *analyzeCmd) getRunImage(stackMD platform.StackMetadata) (imgutil.Image, error) {
+	if a.runImageRef == "" {
+		runImageRef, err := lifecycle.ResolveRunImage(stackMD, a.imageName)
+		if err != nil {
+			return nil, err
+		}
+		a.runImageRef = runImageRef
+	}
+
+	var runImage imgutil.Image
+	var err error
+	if a.useDaemon {
+		runImage, err = local.NewImage(
+			a.runImageRef,
+			a.docker,
+			local.FromBaseImage(a.runImageRef),
+		)
+	} else {
+		runImage, err = remote.NewImage(
+			a.runImageRef,
+			a.keychain,
+			remote.FromBaseImage(a.runImageRef),
+		)
+	}
+	return runImage, err
+}
+
 func (a *analyzeCmd) registryImages() []string {
 	var registryImages []string
 	if a.platform06.cacheImageTag != "" {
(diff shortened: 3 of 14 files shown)


jabrown85:
oh jeez - nevermind . this doesn't address #471 yet - I thought that ticket was about making sure build and run images match..but maybe we do this anyway since it is related? 😬 

The `buildpack.toml` check I can work on, but it is a different phase since you only want to validate for the selected group I assume?

jabrown85:
I put in this fix while working on this. https://github.com/buildpacks/spec/pull/225

jabrown85:
@micahyoung I would love your assistance on fixing windows tests. I added skips for the 6 tests I would expect to pass. I am building platform specific run images but I was still seeing failures as if the run image wasn't in the registry. Is there something I need to do to be able to push a _windows_ image into the ggcr registry? If you have any thoughts to try out let me know!

https://github.com/buildpacks/lifecycle/pull/617/checks?check_run_id=2612523589

micahyoung:
@jabrown85 Happy to take a look, as I know this is dragging on for ya. I'll see if I can get your current branch passing locally and send you a patch.

natalieparellano:
Did we answer the question of whether stack.toml should be required?

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1053,247 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("stacks do not match", func() {
```
natalieparellano, line 1101: At first glance this seems like a lot of logic to be testing at the acceptance level. I wonder if we could pull the stack validation into a separate unit-testable service. I feel like our acceptance tests should really only be testing that the correct file is being read, args are respected, etc. But I haven't read through the whole PR yet so let me know if there is something I am missing here!

jabrown85: Yeah - a lot of the validation in other areas is in the `cmd/`, but I agree it feels like too much. Both the `exporter` and `rebaser` do the run image resolution in the `cmd/`. Do you want me to take a go at moving things around?

jabrown85: I took a swing at breaking things apart in 50dd6af611b70a935607791d9ebbf4e68822a4a4 and ec87c177d14e331536dc789acda5ba67ab6a690e. Let me know what you think. 

I find it hard to move things around in this repo. The line of what belongs in the cmd vs what belongs in the analyzer is blurry at best. We have things like `dockerClient` that we tend to not leak into the analyzer, for instance. But that seam makes it difficult to write something like a `ImageResolver` that has an interface like `resolver.Resolve(runImageRef)` that abstracts away the remote vs local. So that local vs remote logic ends up in each command. 

Maybe giving `lifecycle.Analyzer` an `imageResolver: defaultImageResolver(useDaemon, dockerClient, regCreds)` and push _all_ the image resolution inside could work out. Make the cmds as anemic as possible so that we can unit test. We could also have `fileResolver: defaultFileResolver()` that we use to read files that we can mock out as well to push all the reads into the analyzer and out of the cmd.

This feels like a larger discussion though...

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1053,151 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				assertAnalyzedMetadata(t, filepath.Join(copyDir, "some-other-layers", "analyzed.toml")) // analyzed.toml is written at the provided -layers directory: /some-other-layers
 			})
 		})
+
+		when("validating stack", func() {
+			it.Before(func() {
+				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not validate stack")
+			})
+
+			when("stack metadata is present", func() {
```
yaelharel, line 1062: What should happen when the stack metadata doesn't exist? Do we want to test it?

cmd/lifecycle/analyzer.go
```diff
@@ -208,6 +214,51 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 	return analyzedMD, nil
 }
 
+func (a *analyzeCmd) validateStack() error {
+	if !a.supportsStackValidation() {
+		return nil
+	}
+
+	var stackMD platform.StackMetadata
+	if _, err := toml.DecodeFile(a.stackPath, &stackMD); err != nil && !os.IsNotExist(err) {
```
yaelharel, line 223: What should happen if the file doesn't exist? Don't we want to handle it or at least add a comment about it?

stack_validation.go
```diff
@@ -0,0 +1,44 @@
...
+func getBuildStack(stackMD platform.StackMetadata) (string, error) {
+	buildStackID := os.Getenv(cmd.EnvStackID)
+	if buildStackID == "" {
+		buildStackID = stackMD.BuildImage.StackID
+	}
+
+	if buildStackID == "" {
+		return "", errors.New("CNB_STACK_ID is required when there is no stack metadata available")
+	}
+	return buildStackID, nil
```
yaelharel, line 43: I found this code a bit confusing. What do you think about:

```suggestion
	var buildStackID string

	if buildStackID = os.Getenv(cmd.EnvStackID); buildStackID != "" {
		return buildStackID, nil
	}
	if buildStackID = stackMD.BuildImage.StackID; buildStackID != "" {
		return buildStackID, nil
	}
	return  "", errors.New("CNB_STACK_ID is required when there is no stack metadata available")
```

or if we don't care about calling `os.Getenv` twice, we can change it to:

```suggestion
	if os.Getenv(cmd.EnvStackID) != "" {
		return os.Getenv(cmd.EnvStackID), nil
	}
	if stackMD.BuildImage.StackID != "" {
		return stackMD.BuildImage.StackID, nil
	}
	return  "", errors.New("CNB_STACK_ID is required when there is no stack metadata available")

```

stack_validation.go
```diff
@@ -0,0 +1,44 @@
...
+		return err
+	}
+
+	runStackID, err := runImage.Label(platform.StackIDLabel)
+	if err != nil {
+		return errors.Wrap(err, "get run image label")
+	}
+	if runStackID == "" {
+		return errors.New("get run image label: io.buildpacks.stack.id")
+	}
```
yaelharel, line 26: I would put these lines under a `getRunStack` function similar to what you did for the `buildStackID`.

stack_validation_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/buildpacks/lifecycle/platform"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+)
+
+func TestStackValidation(t *testing.T) {
+	spec.Run(t, "StackValidation", testStackValidation, spec.Report(report.Terminal{}))
+}
+
+func testStackValidation(t *testing.T, when spec.G, it spec.S) {
+	when(".ValidateStack", func() {
```
yaelharel, line 21:
```suggestion
	when("ValidateStack", func() {
```

yaelharel: I saw that we're adding this `.` sometimes (for example in `utils_test.go`), but I'm not sure why we're doing it...

utils_test.go
```diff
@@ -145,4 +146,38 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
 			}
 		})
 	})
+
+	when(".ResolveRunImage", func() {
+		when("when there are no mirrors", func() {
```
yaelharel, line 151:
```suggestion
		when("there are no mirrors", func() {
```

utils_test.go
```diff
@@ -145,4 +146,38 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
...
+			it("should return run-image", func() {
+				md := platform.StackMetadata{RunImage: platform.StackRunImageMetadata{Image: "company/run:focal"}}
+				dstImage := "someregistry/whatever"
+				res, err := lifecycle.ResolveRunImage(md, dstImage)
+				h.AssertNil(t, err)
+				h.AssertEq(t, res, md.RunImage.Image)
+			})
+		})
+
+		when("when there are mirrors", func() {
```
yaelharel, line 161:
```suggestion
		when("there are mirrors", func() {
```

utils_test.go
```diff
@@ -145,4 +146,38 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
...
+					Mirrors: []string{"some.registry/_/run:focal"},
+				}}
+				dstImage := "some.registry/app/web"
+				res, err := lifecycle.ResolveRunImage(md, dstImage)
+				h.AssertNil(t, err)
+				h.AssertEq(t, res, "some.registry/_/run:focal")
+			})
+		})
+
+		when("when there is no run image defined", func() {
```
yaelharel, line 174:
```suggestion
		when("there is no run image defined", func() {
```

utils_test.go
```diff
@@ -145,4 +146,38 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
 			}
 		})
 	})
+
+	when(".ResolveRunImage", func() {
```
yaelharel, line 150: Please see https://github.com/buildpacks/lifecycle/pull/617/files#r645779179.

acceptance/analyzer_test.go
```diff
@@ -249,6 +277,8 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("analyzed path is provided", func() {
 			it("writes analyzed.toml at the provided path", func() {
+				h.SkipIf(t, runtime.GOOS == "windows", "Not yet working on Windows")
```
yaelharel, line 280: What do we need in order to make it run on Windows?

natalieparellano: Hmm... I recall there was some issue, but I want to say this could be removed now.

stack_validation_test.go
```diff
@@ -0,0 +1,82 @@
...
+			it.Before(func() {
+				os.Setenv("CNB_STACK_ID", "my-stack")
+			})
+
+			it.After(func() {
+				h.AssertNil(t, os.Unsetenv("CNB_STACK_ID"))
+			})
+
+			it("should use that value", func() {
+				md := platform.StackMetadata{BuildImage: platform.StackBuildImageMetadata{StackID: ""}}
```
natalieparellano, line 63:
```suggestion
				md := platform.StackMetadata{BuildImage: platform.StackBuildImageMetadata{StackID: "my-other-stack"}}
```

stack_validation_test.go
```diff
@@ -0,0 +1,82 @@
...
+		when("CNB_STACK_ID is present", func() {
+			it.Before(func() {
+				os.Setenv("CNB_STACK_ID", "my-stack")
+			})
+
+			it.After(func() {
+				h.AssertNil(t, os.Unsetenv("CNB_STACK_ID"))
+			})
+
+			it("should use that value", func() {
```
natalieparellano, line 62:
```suggestion
			it("prefers that value", func() {
```

utils_test.go
```diff
@@ -145,4 +146,38 @@ func testUtils(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertEq(t, res, "some.registry/_/run:focal")
+			})
+		})
+
+		when("when there is no run image defined", func() {
+			it("should fail", func() {
+				md := platform.StackMetadata{}
+				dstImage := "someregistry/whatever"
+				_, err := lifecycle.ResolveRunImage(md, dstImage)
+				h.AssertError(t, err, "CNB_RUN_IMAGE is required when there is no stack metadata available")
```
natalieparellano, line 179: Could we add a test that uses `CNB_RUN_IMAGE`?

acceptance/analyzer_test.go
```diff
@@ -73,6 +73,33 @@ func TestAnalyzer(t *testing.T) {
...
+		"--build-arg", "stackid=io.buildpacks.stacks.bionic",
+	)
+	buildAuthRegistryImage(
+		t,
+		"company/stack:centos",
+		runImageContext,
+		"-f", filepath.Join(runImageContext, dockerfileName),
+		"--build-arg", "stackid=io.company.centos",
+	)
+
```
natalieparellano, line 92:
```suggestion

// build run image into daemon
```

natalieparellano: Are we cleaning up this image?

acceptance/analyzer_test.go
```diff
@@ -643,6 +678,8 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("registry case", func() {
 			it("writes analyzed.toml", func() {
+				h.SkipIf(t, runtime.GOOS == "windows", "Not yet working on Windows")
```
natalieparellano, line 681: Could this be removed?

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1053,151 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+
+		when("validating stack", func() {
+			it.Before(func() {
+				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not validate stack")
+			})
+
+			when("stack metadata is present", func() {
+				when("stacks match", func() {
+					it("passes validation", func() {
+						h.SkipIf(t, runtime.GOOS == "windows", "Not yet working on Windows")
```
natalieparellano, line 1065: Remove?

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1053,151 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--env", "CNB_STACK_PATH=/cnb/mismatch-stack.toml",
+								"--env", "CNB_RUN_IMAGE="+noAuthRegistry.RepoName("company/stack:bionic"),
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("stack metadata does not contain run-image.image", func() {
```
natalieparellano, line 1103: Is this one covered by units now?

acceptance/testdata/analyzer/analyze-image/Dockerfile
```diff
@@ -19,3 +19,11 @@ RUN chown -R $CNB_USER_ID:$CNB_GROUP_ID /layers
 
 # ensure docker config directory is root owned and NOT world readable
 RUN chown -R root /docker-config; chmod -R 700 /docker-config
+
+ARG registry
```
natalieparellano, line 23: Do we need to update the corresponding Dockerfile.windows now?

cmd/lifecycle/analyzer.go
natalieparellano: I think we fall back to `CNB_STACK_ID`.

natalieparellano: What about making a `lifecycle.ResolveStackID` function?

utils_test.go
jabrown85: Actually, no. The `FlagRunImage` is what ends up reading from `CNB_RUN_IMAGE ` to set `runImage`. This message maybe should be updated since it comes from this function that doesn't deal with it, but this helper doesn't deal with resolving the env var override. See `analyzer.getRunImage`.

acceptance/testdata/analyzer/analyze-image/Dockerfile
jabrown85: Probably. I need to figure out how to re-write it for windows, but I uncommented all the tests that were skipping windows so I can see them fail and fix it :)

utils_test.go
natalieparellano: This makes sense. In that case I think we should change the error message (to avoid confusion). 

It would be nice to advise the user how they might supply the run image when it's missing from the stack metadata (perhaps `analyzer.getRunImage` could do this) - at the same time, I wonder if `CNB_RUN_IMAGE is required` is too specific - because they could also pass `-run-image`, no? It seems like we'd want something like `run image not provided via argument, environment, or stack metadata` but that reads a little weirdly to me, not sure how to improve it...

acceptance/testdata/analyzer/analyze-image/Dockerfile
jabrown85: Alright - I finally fixed up windows acceptance tests.

acceptance/analyzer_test.go
jabrown85: b5665e3

acceptance/analyzer_test.go
yaelharel: This conversation has been marked as resolved although there was no comment or the code wasn't removed.
Did I miss something?

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1053,172 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+
+						h.AssertNotNil(t, err)
+						expected := "failed to resolve run image"
+						h.AssertStringContains(t, string(output), expected)
+					})
+				})
+
+				when("run image has mirrors", func() {
+					it("uses expected mirror for run-image", func() {
+						h.SkipIf(t, runtime.GOOS == "windows", "Not yet working on Windows")
```
yaelharel, line 1139: Could this be removed?

cmd/lifecycle/analyzer.go
yaelharel: Regarding what happens if the file doesn't exist:
I think that if the file doesn't exist, `stackMD` will be empty and then in `getRunImage`, if `a.runImageRef == ""`, `ResolveRunImage` will return an error.
Maybe it's worth adding a comment about it?

acceptance/analyzer_test.go
jabrown85: Ah, i thought I got them all. Let's see if i can remove it 😬

utils_test.go
yaelharel: If we deleted this `.` here, I think it's worth removing the `.` in the other `when` in this file.
* ReadOrder
* ReadGroup
* WriteTOML
* TruncateSha

acceptance/analyzer_test.go
jabrown85: Removed. we'll see if it passes without more work in CI

acceptance/analyzer_test.go
yaelharel: I think that there are 5 tests that we're currently skipping on "windows". Why? Can we make them work on "windows"?

yaelharel: Now I see that next to 3 of them we have a comment that they are "not relevant on Windows".
Can we expand these comments so they'll be clearer?

utils_test.go
jabrown85: Do we want to remove `.`? I did it because it was requested, but I also am fine with leaving them and matching the style of the test that already exist. That is why I used the `.` in the first place...

acceptance/analyzer_test.go
jabrown85: All skips have been removed that I see and they all pass 🎉

utils_test.go
yaelharel: I'm not sure. I think this is the only file that is using `.` and I'm not sure why.
Is there a reason for it?

acceptance/analyzer_test.go
yaelharel: I know that it's not specifically related to your PR, but since you're already touching this file, do you mind rewriting the comments about those tests that are "Not relevant on Windows"?
We can totally do it in another PR if you prefer!

utils_test.go
jabrown85: 🤷  - I've seen this pattern before for denoting public method tests. `#Method` or `.Method`

I am happy to remove them all or match the style of the file

acceptance/analyzer_test.go
jabrown85: I think we should do this in another PR

utils_test.go
yaelharel: But we're usually testing public methods and I have never seen this style anywhere else, but maybe I'm wrong and I just missed it.
Anyway, my first thought was about removing it but I don't have a strong opinion about it.

acceptance/analyzer_test.go
yaelharel: Sounds good. Thanks!

stack_validation.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/buildpacks/imgutil"
+	"github.com/pkg/errors"
+
+	"github.com/buildpacks/lifecycle/cmd"
+	"github.com/buildpacks/lifecycle/platform"
+)
+
+func ValidateStack(stackMD platform.StackMetadata, runImage imgutil.Image) error {
+	buildStackID, err := getBuildStack(stackMD)
```
natalieparellano, line 15: This is sooo much clearer now, thank you! :)

stack_validation_test.go
```diff
@@ -0,0 +1,82 @@
...
+		})
+
+		when("no build stack is present", func() {
+			it("should fail", func() {
+				md := platform.StackMetadata{BuildImage: platform.StackBuildImageMetadata{StackID: ""}}
+				runImage := fakes.NewImage("runimg", "", nil)
+				h.AssertNil(t, runImage.SetLabel(platform.StackIDLabel, "my-stack"))
+				err := lifecycle.ValidateStack(md, runImage)
+				h.AssertNotNil(t, err)
+				h.AssertError(t, err, "CNB_STACK_ID is required when there is no stack metadata available")
```
natalieparellano, line 78: It's interesting - the spec is kind of unclear if `CNB_STACK_ID` is always required (regardless of other data) - 

https://github.com/buildpacks/spec/blob/main/platform.md#stack-provided-variables - "SHOULD be set"
https://github.com/buildpacks/spec/blob/main/platform.md#build-image - "platforms MUST ensure"

cmd/lifecycle/analyzer.go
natalieparellano: If the file doesn't exist, it's still possible that `a.runImageRef` is non-empty (specified via command flag or env var).

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				assertAnalyzedMetadata(t, filepath.Join(copyDir, "some-other-layers", "analyzed.toml")) // analyzed.toml is written at the provided -layers directory: /some-other-layers
 			})
 		})
+
+		when("validating stack", func() {
+			it.Before(func() {
+				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not validate stack")
```
natalieparellano, line 1057: This is interesting - technically earlier platform specs did claim to validate the stack id, it just wasn't done in practice. But I guess in practice we're also trying not to change behavior for previous platform apis...

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+			})
+
+			when("stack metadata is present", func() {
+				when("stacks match", func() {
+					it("passes validation", func() {
+						execArgs := []string{ctrPath(analyzerPath)}
+
+						h.DockerRunAndCopy(t,
```
natalieparellano, line 1065: We could just use `h.DockerRun` here, since we don't care about copying things from the container.

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+					})
+				})
+
+				when("stack metadata file is invalid", func() {
+					it("fails validation", func() {
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--network", registryNetwork,
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--env", "CNB_STACK_PATH=/cnb/bad-stack.toml",
```
natalieparellano, line 1105:
```suggestion
							"--env", "CNB_STACK_PATH=/cnb/file-does-not-exist.toml",
```

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+
+						h.AssertNotNil(t, err)
+						expected := "get stack metadata"
+						h.AssertStringContains(t, string(output), expected)
+					})
+				})
+
+				when("run image inaccessible", func() {
+					it("fails validation", func() {
+						cmd := exec.Command(
```
natalieparellano, line 1119: See https://github.com/buildpacks/lifecycle/pull/617/files?file-filters%5B%5D=.go&file-filters%5B%5D=.toml&file-filters%5B%5D=No+extension#r649540215

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+
+						h.AssertNotNil(t, err)
+						expected := "failed to resolve run image"
+						h.AssertStringContains(t, string(output), expected)
+					})
+				})
+
+				when("run image has mirrors", func() {
+					it("uses expected mirror for run-image", func() {
+						execArgs := []string{ctrPath(analyzerPath), "--previous-image=" + noAuthRegistry.RepoName("apprepo/myapp")}
```
natalieparellano, line 1137: Why is this needed?

natalieparellano, line 1137:
```suggestion
						execArgs := []string{ctrPath(analyzerPath), "--previous-image=" + noAuthRegistry.RepoName("apprepo/myapp")} // previous image located on same registry as mirror in /cnb/run-mirror-stack.toml
```

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+
+						h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							ctrPath("/layers/analyzed.toml"),
+							analyzeImage,
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--env", "CNB_STACK_PATH=/cnb/run-mirror-stack.toml",
```
natalieparellano, line 1147:
```suggestion
								"--env", "CNB_STACK_PATH=/cnb/run-mirror-stack.toml", // /cnb/run-mirror-stack.toml points to run image on gcr.io and mirror on test registry
```

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+							h.DockerRunAndCopy(t,
+								containerName,
+								copyDir,
+								ctrPath("/layers/analyzed.toml"),
+								analyzeImage,
+								h.WithFlags(append(
+									dockerSocketMount,
+									"--network", registryNetwork,
+									"--env", "CNB_PLATFORM_API="+platformAPI,
+									"--env", "CNB_STACK_PATH=/cnb/local-bionic-stack.toml",
```
natalieparellano, line 1168: Does this file exist? How are the stacks matching?

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+
+						h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							ctrPath("/layers/analyzed.toml"),
+							analyzeImage,
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--env", "CNB_STACK_PATH=/cnb/missing.toml",
```
natalieparellano, line 1190:
```suggestion
								"--env", "CNB_STACK_PATH=/cnb/file-does-not-exist.toml",
```

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+								)...),
+								h.WithArgs(execArgs...),
+							)
+						})
+					})
+				})
+			})
+
+			when("stack metadata is not present", func() {
+				when("run-image and stack-id are present", func() {
```
natalieparellano, line 1178:
```suggestion
				when("run image and stack id are present", func() {
```

natalieparellano: (the dashes made me think of a flag or something)

acceptance/testdata/analyzer/analyze-image/container/cnb/bad-stack.toml
```diff
@@ -0,0 +1 @@
+[run-images
```
natalieparellano, line 1: Are we using this file?

acceptance/testdata/analyzer/analyze-image/container/cnb/local-bionic-stack.toml
```diff
@@ -0,0 +1,6 @@
+[run-image]
+ image = "localcompany/stack:bionic"
+ mirrors = []
+[build-image]
+ stack-id = "io.buildpacks.stacks.bionic"
+ mixins = []
```
natalieparellano, line 6: Would it make sense to move this and /cnb/missing-run-image.toml to the Dockerfile, so that all the stack tomls are defined in one place? (Having it defined in two places caused me a bit of confusion)

acceptance/testdata/analyzer/analyze-image/container/cnb/missing-run-image.toml
```diff
@@ -0,0 +1,3 @@
+[build-image]
```
natalieparellano, line 1: Are we using this file?

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+								"--env", "CNB_STACK_PATH=/cnb/missing.toml",
+								"--env", "CNB_RUN_IMAGE="+noAuthRegistry.RepoName("company/stack:bionic"),
+								"--env", "CNB_STACK_ID=io.buildpacks.stacks.bionic",
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("run-image and stack-id are not present", func() {
```
natalieparellano, line 1199:
```suggestion
				when("run image and stack id are not provided as arguments or in the environment", func() {
```

acceptance/analyzer_test.go
natalieparellano, line 1168:
```suggestion
									"--env", "CNB_STACK_PATH=/cnb/local-bionic-stack.toml", // /cnb/local-bionic-stack.toml has `io.buildpacks.stacks.bionic` and points to run image `localcompany/stack:bionic` with same stack id
```

acceptance/testdata/analyzer/analyze-image/Dockerfile
```diff
@@ -19,3 +19,39 @@ RUN chown -R $CNB_USER_ID:$CNB_GROUP_ID /layers
...
+ stack-id = \"io.buildpacks.stacks.bionic\"\n\
+ mixins = []\n\
+" > /cnb/mismatch-stack.toml
+
+RUN echo "\
+[run-image]\n\
+ image = \"${registry}/company/stack:bionic\"\n\
+ mirrors = []\n\
+" > /cnb/missing-build-image-stack.toml
+
```
natalieparellano, line 49:
```suggestion
```
(don't think it's used anywhere)

acceptance/analyzer_test.go
natalieparellano: There's a case to be made I think for also testing that we're correctly passing the `-run-image` flag.

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+					it("fails validation", func() {
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--network", registryNetwork,
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--env", "CNB_STACK_PATH=/cnb/missing.toml",
```
natalieparellano, line 1205:
```suggestion
							"--env", "CNB_STACK_PATH=/cnb/file-does-not-exist.toml",
```

natalieparellano:
This is looking great! The acceptance tests require a certain amount of concentration and navigating around to follow, so I added some suggestions for comments that might make it easier to understand.

cmd/lifecycle/analyzer.go
yaelharel: Yeah, I know that it's possible but if `a.runImageRef` is also empty, then we might get an error.
Anyway, I think that maybe it's worth adding a comment about it.

acceptance/testdata/analyzer/analyze-image/container/cnb/bad-stack.toml
yaelharel: acceptance/analyzer_test.go:1105

acceptance/testdata/analyzer/analyze-image/container/cnb/missing-run-image.toml
yaelharel: I don't think so.

acceptance/testdata/analyzer/analyze-image/container/cnb/bad-stack.toml
natalieparellano: @yaelharel thanks! @jabrown85 could we also move this to the Dockerfile?

cmd/lifecycle/analyzer.go
natalieparellano: I'm not sure it's necessary, but I wouldn't be opposed to adding a comment. @jabrown85 WDYT?

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+								"--env", "CNB_RUN_IMAGE="+noAuthRegistry.RepoName("company/stack:bionic"),
+							),
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("stack metadata file is invalid", func() {
+					it("fails validation", func() {
+						cmd := exec.Command(
```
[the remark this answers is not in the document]
jabrown85: It doesn't appear to work without modification. The `DockerRun` doesn't do `CombinedOutput` and the output is empty.

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/testdata/analyzer/analyze-image/Dockerfile
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/testdata/analyzer/analyze-image/container/cnb/missing-run-image.toml
jabrown85: 11a5a76

acceptance/testdata/analyzer/analyze-image/container/cnb/local-bionic-stack.toml
jabrown85: 11a5a76

acceptance/testdata/analyzer/analyze-image/container/cnb/bad-stack.toml
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: See my other comment, the combined output seems to make this not possible. I could be missing something though.

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+
+						h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							ctrPath("/layers/analyzed.toml"),
+							analyzeImage,
+							h.WithFlags(
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--env", "CNB_STACK_PATH=/cnb/mismatch-stack.toml",
```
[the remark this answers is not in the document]
jabrown85: 11a5a76

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+							h.WithArgs(execArgs...),
+						)
+					})
+				})
+
+				when("CNB_RUN_IMAGE is present", func() {
+					it("uses CNB_RUN_IMAGE for validation", func() {
+						execArgs := []string{ctrPath(analyzerPath)}
+
+						h.DockerRunAndCopy(t,
```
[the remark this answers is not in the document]
jabrown85: 11a5a76

acceptance/analyzer_test.go
```diff
@@ -1015,6 +1051,170 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+			when("stack metadata is present", func() {
+				when("stacks match", func() {
+					it("passes validation", func() {
+						execArgs := []string{ctrPath(analyzerPath)}
+
+						h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							ctrPath("/layers/analyzed.toml"),
+							analyzeImage,
```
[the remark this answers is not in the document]
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: 11a5a76

acceptance/analyzer_test.go
jabrown85: c8d41f0

cmd/lifecycle/analyzer.go
jabrown85: I'm open to a suggestion with the comment we think is appropriate. I am far less comment driven than test driven - I find comments fall out of date rather quickly unless it's describing a public API.

jabrown85:
@natalieparellano I'd like to squash everything so I can more easily fix the conflict on `main`. Let me know when you've had a chance to review the later commits. I don't want to take away your ability to review easily

acceptance/analyzer_test.go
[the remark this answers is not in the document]
natalieparellano: Ohh weird... nevermind then! We could probably improve upon `DockerRun` at some point but not required here.

utils_test.go
natalieparellano: Just curious, when do we use `.` vs `#`? (I've seen "object method" vs "static method" in other languages but I'm not sure what is our convention.) That said I don't have an opinion here.

cmd/lifecycle/analyzer.go
natalieparellano: Let's leave it for now, we could add a comment if we find that it's necessary.

natalieparellano:
> @natalieparellano I'd like to squash everything so I can more easily fix the conflict on `main`. Let me know when you've had a chance to review the later commits. I don't want to take away your ability to review easily

@jabrown85 go for it!

natalieparellano (dismissed):

natalieparellano (commented):
Do we need to add logic for the creator?

jabrown85:
@natalieparellano creator should now be good to go after cc21037 and 55c17a5

utils.go
```diff
@@ -64,6 +66,30 @@ func DecodeLabel(image imgutil.Image, label string, v interface{}) error {
 	return nil
 }
 
+func ResolveRunImage(stackMD platform.StackMetadata, destinationImageRef string) (string, error) {
+	runImageRef := stackMD.RunImage.Image
```
natalieparellano, line 70: Is this in danger of throwing a `nil` panic?

jabrown85: I don't think it is possible to panic here. `stackMD` is not a pointer and the definition [here](https://github.com/buildpacks/lifecycle/blob/a7428a55c2a14d8a37e84285b95dc63192e3264e/platform/files.go#L208) has `Runimage` as `StackRunImageMetadata` - also not a pointer. It should always be `""` in the worst case.

natalieparellano: Answer - no, because of https://github.com/buildpacks/lifecycle/blob/162090f6255b289c36fcfac8a884a8e8c945c82e/platform/files.go#L208 (no omit empty)

natalieparellano (approved):

PR #621, opened 2021-05-19
Re-Align analyzer with spec and creator

jabrown85:
See [here](https://github.com/buildpacks/spec/pull/225#pullrequestreview-663490720)

We should still be taking in `<image>` to represent the write location and optionally the `<previous-image>` if no explicitly defined. This re-aligns with the spec and `creator` behavior.

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -113,7 +113,6 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("called without an app image", func() {
 			it("errors", func() {
-				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0, "Platform API >= 0.7 does not accept an image argument")
 				cmd := exec.Command(
 					"docker", "run", "--rm",
 					"--env", "CNB_PLATFORM_API="+platformAPI,
@@ -206,11 +205,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			it("recursively chowns the directory", func() {
 				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
 
-				imgArg := ""
-				// Platform >= 0.7 does not allow an argument, < 7 requires an argument
-				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
-					imgArg = "some-image"
-				}
+				imgArg := "some-image"
 
 				output := h.DockerRun(t,
 					analyzeImage,
@@ -252,16 +247,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				execArgs := []string{
 					ctrPath(analyzerPath),
 					"-analyzed", ctrPath("/some-dir/some-analyzed.toml"),
-					"-previous-image", "some-image",
-				}
-
-				// Platform >= 0.7 does not allow an argument, < 7 requires an argument
-				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
-					execArgs = []string{
-						ctrPath(analyzerPath),
-						"-analyzed", ctrPath("/some-dir/some-analyzed.toml"),
-						"some-image",
-					}
+					"some-image",
 				}
 
 				h.DockerRunAndCopy(t,
@@ -282,16 +268,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				execArgs := []string{
 					ctrPath(analyzerPath),
 					"-daemon",
-					"-previous-image", "some-image",
-				}
-
-				// Platform >= 0.7 does not allow an argument, < 7 requires an argument
-				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
-					execArgs = []string{
-						ctrPath(analyzerPath),
-						"-daemon",
-						"some-image",
-					}
+					"some-image",
 				}
 
 				h.DockerRunAndCopy(t,
@@ -345,7 +322,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						h.WithArgs(
 							ctrPath(analyzerPath),
 							"-daemon",
-							"-previous-image", appImage),
+							appImage),
 					)
 
 					assertNoRestoreOfAppMetadata(t, copyDir, output)
@@ -643,11 +620,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("registry case", func() {
 			it("writes analyzed.toml", func() {
-				execArgs := []string{ctrPath(analyzerPath)}
-				// Platform >= 0.7 does not allow an argument, < 7 requires an argument
-				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
-					execArgs = append(execArgs, "some-image")
-				}
+				execArgs := []string{ctrPath(analyzerPath), "some-image"}
 
 				h.DockerRunAndCopy(t,
 					containerName,
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -789,6 +789,45 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				})
 			})
 
+			when("called with previous image", func() {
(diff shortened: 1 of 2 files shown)


acceptance/analyzer_test.go
```diff
@@ -206,11 +205,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			it("recursively chowns the directory", func() {
 				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
 
-				imgArg := ""
-				// Platform >= 0.7 does not allow an argument, < 7 requires an argument
-				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
-					imgArg = "some-image"
-				}
+				imgArg := "some-image"
```
natalieparellano, line 208: Could we just inline this?

acceptance/analyzer_test.go
```diff
@@ -643,11 +620,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("registry case", func() {
 			it("writes analyzed.toml", func() {
-				execArgs := []string{ctrPath(analyzerPath)}
-				// Platform >= 0.7 does not allow an argument, < 7 requires an argument
-				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
-					execArgs = append(execArgs, "some-image")
-				}
+				execArgs := []string{ctrPath(analyzerPath), "some-image"}
```
natalieparellano, line 623: Could we inline this as well?

acceptance/analyzer_test.go
natalieparellano: Could we change the occurrences of `"-previous-image", "some-image"` in this file to just `"some-image"`, so that it's more realistic?

Also, the only tests left in this file that use `-previous-image` are ones that expect an error. Is there a way to show that we correctly process this flag when the previous image and the destination image aren't the same?

cmd/lifecycle/analyzer.go
```diff
@@ -214,7 +215,8 @@ func (a *analyzeCmd) registryImages() []string {
 		registryImages = append(registryImages, a.platform06.cacheImageTag)
 	}
 	if !a.useDaemon {
-		registryImages = append(registryImages, a.analyzeArgs.imageName)
+		registryImages = append(registryImages, append([]string{a.imageName}, a.additionalTags...)...)
+		registryImages = append(registryImages, a.runImageRef, a.previousImage)
```
natalieparellano, line 219: Is it possible to combine this into one big `append`, so that it's more readable?

jabrown85: Fixed in 942ee2c

acceptance/analyzer_test.go
jabrown85: Fixed in 942ee2c

acceptance/analyzer_test.go
jabrown85: Fixed in 942ee2c

natalieparellano: Ahh the tests are easier to read now! I noticed that we're still not using `-previous-image` anywhere - could we just add a new context like `when("called with previous image"` that uses `some-image` as the previous image and tries to write to `some-new-image` (or something). That way we can be sure that we're correctly parsing this argument.

cmd/lifecycle/analyzer.go
```diff
@@ -175,18 +176,18 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 		img imgutil.Image
 		err error
 	)
-	if aa.imageName != "" {
+	if aa.previousImage != "" {
```
natalieparellano, line 179: I think we could remove this guard as `aa.previousImage` will never be blank (because `a.imageName` will never be blank). We didn't have this guard in 0.11.x.

acceptance/analyzer_test.go
jabrown85: See 0e2047a!

cmd/lifecycle/analyzer.go
jabrown85: Good catch! See 33f945a

acceptance/analyzer_test.go
```diff
@@ -820,6 +789,51 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+						authRegAppImage, appAuthConfig = buildAuthRegistryImage(
+							t,
+							"some-app-image-"+h.RandString(10),
+							filepath.Join("testdata", "analyzer", "app-image"),
+							"--build-arg", "fromImage="+containerBaseImage,
+							"--build-arg", "metadata="+metadata,
+						)
+					})
+
+					it("writes analyzed.toml with previous image identifier", func() {
```
yaelharel, line 811: Are there any other tests that we would like to add that relates to the previous image?
Maybe if it's not accessible / doesn't exists?

jabrown85: The tests we have in place for `image` should cover that, I believe. Since `previousImage = image`

yaelharel: Yeah, I thought about it but since we always passing `image`, I was wondering if we can test it directly, mostly in case something will change in the future.

natalieparellano: I think this is covered by the new acceptance test that Jesse added.

acceptance/analyzer_test.go
```diff
@@ -820,6 +789,51 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+						execArgs := []string{
+							ctrPath(analyzerPath),
+							"-previous-image", authRegAppImage,
+							"some-fake-image",
+						}
```
natalieparellano, line 816: As an aside, I love this. It makes it so much easier to understand how we're exec'ing the analyzer (what we care most about) instead of having to read all of the call to `DockerRunAndCopy`.

acceptance/analyzer_test.go
```diff
@@ -1054,6 +1068,17 @@ func assertAnalyzedMetadata(t *testing.T, path string) {
 	h.AssertNil(t, err)
 }
 
+func getAnalyzedMetadataImageRef(t *testing.T, path string) *platform.AnalyzedMetadata {
```
natalieparellano, line 1071:
```suggestion
func getAnalyzedMetadata(t *testing.T, path string) *platform.AnalyzedMetadata {
```

natalieparellano: (since it returns the whole metadata)

But maybe we could modify `assertAnalyzedMetadata` (which does essentially the same thing) to return the metadata struct.

acceptance/analyzer_test.go
natalieparellano: I see what you're saying - internally we know that we assign `-previous-image` to `Analyzer.Image` and so they should be treated equivalently, but it's not obvious from the tests. However since we're passing `"some-fake-image"` as the positional argument and yet getting metadata for the previous image I think we're getting pretty good coverage from this test.

natalieparellano (approved):
One nit (see https://github.com/buildpacks/lifecycle/pull/621#discussion_r644326925) but overall this is looking great!

acceptance/analyzer_test.go
jabrown85: Fixed in e1f3c71

acceptance/analyzer_test.go
yaelharel: And what about passing two "real" images? One for the `image` and one for the `previousImage`?

natalieparellano: IDK, not sure if there is significant value and it would create more overhead. @jabrown85 WDYT?

jabrown85: I'm fine leaving it as-is, but will add more tests if you two think it is valuable enough!

yaelharel: Personally I think it's better to create another test so our coverage will be better for future changes, but if you both think it's unnecessary, we can keep it as is for now.

natalieparellano:
@yaelharel did you want to review again before we merge?

yaelharel (approved):
The only question that left for me is: https://github.com/buildpacks/lifecycle/pull/621#discussion_r645824176.
As long as both of you think it's unnecessary, I'm ok with merging this PR 😊.

acceptance/analyzer_test.go
natalieparellano: 🤷🏼‍♀️

jabrown85: Added new test in 251ba00.

PR #624, opened 2021-05-28
Don't write sha files when platform API >= 0.7

### LINKED ISSUE #584: [Restore] Layer metadata restorer should not write layer.sha files for platform API >= 0.7
### Description
For earlier platform API, the analyzer would write `<layers>/<buidpack-id>/<layer>.sha` files containing the diffID of each analyzed layer. These files are used by the restorer to verify that the layer metadata written by the analyzer was for the same image / layer that the restorer has access to.

Now that metadata and cache restoration are happening in the same phase in platform API >= 0.7, these files are no longer needed. 

### Proposed solution
Layer metadata restorer should not write layer.sha files for platform API >= 0.7

This could be folded into https://github.com/buildpacks/lifecycle/pull/582, but we'd like to make this a separate unit of work in order to unblock other issues that rely on https://github.com/buildpacks/lifecycle/pull/582.

### Additional context
Spec PR: https://github.com/buildpacks/spec/pull/197

yaelharel:
Fixes #584

### DIFF
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -194,13 +194,14 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 		}
 	}
 
+	useShaFile := true
 	analyzedMD, err := (&lifecycle.Analyzer{
 		Buildpacks:            aa.platform06.group.Group,
 		Cache:                 aa.platform06.cache,
 		Logger:                cmd.DefaultLogger,
 		Platform:              aa.platform,
 		Image:                 img,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers, true),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers, useShaFile),
 	}).Analyze()
 	if err != nil {
 		return platform.AnalyzedMetadata{}, cmd.FailErrCode(err, aa.platform.CodeFor(cmd.AnalyzeError), "analyzer")
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -200,7 +200,7 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 		Logger:                cmd.DefaultLogger,
 		Platform:              aa.platform,
 		Image:                 img,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers, true),
 	}).Analyze()
 	if err != nil {
 		return platform.AnalyzedMetadata{}, cmd.FailErrCode(err, aa.platform.CodeFor(cmd.AnalyzeError), "analyzer")
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -194,14 +194,13 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 		}
 	}
 
-	useShaFile := true
 	analyzedMD, err := (&lifecycle.Analyzer{
 		Buildpacks:            aa.platform06.group.Group,
 		Cache:                 aa.platform06.cache,
 		Logger:                cmd.DefaultLogger,
 		Platform:              aa.platform,
 		Image:                 img,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers, useShaFile),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers),
 	}).Analyze()
 	if err != nil {
 		return platform.AnalyzedMetadata{}, cmd.FailErrCode(err, aa.platform.CodeFor(cmd.AnalyzeError), "analyzer")
--- a/cmd/lifecycle/restorer.go
+++ b/cmd/lifecycle/restorer.go
@@ -116,13 +116,12 @@ func (r *restoreCmd) registryImages() []string {
 }
 
 func (r restoreArgs) restore(layerMetadata platform.LayersMetadata, group buildpack.Group, cacheStore lifecycle.Cache) error {
-	useShaFile := !r.restoresLayerMetadata()
 	restorer := &lifecycle.Restorer{
 		LayersDir:             r.layersDir,
 		Buildpacks:            group.Group,
 		Logger:                cmd.DefaultLogger,
 		Platform:              r.platform,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers, useShaFile),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers),
 		LayersMetadata:        layerMetadata,
 	}
 
--- a/cmd/lifecycle/restorer.go
+++ b/cmd/lifecycle/restorer.go
@@ -121,7 +121,7 @@ func (r restoreArgs) restore(layerMetadata platform.LayersMetadata, group buildp
 		Buildpacks:            group.Group,
 		Logger:                cmd.DefaultLogger,
 		Platform:              r.platform,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers, false),
 		LayersMetadata:        layerMetadata,
 	}
 
--- a/cmd/lifecycle/restorer.go
+++ b/cmd/lifecycle/restorer.go
@@ -121,7 +121,7 @@ func (r restoreArgs) restore(layerMetadata platform.LayersMetadata, group buildp
 		Buildpacks:            group.Group,
 		Logger:                cmd.DefaultLogger,
 		Platform:              r.platform,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers, false),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers, !r.restoresLayerMetadata()),
 		LayersMetadata:        layerMetadata,
 	}
 
--- a/cmd/lifecycle/restorer.go
+++ b/cmd/lifecycle/restorer.go
@@ -116,12 +116,13 @@ func (r *restoreCmd) registryImages() []string {
 }
 
 func (r restoreArgs) restore(layerMetadata platform.LayersMetadata, group buildpack.Group, cacheStore lifecycle.Cache) error {
+	useShaFile := !r.restoresLayerMetadata()
 	restorer := &lifecycle.Restorer{
 		LayersDir:             r.layersDir,
 		Buildpacks:            group.Group,
 		Logger:                cmd.DefaultLogger,
 		Platform:              r.platform,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers, !r.restoresLayerMetadata()),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers, useShaFile),
 		LayersMetadata:        layerMetadata,
 	}
 
--- a/layer_metadata_restorer.go
+++ b/layer_metadata_restorer.go
@@ -62,7 +62,7 @@ func (lss *defaultLayerSHAStore) get(buildpackID string, layer bpLayer) (string,
 		}
 		return data.SHA, nil
 	}
-	return lss.getShaByBuildpackLayers(buildpackID, layer.name()), nil
+	return lss.getShaByBuildpackLayer(buildpackID, layer.name()), nil
 }
 
 func (lss *defaultLayerSHAStore) addLayerToMap(buildpackID, layerName, sha string) {
@@ -74,7 +74,7 @@ func (lss *defaultLayerSHAStore) addLayerToMap(buildpackID, layerName, sha strin
 }
 
 // if the layer exists for the buildpack ID, its SHA will be returned. Otherwise, an empty string will be returned.
-func (lss *defaultLayerSHAStore) getShaByBuildpackLayers(buildpackID, layerName string) string {
+func (lss *defaultLayerSHAStore) getShaByBuildpackLayer(buildpackID, layerName string) string {
 	if layerToSha, buildpackExists := lss.buildpacksToLayersShaMap[buildpackID]; buildpackExists {
 		if sha, layerExists := layerToSha.layerToShaMap[layerName]; layerExists {
 			return sha
--- a/layer_metadata_restorer.go
+++ b/layer_metadata_restorer.go
@@ -81,8 +81,8 @@ func (bls *BuildpackLayersToSha) getShaByBuildpackLayers(buildpackID, layerName 
 }
 
 func (bls *BuildpackLayersToSha) addLayerToMap(buildpackID, layerName, sha string) {
-	_, ok := bls.buildpacksToLayersShaMap[buildpackID]
-	if !ok {
+	_, exists := bls.buildpacksToLayersShaMap[buildpackID]
+	if !exists {
 		bls.buildpacksToLayersShaMap[buildpackID] = initializeLayerToSha()
 	}
 	bls.buildpacksToLayersShaMap[buildpackID].layerToShaMap[layerName] = sha
--- a/layer_metadata_restorer.go
+++ b/layer_metadata_restorer.go
@@ -13,15 +13,13 @@ import (
 
 //go:generate mockgen -package testmock -destination testmock/layer_metadata_restorer.go github.com/buildpacks/lifecycle LayerMetadataRestorer
 type LayerMetadataRestorer interface {
-	UseSHAFiles() bool
 	Restore(buildpacks []buildpack.GroupBuildpack, appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata, layerSHAStore LayerSHAStore) error
 }
 
 type DefaultLayerMetadataRestorer struct {
-	logger      Logger
-	layersDir   string
-	skipLayers  bool
-	useShaFiles bool
+	logger     Logger
+	layersDir  string
+	skipLayers bool
 }
 
 type LayerSHAStore interface {
@@ -29,8 +27,29 @@ type LayerSHAStore interface {
 	get(buildpackID string, layer bpLayer) (string, error)
 }
 
-type defaultLayerSHAStore struct {
-	useShaFiles              bool
+func NewLayerSHAStore(useShaFiles bool) LayerSHAStore {
+	if useShaFiles {
+		return &filesLayerSHAStore{}
+	}
+	return &mapLayerSHAStore{make(map[string]layerToSha)}
+}
+
+type filesLayerSHAStore struct {
+}
+
+func (flss *filesLayerSHAStore) add(buildpackID, sha string, layer *bpLayer) error {
+	return layer.writeSha(sha)
+}
+
+func (flss *filesLayerSHAStore) get(buildpackID string, layer bpLayer) (string, error) {
+	data, err := layer.read()
+	if err != nil {
+		return "", errors.Wrapf(err, "reading layer")
+	}
+	return data.SHA, nil
+}
+
+type mapLayerSHAStore struct {
 	buildpacksToLayersShaMap map[string]layerToSha
 }
 
@@ -38,44 +57,26 @@ type layerToSha struct {
 	layerToShaMap map[string]string
 }
 
-func NewLayerSHAStore(useShaFiles bool) LayerSHAStore {
-	var buildpacksToLayersShaMap map[string]layerToSha
-	if !useShaFiles {
-		buildpacksToLayersShaMap = make(map[string]layerToSha)
-	}
-	return &defaultLayerSHAStore{useShaFiles, buildpacksToLayersShaMap}
-}
-
-func (lss *defaultLayerSHAStore) add(buildpackID, sha string, layer *bpLayer) error {
-	if lss.useShaFiles {
-		return layer.writeSha(sha)
-	}
-	lss.addLayerToMap(buildpackID, layer.name(), sha)
+func (mlss *mapLayerSHAStore) add(buildpackID, sha string, layer *bpLayer) error {
+	mlss.addLayerToMap(buildpackID, layer.name(), sha)
 	return nil
 }
 
-func (lss *defaultLayerSHAStore) get(buildpackID string, layer bpLayer) (string, error) {
-	if lss.useShaFiles {
-		data, err := layer.read()
-		if err != nil {
-			return "", errors.Wrapf(err, "reading layer")
-		}
-		return data.SHA, nil
-	}
-	return lss.getShaByBuildpackLayer(buildpackID, layer.name()), nil
+func (mlss *mapLayerSHAStore) get(buildpackID string, layer bpLayer) (string, error) {
+	return mlss.getShaByBuildpackLayer(buildpackID, layer.name()), nil
 }
 
-func (lss *defaultLayerSHAStore) addLayerToMap(buildpackID, layerName, sha string) {
-	_, exists := lss.buildpacksToLayersShaMap[buildpackID]
+func (mlss *mapLayerSHAStore) addLayerToMap(buildpackID, layerName, sha string) {
+	_, exists := mlss.buildpacksToLayersShaMap[buildpackID]
 	if !exists {
-		lss.buildpacksToLayersShaMap[buildpackID] = layerToSha{make(map[string]string)}
+		mlss.buildpacksToLayersShaMap[buildpackID] = layerToSha{make(map[string]string)}
 	}
-	lss.buildpacksToLayersShaMap[buildpackID].layerToShaMap[layerName] = sha
+	mlss.buildpacksToLayersShaMap[buildpackID].layerToShaMap[layerName] = sha
 }
 
 // if the layer exists for the buildpack ID, its SHA will be returned. Otherwise, an empty string will be returned.
-func (lss *defaultLayerSHAStore) getShaByBuildpackLayer(buildpackID, layerName string) string {
-	if layerToSha, buildpackExists := lss.buildpacksToLayersShaMap[buildpackID]; buildpackExists {
+func (mlss *mapLayerSHAStore) getShaByBuildpackLayer(buildpackID, layerName string) string {
+	if layerToSha, buildpackExists := mlss.buildpacksToLayersShaMap[buildpackID]; buildpackExists {
 		if sha, layerExists := layerToSha.layerToShaMap[layerName]; layerExists {
 			return sha
 		}
@@ -83,17 +84,12 @@ func (lss *defaultLayerSHAStore) getShaByBuildpackLayer(buildpackID, layerName s
 	return ""
 }
 
-func NewLayerMetadataRestorer(logger Logger, layersDir string, skipLayers, useShaFiles bool) LayerMetadataRestorer {
+func NewLayerMetadataRestorer(logger Logger, layersDir string, skipLayers bool) LayerMetadataRestorer {
 	return &DefaultLayerMetadataRestorer{
-		logger:      logger,
-		layersDir:   layersDir,
-		skipLayers:  skipLayers,
-		useShaFiles: useShaFiles,
+		logger:     logger,
+		layersDir:  layersDir,
+		skipLayers: skipLayers,
 	}
-}
-
-func (la *DefaultLayerMetadataRestorer) UseSHAFiles() bool {
-	return la.useShaFiles
 }
 
 func (la *DefaultLayerMetadataRestorer) Restore(buildpacks []buildpack.GroupBuildpack, appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata, layerSHAStore LayerSHAStore) error {
--- a/layer_metadata_restorer.go
+++ b/layer_metadata_restorer.go
@@ -13,7 +13,7 @@ import (
 
 //go:generate mockgen -package testmock -destination testmock/layer_metadata_restorer.go github.com/buildpacks/lifecycle LayerMetadataRestorer
 type LayerMetadataRestorer interface {
-	Restore(buildpacks []buildpack.GroupBuildpack, appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata) error
+	Restore(buildpacks []buildpack.GroupBuildpack, appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata, writeShaToFile bool) (BuildpackLayersToSha, error)
 }
 
 type DefaultLayerMetadataRestorer struct {
@@ -30,16 +30,17 @@ func NewLayerMetadataRestorer(logger Logger, layersDir string, skipLayers bool) 
 	}
 }
 
-func (la *DefaultLayerMetadataRestorer) Restore(buildpacks []buildpack.GroupBuildpack, appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata) error {
+func (la *DefaultLayerMetadataRestorer) Restore(buildpacks []buildpack.GroupBuildpack, appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata, writeShaToFile bool) (BuildpackLayersToSha, error) {
 	if err := la.restoreStoreTOML(appMeta, buildpacks); err != nil {
-		return err
+		return BuildpackLayersToSha{}, err
 	}
 
-	if err := la.restoreLayerMetadata(appMeta, cacheMeta, buildpacks); err != nil {
-		return err
+	buildpackLayersToSha, err := la.restoreLayerMetadata(appMeta, cacheMeta, buildpacks, writeShaToFile)
+	if err != nil {
+		return BuildpackLayersToSha{}, err
 	}
 
-	return nil
+	return buildpackLayersToSha, nil
 }
 
 func (la *DefaultLayerMetadataRestorer) restoreStoreTOML(appMeta platform.LayersMetadata, buildpacks []buildpack.GroupBuildpack) error {
@@ -53,24 +54,63 @@ func (la *DefaultLayerMetadataRestorer) restoreStoreTOML(appMeta platform.Layers
 	return nil
 }
 
-func (la *DefaultLayerMetadataRestorer) restoreLayerMetadata(appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata, buildpacks []buildpack.GroupBuildpack) error {
+type layerToSha struct {
+	layerToShaMap map[string]string
+}
+
+func initializeLayerToSha() layerToSha {
+	return layerToSha{make(map[string]string)}
+}
+
+type BuildpackLayersToSha struct {
+	buildpacksToLayersShaMap map[string]layerToSha
+}
+
+func initializeBuildpackLayersToSha() BuildpackLayersToSha {
+	return BuildpackLayersToSha{make(map[string]layerToSha)}
+}
+
+// if the layer exists for the buildpack ID, its SHA will be returned. Otherwise, an empty string will be returned.
+func (bls *BuildpackLayersToSha) getShaByBuildpackLayers(buildpackID, layerName string) string {
+	if layerToSha, buildpackExists := bls.buildpacksToLayersShaMap[buildpackID]; buildpackExists {
+		if sha, layerExists := layerToSha.layerToShaMap[layerName]; layerExists {
+			return sha
+		}
+	}
+	return ""
+}
+
+func (bls *BuildpackLayersToSha) addLayerToMap(buildpackID, layerName, sha string) {
+	_, ok := bls.buildpacksToLayersShaMap[buildpackID]
+	if !ok {
+		bls.buildpacksToLayersShaMap[buildpackID] = initializeLayerToSha()
+	}
+	bls.buildpacksToLayersShaMap[buildpackID].layerToShaMap[layerName] = sha
+}
+
+func (la *DefaultLayerMetadataRestorer) restoreLayerMetadata(appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata, buildpacks []buildpack.GroupBuildpack, writeShaToFile bool) (BuildpackLayersToSha, error) {
+	var buildpackLayersToSha BuildpackLayersToSha
+	if !writeShaToFile {
+		buildpackLayersToSha = initializeBuildpackLayersToSha()
+	}
+
 	if la.SkipLayers {
 		la.Logger.Infof("Skipping buildpack layer analysis")
-		return nil
+		return BuildpackLayersToSha{}, nil
 	}
 
 	for _, buildpack := range buildpacks {
 		buildpackDir, err := readBuildpackLayersDir(la.LayersDir, buildpack, la.Logger)
 		if err != nil {
-			return errors.Wrap(err, "reading buildpack layer directory")
+			return BuildpackLayersToSha{}, errors.Wrap(err, "reading buildpack layer directory")
 		}
 
 		// Restore metadata for launch=true layers.
 		// The restorer step will restore the layer data for cache=true layers if possible or delete the layer.
 		appLayers := appMeta.MetadataForBuildpack(buildpack.ID).Layers
 		cachedLayers := cacheMeta.MetadataForBuildpack(buildpack.ID).Layers
-		for name, layer := range appLayers {
-			identifier := fmt.Sprintf("%s:%s", buildpack.ID, name)
+		for layerName, layer := range appLayers {
+			identifier := fmt.Sprintf("%s:%s", buildpack.ID, layerName)
 			if !layer.Launch {
 				la.Logger.Debugf("Not restoring metadata for %q, marked as launch=false", identifier)
 				continue
@@ -82,7 +122,7 @@ func (la *DefaultLayerMetadataRestorer) restoreLayerMetadata(appMeta platform.La
 				continue
 			}
 			if layer.Cache {
-				if cacheLayer, ok := cachedLayers[name]; !ok || !cacheLayer.Cache {
+				if cacheLayer, ok := cachedLayers[layerName]; !ok || !cacheLayer.Cache {
 					// The layer is not cache=true in the cache metadata and will not be restored.
 					// Do not write the metadata file so that it is clear to the buildpack that it needs to recreate the layer.
 					// Although a launch=true (only) layer still needs a metadata file, the restorer will remove the file anyway when it does its cleanup (for buildpack apis < 0.6).
@@ -91,15 +131,15 @@ func (la *DefaultLayerMetadataRestorer) restoreLayerMetadata(appMeta platform.La
 				}
 			}
 			la.Logger.Infof("Restoring metadata for %q from app image", identifier)
-			if err := la.writeLayerMetadata(buildpackDir, name, layer); err != nil {
-				return err
+			if err := la.writeLayerMetadata(buildpackDir, layerName, layer, writeShaToFile, buildpack.ID, &buildpackLayersToSha); err != nil {
+				return BuildpackLayersToSha{}, err
 			}
 		}
 
 		// Restore metadata for cache=true layers.
 		// The restorer step will restore the layer data if possible or delete the layer.
-		for name, layer := range cachedLayers {
-			identifier := fmt.Sprintf("%s:%s", buildpack.ID, name)
+		for layerName, layer := range cachedLayers {
+			identifier := fmt.Sprintf("%s:%s", buildpack.ID, layerName)
 			if !layer.Cache {
 				la.Logger.Debugf("Not restoring %q from cache, marked as cache=false", identifier)
 				continue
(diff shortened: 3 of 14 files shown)


restorer.go
```diff
@@ -31,8 +31,10 @@ func (r *Restorer) Restore(cache Cache) error {
 		return err
 	}
 
+	var buildpackLayersToSha BuildpackLayersToSha
 	if r.restoresLayerMetadata() {
-		if err := r.LayerMetadataRestorer.Restore(r.Buildpacks, r.LayersMetadata, cacheMeta); err != nil {
+		var err error
+		if buildpackLayersToSha, err = r.LayerMetadataRestorer.Restore(r.Buildpacks, r.LayersMetadata, cacheMeta, false); err != nil {
```
yaelharel, line 37: I decided to add the BuildpackLayersToSha return value to the `LayerMetadataRestorer.Restore` function, although it's possible to solve the issue without it.
If we would want to not add this return value, we would need to add a few `if-else` conditions to this function as appear [here](https://github.com/buildpacks/lifecycle/blob/main/layer_metadata_restorer.go#L72-L116) in order to decide whether to read the sha from the app layer, the cache layer, or to not read it from anywhere.
Since we call the `restoreLayerMetadata` function anyway in this flow, I decided to just use what we calculate there.

If anyone has different opinion or a better solution for it, I would love to discuss it.

jabrown85: I wonder if something like this would make it so we could not return `buildpackLayersToSha` to the caller but the `LayerMetadataRestorer` could keep the data around after `Restore` and use it during `validateContents`

```
if !r.LayerMetadataRestorer.validateContents(buildpack.ID, layerName, cachedLayer.SHA) {
   r.Logger.Infof("Removing %q, wrong sha", bpLayer.Identifier())
}
```

layers.go
```diff
@@ -124,7 +124,7 @@ type bpLayer struct { // TODO: need to refactor so api and logger won't be part
 	logger Logger
 }
 
-func (bp *bpLayer) read() (platform.BuildpackLayerMetadata, error) {
+func (bp *bpLayer) read(restoreSha bool) (platform.BuildpackLayerMetadata, error) {
```
jabrown85, line 127:
```suggestion
func (bp *bpLayer) read(readShaFile bool) (platform.BuildpackLayerMetadata, error) {
```

layer_metadata_restorer.go
```diff
@@ -53,24 +54,63 @@ func (la *DefaultLayerMetadataRestorer) restoreStoreTOML(appMeta platform.Layers
...
+	if layerToSha, buildpackExists := bls.buildpacksToLayersShaMap[buildpackID]; buildpackExists {
+		if sha, layerExists := layerToSha.layerToShaMap[layerName]; layerExists {
+			return sha
+		}
+	}
+	return ""
+}
+
+func (bls *BuildpackLayersToSha) addLayerToMap(buildpackID, layerName, sha string) {
+	_, ok := bls.buildpacksToLayersShaMap[buildpackID]
```
jabrown85, line 84:
```suggestion
	_, exists := bls.buildpacksToLayersShaMap[buildpackID]
```

layer_metadata_restorer.go
```diff
@@ -53,24 +54,63 @@ func (la *DefaultLayerMetadataRestorer) restoreStoreTOML(appMeta platform.Layers
...
+	if !ok {
+		bls.buildpacksToLayersShaMap[buildpackID] = initializeLayerToSha()
+	}
+	bls.buildpacksToLayersShaMap[buildpackID].layerToShaMap[layerName] = sha
+}
+
+func (la *DefaultLayerMetadataRestorer) restoreLayerMetadata(appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata, buildpacks []buildpack.GroupBuildpack, writeShaToFile bool) (BuildpackLayersToSha, error) {
```
jabrown85, line 91: Having `writeShaToFile` in `restoreLayerMetadata` and in `writeLayerMetadata` makes it so a consumer could actually get themselves into a weird state. Do you think we could hoist `writeShaToFile` to `NewLayerMetadataRestorer` so symmetry between the read and write exists?

restorer.go
yaelharel: Interesting. But what is the advantage of doing this? Don't you think that returning a value from `Restore` is clearer?

jabrown85: I see that as an implementation detail to the `LayerMetadataRestorer`. If we give the `LayerMetadataRestorer` enough information up front, I would like to think the consumers of that interface shouldn't have to see any internal data structures. I'm trying to push more data _into_ the `LayerMetadataRestorer` as opposed to have more in the `restorer` itself.

layer_metadata_restorer.go
yaelharel: I like your idea, but I just wanted to make sure we're on the same page:

The write flow is:
```
Restorer::Restore
LayerMetadataRestorer::Restore (DefaultLayerMetadataRestorer)
DefaultLayerMetadataRestorer::restoreLayerMetadata
DefaultLayerMetadataRestorer::writeLayerMetadata
bpLayer::writeSha
```

The read flow is:
```
Restorer::Restore
bpLayer::read
```

Adding `writeShaToFile` to `NewLayerMetadataRestorer` is a good idea in order to reduce the number of variables to:
```
LayerMetadataRestorer::Restore (DefaultLayerMetadataRestorer)
DefaultLayerMetadataRestorer::restoreLayerMetadata
DefaultLayerMetadataRestorer::writeLayerMetadata
```

We'll still need to use this extra variable when calling `bpLayer::writeSha` and when deciding whether to call `bpLayer::read`.

Are you ok with this or did you mean something else?

layers.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/7934b649bc47e828d91f72ac2217d6b9f7c54736.

layer_metadata_restorer.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/7934b649bc47e828d91f72ac2217d6b9f7c54736.

restorer.go
yaelharel: I like the idea of not exposing internal data structures, but for code readability and keeping one flow for old and new API versions, I would replace [this line](https://github.com/buildpacks/lifecycle/pull/624/files#diff-98f352cc6ef347aa505fe6d5f3e01c3a47ee0b5d1a91d1e918343284c893b51eR88) with `r.LayerMetadataRestorer.getShaByBuildpackLayers(buildpack.ID, layerName)`.
But then I'm wondering again if it would be clearer to the reader.
Maybe even though it will be a little bit less clearer than the current code (from my point of view), it's better to do that in order to prevent exposing the `BuildpackLayersToSha` data structure.

What do you think?

layer_metadata_restorer.go
jabrown85: Yeah - that is what I had in mind

restorer.go
jabrown85: Maybe? It's hard to say without trying it out..I really would like to see this type of branching go away from `restorer.go` entirely. This is essentially platform API version branching.

```
restoreSha := !r.restoresLayerMetadata()
var layerSha string
if restoreSha {
  layerSha = data.SHA
} else {
  layerSha = buildpackLayersToSha.getShaByBuildpackLayers(buildpack.ID, layerName)
}
if layerSha != cachedLayer.SHA {
  r.Logger.Infof("Removing %q, wrong sha", bpLayer.Identifier())
  ...
}
```

In favor of something like this. 
```
if !someSmallerUnitOfWork.CacheIsValid(bpLayer, cacheLayer) {
  r.Logger.Infof("Removing %q, wrong sha", bpLayer.Identifier())
  ...
}
```

The `someSmallerUnitOfWork` here may not be `LayerMetadataRestorer`, but ideally the platform version swapping behavior can be hoisted out to where we build `restorer` and `someSmallerUnitOfWork` in the first place. But this is not a blocker - just thinking out loud really.

yaelharel: I can try to do something. Let me work on it and I'll make this change in a separate PR so we can revert it if we don't like it.

layer_metadata_restorer.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/8a9af603ec9565b21ee5e43723c5c52f4263ebd1.
I'm going to mark this conversation as resolved but please feel free to reopen it if you think that I didn't address your feedback.
Thanks!

jabrown85 (approved):

restorer.go
yaelharel: What do you think about https://github.com/buildpacks/lifecycle/pull/624/commits/a26294188114c6d1fc9f6001d4427fc0c2df2e27?
I put the new function in `LayerMetadataRestorer` in the meantime although it's not the best place for it.

yaelharel: Please correct me if I'm wrong but I think that the main problem of putting the new function in `LayerMetadataRestorer` is that the `analyzer` doesn't use it.
So maybe we can create another interface for the `restorer` that will use the `LayerMetadataRestorer` interface and expand it with the new function.
Similar to `ReadWriteCloser` in the [io package](https://golang.org/pkg/io/#ReadWriteCloser).

jabrown85: I like the changes! The tests are easier to read and reason about IMO. `analyzer` does have use `LayerMetadataRestorer` - did you mean something else?

yaelharel: I meant that the `analyzer` doesn't use the `CacheIsValid` function.

restorer_test.go
```diff
@@ -336,39 +385,81 @@ func testRestorerBuilder(buildpackAPI, platformAPI string) func(t *testing.T, wh
...
+                   },
+                   "cache": true,
+                   "launch": true,
+                   "sha": "%s"
+               }
+           }
+       }
+   ]
+}
+`, wrongSha))
```
natalieparellano, line 417: I find this variable name a bit confusing, as technically the SHA in the app is the "correct" one. Could we just call it `appSHA`?

natalieparellano (commented):
Pardon the many comments, some of them are me just trying to understand things better :)

The location of `CacheIsValid` is leading to some weirdness IMO. I missed out on some of the earlier discussion so maybe we could sync up at some point.

natalieparellano (commented):
Apologies if these are duplicate comments, I accidentally reviewed in two tabs so trying not to lose anything...

layers.go
```diff
@@ -20,7 +20,7 @@ import (
 
 type bpLayersDir struct {
 	path      string
-	layers    []bpLayer
+	layers    []BpLayer
```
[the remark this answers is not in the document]
natalieparellano: See https://github.com/buildpacks/lifecycle/pull/624#discussion_r643461761 - this **might** fix it.

layer_metadata_restorer.go
```diff
@@ -14,19 +14,24 @@ import (
 //go:generate mockgen -package testmock -destination testmock/layer_metadata_restorer.go github.com/buildpacks/lifecycle LayerMetadataRestorer
 type LayerMetadataRestorer interface {
 	Restore(buildpacks []buildpack.GroupBuildpack, appMeta platform.LayersMetadata, cacheMeta platform.CacheMetadata) error
+	CacheIsValid(buildpackID, cachedLayerSha string, layer BpLayer) (bool, error)
```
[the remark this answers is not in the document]
yaelharel: Yeah, I thought about it but since it's still in progress and I wanted to make sure that we're all ok with this change, I didn't add it yet. Some of my thoughts appear [here](https://github.com/buildpacks/lifecycle/pull/624#discussion_r643423966).

layer_metadata_restorer.go
```diff
@@ -69,62 +112,92 @@ func (la *DefaultLayerMetadataRestorer) restoreLayerMetadata(appMeta platform.La
...
+		return layer.writeSha(metadata.SHA)
+	}
+	la.buildpacksToLayersSha.addLayerToMap(buildpackID, layerName, metadata.SHA)
+	return nil
+}
+
+func (la *DefaultLayerMetadataRestorer) CacheIsValid(buildpackID, cachedLayerSha string, layer BpLayer) (bool, error) {
+	var layerSha string
+	if la.useShaFile {
+		data, err := layer.read(la.useShaFile)
```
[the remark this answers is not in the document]
yaelharel: It used to be: https://github.com/buildpacks/lifecycle/blob/main/restorer.go#L76

Maybe renaming `layer` will make it clearer?

layer_metadata_restorer.go
```diff
@@ -69,62 +112,92 @@ func (la *DefaultLayerMetadataRestorer) restoreLayerMetadata(appMeta platform.La
...
+		if err != nil {
+			return false, errors.Wrapf(err, "reading layer")
+		}
+		layerSha = data.SHA
+	} else {
+		layerSha = la.buildpacksToLayersSha.getShaByBuildpackLayers(buildpackID, layer.name())
```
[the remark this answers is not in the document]
yaelharel: I have a feeling that it relates to the name of `buildpacksToLayerSha` and the function `getShaByBuildpackLayers`, but I don't have a better name for it.
Any ideas?

layer_metadata_restorer.go
```diff
@@ -45,22 +50,60 @@ func (la *DefaultLayerMetadataRestorer) Restore(buildpacks []buildpack.GroupBuil
...
-	if la.SkipLayers {
-		la.Logger.Infof("Skipping buildpack layer analysis")
+	if la.skipLayers {
+		la.logger.Infof("Skipping buildpack layer analysis")
 		return nil
 	}
 
+	if !la.useShaFile {
+		la.buildpacksToLayersSha = initializeBuildpackLayersToSha()
+	}
```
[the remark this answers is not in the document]
yaelharel: We can but I don't like it because we call `NewLayerMetadataRestorer` even for `platform API < 0.7` and then I thought that maybe it's better that the map won't be initialized at all.
I don't have strong opinion on it, just didn't want to touch it if not necessary.

layer_metadata_restorer.go
```diff
@@ -45,22 +50,60 @@ func (la *DefaultLayerMetadataRestorer) Restore(buildpacks []buildpack.GroupBuil
...
 		}
 	}
 	return nil
 }
 
+type layerToSha struct {
+	layerToShaMap map[string]string
+}
+
+func initializeLayerToSha() layerToSha {
```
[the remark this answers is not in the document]
yaelharel: `initializeLayerToSha` is being called `addLayerToMap`, and it may happen more than once.
I think you wanted to put your comment on `initializeBuildpackLayersToSha`.
I don't have a strong opinion on it but I think it's clearer this way, especially if it's part of `restoreLayerMetadata` (and not as part of `NewLayerMetadataRestorer` as you suggested [here](https://github.com/buildpacks/lifecycle/pull/624#discussion_r643450316)).

restorer.go
```diff
@@ -64,27 +64,20 @@ func (r *Restorer) Restore(cache Cache) error {
...
 			if !exists {
 				r.Logger.Infof("Removing %q, not in cache", bpLayer.Identifier())
 				if err := bpLayer.remove(); err != nil {
 					return errors.Wrapf(err, "removing layer")
 				}
 				continue
 			}
-			data, err := bpLayer.read()
+
+			restoreLayer, err := r.LayerMetadataRestorer.CacheIsValid(buildpack.ID, cachedLayer.SHA, bpLayer)
```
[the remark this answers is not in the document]
yaelharel: As you can see in [this commit](https://github.com/buildpacks/lifecycle/pull/624/commits/a26294188114c6d1fc9f6001d4427fc0c2df2e27), the purpose of creating `CacheIsValid` was to hide [this part](https://github.com/buildpacks/lifecycle/pull/624/commits/a26294188114c6d1fc9f6001d4427fc0c2df2e27#diff-98f352cc6ef347aa505fe6d5f3e01c3a47ee0b5d1a91d1e918343284c893b51eL79-L102).
As I wrote [here](https://github.com/buildpacks/lifecycle/pull/624#discussion_r643423966), I'm not sure if this is the best direction and I have some thoughts about it.

restorer_test.go
```diff
@@ -88,38 +87,58 @@ func testRestorerBuilder(buildpackAPI, platformAPI string) func(t *testing.T, wh
...
 				when("there is a cache=false layer", func() {
 					it.Before(func() {
-						var meta string
+						var meta, sha string
 						if api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) < 0 {
 							meta = "cache=false"
 						}
-						h.AssertNil(t, writeLayer(layersDir, "buildpack.id", "cache-false", meta, "cache-false-layer-sha"))
+						if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
+							sha = "cache-false-layer-sha"
```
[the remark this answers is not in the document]
yaelharel: Do you want to add this comment in every place where the sha is an empty string?
I'm wondering if it's necessary because it appears in many places in the tests and I think that it's pretty clear in the `writeLayer` function. Maybe we can add a comment in the `writeLayer` function instead?

restorer_test.go
yaelharel: Interesting. I understand what you're saying but I wanted to make it clear that it's a different sha. I took the name from the `it` of this test `there is a cache=true layer with wrong sha`.

layer_metadata_restorer.go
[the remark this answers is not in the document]
natalieparellano: But `NewLayerMetadataRestorer` has access to `useShaFile`, no?

layer_metadata_restorer.go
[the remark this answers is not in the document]
natalieparellano: Does it happen more than once? I missed that. Personally I think having the separate function makes the code harder to read.

restorer_test.go
natalieparellano: Maybe `otherSHA`?

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: You're right. I forgot that I added it as part of one of the commits that I made yesterday.
I can move the `initializeBuildpackLayersToSha` there. Thanks.

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: It happens for every `buildpackID` that is being added to the map.
`restoreLayerMetadata` is looping over `buildpacks`, then inside we're looping over the layers and calling to `writeLayerMetadata` when needed. There we call `addLayerToMap` that will call `initializeLayerToSha` for the first layer of a `buildpackID`.
I think that if you'll review it through an IDE (instead of Github), it'll be easier for you to see it.
Personally, I think it's clearer to have it as a separate function.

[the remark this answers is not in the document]
natalieparellano: Ohh I just meant that `initializeLayerToSha` appears only twice in the code (where it's declared, and where it's invoked). But I am happy to keep it as-is if you feel strongly about it.

restorer.go
natalieparellano: @yaelharel and I had a discussion offline and we agree that it's weird that `CacheIsValid` is part of `LayerMetadataRestorer`.

I think this logic:
```
	var layerSha string
	if la.useShaFile {
		data, err := layer.read(la.useShaFile)
		if err != nil {
			return false, errors.Wrapf(err, "reading layer")
		}
		layerSha = data.SHA
	} else {
		layerSha = la.buildpacksToLayersSha.getShaByBuildpackLayers(buildpackID, layer.name())
	}
```
could go into a separate service, call it `LayerSHAStore`. It could have `Add` and `Get` methods. You could optionally pass this into `LayerMetadataRestorer.Restore`. The analyzer on older platform apis could ignore it (since it doesn't need it) and the restorer on newer platform apis could use it.

So to summarize you'd have `LayerMetadataRestorer.Restore()` internally calling `LayerSHAStore.Add()` if provided by the caller and restorer calling `LayerSHAStore.Get()`. 

WDYT?

jabrown85: That makes sense to me.

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: yeah, both functions appear only twice in the code - where they declared and where they invoked.

restorer.go
yaelharel: Great! I think we're all on the same page.
I'll work on making these changes.

natalieparellano: Copying my edit for visibility:

> Edit: I just realized that restorer on older platform apis will not be able to use `LayerSHAStore.Get()` since it's not calling `LayerMetadataRestorer.Restore()`. It will have to read the layer.sha directly. I'm not sure if that makes things better or worse.

So you'd have something like 
```
restoreSha := !r.restoresLayerMetadata()
var layerSha string
if restoreSha {
  layerSha = data.SHA
} else {
  layerSha = LayerSHAStore.Get(buildpack.ID, layerName)
}
if layerSha != cachedLayer.SHA {
  r.Logger.Infof("Removing %q, wrong sha", bpLayer.Identifier())
  ...
}
```

which from https://github.com/buildpacks/lifecycle/pull/624#discussion_r643257297 we were trying to avoid.

natalieparellano: I just want to add that despite retaining the branching I like this option "best" for now. We could potentially explore creating a new service, say `CacheValidator`, to wrap the if/else but I think having the `LayerSHAStore` abstraction would be a convenient way to pass the data from `LayerMetadataRestorer` to `CacheValidator` anyway so we might as well go down this route and potentially improve upon. Just my two cents!

layers.go
```diff
@@ -137,17 +137,20 @@ func (bp *bpLayer) read() (platform.BuildpackLayerMetadata, error) {
...
 		}
 	}
-	sha, err := ioutil.ReadFile(bp.path + ".sha")
-	if err != nil {
-		if os.IsNotExist(err) {
-			return platform.BuildpackLayerMetadata{LayerMetadata: platform.LayerMetadata{SHA: ""}, LayerMetadataFile: layerMetadataFile}, nil
+	var sha string
+	if readShaFile {
+		shaBytes, err := ioutil.ReadFile(bp.path + ".sha")
+		if err != nil && !os.IsNotExist(err) {
```
natalieparellano, line 143: Looking at this again... if the layer.sha file does not exist we're happily returning `platform.BuildpackLayerMetadata{LayerMetadata: platform.LayerMetadata{SHA: sha}, LayerMetadataFile: layerMetadataFile}, nil` ...so why do we need to guard this logic on `if readShaFile`? I think we could just take it out.

layer_metadata_restorer.go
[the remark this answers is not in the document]
natalieparellano: Possibly relevant: https://github.com/buildpacks/lifecycle/pull/624/files#r644317482

layers.go
yaelharel: I agree that it should work if we delete the `if readShaFile` condition.
We can also add a comment next to the `if err != nil && !os.IsNotExist(err) {`.
But I think it's more readable the way it's now.
What do you think?

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: I don't think that it's too relevant unless what is bothering you here is the argument that we're passing to the `read` function. We'll need to call the `read` function anyway and to wrap this call with an if condition (or at least the `else` part).

layers.go
yaelharel: After thinking about it again, I decided to remove the variable as you suggested and just add a clarification comment. Thanks!

yaelharel: Although...maybe it's better for the future to split this function to 2 functions because most of the times when we call it, we don't even care about the sha file.

layer_metadata_restorer_test.go
```diff
@@ -32,15 +32,18 @@ func testLayerMetadataRestorer(t *testing.T, when spec.G, it spec.S) {
...
 
 	it.Before(func() {
 		var err error
 
 		layerDir, err = ioutil.TempDir("", "lifecycle-layer-dir")
 		h.AssertNil(t, err)
-		logger := &log.Logger{Handler: &discard.Handler{}}
-		layerAnalyzer = lifecycle.NewLayerMetadataRestorer(logger, layerDir, skipLayers)
+		useShaFile = true
+		logger = log.Logger{Handler: &discard.Handler{}}
```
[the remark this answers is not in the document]
yaelharel: I created the `logger` so we can define it only once and use it when recreating `layerAnalyzer`.
Since I moved the `Removing %q, wrong sha` logic back to the `Restorer`, we can't test it here.

cmd/lifecycle/analyzer.go
```diff
@@ -200,7 +200,7 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 		Logger:                cmd.DefaultLogger,
 		Platform:              aa.platform,
 		Image:                 img,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, aa.layersDir, aa.platform06.skipLayers, true),
```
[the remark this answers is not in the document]
yaelharel: Regarding creating a variable, I didn't think it's necessary because when you use an IDE, it usually shows you what variable you're setting, but I added it.

Regarding the analyzer acceptance tests, I think that we're already testing that there are no sha files [here](https://github.com/buildpacks/lifecycle/blob/main/acceptance/analyzer_test.go#L1074-L1087). Do you want to add something else?

layer_metadata_restorer.go
```diff
@@ -14,19 +14,24 @@ import (
...
 
 type DefaultLayerMetadataRestorer struct {
-	Logger     Logger
-	LayersDir  string
-	SkipLayers bool
+	logger                Logger
+	layersDir             string
+	skipLayers            bool
+	useShaFile            bool
+	buildpacksToLayersSha buildpackLayersToSha
```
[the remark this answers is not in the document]
yaelharel: I'm going to "hide" this struct in an interface so maybe it's better to keep it very explicit. Please let me know what you think after my next commit.

layer_metadata_restorer.go
```diff
@@ -45,22 +50,60 @@ func (la *DefaultLayerMetadataRestorer) Restore(buildpacks []buildpack.GroupBuil
...
+type buildpackLayersToSha struct {
+	buildpacksToLayersShaMap map[string]layerToSha
+}
+
+func initializeBuildpackLayersToSha() buildpackLayersToSha {
+	return buildpackLayersToSha{make(map[string]layerToSha)}
+}
+
+// if the layer exists for the buildpack ID, its SHA will be returned. Otherwise, an empty string will be returned.
+func (bls *buildpackLayersToSha) getShaByBuildpackLayers(buildpackID, layerName string) string {
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624#discussion_r645217551.

layer_metadata_restorer.go
```diff
@@ -45,22 +50,60 @@ func (la *DefaultLayerMetadataRestorer) Restore(buildpacks []buildpack.GroupBuil
...
+	if layerToSha, buildpackExists := bls.buildpacksToLayersShaMap[buildpackID]; buildpackExists {
+		if sha, layerExists := layerToSha.layerToShaMap[layerName]; layerExists {
+			return sha
+		}
+	}
+	return ""
+}
+
+func (bls *buildpackLayersToSha) addLayerToMap(buildpackID, layerName, sha string) {
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624#discussion_r645217551.

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: I moved it to `NewLayerSHAStore`. Please see my coming commit.

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: I changed this code as part of my last commit.

layers.go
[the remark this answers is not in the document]
yaelharel: The implementation was changed so it's private again :)

restorer_test.go
```diff
@@ -439,18 +542,55 @@ func testRestorerBuilder(buildpackAPI, platformAPI string) func(t *testing.T, wh
...
 				when("there are multiple cache=true layers", func() {
-					var typesMeta, cacheOnlyMeta, cacheLaunchMeta, escapedMeta string
+					var cacheOnlyMeta, cacheLaunchMeta, escapedMeta string
 
 					it.Before(func() {
+						var typesMeta, cacheOnlySha, cacheLunchSha, escapedSha string
+
+						if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0 {
+							cacheOnlySha = cacheOnlyLayerSHA
+							cacheLunchSha = cacheLaunchLayerSHA
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624#discussion_r643580066. I'm not sure that it's really necessary and if it's, probably it's better to write it everywhere.

restorer_test.go
yaelharel: I changed it to `madeUpSha` because we're setting it to `some-made-up-sha`. What do you think?

layers.go
yaelharel: I decided to take you advice and to remove the `readShaFile` completely from this function.
I added a comment next to the `if` condition.
Please take a look and let me know what you think.

restorer.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

layer_metadata_restorer_test.go
```diff
@@ -32,15 +32,18 @@ func testLayerMetadataRestorer(t *testing.T, when spec.G, it spec.S) {
...
 	)
 
 	it.Before(func() {
 		var err error
 
 		layerDir, err = ioutil.TempDir("", "lifecycle-layer-dir")
 		h.AssertNil(t, err)
-		logger := &log.Logger{Handler: &discard.Handler{}}
-		layerAnalyzer = lifecycle.NewLayerMetadataRestorer(logger, layerDir, skipLayers)
+		useShaFile = true
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

layer_metadata_restorer_test.go
```diff
@@ -348,8 +369,11 @@ func testLayerMetadataRestorer(t *testing.T, when spec.G, it spec.S) {
 				})
 			})
 
-			when("SkipLayers is true", func() {
-				skipLayers = true
+			when("skipLayers is true", func() {
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

cmd/lifecycle/restorer.go
```diff
@@ -121,7 +121,7 @@ func (r restoreArgs) restore(layerMetadata platform.LayersMetadata, group buildp
 		Buildpacks:            group.Group,
 		Logger:                cmd.DefaultLogger,
 		Platform:              r.platform,
-		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers),
+		LayerMetadataRestorer: lifecycle.NewLayerMetadataRestorer(cmd.DefaultLogger, r.layersDir, r.skipLayers, !r.restoresLayerMetadata()),
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d and let me know if it's still weird to you.

layer_metadata_restorer.go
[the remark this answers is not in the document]
yaelharel: I hide this implementation in the `LayerSHAStore` interface. Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d and let me know what you think.

layer_metadata_restorer.go
```diff
@@ -14,19 +14,24 @@ import (
...
-	SkipLayers bool
+	logger                Logger
+	layersDir             string
+	skipLayers            bool
+	useShaFile            bool
+	buildpacksToLayersSha buildpackLayersToSha
 }
 
-func NewLayerMetadataRestorer(logger Logger, layersDir string, skipLayers bool) LayerMetadataRestorer {
+func NewLayerMetadataRestorer(logger Logger, layersDir string, skipLayers, useShaFile bool) LayerMetadataRestorer {
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

layer_metadata_restorer.go
```diff
@@ -14,19 +14,24 @@ import (
...
 	return &DefaultLayerMetadataRestorer{
-		Logger:     logger,
-		LayersDir:  layersDir,
-		SkipLayers: skipLayers,
+		logger:                logger,
+		layersDir:             layersDir,
+		skipLayers:            skipLayers,
+		useShaFile:            useShaFile,
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

restorer.go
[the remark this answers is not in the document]
yaelharel: The code was changed as part of https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d. Please take a look and let me know what you think.

restorer_test.go
```diff
@@ -88,38 +87,58 @@ func testRestorerBuilder(buildpackAPI, platformAPI string) func(t *testing.T, wh
...
 
 						h.AssertPathDoesNotExist(t, filepath.Join(layersDir, "buildpack.id", "cache-true.toml"))
-						h.AssertPathDoesNotExist(t, filepath.Join(layersDir, "buildpack.id", "cache-true.sha"))
+					})
+
+					it("removes sha file", func() {
+						h.SkipIf(t, api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) >= 0, "Not possible to determine if layers not in cache are launch-only or the result of incorrect metadata")
+						h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0, "sha file isn't created")
+
+						h.AssertPathDoesNotExist(t, filepath.Join(layersDir, "buildpack.id", "cache-true.sha")) // not really needed for platformAPI >= 0.7
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

restorer_test.go
```diff
@@ -132,38 +151,58 @@ func testRestorerBuilder(buildpackAPI, platformAPI string) func(t *testing.T, wh
...
 					})
 
-					it("removes metadata and sha file", func() {
+					it("removes metadata file", func() {
 						h.SkipIf(t, api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) >= 0, "Not possible to determine if layers not in cache are launch-only or the result of incorrect metadata")
 
 						h.AssertPathDoesNotExist(t, filepath.Join(layersDir, "buildpack.id", "cache-true.toml"))
+					})
+
+					it("removes metadata and sha file", func() {
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

restorer_test.go
```diff
@@ -336,39 +385,81 @@ func testRestorerBuilder(buildpackAPI, platformAPI string) func(t *testing.T, wh
...
 					})
 
-					it("removes metadata and sha file", func() {
+					it("removes metadata file", func() {
 						h.SkipIf(t, api.MustParse(buildpackAPI).Compare(api.MustParse("0.6")) >= 0, "Not possible to determine if layers not in cache are launch-only or the result of incorrect metadata")
-
 						h.AssertPathDoesNotExist(t, filepath.Join(layersDir, "buildpack.id", "cache-layer-not-in-cache.toml"))
+					})
+
+					it("removes metadata and sha file", func() {
```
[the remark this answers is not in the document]
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.

yaelharel:
@natalieparellano @jabrown85, please review my last commit https://github.com/buildpacks/lifecycle/pull/624/commits/f584e397d9e441ee1860957da6c18e023d65a89d.
I marked some of the conversations as resolved because I think that I addressed the feedback or the comment wasn't relevant anymore. Please feel free to reopen them.
I kept some of the conversations open with questions or since I wanted to give you the opportunity to take a look easily and make sure that I really solved them. Please comment or mark them as resolved if you think that no other changes are necessary.
Thanks!

layer_metadata_restorer_test.go
```diff
@@ -28,19 +28,24 @@ func testLayerMetadataRestorer(t *testing.T, when spec.G, it spec.S) {
...
 		var err error
 
 		layerDir, err = ioutil.TempDir("", "lifecycle-layer-dir")
 		h.AssertNil(t, err)
-		logger := &log.Logger{Handler: &discard.Handler{}}
-		layerAnalyzer = lifecycle.NewLayerMetadataRestorer(logger, layerDir, skipLayers)
+		useShaFiles = true // notice - the default for platform API >= 0.7 is false (it's set to false in some of the tests)
+		logger = log.Logger{Handler: &discard.Handler{}}
+		layerAnalyzer = lifecycle.NewLayerMetadataRestorer(&logger, layerDir, skipLayers, useShaFiles)
+		layerSHAStore = lifecycle.NewLayerSHAStore(layerAnalyzer.UseSHAFiles())
```
jabrown85, line 48:
```suggestion
		layerSHAStore = lifecycle.NewLayerSHAStore(useShaFiles)
```

layer_metadata_restorer_test.go
```diff
@@ -182,8 +187,27 @@ func testLayerMetadataRestorer(t *testing.T, when spec.G, it spec.S) {
 				})
 			})
 
+			when("restoring sha files is not needed", func() {
+				it.Before(func() {
+					useShaFiles = false
+					layerAnalyzer = lifecycle.NewLayerMetadataRestorer(&logger, layerDir, skipLayers, useShaFiles)
+					layerSHAStore = lifecycle.NewLayerSHAStore(layerAnalyzer.UseSHAFiles())
```
jabrown85, line 194:
```suggestion
					layerSHAStore = lifecycle.NewLayerSHAStore(useShaFiles)
```

jabrown85 (approved):
I like the recent changes! The bending of the code to handle so many simultaneous platform versions makes what seems like a reasonably small change so complicated.

I really want to experiment with pushing that complexity away via duplication on our next platform version.

layer_metadata_restorer_test.go
yaelharel: I know that it's possible to make this change, but don't you think it's clearer the to keep it as is because this is how we do it in the real code?
This is how we keep both values sync for sure.

layer_metadata_restorer_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624#discussion_r645665123.

layer_metadata_restorer_test.go
jabrown85: Eh, I don't feel strongly about it. If you do, keep it as is 😄

yaelharel: I also don't have strong opinion about it. Just trying to make our code as clear as possible to everyone 😊
@natalieparellano, what do you think?

cmd/lifecycle/analyzer.go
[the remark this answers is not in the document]
natalieparellano: > Regarding the analyzer acceptance tests, I think that we're already testing that there are no sha files here. Do you want to add something else?

I missed that. Great!

restorer_test.go
natalieparellano: 🤷🏼  I prefer `otherSHA` or `appSHA`, but I won't die on this hill :)

layers.go
natalieparellano: Great!

layer_metadata_restorer_test.go
natalieparellano: See https://github.com/buildpacks/lifecycle/commit/f584e397d9e441ee1860957da6c18e023d65a89d#r51743291 (idea for making the layer SHA store "dumber")

restorer_test.go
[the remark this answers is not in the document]
natalieparellano: Ohh fine, let's not duplicate the comment. I wish I didn't have to navigate to the function declaration to understand that an empty sha string results in non-existent sha file, but maybe there isn't a better solution.

cmd/lifecycle/analyzer.go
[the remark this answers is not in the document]
natalieparellano: Sorry, I should have said `restorer` acceptance tests! For newer platform api...

restorer_test.go
yaelharel: 😁 No problem, I'll change it to `otherSHA`.

layer_metadata_restorer.go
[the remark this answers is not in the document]
natalieparellano: I'm going to mark this conversation as resolved in favor of https://github.com/buildpacks/lifecycle/pull/624#discussion_r645813464

layer_metadata_restorer.go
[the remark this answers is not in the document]
natalieparellano: Marking this as resolved in favor of #624 (comment)

restorer_test.go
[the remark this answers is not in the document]
yaelharel: I added a comment in the function to make it clearer.

layer_metadata_restorer.go
```diff
@@ -13,29 +13,95 @@ import (
...
+func (lss *defaultLayerSHAStore) addLayerToMap(buildpackID, layerName, sha string) {
+	_, exists := lss.buildpacksToLayersShaMap[buildpackID]
+	if !exists {
+		lss.buildpacksToLayersShaMap[buildpackID] = layerToSha{make(map[string]string)}
+	}
+	lss.buildpacksToLayersShaMap[buildpackID].layerToShaMap[layerName] = sha
+}
+
+// if the layer exists for the buildpack ID, its SHA will be returned. Otherwise, an empty string will be returned.
+func (lss *defaultLayerSHAStore) getShaByBuildpackLayers(buildpackID, layerName string) string {
```
natalieparellano, line 77:
```suggestion
func (lss *defaultLayerSHAStore) getShaByBuildpackLayer(buildpackID, layerName string) string {
```

natalieparellano: (nit)

layer_metadata_restorer_test.go
yaelharel: Marking this conversation as resolved following https://github.com/buildpacks/lifecycle/pull/624#commitcomment-51744864.

layer_metadata_restorer.go
```diff
@@ -13,29 +13,95 @@ import (
...
+	logger      Logger
+	layersDir   string
+	skipLayers  bool
+	useShaFiles bool
 }
 
-func NewLayerMetadataRestorer(logger Logger, layersDir string, skipLayers bool) LayerMetadataRestorer {
+type LayerSHAStore interface {
+	add(buildpackID, sha string, layer *bpLayer) error
+	get(buildpackID string, layer bpLayer) (string, error)
```
natalieparellano, line 29: This is possible, but is there any particular reason to do this? Making the methods public would allow us to more easily extract a new package if we needed to.

layer_metadata_restorer.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/cb53965d3e716161d3a6131385120c3967d1be80.

restorer_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/cb53965d3e716161d3a6131385120c3967d1be80.

layer_metadata_restorer.go
```diff
@@ -13,29 +13,95 @@ import (
...
-	LayersDir  string
-	SkipLayers bool
+	logger      Logger
+	layersDir   string
+	skipLayers  bool
+	useShaFiles bool
 }
 
-func NewLayerMetadataRestorer(logger Logger, layersDir string, skipLayers bool) LayerMetadataRestorer {
+type LayerSHAStore interface {
```
natalieparellano, line 27: Could we add a docs comment here explaining how this "store" is using the filesystem OR memory?

layer_metadata_restorer.go
```diff
@@ -13,29 +13,95 @@ import (
...
+	add(buildpackID, sha string, layer *bpLayer) error
+	get(buildpackID string, layer bpLayer) (string, error)
+}
+
+type defaultLayerSHAStore struct {
+	useShaFiles              bool
```
natalieparellano, line 33: I wonder instead of having this boolean, we could have two different implementations of the interface - one using the filesystem, the other using memory. When the layer metadata restorer is constructed, we could pass in the right one.

layer_metadata_restorer_test.go
natalieparellano: If we do https://github.com/buildpacks/lifecycle/pull/624/files#r645837750 this question goes away.

layer_metadata_restorer.go
yaelharel: Unless you would like to extract them to a new package now, I think that it's better to keep them private for now. If we decide to change it in the future, it's very easy to refactor that.

natalieparellano: I guess so... I don't really see any precedent for private methods on an interface anywhere else in our code.

natalieparellano: Eventually I think it makes sense for layer metadata restorer and layer sha store to be part of an internal package. We could do that in a separate PR though.

layer_metadata_restorer_test.go
natalieparellano: (nit) This wasn't introduced in this PR, but could we change the name of `layerAnalyzer` in the test file to `layerMetadataRestorer` (or some abbreviated version of it)? `layerAnalyzer` is a hold out from when a previous PR was architected differently.

layer_metadata_restorer.go
jabrown85: I like that suggestion @natalieparellano !

layer_metadata_restorer.go
yaelharel: I saw that most interfaces have public methods although they don't really need it.
As I wrote, I don't like it.
I don't mind moving "layer metadata restorer and "layer sha store" to a different package (maybe it's better in a separate PR? I'm not sure) and then change these functions to be public.
Are you ok with that?

layer_metadata_restorer_test.go
yaelharel: Renamed `layerAnalyzer` to `layerMetadataRestorer` in https://github.com/buildpacks/lifecycle/pull/624/commits/debb186ed16488b89d25430039aa293877d0f6c8.
Going to mark this conversation as resolved and to track the changes that relates to the `useShaFile` boolean as part of https://github.com/buildpacks/lifecycle/pull/624/files#r645837750.

layer_metadata_restorer.go
yaelharel: I like it too! Please see https://github.com/buildpacks/lifecycle/pull/624/commits/76b95b3821257dd6127e05958f303dcab437fab8.

layer_metadata_restorer.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/624/commits/76b95b3821257dd6127e05958f303dcab437fab8. Personally I think that a comment is unnecessary now but please let me know if you think otherwise.

layer_metadata_restorer.go
natalieparellano: Made an issue: https://github.com/buildpacks/lifecycle/issues/629

natalieparellano (approved):

natalieparellano:
I love the new changes!

PR #626, opened 2021-06-08
Add GHA for shipping release candidates

### LINKED ISSUE #561: lifecycle should ship release candidates
Motivation: enable further validation of the lifecycle prior to shipping.

We could cut an RC version similar to `pack`: https://github.com/buildpacks/pack/releases/tag/v0.18.0-rc1

Note that we should **not** push semver tagged lifecycle images to `buildpacksio` without checking with downstream consumers first. Example: https://github.com/paketo-buildpacks/packit/issues/144. We could consider pushing to an alternative namespace.

We could also consider specifying the amount of time that we will require for code freeze before shipping. I believe on `pack` it's 5 days.

natalieparellano:
Resolves #561 

Note that publishing a pre-release doesn't re-tag the associated lifecycle image. I was concerned that consumers of the lifecycle might mistake it for a "real" release (see [Paketo issue](https://github.com/paketo-buildpacks/packit/issues/144), this has been fixed, but there could be other consumers with this problem). [pack](https://github.com/buildpacks/pack/blob/e53adb9c9e55b33e830f2e92bc4741cc32dc24f7/.github/workflows/delivery-docker.yml#L6) doesn't appear to ship pack images for pre-releases. 

However it's worth noting that potentially just publishing a pre-release on GitHub could confuse some people - perhaps we should do some socializing of this change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
           TEST_COVERAGE: 1
         run: make test
       - name: Upload Coverage
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v1.0.5
         with:
           file: ./out/tests/coverage-unit.txt
           flags: unit,os_linux
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -108,7 +108,55 @@ jobs:
           unzip -d artifact-windows artifact-windows.zip
           lifecycle_path=$(ls artifact-windows/lifecycle-*windows.x86-64.tgz)
           echo "ARTIFACT_WINDOWS_PATH=$PWD/$lifecycle_path" >> $GITHUB_ENV
+      - name: Create Pre Release
+        if: contains(${{ env.LIFECYCLE_VERSION }}, 'rc')
+        id: create_release
+        uses: actions/create-release@latest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: v${{ env.LIFECYCLE_VERSION }}
+          release_name: lifecycle v${{ env.LIFECYCLE_VERSION }}
+          draft: true
+          prerelease: true
+          body: |
+            # lifecycle v${{ env.LIFECYCLE_VERSION }}
+
+            Welcome to `v${{ env.LIFECYCLE_VERSION }}`, a **beta** pre-release of the Cloud Native Buildpacks Lifecycle.
+
+            ##  Prerequisites
+
+            The lifecycle runs as a normal user in a series of unprivileged containers. To export images and cache image layers, it requires access to a Docker daemon **or** Docker registry.
+
+            ## Install
+
+            Extract the `.tgz` file and copy the lifecycle binaries into a [build stack base image](https://github.com/buildpack/spec/blob/master/platform.md#stacks). The build image can then be orchestrated by a platform implementation such as the [pack CLI](https://github.com/buildpack/pack) or [tekton](https://github.com/tektoncd/catalog/blob/master/task/buildpacks/0.1/README.md).
+
+            ## Lifecycle Image
+
+            An OCI image containing the lifecycle binaries is available at buildpacksio/lifecycle:${{ github.sha }}.
+      - name: Upload Pre Release Asset - linux
+        if: contains(${{ env.LIFECYCLE_VERSION }}, 'rc')
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           TEST_COVERAGE: 1
         run: make test
       - name: Upload Coverage
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v1.0.5
```
natalieparellano, line 33: Fixes issue with uploading coverage: https://github.com/codecov/codecov-action/issues/43

.github/workflows/draft-release.yml
```diff
@@ -108,7 +111,55 @@ jobs:
...
+
+            The lifecycle runs as a normal user in a series of unprivileged containers. To export images and cache image layers, it requires access to a Docker daemon **or** Docker registry.
+
+            ## Install
+
+            Extract the `.tgz` file and copy the lifecycle binaries into a [build stack base image](https://github.com/buildpack/spec/blob/master/platform.md#stacks). The build image can then be orchestrated by a platform implementation such as the [pack CLI](https://github.com/buildpack/pack) or [tekton](https://github.com/tektoncd/catalog/blob/master/task/buildpacks/0.1/README.md).
+
+            ## Lifecycle Image
+
+            An OCI image containing the lifecycle binaries is available at buildpacksio/lifecycle:${{ env.SHORT_SHA }}.
```
natalieparellano, line 140: Added this since we aren't re-tagging the image. But I feel that it's a nice addition regardless.

.github/workflows/draft-release.yml
```diff
@@ -108,7 +111,55 @@ jobs:
...
+        id: create_pre_release
+        uses: actions/create-release@latest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: v${{ env.LIFECYCLE_VERSION }}
+          release_name: lifecycle v${{ env.LIFECYCLE_VERSION }}
+          draft: true
+          prerelease: true
+          body: |
```
natalieparellano, line 125: This duplication is 😐  but I don't see a way to fix it easily.

natalieparellano:
Tested on fork:

https://github.com/natalieparellano/lifecycle/actions/runs/919043320 produced https://github.com/natalieparellano/lifecycle/releases/tag/v0.99.0-rc.1

https://github.com/natalieparellano/lifecycle/actions/runs/916200178 produced https://github.com/natalieparellano/lifecycle/releases/tag/v0.99.0 and was followed by https://github.com/natalieparellano/lifecycle/actions/runs/916216614

.github/workflows/draft-release.yml
yaelharel: Can we put the text in a file and stream it in twice?

.github/workflows/draft-release.yml
```diff
@@ -108,7 +111,55 @@ jobs:
...
+
+            ## Install
+
+            Extract the `.tgz` file and copy the lifecycle binaries into a [build stack base image](https://github.com/buildpack/spec/blob/master/platform.md#stacks). The build image can then be orchestrated by a platform implementation such as the [pack CLI](https://github.com/buildpack/pack) or [tekton](https://github.com/tektoncd/catalog/blob/master/task/buildpacks/0.1/README.md).
+
+            ## Lifecycle Image
+
+            An OCI image containing the lifecycle binaries is available at buildpacksio/lifecycle:${{ env.SHORT_SHA }}.
+      - name: Upload Pre Release Asset - linux
+        if: contains(env.LIFECYCLE_VERSION, 'rc')
```
yaelharel, line 142: I think it's worth adding a comment at the end of this line, like an example of how a release candidate branch name should look like.

.github/workflows/draft-release.yml
natalieparellano: Maybe... but it gets tricky because we're also interpolating variables in the text.

yaelharel: I'm ok with keeping it as is but maybe if we're doing that, it's worth adding a comment next to each `body` so if we decide in the future to edit this text, we won't forget to do it in both places.

yaelharel (approved):
Added two small suggestions for adding comments, but other than that, it looks good! 😊

.github/workflows/draft-release.yml
natalieparellano: Since we have https://github.com/buildpacks/lifecycle/pull/626/files#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23cR5, not sure a comment would add much value here.

.github/workflows/draft-release.yml
natalieparellano: Concluded that improving this flow isn't worth the effort at this time.

Issue #435, opened 2020-10-09
Support ARM


ekcasey:
GOAL: Support buildpack builds for ARM images

** WARNING** Not yet ready to implement, may require RFCs, spec changes and/or imgutil changes.

# Lifecycle Behavior
## `analyzer` 
1. If the previous image is a multiarch image, select the child manifest that matches the runtime architecture of the build environment.
2. Warn and proceed without analyzing the previous image if there the image exists but has the wrong architecture.

Because `analyzer` may run in a lifecycle image rather than in the build environment the runtime architecture of the analyze phase may not match the build environment. We need an RFC to either:
1. Record the information in `detector` and pass it to `analyzer`.
2. Accept an `-arch` flag and fail if the previous image is a multiarch image and the `-arch` flag is not provided by the platform

## `exporter`
1. If the `run-image` is a multiarch image, select the child manifest that matches the runtime architecture of the build environment.
2. Fail if there is no `run-image` with the correct architecture
2. If there was a previous image, validate that the previous image architecture and the run image architecture match.

Because `exporter` may run in a lifecycle image rather than in the build environment the runtime architecture of the analyze phase may not match the build environment. We need an RFC to either:
1. Record the information in a previous phase and pass it to `analyzer`.
2. Accept an `-arch` flag and fail if the previous image is a multiarch image and the `-arch` flag is not provided by the platform

## `rebaser`
1. If the `run-image` is a multiarch image, select the child manifest that matches the **app image architecture**
2. If the `run-image` is *not* a multiarch image `rebaser` should validate that the architecture matches the app-image before rebasing.

# Lifecycle Artifacts
## Publish ARM lifecycle artifacts
Add the following new Makefile targets:
* `make build-linux-x86-64`
* `make build-linux-arm64`
* `make build-windows-x86-64`
* `make build-windows-arm64`
* `make package-linux-x86-64`
* `make package-linux-arm64`
* `make package-windows-x86-64`
* `make package-windows-arm64`
Binaries should end up in `out/<os>/<arch>`.

And update the existing targets:
* `make build-linux` -> `make build-linux-x86-64` && `make build-linux-arm64`
* `make build-windows` -> `make build-windows-x86-64` && `make build-windows-arm64`
* `make package-linux` -> `make package-linux-x86-64` && `make package-linux-arm64`
* `make package-windows` -> `make package-windows-x86-64` && `make package-windows-arm64`

## Local ARM image builds
Add the following new Makefile targets:
[...]

PR #628, opened 2021-06-09
Update dependencies

imjasonh:
Picks up https://github.com/buildpacks/imgutil/pull/124, upgrades the docker/docker dep to v20.10.7, drops pinned x/sys dep

Signed-off-by: Jason Hall <jasonhall@redhat.com>

cc @natalieparellano

PR #630, opened 2021-06-11
Use syscall package instead of cgo

imjasonh:
The use of cgo was introduced in https://github.com/buildpacks/lifecycle/pull/268, instead of using golang.org/x/sys/unix. I believe syscall.Setres{g,u}id is the platform-independent equivalent, but I admit I don't know enough about this to be confident this isn't a regression.

If this is acceptable, it could be a step toward unblocking https://github.com/buildpacks/lifecycle/issues/435

@ekcasey

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,6 @@ docker-compilation-image-linux:
 
 $(BUILD_DIR)/linux/lifecycle/lifecycle: export GOOS:=linux
 $(BUILD_DIR)/linux/lifecycle/lifecycle: OUT_DIR:=$(BUILD_DIR)/$(GOOS)/lifecycle
-$(BUILD_DIR)/linux/lifecycle/lifecycle: GOENV:=GOARCH=$(GOARCH) CGO_ENABLED=1
 $(BUILD_DIR)/linux/lifecycle/lifecycle: docker-compilation-image-linux
 $(BUILD_DIR)/linux/lifecycle/lifecycle: $(GOFILES)
 $(BUILD_DIR)/linux/lifecycle/lifecycle:
--- a/priv/user_linux.go
+++ b/priv/user_linux.go
@@ -11,27 +11,6 @@ import (
 	"syscall"
 )
 
-/*
-#cgo LDFLAGS: --static
-#define _GNU_SOURCE
-#include <unistd.h>
-#include <errno.h>
-
-static int
-csetresuid(uid_t ruid, uid_t euid, uid_t suid) {
-  int ec = setresuid(ruid, euid, suid);
-  return (ec < 0) ? errno : 0;
-}
-
-static int
-csetresgid(gid_t rgid, gid_t egid, gid_t sgid) {
-  int ec = setresgid(rgid, egid, sgid);
-  return (ec < 0) ? errno : 0;
-}
-
-*/
-import "C"
-
 // EnsureOwner recursively chowns a dir if it isn't writable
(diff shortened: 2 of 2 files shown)


micahyoung:
I personally really like this if it works equivalently. In addition to opening up `arm`, it would even simplify building for `amd64`, which currently when building with `make build-linux`, uses a Docker container ([Makefile](https://github.com/buildpacks/lifecycle/blob/main/Makefile#L70-L78)). 

I did some a quick search and it appears there was a bug related to multithreading in `Setuid` which [appears](https://github.com/golang/go/issues/1435) to have been fixed and merged into 1.16 in Dec 2020. We may need to bump lifecycle's Golang to 1.16 (and related build images, etc).

But we'll see what Emily says. In the meantime, I triggered a build and it is looking green.

natalieparellano:
IIUC the validation for this would involve running a phase like the analyzer a bunch of times as root and then observing if all the threads have the correct user when we drop privileges. @ekcasey I recall you did this validation ages ago, can you remember how many "trials" were sufficient to demonstrate that it was working?

natalieparellano:
After speaking with @ekcasey it seems that previously, when we tried to use the syscall package to drop privileges, the issue manifested as failed builds, because files or directories were getting written as the wrong user. To test this, I added the following test to analyzer acceptance:

```
		it("drops privileges", func() {

			output := h.DockerRun(t,
				analyzeImage,
				h.WithFlags(
					"--network", registryNetwork,
					"--env", "CNB_PLATFORM_API="+platformAPI,
				),
				h.WithBash(
					fmt.Sprintf("%s -analyzed /some-dir/some-analyzed.toml %s; ls -al /some-dir",
						ctrPath(analyzerPath),
						noAuthRegistry.RepoName("some-image"),
					),
				),
			)

			h.AssertMatch(t, output, "2222 3333 .+ some-analyzed.toml")
		})
```

...and it fails (non-deterministically) when applied to this branch 😭  (but passes on `main`)

```
    analyzer_test.go:281: Expected: 'Warning: Not restoring cached layer metadata, no cache flag specified.
        Previous image with name "192.168.1.43:49946/some-image" not found
        total 12
        drwxr-xr-x 1 2222 3333 4096 Jul  8 20:26 .
        drwxr-xr-x 1 root root 4096 Jul  8 20:26 ..
        -rw-r--r-- 1 2222 3333    0 Mar 26 15:37 .gitkeep
        -rw-r--r-- 1 root root  247 Jul  8 20:26 some-analyzed.toml
        ' to match regex '2222 3333 .+ some-analyzed.toml'
```

I'm noticing that this branch still uses go 1.15, so maybe rebasing onto main (with 1.16) will help. I put up a [PR](https://github.com/buildpacks/lifecycle/pull/658) to add the mentioned test, hopefully when that is merged it will make this one easier to test.

imjasonh:
> I'm noticing that this branch still uses go 1.15, so maybe rebasing onto main (with 1.16) will help. I put up a [PR](https://github.com/buildpacks/lifecycle/pull/658) to add the mentioned test, hopefully when that is merged it will make this one easier to test.

https://github.com/buildpacks/lifecycle/pull/628 updated CI to use Go 1.15 instead, I've rebased this PR on `main` to pick this up 🤞 

Thanks for adding the acceptance test! 👍

natalieparellano:
Looks like it's working! I've had 20+ green tests, where before the failure rate was approximately 4/5 🎉

ekcasey (approved):

PR #639, opened 2021-06-15
Validate mixins

### LINKED ISSUE #578: [Detect] Fail fast if buildpack.toml mixins are not satisfied
By reading application buildpacks' buildpack.toml, the detector can know what mixins are being statically requested.

By reading stack buildpacks' buildpack.toml, the detector can know what mixins are being provided. The detector can also know what mixins are pre-installed on the stack by reading `analyzed.toml`.

Therefore, if all statically requested mixins are NOT satisfied by stack buildpacks and/or the stack, the detector should fail.

Note that the build could still fail if buildpacks dynamically request mixins in the build plan that are not satisfied. But failing fast seems like the right thing to do.

natalieparellano:
Resolves #578


natalieparellano:
If you set aside the platform -> common changes (which I’m not in love with, because things like platform.LayersMetadata read better than common.LayersMetadata) the most substantive changes are -

detector.go - `MixinValidator` interface
platform/common/analyzed.go - moved structs for analyzed.toml here, added `AnalyzedMetadata` and `AnalyzedMetadataBuilder` interfaces
platform/pre06/analyzed.go - lots of boilerplate code to implement `DecodeAnalyzedMetadataFile` and `NewAnalyzedMetadataBuilder`
platform/v07/analyzed.go - more boilerplate for the new api
platform/common/exit.go - moved codes here to avoid circular dependency
stack.go & stack_test.go - stack validation logic

I actually like how the code reads. My main concern is that it's an awful lot of code to write (for just analyzed.toml). I would love to hear what others think.

natalieparellano:
Received some good feedback in sub team sync - will take a stab at making this simpler.

natalieparellano:
Given that this validation is no longer needed, I'm going to close this PR.

PR #641, opened 2021-06-24
Fix: launcher.go swallows downstream error messages

phil9909:
launcher.go currently swallows error messages from `os.Stat`, `tar.FileInfoHeader` and `os.Open`.
We should change this, as the downstream errors can have different reasons, which are interesting to the enduser. E.g. `os.Stat` could return an error when the file is not found or when permissions are wrong.


jabrown85 (approved):

natalieparellano (approved):
@phil9909 thanks for the PR!

phil9909:
That was fast, thanks :)

PR #646, opened 2021-06-25
Adding registry credentials validations - Issue 572

### LINKED ISSUE #572: [Analyze] Validate registry credentials
As outlined in RFC: https://github.com/buildpacks/rfcs/blob/main/text/0075-move-analyze-phase.md

Analyzer should
* Ensure registry read/write access to `<tags>`, `<cache-image>`
* Ensure registry read access to `<previous-image>`, `<run-image>`

jjbustamante:
## Summary

As part of the [Stack Buildpacks](https://github.com/buildpacks/rfcs/blob/app-mixins/text/0000-app-mixins-stack-buildpacks.md) feature, these validations improve the *fast failing builds* concept. Checking the registry credentials earlier in the process will avoid executing build processes and fail the build during the interaction with the registry.

The following validations were added when the analyzer is executed without the `daemon` flag:

**API version >= 0.7** 

- Ensure registry read/write access to `<tags>`, `<cache-image>`, `<output-image>`
- Ensure registry read access to `<previous-image>`, `<run-image>`

API Version < 0.7 will not be affected by this new valiations

## Documentation

- Should this change be documented?
    - [x] Yes

## Related
Resolves #572

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -33,7 +33,7 @@ var (
 	analyzeImage         = "lifecycle/acceptance/analyzer"
 	analyzerPath         = "/cnb/lifecycle/analyzer"
 	cacheFixtureDir      = filepath.Join("testdata", "analyzer", "cache-dir")
-	daemonOS             string
+	daemonOS, daemonArch string
 	noAuthRegistry       *ih.DockerRegistry
 	authRegistry         *ih.DockerRegistry
 	registryNetwork      string
@@ -45,6 +45,10 @@ func TestAnalyzer(t *testing.T) {
 	info, err := h.DockerCli(t).Info(context.TODO())
 	h.AssertNil(t, err)
 	daemonOS = info.OSType
+	daemonArch = info.Architecture
+	if daemonArch == "x86_64" {
+		daemonArch = "amd64"
+	}
 
 	// Setup registry
 
@@ -75,7 +79,7 @@ func TestAnalyzer(t *testing.T) {
 
 	// Setup test container
 
-	h.MakeAndCopyLifecycle(t, daemonOS, analyzerBinaryDir)
+	h.MakeAndCopyLifecycle(t, daemonOS, daemonArch, analyzerBinaryDir)
 	h.DockerBuild(t,
 		analyzeImage,
 		analyzeDockerContext,
@@ -289,17 +293,21 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		it("drops privileges", func() {
 			h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
+			imageName := authRegistry.RepoName("some-image")
+			authConfig, err := auth.BuildEnvVar(authn.DefaultKeychain, imageName)
+			h.AssertNil(t, err)
 
 			output := h.DockerRun(t,
 				analyzeImage,
 				h.WithFlags(
 					"--network", registryNetwork,
 					"--env", "CNB_PLATFORM_API="+platformAPI,
+					"--env", "CNB_REGISTRY_AUTH="+authConfig,
 				),
 				h.WithBash(
 					fmt.Sprintf("%s -analyzed /some-dir/some-analyzed.toml %s; ls -al /some-dir",
 						ctrPath(analyzerPath),
-						noAuthRegistry.RepoName("some-image"),
+						imageName,
 					),
 				),
 			)
@@ -983,7 +991,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					})
 				})
 
-				when("do have read access to registry", func() {
+				when("no read access", func() {
 					it("throws read error accessing previous image", func() {
 						cmd := exec.Command(
 							"docker", "run", "--rm",
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -204,11 +204,11 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 		when("the provided layers directory isn't writeable", func() {
 			it("recursively chowns the directory", func() {
 				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
-
 				output := h.DockerRun(t,
 					analyzeImage,
 					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI),
-					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s some-image; ls -al /layers", analyzerPath)),
+					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s %s; ls -al /layers", analyzerPath,
+						noAuthRegistry.RepoName("some-image"))),
 				)
 
 				h.AssertMatch(t, output, "2222 3333 .+ \\.")
@@ -245,7 +245,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				execArgs := []string{
 					ctrPath(analyzerPath),
 					"-analyzed", ctrPath("/some-dir/some-analyzed.toml"),
-					"some-image",
+					noAuthRegistry.RepoName("some-image"),
 				}
 
 				h.DockerRunAndCopy(t,
@@ -323,7 +323,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				execArgs := []string{
 					ctrPath(analyzerPath),
 					"-daemon",
-					"some-image",
+					noAuthRegistry.RepoName("some-image"),
 				}
 
 				h.DockerRunAndCopy(t,
@@ -377,7 +377,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						h.WithArgs(
 							ctrPath(analyzerPath),
 							"-daemon",
-							appImage),
+							noAuthRegistry.RepoName(appImage)),
 					)
 
 					assertNoRestoreOfAppMetadata(t, copyDir, output)
@@ -433,12 +433,12 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			when("cache is provided", func() {
 				when("cache image case", func() {
 					when("cache image is in a daemon", func() {
-						var cacheImage string
+						var cacheImage, basicAuth string
 
 						it.Before(func() {
 							metadata := minifyMetadata(t, filepath.Join("testdata", "analyzer", "cache_image_metadata.json"), platform.CacheMetadata{})
 							cacheImage = "some-cache-image-" + h.RandString(10)
-
+							basicAuth, _ = auth.BuildEnvVar(authn.DefaultKeychain, authRegistry.RepoName(cacheImage))
 							cmd := exec.Command(
 								"docker",
 								"build",
@@ -465,12 +465,13 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 								h.WithFlags(append(
 									dockerSocketMount,
 									"--env", "CNB_PLATFORM_API="+platformAPI,
+									"--env", "CNB_REGISTRY_AUTH="+basicAuth,
 								)...),
 								h.WithArgs(
 									ctrPath(analyzerPath),
 									"-daemon",
-									"-cache-image", cacheImage,
-									"some-image",
+									"-cache-image", authRegistry.RepoName(cacheImage),
+									authRegistry.RepoName("some-image"),
 								),
 							)
 
@@ -577,12 +578,14 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 									copyDir,
 									ctrPath("/layers"),
 									analyzeImage,
-									h.WithFlags(
+									h.WithFlags(append(
+										dockerSocketMount,
 										"--network", registryNetwork,
 										"--env", "CNB_PLATFORM_API="+platformAPI,
-									),
+									)...),
 									h.WithArgs(
 										ctrPath(analyzerPath),
+										"-daemon",
 										"-cache-image",
 										noAuthRegCacheImage,
 										"some-image",
@@ -683,7 +686,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					h.WithFlags(
 						"--env", "CNB_PLATFORM_API="+platformAPI,
 					),
-					h.WithArgs(ctrPath(analyzerPath), "some-image"),
+					h.WithArgs(ctrPath(analyzerPath), noAuthRegistry.RepoName("some-image")),
 				)
 
 				assertAnalyzedMetadata(t, filepath.Join(copyDir, "analyzed.toml"))
@@ -1017,25 +1020,24 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						// Don't attempt to remove the image, as it's stored in the test registry, which is ephemeral.
 						// Attempting to remove the image sometimes produces `No such image` flakes.
 
-						it("restores cache metadata", func() {
+						it("throw read/write error", func() {
 							h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0, "Platform API >= 0.7 does not read from the cache")
-							output := h.DockerRunAndCopy(t,
-								containerName,
-								copyDir,
-								ctrPath("/layers"),
+							cmd := exec.Command(
+								"docker", "run", "--rm",
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--name", containerName,
 								analyzeImage,
-								h.WithFlags(
-									"--network", registryNetwork,
-									"--env", "CNB_PLATFORM_API="+platformAPI,
-								),
-								h.WithArgs(
-									ctrPath(analyzerPath),
-									"-cache-image", noAuthRegCacheImage,
-									"some-image",
-								),
-							)
+								ctrPath(analyzerPath),
+								"-cache-image",
+								noAuthRegCacheImage,
+								"some-image",
+							) // #nosec G204
+							output, err := cmd.CombinedOutput()
 
-							assertLogsAndRestoresCacheMetadata(t, copyDir, output)
+							h.AssertNotNil(t, err)
+							expected := "failed to : read/write image " + noAuthRegCacheImage + " from/to the registry"
+							h.AssertStringContains(t, string(output), expected)
 						})
 					})
 				})
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -1094,6 +1094,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			})
 
 			when("called with tag", func() {
+				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not use tag flag")
 				when("have read/write access to registry", func() {
 					var imageName, authConfig string
 					var err error
@@ -1113,12 +1114,11 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 							"--name", containerName,
 							analyzeImage,
 							ctrPath(analyzerPath),
-							"-tag", authRegistry.RepoName("my-tag"),
+							"-tag", noAuthRegistry.RepoName("my-tag"),
 							imageName,
 						) // #nosec G204
 						output, err := cmd.CombinedOutput()
-
-						h.AssertNil(t, err)
+						h.AssertNotNil(t, err)
 						expected := "Previous image with name \"" + imageName + "\" not found"
 						h.AssertStringContains(t, string(output), expected)
 					})
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -442,11 +442,12 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			when("cache is provided", func() {
 				when("cache image case", func() {
 					when("cache image is in a daemon", func() {
-						var cacheImage string
+						var cacheImage, basicAuth string
 
 						it.Before(func() {
 							metadata := minifyMetadata(t, filepath.Join("testdata", "analyzer", "cache_image_metadata.json"), platform.CacheMetadata{})
 							cacheImage = "some-cache-image-" + h.RandString(10)
+							basicAuth, _ = auth.BuildEnvVar(authn.DefaultKeychain, authRegistry.RepoName(cacheImage))
 							cmd := exec.Command(
 								"docker",
 								"build",
@@ -473,12 +474,13 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 								h.WithFlags(append(
 									dockerSocketMount,
 									"--env", "CNB_PLATFORM_API="+platformAPI,
+									"--env", "CNB_REGISTRY_AUTH="+basicAuth,
 								)...),
 								h.WithArgs(
 									ctrPath(analyzerPath),
 									"-daemon",
-									"-cache-image", cacheImage,
-									"some-image",
+									"-cache-image", authRegistry.RepoName(cacheImage),
+									authRegistry.RepoName("some-image"),
 								),
 							)
 
@@ -1047,7 +1049,6 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 						// Don't attempt to remove the image, as it's stored in the test registry, which is ephemeral.
 						// Attempting to remove the image sometimes produces `No such image` flakes.
-
 						it("throws read/write error accessing cache image", func() {
 							h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0, "Platform API >= 0.7 does not read from the cache")
 							cmd := exec.Command(
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -603,7 +603,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						})
 
 						when("no auth registry", func() {
-							var noAuthRegCacheImage string
+							var readOnlyRegCacheImage string
 
 							it.Before(func() {
 								metadata := minifyMetadata(t, filepath.Join("testdata", "analyzer", "cache_image_metadata.json"), platform.CacheMetadata{})
@@ -617,7 +617,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 									"--build-arg", "metadata="+metadata,
 								)
 
-								noAuthRegCacheImage = readOnlyRegistry.RepoName(imageName)
+								readOnlyRegCacheImage = readOnlyRegistry.RepoName(imageName)
 							})
 
 							// Don't attempt to remove the image, as it's stored in the test registry, which is ephemeral.
@@ -639,7 +639,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 										ctrPath(analyzerPath),
 										"-daemon",
 										"-cache-image",
-										noAuthRegCacheImage,
+										readOnlyRegCacheImage,
 										"some-image",
 									),
 								)
@@ -1082,7 +1082,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					})
 
 					when("no auth registry", func() {
-						var noAuthRegCacheImage string
+						var readOnlyRegCacheImage string
 
 						it.Before(func() {
 							metadata := minifyMetadata(t, filepath.Join("testdata", "analyzer", "cache_image_metadata.json"), platform.CacheMetadata{})
@@ -1096,7 +1096,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 								"--build-arg", "metadata="+metadata,
 							)
 
-							noAuthRegCacheImage = readOnlyRegistry.RepoName(imageName)
+							readOnlyRegCacheImage = readOnlyRegistry.RepoName(imageName)
 						})
 
 						// Don't attempt to remove the image, as it's stored in the test registry, which is ephemeral.
@@ -1114,7 +1114,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 								),
 								h.WithArgs(
 									ctrPath(analyzerPath),
-									"-cache-image", noAuthRegCacheImage,
+									"-cache-image", readOnlyRegCacheImage,
 									"some-image",
 								),
 							)
@@ -1132,13 +1132,13 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 								analyzeImage,
 								ctrPath(analyzerPath),
 								"-cache-image",
-								noAuthRegCacheImage,
+								readOnlyRegCacheImage,
 								readOnlyRegistry.RepoName("some-image"),
 							) // #nosec G204
 							output, err := cmd.CombinedOutput()
 
 							h.AssertNotNil(t, err)
-							expected := "failed to : ensure registry read/write access to " + noAuthRegCacheImage
+							expected := "failed to : ensure registry read/write access to " + readOnlyRegCacheImage
 							h.AssertStringContains(t, string(output), expected)
 						})
 					})
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -941,11 +941,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 							"--name", containerName,
 							analyzeImage,
 							ctrPath(analyzerPath),
-<<<<<<< HEAD
 							"-previous-image", "my-previous-image",
-=======
-							"-previous-image","my-previous-image",
->>>>>>> 7397de4 (adding more unit tests and also adding the network flag to verify if the test failures are fixed)
 							"some-image",
 						) // #nosec G204
 						output, err := cmd.CombinedOutput()
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -206,7 +206,8 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
 				output := h.DockerRun(t,
 					analyzeImage,
-					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI),
+					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI,
+						"--network", registryNetwork),
 					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s %s; ls -al /layers", analyzerPath,
 						noAuthRegistry.RepoName("some-image"))),
 				)
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -1094,7 +1094,6 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			})
 
 			when("called with tag", func() {
-				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not use tag flag")
 				when("have read/write access to registry", func() {
 					var imageName, authConfig string
 					var err error
@@ -1114,11 +1113,11 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 							"--name", containerName,
 							analyzeImage,
 							ctrPath(analyzerPath),
-							"-tag", noAuthRegistry.RepoName("my-tag"),
+							"-tag", authRegistry.RepoName("my-tag"),
 							imageName,
 						) // #nosec G204
 						output, err := cmd.CombinedOutput()
-						h.AssertNotNil(t, err)
+						h.AssertNil(t, err)
 						expected := "Previous image with name \"" + imageName + "\" not found"
 						h.AssertStringContains(t, string(output), expected)
 					})
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -206,10 +206,15 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
 				output := h.DockerRun(t,
 					analyzeImage,
-					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI,
-						"--network", registryNetwork),
-					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s %s; ls -al /layers", analyzerPath,
-						noAuthRegistry.RepoName("some-image"))),
+					h.WithFlags(
+						"--env", "CNB_PLATFORM_API="+platformAPI,
+						"--network", registryNetwork,
+					),
+					h.WithBash(
+						fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s %s; ls -al /layers",
+							analyzerPath,
+							noAuthRegistry.RepoName("some-image")),
+					),
 				)
 
 				h.AssertMatch(t, output, "2222 3333 .+ \\.")
@@ -933,7 +938,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				})
 
 				when("do have read access to registry", func() {
-					it("throw read error", func() {
+					it("throws read error accessing previous image", func() {
 						h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not use tag flag")
 						cmd := exec.Command(
 							"docker", "run", "--rm",
@@ -1044,7 +1049,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						// Don't attempt to remove the image, as it's stored in the test registry, which is ephemeral.
 						// Attempting to remove the image sometimes produces `No such image` flakes.
 
-						it("throw read/write error", func() {
+						it("throws read/write error accessing cache image", func() {
 							h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0, "Platform API >= 0.7 does not read from the cache")
 							cmd := exec.Command(
 								"docker", "run", "--rm",
@@ -1090,7 +1095,6 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			})
 
 			when("called with tag", func() {
-				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not use tag flag")
 				when("have read/write access to registry", func() {
 					var imageName, authConfig string
 					var err error
@@ -1100,7 +1104,8 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						h.AssertNil(t, err)
 					})
 
-					it("pass through read/write validation and failed finding image name", func() {
+					it("passes read/write validation and does not error if there is no previous image", func() {
+						h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not use tag flag")
 						cmd := exec.Command(
 							"docker", "run", "--rm",
 							"--network", registryNetwork,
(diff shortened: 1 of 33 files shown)


cmd/lifecycle/analyzer.go
```diff
@@ -120,6 +120,30 @@ func (a *analyzeCmd) Privileges() error {
...
+						return cmd.FailErr(err)
+					}
+				}
+			}
+		} else {
+			err := a.verifyReadWriteAccess(a.platform06.cacheImageTag)
+			if err != nil {
+				return cmd.FailErr(err)
+			}
+		}
```
natalieparellano, line 146: We could reduce the complexity of this code by just iterating over `a.registryImages()`. It might also be worthwhile to make an `a.validateRegistryAccess` function, so that it could also be invoked by the creator.

natalieparellano: cc @jabrown85 for moving logic out of `cmd` package!

cmd/lifecycle/analyzer.go
```diff
@@ -225,3 +249,23 @@ func (a *analyzeCmd) restoresLayerMetadata() bool {
 func (a *analyzeCmd) platformAPIVersionGreaterThan06() bool {
 	return api.MustParse(a.platform.API()).Compare(api.MustParse("0.7")) >= 0
 }
+
+func (a *analyzeCmd) verifyReadAccess(name string) error {
```
jabrown85, line 253:
```suggestion
func (a *analyzeCmd) verifyReadAccess(imageRef string) error {
```

cmd/lifecycle/analyzer.go
jabrown85: Hmm. Yeah. The different scopes required make this kind of annoying to deal with. Maybe introducing a struct (or interface) that the `image` package could rely on. That way creator could pass `createCmd` and analyzer could pass `analyzeArgs `.


```
"github.com/buildpacks/lifecycle/image"

// the proposed struct, but we could also introduce an interface with methods like ReadableImages() WriteableImages() that were implemented in both creator and analyzer cmds/args
type registryInputs struct {
  cacheImagetag string
  previousImage string
  additionalTags []string
}

// in analyzer
if err := image.VerifyRegistryAccess(a); err != nil {
		return cmd.FailErr(err)
}

// in creator
if err := image.VerifyRegistryAccess(c); err != nil {
		return cmd.FailErr(err)
}


VerifyRegistryAccess(regInputs registryInputs) {...}
```

cmd/lifecycle/analyzer.go
jjbustamante: Done in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/d264cb1197460b96220754740341e1a28520c062)

cmd/lifecycle/analyzer.go
```diff
@@ -225,3 +249,23 @@ func (a *analyzeCmd) restoresLayerMetadata() bool {
...
+	if name != "" {
+		img, _ := remote.NewImage(name, a.keychain)
+		if !img.CheckReadAccess() {
+			return errors.New(fmt.Sprintf("read image %s from the registry", name))
+		}
+	}
+	return nil
+}
+
+func (a *analyzeCmd) verifyReadWriteAccess(name string) error {
```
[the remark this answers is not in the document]
jjbustamante: Done in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/d264cb1197460b96220754740341e1a28520c062)

cmd/lifecycle/analyzer.go
```diff
@@ -120,6 +120,30 @@ func (a *analyzeCmd) Privileges() error {
 		if err != nil {
 			return cmd.FailErr(err, "initialize docker client")
 		}
+	} else {
+		if a.platformAPIVersionGreaterThan06() {
```
natalieparellano, line 124: We'll need to do this for the `creator` also (see https://github.com/buildpacks/lifecycle/blob/771ac0b9d0d56cd3baaf28355498a0a636ea6182/cmd/lifecycle/creator.go#L261)

acceptance/analyzer_test.go
```diff
@@ -204,11 +204,12 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
 				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
-
 				output := h.DockerRun(t,
 					analyzeImage,
-					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI),
-					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s some-image; ls -al /layers", analyzerPath)),
+					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI,
+						"--network", registryNetwork),
+					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s %s; ls -al /layers", analyzerPath,
+						noAuthRegistry.RepoName("some-image"))),
```
natalieparellano, line 212:
```suggestion
					h.WithFlags(
						"--env", "CNB_PLATFORM_API="+platformAPI,
						"--network", registryNetwork,
					),
					h.WithBash(
						fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s %s; ls -al /layers",
							analyzerPath,
							noAuthRegistry.RepoName("some-image")),
					),
```

natalieparellano: This was easier to parse in the IDE, but looking at it in GitHub I could not figure out what was going on!

acceptance/analyzer_test.go
```diff
@@ -520,12 +523,14 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
-									h.WithFlags(
+									h.WithFlags(append(
+										dockerSocketMount,
 										"--network", registryNetwork,
 										"--env", "CNB_PLATFORM_API="+platformAPI,
-									),
+									)...),
 									h.WithArgs(
 										ctrPath(analyzerPath),
+										"-daemon",
```
natalieparellano, line 533: Nice catch!

acceptance/analyzer_test.go
```diff
@@ -868,6 +874,27 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						})
 					})
 				})
+
+				when("do have read access to registry", func() {
```
natalieparellano, line 878:
```suggestion
				when("no read access", func() {
```

acceptance/analyzer_test.go
```diff
@@ -868,6 +874,27 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+					it("throw read error", func() {
```
natalieparellano, line 879:
```suggestion
					it("throws read error", func() {
```

acceptance/analyzer_test.go
```diff
@@ -868,6 +874,27 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+						h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not use tag flag")
```
natalieparellano, line 880:
```suggestion
						h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not validate registry access")
```

cmd/lifecycle/analyzer.go
```diff
@@ -120,6 +120,30 @@ func (a *analyzeCmd) Privileges() error {
...
+			}
+			if len(a.additionalTags) > 0 {
+				for _, tag := range a.additionalTags {
+					err := a.verifyReadWriteAccess(tag)
+					if err != nil {
+						return cmd.FailErr(err)
+					}
+				}
+			}
+		} else {
```
yaelharel, line 141: Please correct me if I'm wrong but this is a new validation for platform API <= 0.6, right?
Does it mean that a user that is already successfully using an old platform API version will start have failures after upgrading to the new lifecycle version?

acceptance/analyzer_test.go
```diff
@@ -868,6 +874,27 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--network", registryNetwork,
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--name", containerName,
+							analyzeImage,
+							ctrPath(analyzerPath),
+							"-previous-image", "my-previous-image",
+							"some-image",
```
natalieparellano, line 889: I'm curious why this is throwing a read error - shouldn't the analyzer proceed happily if it can't find the previous image? What about `my-previous-image` and `some-image` makes them not readable?

cmd/lifecycle/analyzer.go
yaelharel, line 141: Please correct me if I'm wrong but this is a new validation for platform API <= 0.6, right?
Does it mean that a user that is already successfully using an old platform API version will start have failures after upgrading to the new lifecycle version?

acceptance/analyzer_test.go
```diff
@@ -868,6 +874,27 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+						) // #nosec G204
+						output, err := cmd.CombinedOutput()
+
+						h.AssertNotNil(t, err)
+						expected := "failed to : read image"
```
natalieparellano, line 894: Could we make it more explicit which image is failing to be read?

acceptance/analyzer_test.go
natalieparellano, line 880: Why do we have this skip? Did this come from copy/paste?

cmd/lifecycle/analyzer.go
natalieparellano: @jabrown85 correct me if I'm wrong but I don't think there's harm in doing this for earlier platform API. It would just mean that a build that was going to fail during export will now fail during analyze.

acceptance/analyzer_test.go
```diff
@@ -960,25 +987,24 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						// Don't attempt to remove the image, as it's stored in the test registry, which is ephemeral.
 						// Attempting to remove the image sometimes produces `No such image` flakes.
 
-						it("restores cache metadata", func() {
+						it("throw read/write error", func() {
```
natalieparellano, line 990: I think you could change the name of the context from `when("no auth registry", func() {` to `when("read only registry", func() {`

natalieparellano:
```suggestion
						it("throws read/write error", func() {
```

acceptance/analyzer_test.go
```diff
@@ -960,25 +987,24 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
-								),
-								h.WithArgs(
-									ctrPath(analyzerPath),
-									"-cache-image", noAuthRegCacheImage,
-									"some-image",
-								),
-							)
+								ctrPath(analyzerPath),
+								"-cache-image",
+								noAuthRegCacheImage,
```
natalieparellano, line 1000: I'm still a bit confused as to why we don't have read / write access to the "no auth" registry... should this be called `readOnlyRegCacheImage`?

acceptance/analyzer_test.go
```diff
@@ -1005,6 +1031,59 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+				when("have read/write access to registry", func() {
+					var imageName, authConfig string
+					var err error
+					it.Before(func() {
+						imageName = authRegistry.RepoName("some-image")
+						authConfig, err = auth.BuildEnvVar(authn.DefaultKeychain, imageName)
+						h.AssertNil(t, err)
+					})
+
+					it("pass through read/write validation and failed finding image name", func() {
```
natalieparellano, line 1045:
```suggestion
					it("passes read/write validation and does not error if there is no previous image", func() {
```

acceptance/analyzer_test.go
```diff
@@ -1005,6 +1031,59 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+						output, err := cmd.CombinedOutput()
+
+						h.AssertNil(t, err)
+						expected := "Previous image with name \"" + imageName + "\" not found"
+						h.AssertStringContains(t, string(output), expected)
+					})
+				})
+
+				when("do not have read/write access to registry", func() {
+					it("throw read/write error", func() {
```
natalieparellano, line 1067:
```suggestion
					it("throws read/write error", func() {
```

acceptance/analyzer_test.go
natalieparellano, line 990:
```suggestion
						it("throws read/write error", func() {
```

cmd/lifecycle/analyzer.go
```diff
@@ -120,6 +120,30 @@ func (a *analyzeCmd) Privileges() error {
 		if err != nil {
 			return cmd.FailErr(err, "initialize docker client")
 		}
+	} else {
+		if a.platformAPIVersionGreaterThan06() {
+			err := a.verifyReadAccess(a.previousImage)
+			if err != nil {
+				return cmd.FailErr(err)
+			}
+			err = a.verifyReadAccess(a.runImageRef)
```
natalieparellano, line 129: Is this tested anywhere?

cmd/lifecycle/analyzer.go
natalieparellano: Ohhh I forgot that `registryImages` could include images that you just need read access to, and ones that you need read/write access to... introducing a new struct might be a helpful way to organize this.

acceptance/analyzer_test.go
```diff
@@ -1005,6 +1031,59 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+						h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not use tag flag")
+						cmd := exec.Command(
+							"docker", "run", "--rm",
+							"--network", registryNetwork,
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--name", containerName,
+							analyzeImage,
+							ctrPath(analyzerPath),
+							"-tag", noAuthRegistry.RepoName("my-tag"),
+							noAuthRegistry.RepoName("some-image"),
```
natalieparellano, line 1077: We should also be verifying read/write access to `some-image`.

cmd/lifecycle/analyzer.go
natalieparellano: Oh actually...  maybe I'm wrong - creating the app image would succeed, but then we'd fail to create the cache. So the build would be "failed" but you'd still have an image!

natalieparellano: Maybe just to be on the safe side, we should not validate registry access up front for earlier platform APIs. Curious to know what others think.

natalieparellano:
Relates to https://github.com/buildpacks/lifecycle/pull/647

@jjbustamante added a couple comments and suggestions. I think adding `creator` support and resolving the question of what to do for earlier platform APIs are the most important to figure out right now.

acceptance/analyzer_test.go
jjbustamante: Yeah, it's a copy pace issue 😞

jjbustamante: I removed it in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/04d939f1fc0b51330b0919208870efe1317f1775)

cmd/lifecycle/analyzer.go
jjbustamante: @yaelharel , @natalieparellano right now the PR is running the validation with the following logic:
**API >= 0.7**
-    read access <run-image>
-    read access <previous-image>
-    read/write access to <tags>

**API < 0.7**
-    read/write access to <cache-image>

The <cache-image> tag was removed from earlier 0.7 version but there is a PR open to add it back, right? I think if we merge that one first and the <cache-image> flag is back in version 0.7, then we can just apply the new validation for this version and avoid breaking any previous functionality.

yaelharel: I agree with @natalieparellano. I think that we should just add the validation for new platform API (>= 0.7).

cmd/lifecycle/analyzer.go
jjbustamante: @jabrown85 , @natalieparellano  I found useful the idea with the `interface` because it allows me to simplify the validation logic. It will look like this:

natalieparellano:
@jjbustamante if you rebase onto `main` you'll get https://github.com/buildpacks/lifecycle/pull/647

natalieparellano:
Potentially related spec conversation: https://github.com/buildpacks/spec/pull/234#discussion_r661845162

natalieparellano:
@jjbustamante is it possible to rebase again so that the diff doesn't show commits that are already on main?

cmd/lifecycle/analyzer.go
```diff
@@ -207,21 +220,29 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
...
+	return api.MustParse(a.platform.API()).Compare(api.MustParse("0.7")) >= 0
 }
 
 func (a *analyzeCmd) restoresLayerMetadata() bool {
 	return !a.platformAPIVersionGreaterThan06()
 }
 
-func (a *analyzeCmd) platformAPIVersionGreaterThan06() bool {
-	return api.MustParse(a.platform.API()).Compare(api.MustParse("0.7")) >= 0
+func (a *analyzeCmd) ReadableImages() []string {
```
natalieparellano, line 231:
```suggestion
func (a *analyzeCmd) ReadableRegistryImages() []string {
```

natalieparellano: (this seems more accurate)

cmd/lifecycle/analyzer.go
```diff
@@ -207,21 +220,29 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
...
+			readableImages = append(readableImages, a.previousImageRef, a.runImageRef)
+		}
+	}
+	return readableImages
+}
+func (a *analyzeCmd) WriteableImages() []string {
+	var writeableImages []string
+	if a.platformAPIVersionGreaterThan06() {
+		if !a.useDaemon {
+			writeableImages = append(writeableImages, append([]string{a.cacheImageRef}, a.additionalTags...)...)
```
natalieparellano, line 244: You can use a cache image when building to a daemon (in face, a cache image must always be in a registry, a cache image on the daemon is ignored). Therefore I think this is slightly inaccurate (see `registryImages()`).

Actually it would be great if we could dedup this logic with `registryImages()` - `registryImages()` could be the concatenation of `ReadableImages()` + `WriteableImages()`.

cmd/lifecycle/analyzer.go
```diff
@@ -207,21 +220,29 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
...
 
 func (a *analyzeCmd) restoresLayerMetadata() bool {
 	return !a.platformAPIVersionGreaterThan06()
 }
 
-func (a *analyzeCmd) platformAPIVersionGreaterThan06() bool {
-	return api.MustParse(a.platform.API()).Compare(api.MustParse("0.7")) >= 0
+func (a *analyzeCmd) ReadableImages() []string {
+	var readableImages []string
+	if a.platformAPIVersionGreaterThan06() {
```
natalieparellano, line 233: I think it would make more sense to have the platform version guarding the call to `VerifyRegistryAccess`. Conceivably, we might want to know the readable registry images for earlier platform api (see https://github.com/buildpacks/lifecycle/pull/646/files#r666182517).

image/image.go
```diff
@@ -28,3 +37,44 @@ func ValidateDestinationTags(daemon bool, repoNames ...string) error {
 
 	return nil
 }
+
+func VerifyRegistryAccess(regInputs RegistryInputs, keychain authn.Keychain) error {
+	if len(regInputs.ReadableImages()) > 0 {
```
natalieparellano, line 42: I don't think this guard is necessary. If the array is empty, the loop will just be a no-op.

image/image.go
```diff
@@ -28,3 +37,44 @@ func ValidateDestinationTags(daemon bool, repoNames ...string) error {
...
+		}
+	}
+	return nil
+}
+
+func verifyReadAccess(imageRef string, keychain authn.Keychain) error {
+	if imageRef != "" {
+		img, _ := remote.NewImage(imageRef, keychain)
+		if !img.CheckReadAccess() {
+			return errors.New(fmt.Sprintf("read image %s from the registry", imageRef))
```
natalieparellano, line 66:
```suggestion
			return errors.New(fmt.Sprintf("ensure read access to %s", imageRef))
```
(nit: this seems closer to what we're actually doing...)

image/image.go
```diff
@@ -28,3 +37,44 @@ func ValidateDestinationTags(daemon bool, repoNames ...string) error {
...
+		}
+	}
+	return nil
+}
+
+func verifyReadWriteAccess(imageRef string, keychain authn.Keychain) error {
+	if imageRef != "" {
+		img, _ := remote.NewImage(imageRef, keychain)
+		if !img.CheckReadWriteAccess() {
+			return errors.New(fmt.Sprintf("read/write image %s from/to the registry", imageRef))
```
natalieparellano, line 76:
```suggestion
			return errors.New(fmt.Sprintf("ensure registry read/write access to %s", imageRef))
```

cmd/lifecycle/creator.go
```diff
@@ -260,12 +274,35 @@ func (c *createCmd) Exec() error {
...
+			readableImages = append(readableImages, c.previousImageRef, c.runImageRef)
+		}
+	}
+	return readableImages
+}
+func (c *createCmd) WriteableImages() []string {
+	var writeableImages []string
+	if c.platformAPIVersionGreaterThan06() {
+		if !c.useDaemon {
+			writeableImages = append(writeableImages, append([]string{c.cacheImageRef}, c.additionalTags...)...)
```
natalieparellano, line 304: Same comment as https://github.com/buildpacks/lifecycle/pull/646/files#r666182517

acceptance/analyzer_test.go
```diff
@@ -980,6 +1068,26 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+							cmd := exec.Command(
+								"docker", "run", "--rm",
+								"--network", registryNetwork,
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								"--name", containerName,
+								analyzeImage,
+								ctrPath(analyzerPath),
+								"-cache-image",
+								noAuthRegCacheImage,
+								noAuthRegistry.RepoName("some-image"),
```
natalieparellano, line 1083:
```suggestion
								readOnlyRegistry.RepoName("some-image"),
```

acceptance/analyzer_test.go
natalieparellano:
```suggestion
								readOnlyRegCacheImage,
```

acceptance/analyzer_test.go
natalieparellano: We could also change the context ca. line 1030 to say `read only registry`...

natalieparellano:
I really like the new direction! I think we need to fix the logic for "readable images" so that it includes cache image even when exporting to a daemon. The acceptance tests seem to be missing a few cases still, but I don't think we should sweat over exhaustive coverage here. The read/write ones seem most important to cover.

cmd/lifecycle/analyzer.go
jjbustamante: I refactored the code in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/1ab080726bdfd9210f0b21fe75dae32a42697ef2)

cmd/lifecycle/analyzer.go
jjbustamante: I added the changes in the creator in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/fbdb97d0370bcbe1e9bdc6499c40839ee1255fc3)

cmd/lifecycle/analyzer.go
jjbustamante: In the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/1ab080726bdfd9210f0b21fe75dae32a42697ef2) I added the validations only for API >= 0.7

cmd/lifecycle/analyzer.go
jjbustamante: The `VerifyRegistryAccess` is on the `image` package, I will need to expose the value in the interface but I do not like that approach. I can move the validation like this 

```
func (a *analyzeCmd) Privileges() error {
        // code
	if a.platformAPIVersionGreaterThan06() {
		if err := image.VerifyRegistryAccess(a, a.keychain); err != nil {
			return cmd.FailErr(err)
		}
	}
         // code
}
```

cmd/lifecycle/analyzer.go
jjbustamante: I like the refactor for the `registryImages()` method. Looking at that one, here is my proposal

```
func (aa *analyzeArgs) registryImages() []string {
	var registryImages []string
	registryImages = append(append(registryImages, aa.ReadableRegistryImages()...), aa.WriteableRegistryImages()...)
	if !aa.useDaemon {
		registryImages = append(registryImages, aa.outputImageRef)
	}
	return registryImages
}
```

The only image I can see we are not planning to run read/write validation is  `outputImageRef`

natalieparellano: Actually I think `outputImageRef` is still important to test - it need not exist, but we should still have write access to its registry! It seems to me that `aa.WriteableRegistryImages()` should include it.

cmd/lifecycle/analyzer.go
natalieparellano: That looks great to me!

cmd/lifecycle/analyzer.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/5c900400a2e89ae0d9f8a8d0c02243dec574ceba)

cmd/lifecycle/analyzer.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/5c900400a2e89ae0d9f8a8d0c02243dec574ceba)

cmd/lifecycle/analyzer.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/5c900400a2e89ae0d9f8a8d0c02243dec574ceba)

image/image.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/5c900400a2e89ae0d9f8a8d0c02243dec574ceba)

image/image.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/5c900400a2e89ae0d9f8a8d0c02243dec574ceba)

image/image.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/5c900400a2e89ae0d9f8a8d0c02243dec574ceba)

cmd/lifecycle/creator.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/5c900400a2e89ae0d9f8a8d0c02243dec574ceba)

acceptance/analyzer_test.go
jjbustamante: as you can see the -image-name and -previous-image are actually not pointing to the registry, in this case it is using docker.io as default host url and it is actually returning a 403 respond. basically that's why it has `no read` access

acceptance/analyzer_test.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/c2e71adc20406f06b4f2d3d9b2626778faa93183)

acceptance/analyzer_test.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/2228fae52b7d0bbeeefe01e882486b1d26b30f1c)

acceptance/analyzer_test.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/ba8a18c2765ed0444640debec0d3067f5a555ebb)

cmd/lifecycle/creator.go
```diff
@@ -280,3 +285,26 @@ func (c *createCmd) registryImages() []string {
...
+	return api.MustParse(c.platform.API()).Compare(api.MustParse("0.7")) >= 0
+}
+
+func (c *createCmd) ReadableRegistryImages() []string {
+	var readableImages []string
+	if c.platformAPIVersionGreaterThan06() {
+		if !c.useDaemon {
+			readableImages = append(readableImages, c.previousImageRef, c.runImageRef)
+		}
+	}
```
natalieparellano, line 299:
```suggestion
		if !c.useDaemon {
			readableImages = append(readableImages, c.previousImageRef, c.runImageRef)
		}
```

cmd/lifecycle/creator.go
```diff
@@ -280,3 +285,26 @@ func (c *createCmd) registryImages() []string {
...
+	return readableImages
+}
+func (c *createCmd) WriteableRegistryImages() []string {
+	var writeableImages []string
+	if c.platformAPIVersionGreaterThan06() {
+		if !c.useDaemon {
+			writeableImages = append(writeableImages, append([]string{c.cacheImageRef}, c.additionalTags...)...)
+		}
+	}
```
natalieparellano, line 308:
```suggestion
		if !c.useDaemon {
			writeableImages = append(writeableImages, append([]string{c.cacheImageRef}, c.additionalTags...)...)
	}
```

natalieparellano: This way you could change `registryImages()` to be `ReadableRegistryImages()` + `WriteableRegistryImages()` and avoid duplicating this logic.

cmd/lifecycle/analyzer.go
```diff
@@ -224,3 +223,21 @@ func (a *analyzeCmd) platformAPIVersionGreaterThan06() bool {
 func (a *analyzeCmd) restoresLayerMetadata() bool {
 	return !a.platformAPIVersionGreaterThan06()
 }
+
+func (aa *analyzeArgs) ReadableRegistryImages() []string {
+	var readableImages []string
+	if !aa.useDaemon {
+		readableImages = append(readableImages, aa.previousImageRef, aa.runImageRef)
+	}
+	return readableImages
```
natalieparellano, line 232: I wonder if it's worth making a small utility function to filter out empty strings from the list returned. I see that you have [this check](https://github.com/buildpacks/lifecycle/pull/646/files#diff-c0efa96b114546b626552e11d133bf72b7e0fe50d145eb4a750ca50dfe208b2aR58) but it would be nice to not have to worry about this every time the method is invoked.

acceptance/analyzer_test.go
```diff
@@ -206,13 +206,28 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 		})
 
 		when("the provided layers directory isn't writeable", func() {
+			var imageName, authConfig string
+			var err error
```
natalieparellano, line 210: nit: You could move the `err` declaration inside the `it.Before`, as it's not needed in other scopes.

acceptance/analyzer_test.go
```diff
@@ -206,13 +206,28 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+			it.Before(func() {
+				imageName = authRegistry.RepoName("some-image")
+				authConfig, err = auth.BuildEnvVar(authn.DefaultKeychain, imageName)
+				h.AssertNil(t, err)
+			})
```
natalieparellano, line 215: I feel like there's some refactoring we could do to avoid having to repeat this for every test (like moving it to a parent context). But we don't have to do that as part of this PR.

natalieparellano: Could we just use "read only registry" for this particular test? I think any test that was using "some-image" (or another non-existent image) before could use the read only registry.

acceptance/analyzer_test.go
natalieparellano: Ohh I see... so that must mean there is some image on docker.io called "some-image" that we don't have access to? I would rather not be dependent on the state of Docker Hub for these tests to work (what if someone makes that image public? It's unlikely to happen, but still...) - could we make them images in the test auth registry?

cmd/lifecycle/analyzer.go
jjbustamante: @natalieparellano something like [this](https://play.golang.org/p/UfbbmgoPUIm)?

natalieparellano: That would work!

acceptance/analyzer_test.go
jjbustamante: @natalieparellano I created the following [PR](https://github.com/buildpacks/imgutil/pull/125) in imgutil to help me to solve the issue. The main problem is our current implementation of the dockerRegistry only have two behaviors: **read only** or **basic auth,** in this case `some-image` requires read/write access and `previous-image` requires read access, I can't handle these behaviors. 

I implemented a new custom handler in the docker registry to do what I need, it is not perfect, I think we can refactor to remove duplicated code but I would like you to take a look and give me your thoughts. Using the changes from the imgutil I can add the following code to this test.

```
when("no read access", func() {
					var customRegistry *ih.DockerRegistry
					var previousImage string

					it.Before(func() {
						customDockerConfigDir, err := ioutil.TempDir("", "test.docker.config.custom.dir")
						h.AssertNil(t, err)
						defer os.RemoveAll(customDockerConfigDir)

						var customPrivileges = make(map[string]ih.ImagePrivileges)
						customPrivileges["some-image"] = ih.ImagePrivileges{Readable: true, Writeable: true}
						customPrivileges["my-previous-image"] = ih.ImagePrivileges{Readable: false, Writeable: false}

						sharedRegHandler := registry.New(registry.Logger(log.New(ioutil.Discard, "", log.Lshortfile)))
						customRegistry = ih.NewDockerRegistry(ih.WithAuth(customDockerConfigDir), ih.WithSharedHandler(sharedRegHandler),
							ih.WithCustomPrivileges(customPrivileges))

						customRegistry.Start(t)
						defer customRegistry.Stop(t)

						imageName = customRegistry.RepoName("some-image")
						authConfig, err = auth.BuildEnvVar(authn.DefaultKeychain, imageName)
						previousImage = customRegistry.RepoName("my-previous-image")
						h.AssertNil(t, err)
					})

					it("throws read error accessing previous image", func() {
						cmd := exec.Command(
							"docker", "run", "--rm",
							"--network", registryNetwork,
							"--env", "CNB_PLATFORM_API="+platformAPI,
							"--env", "CNB_REGISTRY_AUTH="+authConfig,
							"--name", containerName,
							analyzeImage,
							ctrPath(analyzerPath),
							"-previous-image", previousImage,
							imageName,
						) // #nosec G204
						output, err := cmd.CombinedOutput()

						h.AssertNotNil(t, err)
						expected := "failed to : ensure registry read access to " + previousImage
						h.AssertStringContains(t, string(output), expected)
					})
				})
```

cmd/lifecycle/creator.go
jjbustamante: Ooops I forgot to do this here in the creater.

acceptance/analyzer_test.go
natalieparellano: I love this idea!!! This will make the tests so much simpler & easier to understand!

My one nit: could we bake the privileges & image type into the name? So something like `some-readable-writeable-app-image` or `some-readable-cache-image` which I **think** would make the tests easier to follow.

Taking a look at the imgutil PR...

cmd/lifecycle/analyzer.go
```diff
@@ -224,3 +223,21 @@ func (a *analyzeCmd) platformAPIVersionGreaterThan06() bool {
...
+}
+func (aa *analyzeArgs) WriteableRegistryImages() []string {
+	var writeableImages []string
+	if aa.cacheImageRef != "" {
+		writeableImages = append(writeableImages, aa.cacheImageRef)
+	}
```
yaelharel, line 238: Why do you only check that this variable is an empty string?
Unless I'm missing something you could either check that every value that you add to the `writableImages` array isn't an empty string (and the same for the `readableImages` array in `ReadableRegistryImages`) and delete the empty string condition in `verifyReadAccess` and `verifyReadWriteAccess` (in `image.go`), or just add the `aa.cacheImageRef` even when it's an empty string.

cmd/lifecycle/analyzer.go
```diff
@@ -224,3 +223,21 @@ func (a *analyzeCmd) platformAPIVersionGreaterThan06() bool {
...
+	if !aa.useDaemon {
+		writeableImages = append(writeableImages, append([]string{aa.outputImageRef}, aa.additionalTags...)...)
```
yaelharel, line 240: Should we add the `outputImageRef` to the PR description?

cmd/lifecycle/analyzer.go
yaelharel: I'm sorry, I've totally missed [this conversation](https://github.com/buildpacks/lifecycle/pull/646/files#r668830740) 😊

natalieparellano: Relevant: https://github.com/buildpacks/lifecycle/pull/646#discussion_r668830740

cmd/lifecycle/creator.go
```diff
@@ -280,3 +285,26 @@ func (c *createCmd) registryImages() []string {
 	}
 	return registryImages
 }
+
+func (c *createCmd) platformAPIVersionGreaterThan06() bool {
```
yaelharel, line 289: Not related to your PR, but should we rename all of these functions to be something like:
`platformAPIVersionIsAtLeast07`?
I think it's a bit weird that we're talking about version `0.7` and the function name indicates `0.6`.
Also, someone can theoretically think that version `0.6.1` counts since is greater than `0.6`.
@natalieparellano @jabrown85

cmd/lifecycle/creator.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/8a82acf4f285ed83dbfb346b7f65f106b548d8aa)

cmd/lifecycle/creator.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/8a82acf4f285ed83dbfb346b7f65f106b548d8aa)

acceptance/analyzer_test.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/8a82acf4f285ed83dbfb346b7f65f106b548d8aa)

cmd/lifecycle/analyzer.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/8a82acf4f285ed83dbfb346b7f65f106b548d8aa)

natalieparellano:
@jjbustamante the current failure is interesting: https://github.com/buildpacks/lifecycle/pull/646/checks?check_run_id=3128273090#step:6:5210

The `no stack metadata found at path` message I think is a red herring (it occurs on linux as well). The real error appears below it: `ERROR: failed to : ensure registry read access to 10.1.0.118:50026/some-app-image-unpchuiall`. I'm not sure why that is occurring or why it's windows-only.

cmd/lifecycle/creator.go
```diff
@@ -276,14 +276,7 @@ func (c *createCmd) Exec() error {
...
 	var registryImages []string
-	if c.cacheImageRef != "" {
-		registryImages = append(registryImages, c.cacheImageRef)
-	}
-	if !c.useDaemon {
-		registryImages = append(registryImages, append([]string{c.outputImageRef}, c.additionalTags...)...)
-		registryImages = append(registryImages, c.runImageRef, c.previousImageRef)
-	}
-	return registryImages
+	return append(append(registryImages, c.ReadableRegistryImages()...), c.WriteableRegistryImages()...)
```
yaelharel, line 279: I think it'll be clearer to not use a nested `append`.
I would either split it to the following two lines:
```
append(registryImages, c.ReadableRegistryImages()...)
return append(registryImages, c.WriteableRegistryImages()...)
```
or I would append the output of `c.ReadableRegistryImages()` and `c.WriteableRegistryImages()` and then append them together to `registryImages`.

cmd/lifecycle/creator.go
```diff
@@ -292,19 +285,16 @@ func (c *createCmd) platformAPIVersionGreaterThan06() bool {
...
 }
 func (c *createCmd) WriteableRegistryImages() []string {
 	var writeableImages []string
-	if c.platformAPIVersionGreaterThan06() {
-		if !c.useDaemon {
-			writeableImages = append(writeableImages, append([]string{c.cacheImageRef}, c.additionalTags...)...)
-		}
+	writeableImages = appendNotEmpty(writeableImages, c.cacheImageRef)
+	if !c.useDaemon {
+		writeableImages = append(writeableImages, appendNotEmpty([]string{c.outputImageRef}, c.additionalTags...)...)
```
yaelharel, line 297: Same comment as https://github.com/buildpacks/lifecycle/pull/646/commits/56d14b18dfe67b24d46995d21395e94847e3d59f#r674434602.
I think that a nested `append` is confusing for the reader.

cmd/lifecycle/main.go
```diff
@@ -104,3 +104,12 @@ func initCache(cacheImageTag, cacheDir string, keychain authn.Keychain) (lifecyc
 	}
 	return cacheStore, nil
 }
+
+func appendNotEmpty(slice []string, elems ...string) []string {
+	for _, v := range elems {
+		if v != "" {
+			slice = append(slice, v)
```
yaelharel, line 111: When calling `append`, if `slice` array is too small to fit all the given values, a bigger array will be allocated. Due to this performance issue, theoretically it would be better to know how many elements we're going to append to `slice`, increase its size (if needed) and then `append` the elements.
Currently we're appending a small number of elements on each call so I'm not sure it's worth such change, but I think we should at least keep it in mind.

acceptance/analyzer_test.go
```diff
@@ -36,9 +36,17 @@ var (
...
 	authRegistry         *ih.DockerRegistry
+	customRegistry       *ih.DockerRegistry
 	registryNetwork      string
 )
 
+const (
+	readWriteImage = "image-readable-writable"
+	onlyReadImage  = "image-readable"
+	onlyWriteImage = "image-writable"
+	noAccessImage  = "noAccessImage"
```
yaelharel, line 47: The same comment as https://github.com/buildpacks/imgutil/pull/125/files#r671486105.

acceptance/analyzer_test.go
```diff
@@ -207,8 +231,8 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("the provided layers directory isn't writeable", func() {
 			var imageName, authConfig string
-			var err error
 			it.Before(func() {
+				var err error
```
yaelharel, line 235: I think it's better to move this line to the line before the usage of this variable, meaning, swap lines 235 and 236.

jjbustamante:
> @jjbustamante the current failure is interesting: https://github.com/buildpacks/lifecycle/pull/646/checks?check_run_id=3128273090#step:6:5210
> 
> The `no stack metadata found at path` message I think is a red herring (it occurs on linux as well). The real error appears below it: `ERROR: failed to : ensure registry read access to 10.1.0.118:50026/some-app-image-unpchuiall`. I'm not sure why that is occurring or why it's windows-only.

Checking the error

```
 docker.go:57: Expected nil: failed to execute command: [docker run --env CNB_PLATFORM_API=0.7 --env CNB_REGISTRY_AUTH={} --network default --name test-container-qbpkuojwhv lifecycle/acceptance/analyzer c:\cnb\lifecycle\analyzer -previous-image 10.1.0.118:50026/some-app-image-unpchuiall 10.1.0.118:50026/some-app-image-endizynstn], exit status 1, , no stack metadata found at path 'c:\cnb\stack.toml'
5211
        ERROR: failed to : ensure registry read access to 10.1.0.118:50026/some-app-image-unpchuiall
```

I can see `--env CNB_REGISTRY_AUTH={} ` is empty, probably that's the reason

cmd/lifecycle/creator.go
jjbustamante: I used two appends in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/7b1d75feefd76dc76fb94799ed2756eeda7da14c)

cmd/lifecycle/creator.go
jjbustamante: I used two appends in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/7b1d75feefd76dc76fb94799ed2756eeda7da14c)

acceptance/analyzer_test.go
jjbustamante: Addressed in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/7b1d75feefd76dc76fb94799ed2756eeda7da14c)

cmd/lifecycle/analyzer.go
jjbustamante: Done

cmd/lifecycle/creator.go
jjbustamante: I agreed but there are some other references in the code that mentioned **platform06** I suggest we should do a full review of those terms and renaming them with a global strategy and avoid doing the change in just one place and keep the other ones.

cmd/lifecycle/analyzer.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/lifecycle/pull/646/commits/c76fda95682b294ec1cf7747f7d9203fa2a6f8e8)

cmd/lifecycle/creator.go
yaelharel: Of course! This is why I wrote that it's unrelated to your PR 😊

natalieparellano (approved):
Looks good! We talked about further refactoring of the analyzer acceptance tests (to make use of the custom registry, and also to reduce unnecessary duplication). But we would like to do that after merging https://github.com/buildpacks/lifecycle/pull/668/files, to make everyone's life easier.

PR #647, opened 2021-06-28
Updates for platform 0.7

natalieparellano:
Please see associated spec PR: https://github.com/buildpacks/spec/pull/234

### DIFF
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -29,41 +29,39 @@ type analyzeCmd struct {
 }
 
 type analyzeArgs struct {
+	cacheImageTag string //nolint - Platform API >= 0.7
 	imageName     string
 	layersDir     string
-	orderPath     string //nolint - Platform API >= 0.7
 	previousImage string //nolint - Platform API >= 0.7
 	runImageRef   string //nolint - Platform API >= 0.7
 	stackPath     string //nolint - Platform API >= 0.7
 	useDaemon     bool
 
-	platform06     analyzeArgsPlatform06
 	additionalTags cmd.StringSlice        //nolint Platform API >= 0.7
 	docker         client.CommonAPIClient // construct if necessary before dropping privileges
 	keychain       authn.Keychain
 	platform       cmd.Platform
+	platform06     analyzeArgsPlatform06
 }
 
 type analyzeArgsPlatform06 struct {
-	cacheDir      string
-	cacheImageTag string
-	groupPath     string
-	skipLayers    bool
-	cache         lifecycle.Cache
-	group         buildpack.Group
+	cacheDir   string
+	groupPath  string
+	skipLayers bool
+	cache      lifecycle.Cache
+	group      buildpack.Group
 }
 
 func (a *analyzeCmd) DefineFlags() {
 	cmd.FlagAnalyzedPath(&a.analyzedPath)
+	cmd.FlagCacheImage(&a.cacheImageTag)
 	cmd.FlagLayersDir(&a.layersDir)
 	if a.platformAPIVersionGreaterThan06() {
-		cmd.FlagOrderPath(&a.orderPath)
 		cmd.FlagPreviousImage(&a.previousImage)
 		cmd.FlagRunImage(&a.runImageRef)
 		cmd.FlagStackPath(&a.stackPath)
 		cmd.FlagTags(&a.additionalTags)
 	} else {
-		cmd.FlagCacheImage(&a.platform06.cacheImageTag)
 		cmd.FlagCacheDir(&a.platform06.cacheDir)
 		cmd.FlagGroupPath(&a.platform06.groupPath)
 		cmd.FlagSkipLayers(&a.platform06.skipLayers)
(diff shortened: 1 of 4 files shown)


cmd/lifecycle/analyzer.go
```diff
@@ -15,55 +15,54 @@ import (
...
+	"github.com/buildpacks/lifecycle/image"
 	"github.com/buildpacks/lifecycle/platform"
 	"github.com/buildpacks/lifecycle/priv"
 )
 
 type analyzeCmd struct {
 	//flags: inputs
 	analyzeArgs
-	uid, gid int
+	stackPath string
```
natalieparellano, line 26: Not needed by creator, so moving it here from `analyzeArgs`

cmd/lifecycle/analyzer.go
```diff
@@ -15,55 +15,54 @@ import (
...
-	previousImage string //nolint - Platform API >= 0.7
-	runImageRef   string //nolint - Platform API >= 0.7
-	stackPath     string //nolint - Platform API >= 0.7
-	useDaemon     bool
-
-	platform06     analyzeArgsPlatform06
-	additionalTags cmd.StringSlice        //nolint Platform API >= 0.7
+	cacheImageRef    string
+	layersDir        string
+	outputImageRef   string
```
natalieparellano, line 36: Renamed from `imageName` ...now that analyzer can care about both output & previous images, it was too confusing.

cmd/lifecycle/analyzer.go
```diff
@@ -77,21 +76,26 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
...
 	}
+
 	if args[0] == "" {
 		return cmd.FailErrCode(errors.New("image argument is required"), cmd.CodeInvalidArgs, "parse arguments")
 	}
-	a.imageName = args[0]
-
-	if a.previousImage == "" {
-		a.previousImage = a.imageName
-	}
```
natalieparellano, line 87: Re-ordered this code so that it's easier to compare with creator.

cmd/lifecycle/analyzer.go
```diff
@@ -77,21 +76,26 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
...
 	}
 
+	if a.previousImageRef == "" {
+		a.previousImageRef = a.outputImageRef
+	}
+
+	if err := image.ValidateDestinationTags(a.useDaemon, append(a.additionalTags, a.outputImageRef)...); err != nil {
+		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "validate image tag(s)")
+	}
+
```
natalieparellano, line 98: This is something that the exporter currently does. I wonder if we want to move it here in the spirit of failing fast.

cmd/lifecycle/analyzer.go
```diff
@@ -130,6 +136,17 @@ func (a *analyzeCmd) Privileges() error {
 	return nil
 }
 
+func (aa *analyzeArgs) registryImages() []string {
```
natalieparellano, line 139: Needed to change the receiver to `analyzeArgs` (from `analyzeCmd`) so I moved it up in the file.

cmd/lifecycle/analyzer.go
natalieparellano: The spec has this wording for exporter:

> If `<daemon>` is false and more than one `<image>` is provided they MUST refer to the same registry

This could be added to the associated spec PR.

cmd/lifecycle/creator.go
```diff
@@ -154,12 +161,16 @@ func (c *createCmd) Exec() error {
...
-			layersDir: c.layersDir,
-			platform:  c.platform,
-			useDaemon: c.useDaemon,
-			docker:    c.docker,
+			additionalTags:   c.additionalTags,
+			cacheImageRef:    c.cacheImageRef,
+			docker:           c.docker,
+			keychain:         c.keychain,
+			layersDir:        c.layersDir,
+			outputImageRef:   c.outputImageRef,
```
natalieparellano, line 169: This might not be needed... it is only used to validate registry access and I think that happens before `analyze()` is called. But at the same time, it doesn't hurt to have it and it might save us from creating bugs later.

natalieparellano: See https://github.com/buildpacks/lifecycle/pull/646

cmd/lifecycle/analyzer.go
jabrown85: I like this!

jabrown85 (approved):
Thanks for all the context. This looks pretty good to me. The renames help readability quite a bit!

Do you think the existing test coverage is enough? I know there are a few additional validations introduced here, but they come from other phases and are now happening earlier. 

We also brought back `cache-image` flag and maybe some of the tests can be not-skipped that were skipped before?

What do you think?

natalieparellano:
Good point about test coverage! I think we could add some acceptance tests for (1) ensuring destination tags are all on the same registry, and (2) that we resolve the correct run image. I think the tests for cache-image would need to come in or after #646 , since we can't assert on the effect of this flag without actually using it. The current skip language still seems relevant: `h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0, "Platform API >= 0.7 does not read from the cache")`.

natalieparellano:
On second thought, I think testing that we've resolved the run image is not testable at the moment for the same reason cache-image is not testable (it has no effect until #646).

cmd/lifecycle/analyzer.go
natalieparellano: Added to spec PR: https://github.com/buildpacks/spec/pull/234/commits/5718cbad8c3bd21f5f0637a72f0bdf70ef19bb8a

yaelharel (approved):
I really like that you renamed some of the variables! It makes the code more readable!

PR #648, opened 2021-06-29
Adds acceptance tests for builder

### LINKED ISSUE #466: [Builder] Acceptance Tests
Add acceptances tests for [builder](https://github.com/buildpacks/spec/blob/main/platform.md#builder).

haliliceylan:
Resolves #466


haliliceylan:
For next step, i will add create necessary toml files to use to hello_world buildpack. I will test some buildpack.toml parsing functions here: 

https://github.com/buildpacks/lifecycle/blob/1d7a850cf03a1e16b6be5f871419d41091da4f29/cmd/lifecycle/builder.go#L72-L74

acceptance/builder_test.go
```diff
@@ -0,0 +1,184 @@
...
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "DELETE_FILE=/layers/plan.toml",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				h.AssertNotNil(t, err)
+				expected := "failed to parse detect plan"
+				h.AssertStringContains(t, string(output), expected)
+			})
```
haliliceylan, line 148: I used `/cnb/delete_file_then_run.sh`file for check default location error is it ok ?

acceptance/builder_test.go
```diff
@@ -0,0 +1,184 @@
...
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "DELETE_FILE=/layers/group.toml",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				h.AssertNotNil(t, err)
+				expected := "failed to read buildpack group"
+				h.AssertStringContains(t, string(output), expected)
+			})
```
haliliceylan, line 97: I used /cnb/delete_file_then_run.shfile for check default location error is it ok ?

acceptance/builder_test.go
haliliceylan: It is like deleting file from env variable and passing all arguments to builder exetuable.

acceptance/builder_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+var (
+	builderDockerContext = filepath.Join("testdata", "builder")
+	builderBinaryDir     = filepath.Join("testdata", "builder", "container", "cnb", "lifecycle")
+	builderImage         = "lifecycle/acceptance/builder"
+	builderUserID        = "1234"
+)
+
+func TestBuilder(t *testing.T) {
+	h.SkipIf(t, runtime.GOOS == "windows", "builder acceptance tests are not yet supported on Windows")
```
natalieparellano, line 27: A bit of context to why Windows tests aren't supported for some phases - for phases that require access to a registry, enabling the test container (where the builder binary is running) to connect with the test registry container was challenging because of networking, but we found a way around this (see analyzer acceptance tests). Also the build phase doesn't need to connect to a registry, so this concern might not be an issue. 

However given that it's a bit complicated to get things working cross-OS, it might make sense to start with Linux tests for now and make them compatible with Windows later.

acceptance/builder_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	rand.Seed(time.Now().UTC().UnixNano())
+
+	//h.MakeAndCopyLifecycle(t, "linux", builderBinaryDir)
```
natalieparellano, line 31: I suspect this is commented out for speed of development, since building the binaries takes time. Our Makefile is not supposed to recompile the binaries if nothing has changed, but I think something is wrong with it... when you run `make build-linux` on the command line you'll see that the second invocation is a no-op, but that's not true when exec'ing from the acceptance tests. I'm not sure what the issue is...

acceptance/testdata/builder/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:bionic
+
+ARG cnb_uid=1234
+ARG cnb_gid=1000
+ARG cnb_platform_api
+
+ENV CNB_PLATFORM_API=${cnb_platform_api}
```
natalieparellano, line 7: We could probably remove this and just set the `CNB_PLATFORM_API` env var when running the container (using `--env`).

acceptance/builder_test.go
```diff
@@ -0,0 +1,184 @@
...
+		// .../cmd/lifecycle/builder.go#116-118
+		when("read buildpack group file", func() {
+			it("no default group toml file in default location", func() {
+				command := exec.Command(
+					"docker",
+					"run",
+					"--entrypoint", "/cnb/delete_file_then_run.sh",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "DELETE_FILE=/layers/group.toml",
```
natalieparellano, line 90: This setup strikes me as a bit convoluted - I wonder if we could pass the group location via the `-group` arg.

natalieparellano: (and then when no file exists at that location, you can make the same assertion)

acceptance/builder_test.go
haliliceylan: Yeah, i think these tests should work fine other OS too. We are using docker for test environment and like you mentioned this phase no need any specific feature depends on OS. I will Add a FIXME comment for not forget later.

acceptance/builder_test.go
haliliceylan: Yeah I'm not changing anything related to lifecycle directly so for testing i commented out this line. But i should not commit...  Normally I would leave this in my local but I committed it because it escaped my notice. Let add `FIXME` comment this line too until end of development. I think in DRAFT PR request there is no CI-CD for tests. So lets skip it until this pr get out of the draft. But i am adding `FIXME`comment anyway.

acceptance/testdata/builder/Dockerfile
haliliceylan: Got it

acceptance/builder_test.go
haliliceylan: Yes it make sense, but also with this way i think we are testing these lines too, am i right ? Because it is also testing default path location.

https://github.com/buildpacks/lifecycle/blob/1d7a850cf03a1e16b6be5f871419d41091da4f29/cmd/lifecycle/builder.go#L48-L50

If u want i can add one more test case for non-exist file test or i can change with that, which one make sense for u ?

acceptance/testdata/builder/Dockerfile
haliliceylan: fixed

acceptance/builder_test.go
natalieparellano: Actually here is how it's done for detector tests: https://github.com/buildpacks/lifecycle/blob/771ac0b9d0d56cd3baaf28355498a0a636ea6182/acceptance/detector_test.go#L87

So the builder image wouldn't have a `/layers/group.toml` but it could have a `/layers/other-group.toml` that is provided with the `-group` flag. There would be two tests - one to show a failure when no `-group` is passed and one to show success when `-group /layers/other-group.toml` is passed.

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
```diff
@@ -0,0 +1,35 @@
...
+echo "     env vars:"
+if compgen -G "${env_dir}/*" > /dev/null; then
+  for var in ${env_dir}/*; do
+    declare "$(basename ${var})=$(<${var})"
+  done
+fi
+export | sed 's/^/       /'
+
+# LAYERS
+echo "     layers_dir: ${layers_dir}"
```
natalieparellano, line 27: I wonder if we can make assertions around this output to verify that the builder was setup correctly: https://github.com/buildpacks/lifecycle/blob/771ac0b9d0d56cd3baaf28355498a0a636ea6182/cmd/lifecycle/builder.go#L85-L91

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/detect
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
natalieparellano, line 1: We could remove this file I think, since we're not testing detect here :)

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
```diff
@@ -0,0 +1,35 @@
...
+
+# PLAN
+echo "     plan_path: ${plan_path}"
+echo "     plan contents:"
+cat ${plan_path} | sed 's/^/       /'
+echo
```
natalieparellano, line 33: It would be a bit trickier to validate that the build plan was passed correctly (see https://github.com/buildpacks/lifecycle/pull/648/files#r661649474 ) as each buildpack only sees a subset (the "buildpack plan") of the complete build plan. But maybe we could assert that the contents of the buildpack plan are a subset of the build plan.

acceptance/builder_test.go
haliliceylan: If I understood correctly, I will delete the `/layers/group.toml` file directly. I will not put group.toml in the default search location. Instead I will create `cnb/group_tomls/always_detect_group.toml` and use that for success scenario. Also for all other test cases I need to pass this file like `CNB_GROUP_PATH=cnb/group_tomls/always_detect_group.toml`.

haliliceylan: I did some changes according to my latest comment. Please check

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
haliliceylan: Yeah, it sounds good to me. I have some problems with `plan.toml`and `group.toml`. I am using these toml files as empty right now, and builder is not even using buildpack. I need to look some docs and fill these tomls correctly. After that i will coverage some basic test cases ([like `buildpack.toml` api test](https://github.com/buildpacks/lifecycle/pull/648#issuecomment-870823450)). Then i can do these assertions.

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/detect
haliliceylan: Deleted.

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
haliliceylan: 👍

acceptance/builder_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+		// .../cmd/lifecycle/builder.go#116-118
+		when("read buildpack group file", func() {
+			it("no default group toml file in default location", func() {
+				command := exec.Command(
+					"docker",
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
```
yaelharel, line 92: I think that it's better that `always_detect_plan.toml` won't be an empty file like `empty_plan.toml`.

acceptance/builder_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+		// .../cmd/lifecycle/builder.go#121
+		when("error during parse plan", func() {
+			it("no default plan.toml file in default location", func() {
+				command := exec.Command(
+					"docker",
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
```
yaelharel, line 144: The same comment as https://github.com/buildpacks/lifecycle/pull/648/files#r662581113.

acceptance/builder_test.go
haliliceylan: Done 👍

acceptance/builder_test.go
haliliceylan: Done 👍

haliliceylan:
> For next step, i will add create necessary toml files to use to hello_world buildpack (i am using empty files now). I will test some buildpack.toml parsing functions here:
> 
> https://github.com/buildpacks/lifecycle/blob/1d7a850cf03a1e16b6be5f871419d41091da4f29/cmd/lifecycle/builder.go#L72-L74

I finished this case too.

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
haliliceylan: I will start this one at daytime, i think all other test cases are done for me. We need to cover some success scenarios and some test cases need to check output (like structure, inside files, toml files keys).

haliliceylan:
I added two test case for these lines;

https://github.com/buildpacks/lifecycle/blob/1d7a850cf03a1e16b6be5f871419d41091da4f29/cmd/lifecycle/builder.go#L48-L54

acceptance/builder_test.go
```diff
@@ -0,0 +1,248 @@
...
+				"root",
+				"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+				builderImage,
+			)
+			output, err := command.CombinedOutput()
+			h.AssertNotNil(t, err)
+			expected := "failed to build: refusing to run as root"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
```
yaelharel, line 81:
```suggestion
	})
	
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,248 @@
...
+				)
+				output, err := command.CombinedOutput()
+				h.AssertNotNil(t, err)
+				expected := "failed to parse detect plan: Near line"
+				h.AssertStringContains(t, string(output), expected)
+			})
+
+		})
+	})
+
```
yaelharel, line 206: I don't think it's mandatory but maybe it's worth adding another `when` to test https://github.com/buildpacks/lifecycle/blob/main/cmd/lifecycle/builder.go#L72-L74.

acceptance/builder_test.go
```diff
@@ -0,0 +1,248 @@
...
+	when("PlaceHolders need to replace with defaults", func() {
+
+		// .../cmd/lifecycle/builder.go#48-50
```
yaelharel, line 209: I think it's better to specify the `Args` function in the comment and not specific line numbers because they can change on every commit.

acceptance/builder_test.go
```diff
@@ -0,0 +1,248 @@
...
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				h.AssertNotNil(t, err)
+				expected := "failed to read buildpack group: open /layers/group.toml"
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+
+		// .../cmd/lifecycle/builder.go#52-54
```
yaelharel, line 228: Please see https://github.com/buildpacks/lifecycle/pull/648/files#r663206073

acceptance/builder_test.go
```diff
@@ -0,0 +1,248 @@
...
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				h.AssertNotNil(t, err)
+				expected := "failed to parse detect plan: open /layers/plan.toml"
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+	})
+
```
yaelharel, line 247: I think it's important to add some tests for good path that will test https://github.com/buildpacks/lifecycle/blob/main/cmd/lifecycle/builder.go#L78-L112 and more specifically will test https://github.com/buildpacks/lifecycle/blob/main/cmd/lifecycle/builder.go#L97 and its output https://github.com/buildpacks/lifecycle/blob/main/cmd/lifecycle/builder.go#L108-L110

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
haliliceylan: Done 👍 , please check this commit https://github.com/buildpacks/lifecycle/pull/648/commits/7233dad5192fb8318d0ec4805e74b19296e138e5

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
haliliceylan: done https://github.com/buildpacks/lifecycle/commit/7233dad5192fb8318d0ec4805e74b19296e138e5

acceptance/builder_test.go
haliliceylan: fixed

acceptance/builder_test.go
haliliceylan: okey i will change all comment lines like that

haliliceylan: Done

acceptance/builder_test.go
haliliceylan: done too

acceptance/builder_test.go
haliliceylan: Creating tests for these 3 scenarios sounds great. I will create these tests for the next step.

acceptance/builder_test.go
haliliceylan: I think we have already test for these lines, please check
https://github.com/buildpacks/lifecycle/pull/648/files#diff-ccbcd4a86ccafc5db2bca8335c8b8d9a1c4def69e661ab586d0d3471f9cd9f66R137-R151

acceptance/builder_test.go
```diff
@@ -0,0 +1,371 @@
...
+			expected := "failed to build: refusing to run as root"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("error on reading Data", func() {
+
+		// .../cmd/lifecycle/builder.go#readData
+		when("read buildpack group file", func() {
+			it("no default group toml file in default location", func() {
```
natalieparellano, line 87:
```suggestion
		when("group", func() {
			when("no default group toml file in default location", func() {
				it("errors", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,371 @@
...
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				h.AssertNotNil(t, err)
+				expected := "failed to read buildpack group"
+				h.AssertStringContains(t, string(output), expected)
+			})
+
+			//TODO: check some output file for this case not except any error message
+			it("empty group toml file", func() {
```
natalieparellano, line 103:
```suggestion
			when("empty group toml file", func() {
				it("<something>", func() {
```

acceptance/builder_test.go
natalieparellano: Is plan path even needed? What happens if you don't pass this?

natalieparellano: Oh... I see... it needs to exist, even if it's just empty.

acceptance/builder_test.go
natalieparellano: I didn't add this comment everywhere,  but it would be great if the other tests could also follow this pattern.

haliliceylan:
for last WIP commit, we can use this shell command
`go test -v -count=1 -tags=acceptance acceptance/acceptance_test.go acceptance/builder_test.go`

also during debug maybe we need to uncomment or delete some lines which has `FIXME `comment lines;

I commented most of build/release command for speed up development. Because i am not changing any code itself of lifecycle.

acceptance/builder_test.go
haliliceylan: done, please check

acceptance/builder_test.go
haliliceylan: done please check,

haliliceylan:
i also fixed when-it functions explanations too according to today feedbacks

acceptance/builder_test.go
haliliceylan: i think i will parse `/layers/config/metadata.toml` then assert content according to my buildpack and other inputs. Please add any other suggestions you can think of.

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					h.AssertNotNil(t, err)
+					expected := "failed to read buildpack group"
+					h.AssertStringContains(t, string(output), expected)
+				})
+
+			})
+
+			//TODO: check some output file for this case not except any error message
+			when("empty group toml file", func() {
+				it("success", func() {
```
natalieparellano, line 106:
```suggestion
				it("succeeds", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+						"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+						"--env", "CNB_GROUP_PATH=/cnb/group_tomls/empty_group.toml",
+						"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+						builderImage,
+					)
+					_, err := command.CombinedOutput()
+					//print(string(output), err)
+					h.AssertNil(t, err)
+					//expected := "failed to read buildpack order file"
+					//h.AssertStringContains(t, string(output), expected)
```
natalieparellano, line 120: I think you could check for a metadata.toml file to show that the build was completed.

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+			output, err := command.CombinedOutput()
+			h.AssertNotNil(t, err)
+			expected := "failed to build: refusing to run as root"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("error on reading Data", func() {
+		// .../cmd/lifecycle/builder.go#readData
+		when("read buildpack group file", func() {
```
natalieparellano, line 85:
```suggestion
		when("read buildpack group file", func() {
```
```suggestion
		when("parse group.toml", func() {
```
(for consistency with line 163)

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				builderImage,
+			)
+			output, err := command.CombinedOutput()
+			h.AssertNotNil(t, err)
+			expected := "failed to build: refusing to run as root"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("error on reading Data", func() {
```
natalieparellano, line 83:
```suggestion
	when("error on reading data", func() {
```

natalieparellano, line 83:
```suggestion
	when("invalid input files", func() {
```

acceptance/builder_test.go
natalieparellano, line 85:
```suggestion
		when("group.toml", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+		// .../cmd/lifecycle/builder.go#readData
+		when("read buildpack group file", func() {
+			when("no default group toml file in default location", func() {
```
natalieparellano, line 86:
```suggestion
			when("not found", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to read buildpack group"
+					h.AssertStringContains(t, string(output), expected)
+				})
+
+			})
+
+			//TODO: check some output file for this case not except any error message
+			when("empty group toml file", func() {
```
natalieparellano, line 105:
```suggestion
			when("empty", func() {
```

acceptance/builder_test.go
natalieparellano: (this makes the test output a bit terser and easier to read!)

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					)
+					_, err := command.CombinedOutput()
+					//print(string(output), err)
+					h.AssertNil(t, err)
+					//expected := "failed to read buildpack order file"
+					//h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+			when("invalid group toml file", func() {
```
natalieparellano, line 124:
```suggestion
			when("invalid", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					)
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to read buildpack group: Near line"
+					h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+			// .../cmd/lifecycle/builder.go#Exec
+			when("invalid builpack api group toml file", func() {
```
natalieparellano, line 143:
```suggestion
			when("invalid builpack api", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to : parse buildpack API"
+					h.AssertStringContains(t, string(output), expected)
+				})
+			})
+		})
+
+		// .../cmd/lifecycle/builder.go#readData
+		when("parse plan.toml", func() {
```
natalieparellano, line 163:
```suggestion
		when("plan.toml", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+			when("no default plan.toml file in default location", func() {
```
natalieparellano, line 164:
```suggestion
			when("not found", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					)
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to parse detect plan"
+					h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+			//TODO: check some output file for this case not except any error message
+			when("empty parse plan.toml file", func() {
```
natalieparellano, line 182:
```suggestion
			when("empty", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					)
+					_, err := command.CombinedOutput()
+					//print(string(output), err)
+					h.AssertNil(t, err)
+					//expected := "failed to read buildpack order file"
+					//h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+			when("invalid parse plan.toml file", func() {
```
natalieparellano, line 201:
```suggestion
			when("invalid", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+
+		// .../cmd/lifecycle/builder.go#Args
+		when("groupPath is equals to PlaceHolder groupPath", func() {
+			it("will replace placeholder with default groupPath", func() {
+				command := exec.Command(
+					"docker",
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=<layers>/group.toml",
```
natalieparellano, line 232: This test is a bit weird to me, because even though it does demonstrate that the code is correct, I don't think anyone would provide `<layers>/group.toml` as the group path in "real life". 

What if instead, we set the `CNB_LAYERS_DIR` to be something other than `/layers`, and provide a group.toml in that directory. So there would be a `/some-other-layers/group.toml` in the fixture, and we could assert that that group.toml was used?

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+
+		// .../cmd/lifecycle/builder.go#Args
+		when("planPath is equals to PlaceHolder planPath", func() {
+			it("will replace placeholder with default planPath", func() {
+				command := exec.Command(
+					"docker",
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_PLAN_PATH=<layers>/plan.toml",
```
natalieparellano, line 251: Same comment as https://github.com/buildpacks/lifecycle/pull/648/files#r669948392

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					h.AssertNotNil(t, err)
+					expected := "failed to parse detect plan: Near line"
+					h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+		})
+	})
+
+	when("PlaceHolders need to replace with defaults", func() {
```
natalieparellano, line 222:
```suggestion
	when("determining the location of input files", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+
+		// .../cmd/lifecycle/builder.go#Args
+		when("groupPath is equals to PlaceHolder groupPath", func() {
+			it("will replace placeholder with default groupPath", func() {
```
natalieparellano, line 226:
```suggestion
			it("will look for group.toml in the provided <layers> directory", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				})
+			})
+
+		})
+	})
+
+	when("PlaceHolders need to replace with defaults", func() {
+
+		// .../cmd/lifecycle/builder.go#Args
+		when("groupPath is equals to PlaceHolder groupPath", func() {
```
natalieparellano, line 225:
```suggestion
		when("group.toml path is not specified", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				)
+				output, err := command.CombinedOutput()
+				h.AssertNotNil(t, err)
+				expected := "failed to read buildpack group: open /layers/group.toml"
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+
+		// .../cmd/lifecycle/builder.go#Args
+		when("planPath is equals to PlaceHolder planPath", func() {
```
natalieparellano, line 244:
```suggestion
		when("plan.toml path is not specified", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+			it("will replace placeholder with default planPath", func() {
```
natalieparellano, line 245:
```suggestion
			it(""will look for plan.toml in the provided <layers> directory", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+					"--env", "CNB_APP_DIR=/env_folders/different_cnb_app_dir_from_env",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNil(t, err) //due to not exist directory
```
natalieparellano, line 280: I think rather than passing a directory that does not exist, we could pass a real directory, and have the buildpack print its current working directory when bin/build is run.

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+					"--env", "CNB_BUILDPACKS_DIR=/env_folders/different_buildpack_dir_from_env",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNil(t, err) //due to not exist directory
```
natalieparellano, line 300: Similar to other comment (https://github.com/buildpacks/lifecycle/pull/648/files#r669952363) we could pass a real directory, and the buildpack found there could print "Hello from other directory" which we could assert on.

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				h.AssertNotNil(t, err)
+				expected := "failed to parse detect plan: open /layers/plan.toml"
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+
+	})
+
+	/// .../cmd/lifecycle/builder.go#build
+	when("Builder args are successfully transmitted to in build script", func() {
```
natalieparellano, line 265: I think we could remove this parent context.

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+		when("CNB_APP_DIR changed", func() {
+			it("CNB_APP_DIR is successfully transmitted to in build script", func() {
```
natalieparellano, line 267:
```suggestion
			it("sets the buildpacks' working directory to CNB_APP_DIR", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNil(t, err) //due to not exist directory
+				expected := "CNB_APP_DIR: /env_folders/different_cnb_app_dir_from_env"
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+
+		when("CNB_BUILDPACKS_DIR changed", func() {
+			it("CNB_BUILDPACKS_DIR is successfully transmitted to in build script", func() {
```
natalieparellano, line 287:
```suggestion
			it("uses buildpacks from CNB_BUILDPACKS_DIR", func() {
```

acceptance/builder_test.go
natalieparellano: Ohh you made this change already, nevermind!

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/differrent_plan_from_env.toml",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNil(t, err) //due to not exist directory
```
natalieparellano, line 340: I think this comment might be outdated :)

acceptance/builder_test.go
natalieparellano: Though see https://github.com/buildpacks/lifecycle/pull/648/files#r669954888

acceptance/builder_test.go
natalieparellano: See https://github.com/buildpacks/lifecycle/pull/648/files#r669954888

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+					"--env", "CNB_LAYERS_DIR=/tmp/different_layers_path_dir_from_env",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNil(t, err) //due to not exist directory
```
natalieparellano, line 320: See https://github.com/buildpacks/lifecycle/pull/648/files#r669954888

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				expected := "CNB_BUILDPACK_DIR: /env_folders/different_buildpack_dir_from_env"
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+
+		when("CNB_LAYERS_DIR", func() {
+			it("CNB_LAYERS_DIR is successfully transmitted to in build script", func() {
```
natalieparellano, line 307:
```suggestion
			it("CNB_LAYERS_DIR is successfully transmitted to build script", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+					"--env", "CNB_LAYERS_DIR=/tmp/different_layers_path_dir_from_env",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNil(t, err) //due to not exist directory
+				expected := "layers_dir: /tmp/different_layers_path_dir_from_env"
```
natalieparellano, line 321: The layers directory provided to an individual buildpack is actually going to be a sub directory of `CNB_LAYERS_DIR`.

natalieparellano:
```suggestion
				expected := "layers_dir: /tmp/different_layers_path_dir_from_env/hello_world"
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+
+		when("CNB_PLAN_PATH", func() {
+			it("CNB_PLAN_PATH is successfully transmitted to in build script", func() {
```
natalieparellano, line 327:
```suggestion
			it("provides the buildpack a filtered version of the plan found at CNB_PLAN_PATH", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNil(t, err) //due to not exist directory
+				expected := "different_plan_from_env.toml_reqires_subset_content"
+				h.AssertStringContains(t, string(output), expected)
+			})
+		})
+
+		when("CNB_PLATFORM_DIR", func() {
+			it("CNB_PLATFORM_DIR is successfully transmitted to in build script", func() {
```
natalieparellano, line 347:
```suggestion
			it("CNB_PLATFORM_DIR is successfully transmitted to build script", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+					"--env", "CNB_PLATFORM_DIR=/different_platform_dir_from_env",
+					builderImage,
+				)
+				output, err := command.CombinedOutput()
+				//print(string(output), err)
+				h.AssertNotNil(t, err) //due to not exist directory
```
natalieparellano, line 360: I don't think this actually shows that `CNB_PLATFORM_DIR` made it all the way to the build script, as we'd be erroring out before any buildpacks are run. 

We could instead pass a real platform directory, and assert that the env was set correctly. Some context - if there's a /platform/env/MY_VAR file with contents MY_VAL, then buildpacks will have `MY_VAR=MY_VAL` in their environment.

acceptance/testdata/builder/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:bionic
+
+ARG cnb_uid=1234
+ARG cnb_gid=1000
+
+ENV CNB_ASSETS=/some/asset/path
+
```
natalieparellano, line 7: We've put a pause on asset packages for now, so I think this could be removed.

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+```bash
+pack build sample-hello-world-app --builder cnbs/sample-builder:bionic --buildpack .
+```
+
+THEN
+
+```
+docker run sample-hello-world-app echo hello
+```
```
natalieparellano, line 21: For maintainability, I think we could delete this file as it's not used anywhere.

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+echo "---> Hello World buildpack"
```
natalieparellano, line 4: Similar to https://github.com/buildpacks/lifecycle/pull/648/files#r669969132, it would be great if we could delete from this file anything that's not used by the tests.

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
natalieparellano: Looks good!

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/buildpack.toml
```diff
@@ -0,0 +1,19 @@
...
+
+# Stacks that the buildpack will work with
+[[stacks]]
+id = "io.buildpacks.samples.stacks.bionic"
+
+[[stacks]]
+id = "io.buildpacks.samples.stacks.alpine"
+
+[[stacks]]
+id = "io.buildpacks.stacks.bionic"
```
natalieparellano, line 19: I think this is all not used by the tests, and could be removed.

acceptance/testdata/builder/container/cnb/group_tomls/always_detect_group.toml
```diff
@@ -0,0 +1,10 @@
+[[group]]
+  api = "0.2"
+  homepage = "https://github.com/buildpacks/samples/tree/main/buildpacks/hello-world"
+  id = "hello_world"
+  version = "0.0.1"
+
+  [[entries.requires]]
+      name = "<dependency name>"
+      [entries.requires.metadata]
+        # arbitrary data describing the required dependency
```
natalieparellano, line 10: nit: missing newline

acceptance/testdata/builder/container/cnb/group_tomls/invalid_buildpack_api_group.toml
```diff
@@ -0,0 +1,5 @@
+[[group]]
+  api = "0.2wrongapiblabla"
+  homepage = "https://github.com/buildpacks/samples/tree/main/buildpacks/hello-world"
+  id = "hello_world"
+  version = "0.0.1"
```
natalieparellano, line 5: nit: missing newline

acceptance/testdata/builder/container/cnb/plan_tomls/always_detect_plan.toml
```diff
@@ -0,0 +1,5 @@
+[[entries]]
+
+  [[entries.providers]]
+    id = "hello_world"
+    version = "0.0.1"
```
natalieparellano, line 5: nit: missing newline

acceptance/testdata/builder/container/cnb/plan_tomls/differrent_plan_from_env.toml
```diff
@@ -0,0 +1,10 @@
+[[entries]]
+
+  [[entries.providers]]
+    id = "hello_world"
+    version = "0.0.1"
+  
+  [[entries.requires]]
+      name = "different_plan_from_env.toml_reqires_subset_content"
+      [entries.requires.metadata]
+        # arbitrary data describing the required dependency
```
natalieparellano, line 10: nit: missing newline

acceptance/testdata/builder/container/cnb/plan_tomls/wrong_plan.toml
```diff
@@ -0,0 +1,3 @@
+I am a plan.toml file which not correct as context and syntax.
+
+For more info please look at https://github.com/buildpacks/spec/blob/main/platform.md#plantoml-toml
```
natalieparellano, line 3: nit: missing newline

acceptance/builder_test.go
```diff
@@ -0,0 +1,367 @@
...
+		when("CNB_PLAN_PATH", func() {
+			it("CNB_PLAN_PATH is successfully transmitted to in build script", func() {
+
+				command := exec.Command(
+					"docker",
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/differrent_plan_from_env.toml",
```
natalieparellano, line 335:
```suggestion
					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/different_plan_from_env.toml",
```

acceptance/testdata/builder/container/cnb/group_tomls/wrong_group.toml
```diff
@@ -0,0 +1,3 @@
+I am a group.toml file which not correct as context and syntax.
+
+For more info please look at https://github.com/buildpacks/spec/blob/main/platform.md#grouptoml-toml
```
natalieparellano, line 3: nit: missing newline

acceptance/testdata/builder/container/env_folders/different_buildpack_dir_from_env/hello_world/0.0.1/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+```bash
+pack build sample-hello-world-app --builder cnbs/sample-builder:bionic --buildpack .
+```
+
+THEN
+
+```
+docker run sample-hello-world-app echo hello
+```
```
natalieparellano, line 21: Same comment as https://github.com/buildpacks/lifecycle/pull/648/files#r669970050

acceptance/testdata/builder/container/env_folders/different_buildpack_dir_from_env/hello_world/0.0.1/bin/build
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+echo "---> Hello World buildpack"
```
natalieparellano, line 4: Same comment as https://github.com/buildpacks/lifecycle/pull/648/files#r669969357 (that will also make the duplication less painful)

acceptance/testdata/builder/container/env_folders/different_buildpack_dir_from_env/hello_world/0.0.1/buildpack.toml
```diff
@@ -0,0 +1,19 @@
...
+
+# Stacks that the buildpack will work with
+[[stacks]]
+id = "io.buildpacks.samples.stacks.bionic"
+
+[[stacks]]
+id = "io.buildpacks.samples.stacks.alpine"
+
+[[stacks]]
+id = "io.buildpacks.stacks.bionic"
```
natalieparellano, line 19: Same comment as https://github.com/buildpacks/lifecycle/pull/648/files#r669970050

acceptance/builder_test.go
haliliceylan: For 2 positive test I am checking metadata.toml now.

acceptance/builder_test.go
haliliceylan: fixed

acceptance/builder_test.go
haliliceylan: fixed

acceptance/testdata/builder/Dockerfile
haliliceylan: fixed

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/README.md
haliliceylan: i deleted two readme file from buildpacks.

acceptance/testdata/builder/container/cnb/group_tomls/invalid_buildpack_api_group.toml
haliliceylan: fixed

acceptance/testdata/builder/container/cnb/group_tomls/wrong_group.toml
haliliceylan: fixed

acceptance/testdata/builder/container/cnb/plan_tomls/always_detect_plan.toml
haliliceylan: fixed

acceptance/testdata/builder/container/cnb/plan_tomls/wrong_plan.toml
haliliceylan: fixed

acceptance/testdata/builder/container/env_folders/different_buildpack_dir_from_env/hello_world/0.0.1/README.md
haliliceylan: deleted

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/buildpack.toml
haliliceylan: I deleted most of parts of buildpack.toml files please check again. We need some keys from this file. When i delete this file completely tests are failing.

acceptance/testdata/builder/container/env_folders/different_buildpack_dir_from_env/hello_world/0.0.1/buildpack.toml
haliliceylan: i added comment other conversation, please check

acceptance/testdata/builder/container/cnb/group_tomls/always_detect_group.toml
haliliceylan: fixed

acceptance/testdata/builder/container/cnb/buildpacks/hello_world/0.0.1/bin/build
haliliceylan: deleted most parts

acceptance/testdata/builder/container/env_folders/different_buildpack_dir_from_env/hello_world/0.0.1/bin/build
haliliceylan: deleted most parts

acceptance/builder_test.go
haliliceylan: we are already printing CNB_BUILDPACK_DIR: env variable and asserting this one, if u want i can add "Hello from other directory" string, i also update build script for this buildpack. please check again.

acceptance/builder_test.go
haliliceylan: yeah i noticed this one and fixed. Thanks 👍

acceptance/builder_test.go
haliliceylan: fixed too

acceptance/builder_test.go
haliliceylan: i did not find correct lines, can you review again this one ? also link is not working too. I am not able to see your comment.

acceptance/builder_test.go
haliliceylan: Wow, this test is sounds great for me let me add.

haliliceylan: I created "acceptance/testdata/builder/container/layers/different_layer_dir_from_env" and i used this folder for tests. Due to permission errors i have to use this directory. Please see this line in dockerfile. https://github.com/buildpacks/lifecycle/pull/648/files#diff-9edd37c29964124e391169e05cb6d5917c38c1dab1922047eaf5fefa2d77dbcaR13

haliliceylan:
I fixed lint problems. I do not know why unit tests are failing, i did not change anything except `acceptance` folder. Please check what can i do for tests.

acceptance/builder_test.go
haliliceylan: please check this commit, https://github.com/buildpacks/lifecycle/pull/648/commits/5840857629d7f90d3ffe40bbdeab38fa26dd0d5d

haliliceylan:
s I don't understand, I can't succeed in the tests, I haven't changed any source code, but the unit and other acceptance tests fail. Could it be because I'm using a Mac Osx environment?

haliliceylan:
image
After the running acceptance test, These containers remain undeleted, I don't understand if it's because of the codes I added or is it a general problem

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+						"run",
+						"--rm",
+						"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+						"--env", "CNB_GROUP_PATH=/cnb/group_tomls/invalid_buildpack_api_group.toml",
+						"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+						builderImage,
+					)
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to : parse buildpack API"
```
yaelharel, line 196: This following error comes from https://github.com/buildpacks/lifecycle/blob/3297dacad3f0d58bcbbb44dff0d896b9ad1c28c1/cmd/version.go#L70 right?
Maybe it's worth changing the expected to the full error.

acceptance/builder_test.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/648/files#r672322686.
Moreover, is possible to add a test to one of the deprecation cases that appear in the following part of code?
https://github.com/buildpacks/lifecycle/blob/3297dacad3f0d58bcbbb44dff0d896b9ad1c28c1/cmd/version.go#L74-L90

acceptance/testdata/builder/container/layers/different_layer_dir_from_env/group.toml
```diff
@@ -0,0 +1,11 @@
+[[group]]
```
yaelharel, line 1: I don't think that you need the `.gitkeep` in this folder because it contains other files (this file and `plan.toml`)

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+			)
+			output, err := command.CombinedOutput()
+			//print(string(output), err)
+			h.AssertNil(t, err) //we have real directory
+			expected := "CNB_BUILDPACK_DIR: /env_folders/different_buildpack_dir_from_env"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("CNB_LAYERS_DIR", func() {
```
yaelharel, line 340:
```suggestion
	when("CNB_LAYERS_DIR is set", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+			)
+			output, err := command.CombinedOutput()
+			//print(string(output), err)
+			h.AssertNil(t, err) //we have real directory
+			expected := "LAYERS_DIR: /layers/different_layer_dir_from_env/hello_world"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("CNB_PLAN_PATH", func() {
```
yaelharel, line 360:
```suggestion
	when("CNB_PLAN_PATH is set", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+			)
+			output, err := command.CombinedOutput()
+			//print(string(output), err)
+			h.AssertNil(t, err) //we have real directory
+			expected := "different_plan_from_env.toml_reqires_subset_content"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("CNB_PLATFORM_DIR", func() {
```
yaelharel, line 379:
```suggestion
	when("CNB_PLATFORM_DIR is set", func() {
```

natalieparellano:
@haliliceylan I believe with the test failures, you're running into a change that was made to `main` - `MakeAndCopyLifecycle` now takes an "architecture" argument. I believe if you make the changes that are shown [here](https://github.com/buildpacks/lifecycle/pull/659/files?file-filters%5B%5D=.go#diff-c4e4b4a55ebbd71443c5260020b287f3018cb7bb4c65c79e74554b7f97c45656) for analyzer, you should be all set.

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/platform"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+)
+
+var (
+	builderDockerContext = filepath.Join("testdata", "builder")
+	builderBinaryDir     = filepath.Join("testdata", "builder", "container", "cnb", "lifecycle")
+	builderImage         = "lifecycle/acceptance/builder"
+	//builderUserID        = "1234"
```
natalieparellano, line 27: Is this needed?

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to parse detect plan"
+					h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+			//TODO: check some output file for this case not except any error message
+			when("empty", func() {
+				it("success", func() {
```
natalieparellano, line 223:
```suggestion
				it("succeeds", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+						builderImage,
+					)
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to parse detect plan"
+					h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+			//TODO: check some output file for this case not except any error message
```
natalieparellano, line 221: Is this still needed?

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_LAYERS_DIR=/layers/different_layer_dir_from_env",
+					"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+					builderImage,
+				)
+				_, err := command.CombinedOutput()
+				h.AssertNil(t, err)
+				//expected := "failed to read buildpack group: open /layers/group.toml"
```
natalieparellano, line 275: Ah - so if the layers dir is set to `/layers/different_layer_dir_from_env` the builder is going to try to use `/layers/different_layer_dir_from_env/group.toml`. If you put a real group file in your fixture (acceptance/testdata/builder/container/layers/different_layer_dir_from_env/group.toml) you could see that the file was used by either (a) referencing a non-existent buildpack, and checking the error message, or (b) referencing a real buildpack, but a different buildpack from `layers/group.toml`, and checking the output from the buildpack's bin/build. I would lean toward (b), but I think either could work.

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+					"run",
+					"--rm",
+					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+					"--env", "CNB_LAYERS_DIR=/layers/different_layer_dir_from_env",
+					"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+					builderImage,
+				)
+				_, err := command.CombinedOutput()
+				h.AssertNil(t, err)
+				//expected := "failed to parse detect plan: open /layers/plan.toml"
```
natalieparellano, line 294: Same comment as https://github.com/buildpacks/lifecycle/pull/648/files#r672620621

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+				"run",
+				"--rm",
+				"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+				"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+				"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+				"--env", "CNB_APP_DIR=/env_folders/different_cnb_app_dir_from_env",
+				builderImage,
+			)
+			output, err := command.CombinedOutput()
+			//print(string(output), err)
```
natalieparellano, line 313: I think this could be removed

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+				"run",
+				"--rm",
+				"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+				"--env", "CNB_GROUP_PATH=/cnb/group_tomls/always_detect_group.toml",
+				"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+				"--env", "CNB_BUILDPACKS_DIR=/env_folders/different_buildpack_dir_from_env",
+				builderImage,
+			)
+			output, err := command.CombinedOutput()
+			//print(string(output), err)
```
natalieparellano, line 333: I think this could be removed

acceptance/builder_test.go
natalieparellano: What you did works!

acceptance/builder_test.go
natalieparellano: That works!

acceptance/testdata/builder/container/env_folders/different_buildpack_dir_from_env/hello_world/0.0.1/buildpack.toml
natalieparellano: Looks good!

acceptance/builder_test.go
```diff
@@ -0,0 +1,418 @@
...
+						builderImage,
+					)
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to read buildpack group"
+					h.AssertStringContains(t, string(output), expected)
+				})
+			})
+
+			//TODO: check some output file for this case not except any error message
```
natalieparellano, line 145: Is this still needed?

acceptance/builder_test.go
natalieparellano: Ohh I think I was saying that the comment was outdated, because we're expecting a nil error. But it looks like you've updated it already!

acceptance/builder_test.go
haliliceylan: we need to add `metadata.toml` checking.

acceptance/testdata/builder/container/layers/different_layer_dir_from_env/group.toml
haliliceylan: deleted

acceptance/builder_test.go
haliliceylan: fixed

acceptance/builder_test.go
haliliceylan: deleted

acceptance/builder_test.go
haliliceylan: deleted

acceptance/builder_test.go
haliliceylan: different buildpack is added. then metadata file is checked.

acceptance/builder_test.go
haliliceylan: different buildpack is added. then metadata file is checked.

acceptance/builder_test.go
haliliceylan: In this file , i did not find any deprecated api version. it passed as nil. I tried pass different versions of platform and Buildpack. But i did not get any deprecated error message. I got it different errors like that `set platform API: platform API version '0.1' is incompatible with the lifecycle`. Is there any suggestion for that ?

https://github.com/buildpacks/lifecycle/blob/3297dacad3f0d58bcbbb44dff0d896b9ad1c28c1/api/apis.go#L7-L10

haliliceylan:
> @haliliceylan I believe with the test failures, you're running into a change that was made to `main` - `MakeAndCopyLifecycle` now takes an "architecture" argument. I believe if you make the changes that are shown [here](https://github.com/buildpacks/lifecycle/pull/659/files?file-filters%5B%5D=.go#diff-c4e4b4a55ebbd71443c5260020b287f3018cb7bb4c65c79e74554b7f97c45656) for analyzer, you should be all set.

I fetched new commits from master repo main. But still some tests are failing. idk what is the wrong

acceptance/builder_test.go
yaelharel: I think it's worth adding this FIXME comment 😊

acceptance/builder_test.go
```diff
@@ -0,0 +1,415 @@
...
+func getBuilderMetadata(t *testing.T, path string) *platform.BuildMetadata {
+	contents, _ := ioutil.ReadFile(path)
+	h.AssertEq(t, len(contents) > 0, true)
+
+	var analyzedMd platform.BuildMetadata
+	_, err := toml.Decode(string(contents), &analyzedMd)
+	h.AssertNil(t, err)
+
+	return &analyzedMd
+}
```
yaelharel, line 415: These two functions are almost the same.
I think it's better to not have duplicated code.
Moreover, I'm not sure why you would want to only `assertBuilderMetadata` without checking the returned metadata from `getBuilderMetadata`.
So probably it's better to keep only `getBuilderMetadata`.

acceptance/builder_test.go
```diff
@@ -0,0 +1,415 @@
...
+						"docker",
+						"run",
+						"--rm",
+						"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
+						"--env", "CNB_PLAN_PATH=/cnb/plan_tomls/always_detect_plan.toml",
+						builderImage,
+					)
+					output, err := command.CombinedOutput()
+					h.AssertNotNil(t, err)
+					expected := "failed to read buildpack group"
```
yaelharel, line 138: I think it's better to change the `expected` string in all of the tests to the full error.
For example, here it'll be:
```
failed to read buildpack group: open /layers/group.toml: no such file or directory
```
On line 174, it'll be:
```
failed to read buildpack group: Near line 1 (last key parsed 'I'): expected key separator '=', but got 'a' instead
```
(you can see more examples in `acceptance/detector_test.go` and `buildpack/detect_test.go`).

In this way, if something changes in the code that influence our tests, we'll be able to catch it more easily.

acceptance/builder_test.go
haliliceylan: added.

acceptance/builder_test.go
yaelharel: I didn't look deeply into the code but from first glance I see that the error that you're getting:`set platform API: platform API version '0.1' is incompatible with the lifecycle` is coming from the [`platformAPIError` function](https://github.com/buildpacks/lifecycle/blob/8f801a55fd4a1c9ee12eadd0b1c5b5cbf5ed98b5/cmd/version.go#L94-L99).
Lifecycle is calling this function in two places in the code: [here](https://github.com/buildpacks/lifecycle/blob/8f801a55fd4a1c9ee12eadd0b1c5b5cbf5ed98b5/cmd/version.go#L54) and [here](https://github.com/buildpacks/lifecycle/blob/8f801a55fd4a1c9ee12eadd0b1c5b5cbf5ed98b5/cmd/version.go#L63).
If you're getting it from the first flow, it's great because it's exactly what we want to test. In that case, maybe you can just add a check of the other errors that lifecycle print [there](https://github.com/buildpacks/lifecycle/blob/8f801a55fd4a1c9ee12eadd0b1c5b5cbf5ed98b5/cmd/version.go#L52-L53).
I would add some logs to check where the error that you're getting comes from.
Moreover, I would also try API `0.3.x` and check what happens.

acceptance/builder_test.go
haliliceylan: fixed in 284958efb31b1a468aa4d0daa153680cb8287c67

acceptance/builder_test.go
haliliceylan: fixed in 7e4d217

haliliceylan:
Can you please check this commit @yaelharel . I changed to test. Because this tests are generating empty config.toml.

f5b6a33

acceptance/builder_test.go
natalieparellano: I am for this check, but I don't think it should be added to the builder acceptance tests because this code path is used by every phase. Maybe we could create an issue to add it to the "version" tests.

acceptance/builder_test.go
```diff
@@ -0,0 +1,412 @@
...
+	builderBinaryDir     = filepath.Join("testdata", "builder", "container", "cnb", "lifecycle")
+	builderImage         = "lifecycle/acceptance/builder"
+)
+
+func TestBuilder(t *testing.T) {
+	//FIXME: test on windows and delete this skip condition if everything is fine
+	h.SkipIf(t, runtime.GOOS == "windows", "builder acceptance tests are not yet supported on Windows")
+
+	rand.Seed(time.Now().UTC().UnixNano())
+	//h.MakeAndCopyLifecycle(t, "linux", "amd64", builderBinaryDir)
```
natalieparellano, line 34: Having this line commented out is causing the linter to fail because `builderBinaryDir` is now unused. I think we could bring it back (and line 40).

acceptance/builder_test.go
```diff
@@ -0,0 +1,412 @@
...
+				md := getBuilderMetadata(t, filepath.Join(copyDir, "layers/different_layer_dir_from_env/config/metadata.toml"))
+
+				h.AssertStringContains(t, md.Buildpacks[0].API, "0.2")
+				h.AssertStringContains(t, md.Buildpacks[0].ID, "hello_world_2")
+				h.AssertStringContains(t, md.Buildpacks[0].Version, "0.0.2")
+			})
+		})
+	})
+
+	when("CNB_APP_DIR changed", func() {
```
natalieparellano, line 308:
```suggestion
	when("CNB_APP_DIR is set", func() {
```
(for consistency)

acceptance/builder_test.go
```diff
@@ -0,0 +1,412 @@
...
+				builderImage,
+			)
+			output, err := command.CombinedOutput()
+			h.AssertNil(t, err)
+			expected := "CNB_APP_DIR: /env_folders/different_cnb_app_dir_from_env"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("CNB_BUILDPACKS_DIR changed", func() {
```
natalieparellano, line 327:
```suggestion
	when("CNB_BUILDPACKS_DIR is set", func() {
```

acceptance/builder_test.go
```diff
@@ -0,0 +1,412 @@
...
+			)
+			output, err := command.CombinedOutput()
+			h.AssertNil(t, err)
+			expected := "CNB_BUILDPACK_DIR: /env_folders/different_buildpack_dir_from_env"
+			h.AssertStringContains(t, string(output), expected)
+		})
+	})
+
+	when("CNB_LAYERS_DIR is set", func() {
+		it("CNB_LAYERS_DIR is successfully transmitted to build script", func() {
```
natalieparellano, line 347:
```suggestion
		it("CNB_LAYERS_DIR is a parent of the buildpack layers dir", func() {
```
(this is more accurate, as it's not set in the env)

acceptance/builder_test.go
natalieparellano: Let's try taking out this Skip and see what happens :)

natalieparellano: (I'm expecting, if there are failures, that we might need to wrap some strings in `ctrPath`)

natalieparellano (commented):
@haliliceylan it's so close! Let's see if we can get everything green and (I hope this isn't too greedy) see about Windows. Looking forward to merging this in!

yaelharel:
Hi @haliliceylan, your changes look great!
I didn't understand your question in the following comment: https://github.com/buildpacks/lifecycle/pull/648#issuecomment-894111705. The commit that you posted (https://github.com/buildpacks/lifecycle/commit/f5b6a3397cb195faba23f3e8841a16ad21d3b74c) isn't part of your PR.

natalieparellano:
I see the failure is here: https://github.com/buildpacks/lifecycle/pull/648/checks?check_run_id=3382748731#step:6:5492

It's trying to use acceptance/testdata/builder/Dockerfile to build the image and failing because there's no Windows version of ubuntu:bionic :)

We need to create a `Dockerfile.windows` similar to https://github.com/buildpacks/lifecycle/blob/main/acceptance/testdata/analyzer/analyze-image/Dockerfile.windows but for the builder.

And then when we build the image, we need to refer to it by `dockerfileName`, similar to [here](https://github.com/buildpacks/lifecycle/blob/dbbce698259699bbde6fdeebfd623cc710bc9c0f/acceptance/analyzer_test.go#L122-L128) so that we can choose the correct Dockerfile depending on the underlying OS.

natalieparellano (approved):

Issue #649, opened 2021-06-30
CNB_ASSETS var should be present in detect env


natalieparellano:
See https://github.com/buildpacks/spec/pull/229#discussion_r661715338

This contradicts the implementation: https://github.com/buildpacks/lifecycle/blob/771ac0b9d0d56cd3baaf28355498a0a636ea6182/acceptance/detector_test.go#L139

We could change the linked test to expect the `CNB_ASSETS` variable.

aemengo:
@natalieparellano I'd like to work on this! _Put me in coach!_

PR #650, opened 2021-07-02
Show CNB_ASSETS env var conditionally during detect

aemengo:
Resolves: https://github.com/buildpacks/lifecycle/issues/649

### DIFF
--- a/acceptance/detector_test.go
+++ b/acceptance/detector_test.go
@@ -121,45 +121,6 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 
-	when("assets env var is set on the builder", func() {
-		var copyDir, containerName string
-		it.Before(func() {
-			containerName = "test-container-" + h.RandString(10)
-			var err error
-			copyDir, err = ioutil.TempDir("", "test-docker-copy-")
-			h.AssertNil(t, err)
-		})
-		it.After(func() {
-			if h.DockerContainerExists(t, containerName) {
-				h.Run(t, exec.Command("docker", "rm", containerName))
-			}
-			os.RemoveAll(copyDir)
-		})
-
-		it("does not appear in the detection environment", func() {
-			h.DockerRunAndCopy(t,
-				containerName,
-				copyDir,
-				"/layers",
-				detectImage,
-				h.WithFlags("--user", userID,
-					"--env", "CNB_ORDER_PATH=/cnb/orders/always_detect_order.toml",
-					"--env", "CNB_BUILDPACKS_DIR=/cnb/custom_buildpacks",
-					"--env", "CNB_APP_DIR=/custom_workspace",
-					"--env", "CNB_GROUP_PATH=./custom_group.toml",
-					"--env", "CNB_PLAN_PATH=./custom_plan.toml",
-					"--env", "CNB_PLATFORM_DIR=/custom_platform",
-					"--env", "CNB_PLATFORM_API="+latestPlatformAPI,
-				),
-				h.WithArgs("-log-level=debug"),
-			)
-
-			logs := h.Run(t, exec.Command("docker", "logs", containerName))
-			h.AssertStringDoesNotContain(t, logs, "CNB_ASSETS")
-			h.AssertStringContains(t, logs, "NO ASSETS")
-		})
-	})
-
 	when("there is a buildpack group that pass detection", func() {
 		var copyDir, containerName string
 
--- a/buildpack/detect_test.go
+++ b/buildpack/detect_test.go
@@ -1,6 +1,7 @@
 package buildpack_test
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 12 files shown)


acceptance/detector_test.go
```diff
@@ -121,45 +121,6 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
-			h.AssertNil(t, err)
-		})
-		it.After(func() {
-			if h.DockerContainerExists(t, containerName) {
-				h.Run(t, exec.Command("docker", "rm", containerName))
-			}
-			os.RemoveAll(copyDir)
-		})
-
-		it("does not appear in the detection environment", func() {
```
natalieparellano, line 139: We could just flip the assertion instead of deleting the entire test... but, maybe the unit tests are sufficient. If we delete this, could we also back out this change? https://github.com/buildpacks/lifecycle/pull/591/files#diff-4d22a0bdf2091cd016a29e3d9de32343eb1460b757600af402decbf49a133d9e

env/build.go
```diff
@@ -57,8 +58,12 @@ func NewBuildEnv(environ []string, platform Platform, buildpack Buildpack) *Env
 // NewDetectEnv returns an detect-time Env from the given environment.
 //
 // Only keys in the BuildEnvIncludelist will be added to the Environment.
-func NewDetectEnv(environ []string) *Env {
```
natalieparellano, line 60: Now that `NewDetectEnv` and `NewBuildEnv` don't behave any differently, we could just remove `NewDetectEnv`. It didn't exist before https://github.com/buildpacks/lifecycle/pull/591

natalieparellano: This unfortunately introduces complexity in that we'll need to also pass a `buildpack` when constructing the env. Here's where we do it for build: https://github.com/buildpacks/lifecycle/pull/591/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yml#diff-5dce70063c4b1b07e611b3fcd32da0445a0b95970028c8992beb9c955e38320fR76

This is unfortunate because the `DetectConfig` (which includes the env) is currently defined on the `Detector` (at the level of all buildpacks, see [here](https://github.com/buildpacks/lifecycle/blob/cf2d0ed8f2059b7c528b75b7da129cbe95c98618/detector.go#L119)) and now it would need to be per-buildpack. For build, I ended up changing the interface to accept the env: https://github.com/buildpacks/lifecycle/pull/591/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yml#diff-e348e4b0ed5d889979d2fb076558d56656b133c84892137fdd110002e8948ec8R63 - but this is made easier by the fact that the logic to "clear" the env already lives in the `buildpack` package (see [here](https://github.com/buildpacks/lifecycle/blob/cf2d0ed8f2059b7c528b75b7da129cbe95c98618/buildpack/build.go#L191-L198)) whereas for detect it lives in the `cmd` package (see [here](https://github.com/buildpacks/lifecycle/blob/cf2d0ed8f2059b7c528b75b7da129cbe95c98618/cmd/lifecycle/detector.go#L98-L99)). 🤷🏼

acceptance/detector_test.go
aemengo: I'm fine keeping it, if desired.

My rationale for removing was that I didn't see any other tests about environment variables being present.

env/build.go
aemengo: Fair enough. I just wanted to know how badly we wanted it 🙂

natalieparellano: Well, that's a good question. I think if I had been more vociferous in raising that question when working on the build phase, we might've concluded that the extra effort wasn't worth it. But it strikes me as a little silly if the detect and build phases worked differently in this respect. Since we have a larger goal of removing logic from `cmd` I would say let's proceed, but let me know if you feel differently.

detector_test.go
```diff
@@ -68,6 +73,10 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
 			bpD2 := testmock.NewMockBuildpack(mockCtrl)
 			bpD1 := testmock.NewMockBuildpack(mockCtrl)
 
+			for _, bp := range []*testmock.MockBuildpack{bpE1, bpA1, bpF1, bpC1, bpB1, bpG1, bpB2, bpC2, bpD2, bpD1} {
+				bp.EXPECT().SupportsAssetPackages().Return(true).AnyTimes()
+			}
```
natalieparellano, line 78: I like this!

acceptance/detector_test.go
aemengo: Removed in latest commit.

env/build.go
aemengo: The extra effort is worth it. I was just being lazy.

Removed in latest commit.

buildpack/detect_test.go
```diff
@@ -118,10 +144,12 @@ func testDetect(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("should fail if buildpacks have both a top level version and a metadata version", func() {
+			mockEnv.EXPECT().WithPlatform(platformDir).Return(append(os.Environ(), "ENV_TYPE=full"), nil)
```
natalieparellano, line 147: This always makes me sad... having to set `EXPECT` on the mock to make the test run even though it doesn't have anything to do with the expectation. That said I'm not sure if there's a way around it. Maaaybe we could make a variable called `someEnv` so that it reads `mockEnv.EXPECT().WithPlatform(platformDir).Return(append(os.Environ(), someEnv), nil)` so that it's a little clearer that we don't care about the env. Just a suggestion.

aemengo: I like it!

Resolved in latest commit.

natalieparellano (approved):

natalieparellano:
@aemengo looks great!

jabrown85 (approved):

PR #652, opened 2021-07-06
Run pack acceptance tests on pushes to main & release/*

natalieparellano:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,6 @@ jobs:
           file: ./out/tests/coverage-unit.txt
           flags: unit,os_linux
           fail_ci_if_error: true
-          verbose: true
   test-windows:
     runs-on: windows-latest
     steps:
@@ -83,7 +82,7 @@ jobs:
       - name: Run Codecov
         run: |
           codecov.exe -f .\out\tests\coverage-unit.txt -v --flag os_windows
-  build:
+  build-and-publish:
     needs:
       - test-linux
       - test-windows
@@ -91,11 +90,25 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 0
+          fetch-depth: 0 # fetch all history for all branches and tags
       - name: Set up go
         uses: actions/setup-go@v2
         with:
           go-version: '1.16'
+      - name: Set version
+        run: |
+          echo "LIFECYCLE_VERSION=$(go run tools/version/main.go)" >> version.txt
+      - uses: actions/upload-artifact@v2
+        with:
+          name: version
+          path: version.txt
+      - name: Set tag
+        run: |
+          echo "LIFECYCLE_IMAGE_TAG=$(git describe --always)" >> tag.txt
+      - uses: actions/upload-artifact@v2
+        with:
+          name: tag
+          path: tag.txt
       - name: Build
         run: |
           make clean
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -35,7 +35,6 @@ jobs:
           file: ./out/tests/coverage-unit.txt
           flags: unit,os_linux
           fail_ci_if_error: true
-          verbose: true
```
natalieparellano, line 38: Unrelated change, to suppress a warning.

.github/workflows/build.yml
```diff
@@ -113,11 +126,110 @@ jobs:
...
+  pack-acceptance-linux:
+    if: github.event_name == 'push'
+    needs: build-and-publish
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: 'buildpacks/pack'
+          path: 'pack'
+          ref: 'main'
```
jabrown85, line 151: Could this reference the latest release? Or do you explicitly want `main`?

.github/workflows/build.yml
```diff
@@ -113,11 +126,110 @@ jobs:
...
+          cat tag.txt >> $GITHUB_ENV
+      - uses: actions/download-artifact@v2
+        with:
+          name: lifecycle-linux-x86-64
+          path: pack
+      - name: Run pack acceptance
+        run: |
+          cd pack
+          git checkout $(git describe --abbrev=0 --tags) # check out the latest tag
+          LIFECYCLE_PATH="../lifecycle-v${{ env.LIFECYCLE_VERSION }}+linux.x86-64.tgz" \
```
jabrown85, line 171: Why does pack need two different references to the same lifecycle?

.github/workflows/post-release.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Set env
         run: |
           echo "LIFECYCLE_VERSION=$(echo ${{ github.event.release.tag_name }} | cut -d "v" -f2)" >> $GITHUB_ENV
-          echo "LIFECYCLE_IMAGE_TAG=$(git describe --always --dirty)" >> $GITHUB_ENV
+          echo "LIFECYCLE_IMAGE_TAG=$(git describe --always)" >> $GITHUB_ENV
```
jabrown85, line 20: Removing `--dirty` makes sense - was it there to guard producing images for a repo with uncommitted code?

jabrown85 (approved):

.github/workflows/post-release.yml
natalieparellano: Way way back when we first decided to auto build & tag lifecycle images, there was the concept of `--dirty` because we thought that would be useful in local development. But it turned out to be more confusing than useful, and we dropped it. I don't think we ever want to publish `<some-sha>-dirty` tagged images to buildpacksio, though it could happen go.sum changes are uncommitted, hence this removal.

.github/workflows/build.yml
natalieparellano: `LIFECYCLE_PATH` is used to create the builder that `pack` uses for trusted builder workflows. `LIFECYCLE_IMAGE` is for the untrusted workflows. I don't think pack lets you grab the lifecycle for a builder from a lifecycle image (though that'd be nice).

.github/workflows/build.yml
natalieparellano: I think this is made superfluous by `git checkout $(git describe --abbrev=0 --tags)`.

PR #659, opened 2021-07-09
Build lifecycle binaries and images for linux/arm64

imjasonh:
This PR makes make targets explicit about the GOARCH being used, in addition to the GOOS.

This changes `make build` to build for `linux/amd64`, `windows/amd64`, and newly, `linux/arm64`. Binaries are built to paths like `out/$GOOS-$GOARCH/lifecycle`, etc.

This also changes the CI image build process to build and attach the `amd64` binary to the build, and to include the `linux/arm64` image in the manifest list.

Acceptance tests were updated to build arch-specific binaries and to look for them in the new correct place.

Open questions:

- This PR doesn't currently change the fact that the `linux/amd64` image is pushed and tagged to `buildpacksio/lifecycle:$FOO-linux`, when it maybe should get pushed to `...linux-amd64`. We can push to both to avoid breaking people, if that's preferred.
- This PR doesn't touch the release processes in `.github/workflows/*release.yaml` -- if this PR lands and doesn't break CI, we can tackle the release changes then.

@natalieparellano

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,6 +107,10 @@ jobs:
           path: out/lifecycle-v*+linux.x86-64.tgz
       - uses: actions/upload-artifact@v2
         with:
+          name: lifecycle-linux-arm64
+          path: out/lifecycle-v*+linux.arm64.tgz
+      - uses: actions/upload-artifact@v2
+        with:
           name: lifecycle-windows-x86-64
           path: out/lifecycle-v*+windows.x86-64.tgz
       - uses: azure/docker-login@v1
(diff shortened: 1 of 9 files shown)


imjasonh:
Progress toward https://github.com/buildpacks/lifecycle/issues/435

cc @ekcasey

jabrown85 (approved):
This looks great! That `Makefile` sure is getting loooong though 😝

imjasonh:
> This looks great! That `Makefile` sure is getting loooong though 😝

I agree! I'm definitely open to ideas to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)ing it, I just didn't want those kinds of cleanups in this PR.

natalieparellano:
Sadly I don't have easy quick access to an arm machine to test this out - but it would be interesting to speculate which of the use cases described in this [Google doc](https://docs.google.com/document/d/1j2EISF1OMIrqJiM826Bm-PYJ9ycCGwdfEKHzet3xIjA/edit#heading=h.2jx1jkccw58p) would be helped by this change. Surely the lifecycle would need additional changes to output arm images (right? I'm just guessing...) but I assume we could build amd images provided that the buildpacks supplied can also run on arm.

imjasonh:
> Sadly I don't have easy quick access to an arm machine to test this out - but it would be interesting to speculate which of the use cases described in this [Google doc](https://docs.google.com/document/d/1j2EISF1OMIrqJiM826Bm-PYJ9ycCGwdfEKHzet3xIjA/edit#heading=h.2jx1jkccw58p) would be helped by this change. Surely the lifecycle would need additional changes to output arm images (right? I'm just guessing...) but I assume we could build amd images provided that the buildpacks supplied can also run on arm.

AIUI, this change just enables the lifecycle binaries to run on Arm machines. For the `build` phase, this means that the builder will execute, e.g., `go build ./cmd/foo` with the env vars `GOOS=linux GOARCH=arm64`, and produce an Arm Go binary. Likewise for Java, etc., I suspect, though I have much less experience with Arm support in non-Go language ecosystems. Builder images might have to take steps to detect they're being run on Arm and produce Arm artifacts, but at least having lifecycle able to run on Arm would be a prerequisite for that.

The `export` phase might need further modifications to emit OCI images with the correct `platform.architecture` set, but this might have already been addressed while implementing Windows support. The `analyze` and `restore` phases might need to take architecture into account when selecting resources to restore. In any case, having Arm lifecycle binaries available will let us test them out and see if anything else needs to be fixed.

.github/workflows/build.yml
```diff
@@ -118,7 +122,12 @@ jobs:
         if: github.event_name == 'push'
         run: |
           LIFECYCLE_IMAGE_TAG=$(git describe --always --dirty)
+          DOCKER_CLI_EXPERIMENTAL=enabled
           go run ./tools/image/main.go -lifecyclePath ./out/lifecycle-v*+linux.x86-64.tgz -tag buildpacksio/lifecycle:${LIFECYCLE_IMAGE_TAG}-linux
+          go run ./tools/image/main.go -lifecyclePath ./out/lifecycle-v*+linux.arm64.tgz -tag buildpacksio/lifecycle:${LIFECYCLE_IMAGE_TAG}-linux-arm64
```
natalieparellano, line 127: I think we'll need further changes to `tools/image/main.go` in order to produce an arm64 image - I'm guessing we'll need to pass `remote.WithDefaultPlatform` [here](https://github.com/buildpacks/lifecycle/blob/f15f80c6731ca9b01545eb8234281a105d3f4577/tools/image/main.go#L94) - though I'm not 100% sure how that will play with `remote.FromBaseImage`.

natalieparellano: It might make sense for "lifecycle image support" to be a separate PR...

imjasonh: Are binaries independently useful outside of being packaged as images? I was hoping to produce an image with this PR that could be evaluated in more depth to flush out bugs before a release.

I've added an `-arch` flag to `tools/image/main.go`, and confirmed that the image's config is correctly set when pushing remotely.

natalieparellano (approved):

natalieparellano:
Thanks @imjasonh for the PR!!!

Issue #656, opened 2021-07-07
Remove run image from exporter


natalieparellano:
### Description
Update exporter code to reflect spec discussion here: https://github.com/buildpacks/spec/pull/234#discussion_r665700935

natalieparellano:
Doing this as part of #662

natalieparellano:
Resolved by #662

PR #662, opened 2021-07-13
Write run image in analyzed.toml and remove -run-image from exporter for platform 0.7+

### LINKED ISSUE #655: Add run image to analyzed.toml
### Description
Update analyzer code to reflect spec discussion here: https://github.com/buildpacks/spec/pull/234#discussion_r665700935

natalieparellano:
Resolves #655 and #656

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -500,7 +500,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 						output, err := cmd.CombinedOutput()
 						h.AssertNotNil(t, err)
 
-						h.AssertStringContains(t, string(output), "failed to : ensure registry read access to some-run-image")
+						h.AssertStringContains(t, string(output), "failed to : ensure registry read access to some-run-image") // TODO: update some-run-image to have explicit permissions when https://github.com/buildpacks/lifecycle/pull/685 is merged
 					})
 
 					when("stack.toml not present", func() {
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"time"
 
@@ -209,10 +210,15 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			it("recursively chowns the directory", func() {
 				h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
 
+				var analyzeFlags []string
+				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0 {
+					analyzeFlags = append(analyzeFlags, []string{"-run-image", "some-run-image"}...)
+				}
+
 				output := h.DockerRun(t,
 					analyzeImage,
 					h.WithFlags("--env", "CNB_PLATFORM_API="+platformAPI),
-					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s some-image; ls -al /layers", analyzerPath)),
+					h.WithBash(fmt.Sprintf("chown -R 9999:9999 /layers; chmod -R 775 /layers; %s %s some-image; ls -al /layers", analyzerPath, flatPrint(analyzeFlags))),
 				)
 
 				h.AssertMatch(t, output, "2222 3333 .+ \\.")
@@ -246,11 +252,14 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("analyzed path is provided", func() {
 			it("writes analyzed.toml at the provided path", func() {
-				execArgs := []string{
-					ctrPath(analyzerPath),
-					"-analyzed", ctrPath("/some-dir/some-analyzed.toml"),
-					"some-image",
+				analyzeFlags := []string{"-analyzed", ctrPath("/some-dir/some-analyzed.toml")}
+				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0 {
+					analyzeFlags = append(analyzeFlags, []string{"-run-image", "some-run-image"}...)
 				}
+
+				var execArgs []string
+				execArgs = append([]string{ctrPath(analyzerPath)}, analyzeFlags...)
+				execArgs = append(execArgs, "some-image")
 
 				h.DockerRunAndCopy(t,
 					containerName,
@@ -268,6 +277,11 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 		it("drops privileges", func() {
 			h.SkipIf(t, runtime.GOOS == "windows", "Not relevant on Windows")
 
+			analyzeArgs := []string{"-analyzed", "/some-dir/some-analyzed.toml"}
+			if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0 {
+				analyzeArgs = append(analyzeArgs, "-run-image", "some-run-image")
+			}
+
 			output := h.DockerRun(t,
 				analyzeImage,
 				h.WithFlags(
@@ -275,8 +289,9 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					"--env", "CNB_PLATFORM_API="+platformAPI,
 				),
 				h.WithBash(
-					fmt.Sprintf("%s -analyzed /some-dir/some-analyzed.toml %s; ls -al /some-dir",
+					fmt.Sprintf("%s %s %s; ls -al /some-dir",
 						ctrPath(analyzerPath),
+						flatPrint(analyzeArgs),
 						noAuthRegistry.RepoName("some-image"),
 					),
 				),
@@ -286,26 +301,82 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 		})
 
 		// TODO: add with or after https://github.com/buildpacks/lifecycle/pull/646
-		when.Pend("run image", func() {
+		when("run image", func() {
 			when("provided", func() {
-				it("validates read access", func() {
+				it("is recorded in analyzed.toml", func() {
 					h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not accept run image")
+
+					h.DockerRunAndCopy(t,
+						containerName,
+						copyDir,
+						ctrPath("/layers/analyzed.toml"),
+						analyzeImage,
+						h.WithFlags(
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+						),
+						h.WithArgs(ctrPath(analyzerPath), "-run-image", "some-run-image", "some-image"),
+					)
+
+					analyzedMD := assertAnalyzedMetadata(t, filepath.Join(copyDir, "analyzed.toml"))
+					h.AssertEq(t, analyzedMD.RunImage.Reference, "some-run-image")
 				})
 			})
 
 			when("not provided", func() {
 				it("falls back to CNB_RUN_IMAGE", func() {
 					h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not accept run image")
+
+					h.DockerRunAndCopy(t,
+						containerName,
+						copyDir,
+						ctrPath("/layers/analyzed.toml"),
+						analyzeImage,
+						h.WithFlags(
+							"--env", "CNB_PLATFORM_API="+platformAPI,
+							"--env", "CNB_RUN_IMAGE=some-run-image",
+						),
+						h.WithArgs(ctrPath(analyzerPath), "some-image"),
+					)
+
+					analyzedMD := assertAnalyzedMetadata(t, filepath.Join(copyDir, "analyzed.toml"))
+					h.AssertEq(t, analyzedMD.RunImage.Reference, "some-run-image")
 				})
 
 				when("CNB_RUN_IMAGE not provided", func() {
 					it("falls back to stack.toml", func() {
 						h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not accept run image")
+
+						h.DockerRunAndCopy(t,
+							containerName,
+							copyDir,
+							ctrPath("/layers/analyzed.toml"),
+							analyzeImage,
+							h.WithFlags(
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+							),
+							h.WithArgs(ctrPath(analyzerPath), "-stack", "/cnb/platform-0.7-stack.toml", "some-image"),
+						)
+
+						analyzedMD := assertAnalyzedMetadata(t, filepath.Join(copyDir, "analyzed.toml"))
+						h.AssertEq(t, analyzedMD.RunImage.Reference, "some-run-image")
 					})
 
 					when("stack.toml not present", func() {
-						it("ignores run image", func() {
+						it("errors", func() {
 							h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 does not accept run image")
+
+							cmd := exec.Command(
+								"docker", "run", "--rm",
+								"--env", "CNB_PLATFORM_API="+platformAPI,
+								analyzeImage,
+								ctrPath(analyzerPath),
+								"some-image",
+							) // #nosec G204
+							output, err := cmd.CombinedOutput()
+
+							h.AssertNotNil(t, err)
+							expected := "-run-image is required when there is no stack metadata available"
+							h.AssertStringContains(t, string(output), expected)
 						})
 					})
 				})
@@ -344,11 +415,14 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("daemon case", func() {
 			it("writes analyzed.toml", func() {
-				execArgs := []string{
-					ctrPath(analyzerPath),
-					"-daemon",
-					"some-image",
+				analyzeFlags := []string{"-daemon"}
+				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0 {
+					analyzeFlags = append(analyzeFlags, []string{"-run-image", "some-run-image"}...)
 				}
+
+				var execArgs []string
+				execArgs = append([]string{ctrPath(analyzerPath)}, analyzeFlags...)
+				execArgs = append(execArgs, "some-image")
 
 				h.DockerRunAndCopy(t,
 					containerName,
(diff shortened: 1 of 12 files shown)


natalieparellano:
TODO: Check that previous image and output image are on the same registry in analyzer

natalieparellano:
Some parts of `cmd` are still untested pending changes from https://github.com/buildpacks/lifecycle/pull/646

acceptance/analyzer_test.go
```diff
@@ -53,10 +54,13 @@ type analyzeFixtures struct {
 	authRegAuthConfig     string
 	authRegCacheImage     string
 	authRegOtherAppImage  string
+	authRegRunImage       string
 	daemonAppImage        string
 	daemonCacheImage      string
+	daemonRunImage        string
```
natalieparellano, line 60: This might not be needed

cmd/lifecycle/exporter.go
```diff
@@ -129,27 +129,70 @@ func (e *exportCmd) Args(nargs int, args []string) error {
...
-	e.analyzedMD, err = parseOptionalAnalyzedMD(cmd.DefaultLogger, e.analyzedPath)
+	e.targetRegistry, err = parseRegistry(e.imageNames[0])
 	if err != nil {
-		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "parse analyzed metadata")
+		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "parse target registry")
+	}
+
+	// run image
+
+	if !e.supportsRunImage() { // TODO: break out into separate function?
```
natalieparellano, line 149: Would love opinions here

aemengo: I'm a big fan of `naked switch` statements for when clauses become complex like here.

But (_once again, stepping back_), I'm a little more concerned about having to handle these "differences across platform APIs" in unique ways. A separate release branch per API version, let's say, would have it's own problems but is easier grasp at a high-level.

```suggestion
	// ...
	return e.populateRunImageRefIfNeeded()
}

func (e *exportCmd) populateRunImageRefIfNeeded() error {
	failError := func(msg string) error {
		return cmd.FailErrCode(errors.New(msg), cmd.CodeInvalidArgs, "parse arguments")
	}

	switch {
	case e.supportsRunImage() && e.runImageRef != "":
		return nil
	case e.supportsRunImage() && e.runImageRef == "":
		runImageRef, err := e.stackMD.BestRunImageMirror(e.targetRegistry)
		if err != nil {
			return failError("-run-image is required when there is no stack metadata available")
		}

		e.runImageRef = runImageRef
		return nil
	case !e.supportsRunImage() && e.runImageRef == "":
		if e.analyzedMD.RunImage == nil || e.analyzedMD.RunImage.Reference == "" {
			return failError("run image not found in analyzed metadata")
		}

		e.runImageRef = e.analyzedMD.RunImage.Reference
		return nil
	case !e.supportsRunImage() && e.runImageRef != "":
		if e.deprecatedRunImageRef != "" {
			return failError("-image is unsupported")
		}

		return failError("-run-image is unsupported")
	default:
		// this code path shouldn't be reached
		// panic("unhandled case")
	}
}
```

natalieparellano: > A separate release branch per API version, let's say, would have it's own problems but is easier grasp at a high-level

I believe @jabrown85 is of a similar mindset. Let's continue to discuss as a team!

natalieparellano: @aemengo thanks for the suggestion! I made some changes in e140e15 - LMK what you think!

aemengo: looks mint! 🌿

platform/v07/platform.go
```diff
@@ -5,22 +5,22 @@ import (
 	"github.com/buildpacks/lifecycle/platform/common"
 )
 
-type Platform struct {
+type v07Platform struct {
```
yaelharel, line 8: Why did you rename all of these structures (`pre06Platform`, `v06Platform` and `v07Platform`)?
I think the previous name was better since the name of the package is based on the platform API (`v07` in this case).

In addition, not specifically related to this PR, but I saw in `platform/platform.go` that we create all of the platform even if we don't need them. I don't think it's a good practice although in the current situation it doesn't really matter because each of these structures is pretty small.

cmd/lifecycle/analyzer.go
```diff
@@ -225,6 +264,30 @@ func (a *analyzeCmd) restoresLayerMetadata() bool {
...
+	if !a.supportsRunImage() && a.runImageRef != "" {
+		return errors.New("-run-image is unsupported")
+	}
+	return nil
+}
+
+func (a *analyzeCmd) populateRunImageIfNeeded(stackMD platform.StackMetadata, targetRegistry string) error {
+	if !a.supportsRunImage() || a.runImageRef != "" {
+		return nil
+	}
```
yaelharel, line 281: I think that adding a function with a meaningful name or even just adding a comment for this if condition will make the code clearer.

cmd/lifecycle/analyzer.go
```diff
@@ -214,6 +244,15 @@ func (aa analyzeArgs) analyze() (platform.AnalyzedMetadata, error) {
 	if err != nil {
 		return platform.AnalyzedMetadata{}, cmd.FailErrCode(err, aa.platform.CodeFor(cmd.AnalyzeError), "analyzer")
 	}
+
+	if aa.runImageRef != "" {
+		ref, err := name.ParseReference(aa.runImageRef, name.WeakValidation)
+		if err != nil {
+			return platform.AnalyzedMetadata{}, cmd.FailErrCode(err, aa.platform.CodeFor(cmd.AnalyzeError), "parse reference for run image")
+		}
```
yaelharel, line 252: This code is very similar to the `parseRegistry` function.
Do you think it's worth creating another helper function that we'll use both here and inside the `parseRegistry` function?

cmd/lifecycle/creator.go
```diff
@@ -300,3 +301,16 @@ func (c *createCmd) WriteableRegistryImages() []string {
 	}
 	return writeableImages
 }
+
+func (c *createCmd) populateRunImageIfNeeded() error {
```
yaelharel, line 305: This is almost the exact function that we have in the `analyzer`. Can we move it to a common place?

cmd/lifecycle/analyzer.go
```diff
@@ -104,15 +119,30 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
...
+	}
+
+	if err := a.populateRunImageIfNeeded(stackMD, targetRegistry); err != nil {
+		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "populate run image")
 	}
 
 	return nil
 }
 
+func parseRegistry(providedRef string) (string, error) {
```
yaelharel, line 138: I'm not sure that this is the best file to put this function since we're also using it in the `creator` and the `exporter`.

cmd/lifecycle/exporter.go
```diff
@@ -129,27 +129,49 @@ func (e *exportCmd) Args(nargs int, args []string) error {
...
+	}
+
+	if err := e.populateRunImageRefIfNeeded(); err != nil {
+		return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "populate run image")
 	}
 
 	return nil
 }
 
+func readStack(stackPath string) (platform.StackMetadata, error) {
```
yaelharel, line 154: Same comment as https://github.com/buildpacks/lifecycle/pull/662/files#r677775980.

cmd/lifecycle/exporter.go
```diff
@@ -129,27 +129,49 @@ func (e *exportCmd) Args(nargs int, args []string) error {
...
+	var (
+		stackMD platform.StackMetadata
+		err     error
```
yaelharel, line 157: Why do we need to define the error here?
I think you can just define it as part of the if condition on line 160.

cmd/lifecycle/exporter.go
```diff
@@ -205,6 +227,35 @@ func (e *exportCmd) registryImages() []string {
...
+		e.runImageRef, err = e.stackMD.BestRunImageMirror(e.targetRegistry)
+		if err != nil {
+			return errors.New("-run-image is required when there is no stack metadata available")
+		}
+	}
+	return nil
+}
+
+func (e *exportCmd) validateRunImageInput() error {
+	switch {
```
yaelharel, line 247: Maybe it's worth adding the relevant image names to the error messages in this switch-case.

acceptance/analyzer_test.go
```diff
@@ -370,21 +386,24 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("analyzed path is provided", func() {
 			it("uses the provided analyzed path", func() {
-				execArgs := []string{
-					ctrPath(analyzerPath),
-					"-analyzed", ctrPath("/some-dir/some-analyzed.toml"),
-					fixtures.authRegAppImage,
+				analyzeFlags := []string{"-analyzed", ctrPath("/some-dir/some-analyzed.toml")}
+				if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0 {
+					analyzeFlags = append(analyzeFlags, []string{"-run-image", fixtures.authRegRunImage}...)
```
yaelharel, line 391: Is there a reason why you're appending an array instead of appending each element (like you're doing on line 420)?
```
analyzeFlags = append(analyzeFlags, "-run-image", fixtures.authRegRunImage)
```

acceptance/analyzer_test.go
```diff
@@ -395,63 +414,117 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+							"--network", registryNetwork,
+							analyzeImage,
+							ctrPath(analyzerPath),
+							"-stack", "/cnb/platform-0.7-stack.toml",
+							authRegistry.RepoName("some-image"),
+						) // #nosec G204
+						output, err := cmd.CombinedOutput()
+						h.AssertNotNil(t, err)
+
+						h.AssertStringContains(t, string(output), "failed to : ensure registry read access to some-run-image")
```
yaelharel, line 503: I'm confused. Can you please explain why we expect an error in this case?

acceptance/analyzer_test.go
```diff
@@ -499,6 +575,16 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 			when("app image exists", func() {
 				it("does not restore app metadata", func() {
 					h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 restores app metadata")
+
+					analyzeFlags := []string{"-daemon"}
+					if api.MustParse(platformAPI).Compare(api.MustParse("0.7")) >= 0 {
+						analyzeFlags = append(analyzeFlags, []string{"-run-image", "some-run-image"}...)
```
yaelharel, line 581: I think you don't need to wrap this with the if condition because we skip the test for platformAPI < 0.
So you can just initialize the flags with both the `-daemon` and the `-run-image` flags.

acceptance/analyzer_test.go
```diff
@@ -499,6 +575,16 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+					}
+
+					var execArgs []string
+					execArgs = append([]string{ctrPath(analyzerPath)}, analyzeFlags...)
+					execArgs = append(execArgs, fixtures.daemonAppImage)
```
yaelharel, line 586: why don't we initialize the `execArgs` in one line?

platform/v07/platform.go
natalieparellano: I found that having the concrete platform have the same name as the interface was confusing. Additionally, there isn't a need for the struct to be public.

cmd/lifecycle/analyzer.go
natalieparellano: What would you call this function?

cmd/lifecycle/analyzer.go
natalieparellano: We can't use `parseRegistry` because that returns the registry, not the reference. What would such a helper look like? I'm not sure how to reduce the duplication. It seems to me that `name.ParseReference` is doing most of the heavy lifting.

cmd/lifecycle/creator.go
natalieparellano: We can't have a common function because `populateRunImageIfNeeded()` updates the receiving struct. 

Eventually, as we move logic out of the `cmd` package, we might have the `<some phase>Cmd` structs implement an interface, and this could be one method on it.

cmd/lifecycle/analyzer.go
natalieparellano: This is true. However we have other common functions that happen to land in one phase or another. We might make a "common.go" file (or something) to hold helper functions that are shared. But I'd rather do that in a separate PR.

cmd/lifecycle/exporter.go
natalieparellano: https://github.com/buildpacks/lifecycle/pull/662/files#r689753934

cmd/lifecycle/exporter.go
natalieparellano: That feels a bit messy to me. I'm not sure having the image names would add much, as the command invocation is already known. I kind of like the terse error message...

acceptance/analyzer_test.go
natalieparellano: Nope! Thanks for pointing that out.

acceptance/analyzer_test.go
natalieparellano: I added a comment on line 497. Does that help?

acceptance/analyzer_test.go
natalieparellano: Good catch!

acceptance/analyzer_test.go
natalieparellano: It's so that we could splat `analyzeFlags`.

cmd/lifecycle/exporter.go
natalieparellano: Updated in 07e41be

acceptance/analyzer_test.go
natalieparellano: Updated in 07e41be

acceptance/analyzer_test.go
natalieparellano: Updated in 07e41be

acceptance/analyzer_test.go
natalieparellano: Updated in 07e41be

cmd/lifecycle/analyzer.go
```diff
@@ -88,8 +89,22 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
...
+	}
+
 	if a.previousImageRef == "" {
 		a.previousImageRef = a.outputImageRef
+	} else if !a.useDaemon {
+		previousRegistry, err := parseRegistry(a.previousImageRef)
+		if err != nil {
+			return cmd.FailErrCode(err, cmd.CodeInvalidArgs, "parse previous registry")
+		}
+		if previousRegistry != targetRegistry {
```
jabrown85, line 104: Why do we need to fail in this case? I'm thinking about a situation where the platform decides to move registries from ECR to GGCR.

cmd/lifecycle/creator.go
jabrown85:
```suggestion
func (c *createCmd) populateRunImage() error {
```

nit: I think dropping the `IfNeeded` reads better to me.

cmd/lifecycle/exporter.go
```diff
@@ -205,6 +226,35 @@ func (e *exportCmd) registryImages() []string {
 	return registryImages
 }
 
+func (e *exportCmd) populateRunImageRefIfNeeded() error {
```
jabrown85, line 229:
```suggestion
func (e *exportCmd) populateRunImageRef error {
```

cmd/lifecycle/exporter.go
```diff
@@ -313,13 +363,13 @@ func (ea exportArgs) initRemoteAppImage(analyzedMD platform.AnalyzedMetadata) (i
...
-		ref, err := name.ParseReference(analyzedMD.Image.Reference, name.WeakValidation)
+		// ensure previous image is on same registry as output image
+		analyzedRegistry, err := parseRegistry(analyzedMD.Image.Reference)
 		if err != nil {
 			return nil, "", cmd.FailErr(err, "parse analyzed registry")
 		}
-		analyzedRegistry := ref.Context().RegistryStr()
-		if analyzedRegistry != ea.registry {
-			return nil, "", fmt.Errorf("analyzed image is on a different registry %s from the exported image %s", analyzedRegistry, ea.registry)
+		if analyzedRegistry != ea.targetRegistry {
```
jabrown85, line 371: I know this wasn't a change, but why does this case produce a hard failure?

natalieparellano: I **think** because of layer re-use. Though I seem to recall @ekcasey mentioned we might be able to copy the layers in the future.

cmd/lifecycle/analyzer.go
natalieparellano: Same answer as here. The failure was always happening in the exporter but I moved it up in the spirit of "failing fast".

cmd/lifecycle/creator.go
natalieparellano: Please see f2ec72b

acceptance/analyzer_test.go
yaelharel: Are we supposed to get a failure in this test case? According to the test name, we're supposed to "falls back to stack.toml" and not "error", no?

cmd/lifecycle/exporter.go
yaelharel: mmmm...I don't have strong opinion on this but I just thought that printing a clear message with as much information as we can would be useful in case of an error.
If you feel differently, please feel free to mark this conversation as resolved.

cmd/lifecycle/analyzer.go
yaelharel: I don't remember what my thoughts were when I wrote this comment 🤷‍♀️.
Reading this again, I think that the code is clear enough.

yaelharel (approved):
I've closed most of the conversations that left open.
The last question that I still have is: https://github.com/buildpacks/lifecycle/pull/662#discussion_r689823492.
Other than that, the code looks great!

acceptance/analyzer_test.go
natalieparellano: It falls back to stack.toml, but it fails because the run image that stack.toml references doesn't exist.

cmd/lifecycle/exporter.go
natalieparellano: If there's no strong opinion, I'll mark as resolved for now to avoid making more changes :)

acceptance/analyzer_test.go
natalieparellano: I'll add a TODO to update this test once we have @jjbustamante's test helper improvements. That will make it clearer.

natalieparellano: Please see 20b0608

yaelharel: I understand that it fails because the run image doesn't exist. My point was that it's confusing that we see a failure in a test that is supposed to pass.
But I'm totally fine with waiting for Juan's test helper improvements.
Thanks!

jabrown85 (approved):

PR #667, opened 2021-07-20
Print the output when decoding the plan toml file of a buildpack fails during detect

### LINKED ISSUE #663: Detect should print buildpack output when parsing plan failed
### Summary
If the lifecycle fails to parse the plan output by a buildpack, we return an error but we do not set the output: https://github.com/buildpacks/lifecycle/blob/3297dacad3f0d58bcbbb44dff0d896b9ad1c28c1/buildpack/detect.go#L79-L90 (notice that we don’t set `Output: out.Bytes()` on line 89). This makes debugging difficult. We should set the output if parsing the plan fails.

---

##### Steps

1. Run detect on a buildpack with a `bin/detect` that succeeds, but outputs an invalid plan.toml
2. Observe that buildpack output is not printed (even with verbose logging)

yaelharel:
Fixes #663

### DIFF
--- a/acceptance/detector_test.go
+++ b/acceptance/detector_test.go
@@ -122,6 +122,30 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 
+	when("there is a buildpack with a bad toml file", func() {
+		it("errors and prints the output", func() {
+			command := exec.Command(
+				"docker",
+				"run",
+				"--rm",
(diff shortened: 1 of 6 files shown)


natalieparellano:
Given that the code changes are all in the `buildpack` package, I'm wondering why we need an acceptance test for this? (Typically the acceptance tests cover things like flag parsing and handoff between `cmd` and `lifecycle` packages...)

natalieparellano (approved):

PR #668, opened 2021-07-21
Refactor analyzer fixtures

natalieparellano:
We're not modifying the image fixtures, so they could be created once per suite instead of once per `it`. This will make test organization easier.

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -11,7 +11,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"strings"
 	"testing"
 	"time"
 
@@ -179,9 +178,9 @@ func (a analyzeFixtures) removeAll(t *testing.T) {
 	h.DockerImageRemove(t, fixtures.daemonCacheImage)
 
 	// remove images that were built locally before being pushed to test registry
-	h.DockerImageRemove(t, strings.Split(fixtures.authRegAppImage, "/")[1])
-	h.DockerImageRemove(t, strings.Split(fixtures.authRegCacheImage, "/")[1])
-	h.DockerImageRemove(t, strings.Split(fixtures.authRegOtherAppImage, "/")[1])
+	h.DockerImageRemove(t, fixtures.authRegAppImage)
+	h.DockerImageRemove(t, fixtures.authRegCacheImage)
+	h.DockerImageRemove(t, fixtures.authRegOtherAppImage)
 }
 
 func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spec.S) {
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"time"
 
@@ -37,7 +38,19 @@ var (
 	noAuthRegistry       *ih.DockerRegistry
 	authRegistry         *ih.DockerRegistry
 	registryNetwork      string
+	fixtures             analyzeFixtures
 )
+
+type analyzeFixtures struct {
+	authRegAppImage       string
+	authRegAuthConfig     string
+	authRegCacheImage     string
+	authRegOtherAppImage  string
+	daemonAppImage        string
+	daemonCacheImage      string
+	readOnlyRegAppImage   string
+	readOnlyRegCacheImage string
+}
 
 func TestAnalyzer(t *testing.T) {
 	rand.Seed(time.Now().UTC().UnixNano())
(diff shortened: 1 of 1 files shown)


acceptance/analyzer_test.go
```diff
@@ -532,13 +564,13 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 										h.WithFlags(append(
 											dockerSocketMount,
 											"--network", registryNetwork,
-											"--env", "CNB_REGISTRY_AUTH="+cacheAuthConfig,
+											"--env", "CNB_REGISTRY_AUTH="+fixtures.authRegAuthConfig,
```
yaelharel, line 566: Before this change we used the second return value from `buildAuthRegistryImage` (which is the `authConfig` instead of the `regRepoName`).

natalieparellano: Since authConfig is per registry, not per repo, we only need to set it once :)

acceptance/analyzer_test.go
```diff
@@ -89,11 +102,88 @@ func TestAnalyzer(t *testing.T) {
...
+		regCacheName,
+		filepath.Join("testdata", "analyzer", "cache-image"),
+		"--build-arg", "fromImage="+containerBaseImage,
+		"--build-arg", "metadata="+cacheMeta,
+	)
+	fixtures.readOnlyRegCacheImage = noAuthRegistry.RepoName(regCacheName)
+
+	// Daemon
+
+	fixtures.daemonAppImage = "some-app-image-" + h.RandString(10)
```
yaelharel, line 151:
```suggestion
	fixtures.daemonAppImage = regAppName
```
Maybe it's worth renaming `regAppName` to something like `someAppImageName`.

acceptance/analyzer_test.go
```diff
@@ -89,11 +102,88 @@ func TestAnalyzer(t *testing.T) {
...
+		"docker",
+		"build",
+		"-t", fixtures.daemonAppImage,
+		"--build-arg", "fromImage="+containerBaseImage,
+		"--build-arg", "metadata="+appMeta,
+		filepath.Join("testdata", "analyzer", "app-image"),
+	) // #nosec G204
+	h.Run(t, cmd)
+
+	fixtures.daemonCacheImage = "some-cache-image-" + h.RandString(10)
```
yaelharel, line 162:
```suggestion
	fixtures.daemonCacheImage = regCacheName
```
Same comment about renaming as [here](https://github.com/buildpacks/lifecycle/pull/668/files#r674153652).

acceptance/analyzer_test.go
```diff
@@ -740,13 +736,13 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 								ctrPath("/layers"),
 								analyzeImage,
 								h.WithFlags(
-									"--env", "CNB_REGISTRY_AUTH="+appAuthConfig,
+									"--env", "CNB_REGISTRY_AUTH="+fixtures.authRegAuthConfig,
```
yaelharel, line 738: Same comment as https://github.com/buildpacks/lifecycle/pull/668/files#r674148205.

acceptance/analyzer_test.go
```diff
@@ -787,13 +783,13 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 								analyzeImage,
 								h.WithFlags(
 									"--network", registryNetwork,
-									"--env", "CNB_REGISTRY_AUTH="+appAuthConfig,
+									"--env", "CNB_REGISTRY_AUTH="+fixtures.authRegAuthConfig,
```
yaelharel, line 785: Same comment as https://github.com/buildpacks/lifecycle/pull/668/files#r674148205.

acceptance/analyzer_test.go
```diff
@@ -876,32 +852,11 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
-							t,
-							"some-app-image-"+h.RandString(10),
-							filepath.Join("testdata", "analyzer", "app-image"),
-							"--build-arg", "fromImage="+containerBaseImage,
-							"--build-arg", "metadata="+metadata,
-						)
-
-						authRegAppOtherImage, appAuthConfig = buildAuthRegistryImage(
-							t,
-							"some-app-image-"+h.RandString(10),
```
yaelharel, line 893: As we can see here, the app name was called `some-app-image` and now we changed it to `some-other-app-image`.
I think there is no problem in this change, but just wanted to double check with you that you're aware of it.

acceptance/analyzer_test.go
```diff
@@ -979,13 +918,13 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 									ctrPath("/layers"),
 									analyzeImage,
 									h.WithFlags(
-										"--env", "CNB_REGISTRY_AUTH="+cacheAuthConfig,
+										"--env", "CNB_REGISTRY_AUTH="+fixtures.authRegAuthConfig,
```
yaelharel, line 920: This is being defined with `appMeta` while before, `cacheAuthConfig` was defined with `cacheMeta`.

yaelharel: If I'm not mistaken, we should use the second return value from line 140.

acceptance/analyzer_test.go
yaelharel, line 566: I think we should use here the second return value from line 140.

yaelharel: Marking as solved following https://github.com/buildpacks/lifecycle/pull/668#discussion_r674152756.

acceptance/analyzer_test.go
yaelharel: Marking as solved following https://github.com/buildpacks/lifecycle/pull/668#discussion_r674152756.

yaelharel (approved):
This refactor is great!
Left a few comments but I'm also ok to merge it as is.

PR #679, opened 2021-07-22
Generate checksums during package

### LINKED ISSUE #674: Generate checksums on release
### Description
We should generate a file with the sha 256 of the lifecycle on release. pack is currently doing this: https://github.com/buildpacks/pack/releases/tag/v0.20.0

Question: should we take the shasum of the tgz file, or of the binary itself? It seems that both could be useful, depending on the context.

yaelharel:
Fixes #674.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -22,7 +22,7 @@ jobs:
             exit 1
           fi
           echo "LIFECYCLE_VERSION=$version" >> $GITHUB_ENV
-      - name: Determine download urls for linux and windows
+      - name: Determine download urls for linux-x86-64 and windows
         id: artifact-urls
         uses: actions/github-script@v3.0.0
         with:
@@ -72,11 +72,11 @@ jobs:
                   .map(artifact => [artifact.name, artifact.archive_download_url]);
                 let urlMap = new Map();
                 tuples.forEach(function(tuple) {
-                  if (tuple[0].includes("linux")) {
+                  if (tuple[0].includes("linux-x86-64")) {
                     if (tuple[0].includes("sha256")) {
-                      urlMap.set("linux_sha256", tuple[1])
+                      urlMap.set("linux_x86_64_sha256", tuple[1])
                     } else {
-                      urlMap.set("linux", tuple[1])
+                      urlMap.set("linux_x86_64", tuple[1])
                     }
                   }
                   if (tuple[0].includes("windows")) {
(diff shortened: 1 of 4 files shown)


yaelharel:
Notes:
* I still need to edit `draft-release.yml`.
* The new "sha" files don't contain the files names as in pack because pack is using the `shasum -a 256` command (and its output contains the file name) while lifecycle is doing it in `tools/packager/main.go`.
```
### pack:
cat pack-v0.20.0-linux.tgz.sha256 
1d879c8f99130b30e080153b1025ee51cfeddfb41612e5c257a89d3a358fc3c2  pack-v0.20.0-linux.tgz
### lifecycle
cat lifecycle-v0.11.2-172+6519163e+linux.x86-64.tgz.sha256 
05f86913fb7ec92e31f15126888b1636794be065ff1de044ff3ed08e1554da3e
``` 
I can add the file name to the file itself but I don't think it's necessary because the user can see it from the file name itself.

tools/packager/main.go
```diff
@@ -112,6 +112,45 @@ func doPackage() error {
...
+	if err != nil {
+		return errors.Wrap(err, "Failed to close gzip writer")
+	}
+
+	err = f.Close()
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("Failed to close -archivePath %s", archivePath))
+	}
+
+	f, err = os.OpenFile(archivePath, os.O_RDONLY, 0777)
```
natalieparellano, line 130: Could we calculate the sha as we're writing the file, so that it's faster?

yaelharel: Thank you for your suggestion.
I have made this change in [this commit](https://github.com/buildpacks/lifecycle/pull/679/commits/f64a1f7c20c52f6fa88ca11b5dcd0ed108155b0a).
I don't think that the performance really matters in this case because in both cases the calculation was very fast, but it appears that this is the convention.

yaelharel:
The Github Actions part was tested on my fork:
* [build](https://github.com/yaelharel/lifecycle/actions/runs/1083516509)
* [draft-release](https://github.com/yaelharel/lifecycle/actions/runs/1083530040)
* [post-release](https://github.com/yaelharel/lifecycle/actions/runs/1083531368)

.github/workflows/draft-release.yml
```diff
@@ -73,17 +73,25 @@ jobs:
                 let urlMap = new Map();
                 tuples.forEach(function(tuple) {
                   if (tuple[0].includes("linux")) {
-                    urlMap.set("linux", tuple[1])
+                    if (tuple[0].includes("sha256")) {
```
natalieparellano, line 76: I think we need to additionally check the architecture in the line above (see https://github.com/buildpacks/lifecycle/issues/682).

yaelharel: Do you want to merge https://github.com/buildpacks/lifecycle/pull/680/files before?

natalieparellano: I don't think we could merge that one before we have a machine to test it. But we could just copy the fix over from there. Anyway this PR introduces a new problem in that the sha file could also be amd64 or arm64.

yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/679/commits/b02eff553070a649a8f38a8bebb1e1336dbb43be.
I'm keeping this conversation open since I'm not sure that I've made all of the changes that you meant.
Please let me know what do you think. Thanks!

natalieparellano:
Verified with https://github.com/yaelharel/lifecycle/releases/tag/v0.99999.0 (LOL version `0.99999.0` - I've been there!)

I think the pack .sha256 files have that formatting because they're using `shasum`. I don't know what is the value of having the filename as part of the file. Maybe we'd want to add a newline (so that it's more readable in a terminal). @jabrown85 do you have any strong preference here?

Worth noting that we'll be using new tags for the linux x86-64 lifecycle image (`linux-x86-64` vs `linux`), but I don't think that matters because the endpoint for the manifest list hasn't changed.

```
$  cat lifecycle-v0.99999.0+linux.x86-64.tgz.sha256
37deb568d2af8ab3f0d475766a19ba0e1420d3e3067a11ed8ce5ce502a51b309
$  shasum -a 256 lifecycle-v0.99999.0+linux.x86-64.tgz  
37deb568d2af8ab3f0d475766a19ba0e1420d3e3067a11ed8ce5ce502a51b309  lifecycle-v0.99999.0+linux.x86-64.tgz

$  cat lifecycle-v0.99999.0+windows.x86-64.tgz.sha256
ec041437b43567c4734df3f0832b4d2af14b5d75801db4a4eccd2003cf8378e5
$  shasum -a 256 lifecycle-v0.99999.0+windows.x86-64.tgz  
ec041437b43567c4734df3f0832b4d2af14b5d75801db4a4eccd2003cf8378e5  lifecycle-v0.99999.0+windows.x86-64.tgz

$  docker run yaelharel/lifecycle:0.99999.0 /cnb/lifecycle/lifecycle -version
0.99999.0+92e9e8d
$  docker run -v lifecycle:/cnb/lifecycle yaelharel/lifecycle:0.99999.0 /cnb/lifecycle/lifecycle -version
0.99999.0+92e9e8d
```

natalieparellano (approved):

yaelharel:
@natalieparellano @jabrown85, thanks for your feedback.
Natalie, you're right. As I wrote in [this comment](https://github.com/buildpacks/lifecycle/pull/679#issuecomment-885227460):

> The new "sha" files don't contain the files names as in pack because pack is using the shasum -a 256 command (and its output contains the file name) while lifecycle is doing it in tools/packager/main.go.

I slightly prefer having a separate sha file for every os, so I'm going to add the newline as you requested.
Jesse, please let me know if you change your mind and you prefer that we'll have a single sha file and I'll make the change accordingly.

yaelharel:
@natalieparellano @jabrown85, please see https://github.com/buildpacks/lifecycle/pull/679/commits/caaeff2fa6b6c6fe4085998b54233f0db29d08d4 for the new line addition.
I didn't re-run the release process on my repository because I don't think it's necessary. If you think differently, please let me know and I'll do so.
Thanks!

PR #684, opened 2021-07-28
Keep connection opened by ping

matejvasek:
`podman` (alternative daemon docker dameon)  is closing connections after 20 seconds without any request. This is causing troubles when `build phase` takes longer than that and builder is trusted.

If builder is trusted then docker client's connection is established at the benign while process is still uid zero. Then the connection is supposed to be kept opened. However `podman` is closing it.

This commit sends `ping` request to the daemon each second while `build phase` is in progress.

**Steps to reproduce the issue**:
1. create simple Java app: `mvn archetype:generate -DarchetypeGroupId=org.apache.maven.archetypes -DarchetypeArtifactId=maven-archetype-quickstart -DarchetypeVersion=1.4 -DgroupId=com.example -DartifactId=test-bp -B`
2. in one terminal run `podman` as a service: `podman  system service unix:///tmp/podman.sock --time=0 --log-level=debug`
3. in another terminal set docker socket via envvar: `export DOCKER_HOST="unix:///tmp/podman.sock"` and then try to build the app: `pack build test-bp --builder paketobuildpacks/builder:full --docker-host=inherit --trust-builder=true --pull-policy=if-not-present -v`

The build fails:
```
ERROR: failed to  image: error during connect: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.40/info": EOF
ERROR: failed to build: executing lifecycle: failed with status code: 1
```
You can also noticed following in the `podman`'s log:
```
DEBU[0045] IdleTracker 0xc000590010:closed 3m+1h/4t connection(s) 
DEBU[0045] IdleTracker 0xc0000111f0:closed 2m+1h/4t connection(s) 
```

### DIFF
--- a/cmd/lifecycle/creator.go
+++ b/cmd/lifecycle/creator.go
@@ -1,7 +1,9 @@
 package main
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	"github.com/docker/docker/client"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -240,6 +242,8 @@ func (c *createCmd) Exec() error {
 		}
 	}
 
+	// send pings to docker daemon while BUILDING to prevent connection closure
+	stopPinging := startPinging(c.docker)
 	cmd.DefaultLogger.Phase("BUILDING")
 	err = buildArgs{
 		buildpacksDir: c.buildpacksDir,
@@ -248,6 +252,8 @@ func (c *createCmd) Exec() error {
 		platform:      c.platform,
 		platformDir:   c.platformDir,
 	}.build(group, plan)
+	stopPinging()
+
 	if err != nil {
 		return err
 	}
@@ -313,4 +319,32 @@ func (c *createCmd) populateRunImage() error {
 		return errors.New("-run-image is required when there is no stack metadata available")
 	}
 	return nil
+}
+
+func startPinging(docker client.CommonAPIClient) (stopPinging func()) {
+	pingCtx, cancelPing := context.WithCancel(context.Background())
+	pingDoneChan := make(chan struct{})
+	go func() {
+		defer func() { close(pingDoneChan) }()
+
+		if docker == nil {
+			return
+		}
+		for {
+			select {
+			case <-time.After(time.Millisecond * 500):
+				_, err := docker.Ping(pingCtx)
+				if err != nil && !errors.Is(err, context.Canceled) {
+					cmd.DefaultLogger.Warnf("ping error: %v", err)
+				}
+			case <-pingCtx.Done():
+				return
+			}
+		}
+	}()
+
+	return func() {
+		cancelPing()
+		<-pingDoneChan
+	}
 }


aemengo:
Is this `ping` request something that the docker daemon presently does? I'm just _slightly_ perturbed about how idiosyncratic this request feels.

Can I also ask the problem that podman was trying to solve with this connection timeout?

matejvasek:
@aemengo I know this looks awful. I also filled PR https://github.com/containers/podman/pull/11072 
This is kinda workaround patching another workaround IMHO. The container is deescalating rights at some point, that leads to possible inability to dial unix socket. So the socket is dialed at beginning (while uid=0) and it should be kept open. The thing is that `docker` is not automatically closing unused connections while `podman` does close unused connection after 20sec.

matejvasek:
>The thing is that docker is not automatically closing unused connections

Also I am not sure if one can relay on this.

matejvasek:
I don't like that the connection is established once on start while running as a root. It shouldn't be needed to be root to dial socket. There is already the `docker` group for that.

So maybe better solution would be to add new flag for `lifecycle` binaries called `--docker-gid`.
Then `pack` could set the flag to numeric id of the `docker` group on given machine. `lifecycle` would call `func Setgroups(gids []int) (err error) ` with the value. This would allow non-root user to dial docker unix socket.

This is however much larger change than the `ping hack`, this would change public API.

wdyt @aemengo

matejvasek:
maybe it wouldn't need new flag, it could get the gid from the mounted socket  via stat

aemengo:
> I don't like that the connection is established once on start while running as a root.

I've never been a stickler for permissions in ephemeral containers, but that's my failing.

> Then the connection is supposed to be kept opened.

Pardon my ignorance, but why must this be the case? From what I see here, ~Thanos~ podman did nothing wrong. Lifecycle should close its connection after use and then _re-dial_ when it needs to talk to the daemon. That's _imho_.

--

Regardless, I think this project has a bad habit of fixing things without reproduction cases. Perhaps a small reproduction case (_for example: a build phase failing on podman_) in your PR would help us get on the same page.

matejvasek:
>Lifecycle should close its connection after use and then re-dial (re-escalate, if needed) when it needs to talk to the daemon.

Totally agree but I am not sure if it's possible to easily and safely elevate right.

Beside using `docker group` I can think of another solution: spawn process under root that would basically work as `socat`, the process would create new docker socket (with non-root access alowed) then the process would just forwarded request to the real docker socket.

matejvasek:
@aemengo I added steps to reproduce.

matejvasek:
cc @ekcasey

matejvasek:
@natalieparellano @aemengo is it conceivable to reopen and merge this PR? The issue with the connection being closed surfaced again.

matejvasek:
@natalieparellano @aemengo PTAL

jabrown85 (approved):

cmd/lifecycle/creator.go
```diff
@@ -247,6 +268,8 @@ func (c *createCmd) Exec() error {
 		platform:      c.platform,
 		platformDir:   c.platformDir,
 	}.build(group, plan)
+	cancelPing()
+	<-pingDoneChan
```
natalieparellano, line 272: Is there a need to block here? If we don't need the channel, that would make the code a bit simpler.

cmd/lifecycle/creator.go
```diff
@@ -239,6 +241,25 @@ func (c *createCmd) Exec() error {
...
+		}
+		for {
+			select {
+			case <-time.After(time.Second * 1):
+				_, _ = c.docker.Ping(pingCtx)
+			case <-pingCtx.Done():
+				return
+			}
+		}
+	}()
```
natalieparellano, line 261: Could we wrap this in a named function, to make the code easier to read? I think something like the below would do the trick (I left out the channel as I didn't see the need for it, but maybe I'm missing something):
```
func (c *createCmd) startPing(pingCtx context.Context) {
	if c.docker == nil {
		return
	}

	go func() {
		for {
			select {
			case <-time.After(time.Second * 1):
				_, _ = c.docker.Ping(pingCtx)
			case <-pingCtx.Done():
				return
			}
		}
	}()
}
```

natalieparellano:
Thank you @matejvasek!

I was able to test this out with podman and it looks good. I left a couple code suggestions.

cmd/lifecycle/creator.go
matejvasek: It's not necessary but it shouldn't block for long, I just like to wait for goroutine if possible.

matejvasek: It should be pretty instantaneous.

matejvasek:
how about:
```diff
diff --git a/cmd/lifecycle/creator.go b/cmd/lifecycle/creator.go
index 0b960754..f648fd12 100644
--- a/cmd/lifecycle/creator.go
+++ b/cmd/lifecycle/creator.go
@@ -242,24 +242,7 @@ func (c *createCmd) Exec() error {
 		}
 	}
 
-	pingCtx, cancelPing := context.WithCancel(context.Background())
-	defer cancelPing()
-	pingDoneChan := make(chan struct{}, 1)
-	// send pings to docker daemon while BUILDING to prevent connection closure
-	go func() {
-		defer func() { pingDoneChan <- struct{}{} }()
-		if c.docker == nil {
-			return
-		}
-		for {
-			select {
-			case <-time.After(time.Millisecond * 500):
-				_, _ = c.docker.Ping(pingCtx)
-			case <-pingCtx.Done():
-				return
-			}
-		}
-	}()
+	stopPing := c.startPing()
 
 	cmd.DefaultLogger.Phase("BUILDING")
 	err = buildArgs{
@@ -269,8 +252,9 @@ func (c *createCmd) Exec() error {
 		platform:      c.platform,
 		platformDir:   c.platformDir,
 	}.build(group, plan)
-	cancelPing()
-	<-pingDoneChan
+
+	stopPing()
+
 	if err != nil {
 		return err
 	}
@@ -337,3 +321,29 @@ func (c *createCmd) populateRunImage() error {
 	}
 	return nil
 }
+
+func (c *createCmd) startPing() (stopPing func()) {
+	pingCtx, cancelPing := context.WithCancel(context.Background())
+	defer cancelPing()
+	pingDoneChan := make(chan struct{}, 1)
+	// send pings to docker daemon while BUILDING to prevent connection closure
+	go func() {
+		defer func() { pingDoneChan <- struct{}{} }()
+		if c.docker == nil {
+			return
+		}
+		for {
+			select {
+			case <-time.After(time.Millisecond * 500):
+				_, _ = c.docker.Ping(pingCtx)
+			case <-pingCtx.Done():
+				return
+			}
+		}
+	}()
+
+	return func() {
+		cancelPing()
+		<-pingDoneChan
+	}
+}
```

matejvasek:
@natalieparellano I believe we must wait for the goroutine.

matejvasek:
@natalieparellano PTAL

natalieparellano (approved):

PR #694, opened 2021-08-19
Install cosign using official GitHub Action

imjasonh:
See https://github.com/sigstore/cosign-installer

This downloads pre-built binaries from official sources, which should save some time since we can avoid building the binary ourselves.

@natalieparellano

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,9 +95,10 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: '1.16'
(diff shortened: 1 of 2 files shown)


natalieparellano (approved):
This is great! Thanks so much!

natalieparellano:
@imjasonh could you set DCO to pass?

imjasonh:
> @imjasonh could you set DCO to pass?

DCO is passing now, sorry about that.

natalieparellano (approved):

PR #695, opened 2021-08-19
Use repository owner from GitHub Actions context

imjasonh:
This should make it slightly easier to run these workflows in a fork.

Unfortunately it [doesn't look like there's an easy context value for just the repo name](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context), so this might still be difficult for people who have a fork with a different name.


natalieparellano (approved):
Thanks @imjasonh !

PR #698, opened 2021-08-26
Add minimal exporter and creator acceptance tests

### LINKED ISSUE #692: Minimal acceptance tests for exporter
### Description
We have https://github.com/buildpacks/lifecycle/issues/467 for adding acceptance tests for the exporter. That issue intends for the tests for be comprehensive, which is a tall order. We should have very minimal acceptance tests for exporter that cover basic things, like:
* it works
* the output app image has the correct os / arch
* the output cache image has the correct os / arch

### Proposed solution
We could copy / paste some setup from the existing restorer acceptance tests (e.g., related to the cache). We should assume that it's a very simple build and maybe it's the first build so we don't have to re-use layers.

### Additional context
The will help #435

natalieparellano:
Fixes #692


jabrown85 (approved):
This looks great!!

PR #700, opened 2021-09-02
Fix bp env bug & remove asset packages

natalieparellano:
When we implemented asset packages, we attempted to only provide the `CNB_ASSETS` env var when BOTH the platform & buildpack apis support it. Because the buildpack api could be different for each buildpack, we decided to provide a new env for each buildpack - we forgot that buildpacks can update the environment for other buildpacks.

Because we put a pause on asset packages, we no longer need to check the platform or the buildpack api when constructing the env, so I've back out all of that code for now.


natalieparellano:
```
$  docker run --env CNB_PLATFORM_API=0.7 -v $PWD:/workspace --mount type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock --user 0 test-builder /cnb/lifecycle/creator -daemon -log-level=debug some-image
```

Before:
```
Paketo Maven Buildpack 5.4.0
  unable to create application layer
  failed to generate expected metadata
  unable to determine java version
  error executing 'javac -version':
   Combined Output: : 
  unable to start PTY
  exec: "javac": executable file not found in $PATH
```

After: successful build :)

natalieparellano:
The current tests are somewhat unsatisfying - in builder_test.go we're testing that `bpA` modifies the environment for `bpB`, but `bpA` is a fake. In buildpack/build_test.go we're testing that we call `AddRootDir` and `AddEnvDir` on the mock env, but we don't show end-to-end that these function calls actually modify the env for the next buildpack. I suspect what we'd **really** like is an acceptance test, maybe that's something we can add once #648 is merged :)

jabrown85 (approved):
Good catch!

Issue #599, opened 2021-04-29
Create helpers to parse API versions


natalieparellano:
### Description
Our code is full of lines like `h.SkipIf(t, api.MustParse(apiStr).Compare(api.MustParse("0.3")) >= 0)`. This makes the code long and hard to read. Let's come up with some better helpers.

PR #705, opened 2021-09-10
Introduce new api version helpers

natalieparellano:
This goes part of the way toward addressing #599

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -93,7 +93,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("called with group", func() {
 			it("errors", func() {
-				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 accepts a -group flag")
+				h.SkipIf(t, api.MustParse(platformAPI).LessThan("0.7"), "Platform API < 0.7 accepts a -group flag")
 				cmd := exec.Command(
 					"docker", "run", "--rm",
 					"--env", "CNB_PLATFORM_API="+platformAPI,
@@ -112,7 +112,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 		when("called with skip layers", func() {
 			it("errors", func() {
-				h.SkipIf(t, api.MustParse(platformAPI).Compare(api.MustParse("0.7")) < 0, "Platform API < 0.7 accepts a -skip-layers flag")
+				h.SkipIf(t, api.MustParse(platformAPI).LessThan("0.7"), "Platform API < 0.7 accepts a -skip-layers flag")
 				cmd := exec.Command(
 					"docker", "run", "--rm",
 					"--env", "CNB_PLATFORM_API="+platformAPI,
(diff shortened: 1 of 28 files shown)


jabrown85 (approved):

acceptance/analyzer_test.go
```diff
@@ -1122,7 +1122,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 
 			it("uses the group path at the layers path and writes analyzed.toml at the layers path", func() {
 				h.SkipIf(t,
-					api.MustParse(platformAPI).Compare(api.MustParse("0.5")) != 0 && api.MustParse(platformAPI).Compare(api.MustParse("0.6")) != 0,
+					api.MustParse(platformAPI).LessThan("0.5") || api.MustParse(platformAPI).AtLeast("0.7"),
```
yaelharel, line 1125: I guess there is no real chance that the following scenario will happen but I'm documenting it just in case:
If we publish new platform APIs like "0.5.x" or "0.6.x", we might need to fix this condition.

natalieparellano: But why? `Compare` takes into account the minor versions: https://github.com/buildpacks/lifecycle/blob/9895acb6a6169013c8eb5f58986e2f49ae478446/api/version.go#L90. The patch version shouldn't matter.

exporter.go
```diff
@@ -114,8 +114,7 @@ func (e *Exporter) Export(opts ExportOptions) (platform.ExportReport, error) {
 		return platform.ExportReport{}, err
 	}
 
-	// platform API > 0.5
-	if e.PlatformAPI.Compare(api.MustParse("0.5")) > 0 {
+	if e.PlatformAPI.AtLeast("0.5") {
```
yaelharel, line 117: According to the previous condition, it should be:
```
e.PlatformAPI.AtLeast("0.6")
```

yaelharel (commented):
I love this PR! It makes our code more readable 😊
I think that the next step will be to get rid of the first part of the comparison lines:
```
api.MustParse(XXX)
```
by saving versions everywhere (or at least in most places) instead of saving strings.
I think that it will also be better in the performance perspective (looking at the `MustParse` and `String` functions of the `Version`).

acceptance/analyzer_test.go
yaelharel: You're right. I was a bit confused with the `Compare` function.

exporter.go
natalieparellano: Good catch! Updated in afc8cc9

exporter_test.go
```diff
@@ -616,12 +616,29 @@ version = "4.5.6"
...
+						exporter.PlatformAPI = api.MustParse("0.5")
+					})
+					it("doesn't set WorkingDir", func() {
+						_, err := exporter.Export(opts)
+						h.AssertNil(t, err)
+
+						val := fakeAppImage.WorkingDir()
+						h.AssertEq(t, val, "")
+					})
+
```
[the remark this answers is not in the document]
natalieparellano: Ahh thank you!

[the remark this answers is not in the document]
natalieparellano: Should be fixed in 731ee95

yaelharel (approved):

PR #720, opened 2021-09-20
Run image should be locked to a digest in analyzed.toml

natalieparellano:

### DIFF
--- a/analyzer_test.go
+++ b/analyzer_test.go
@@ -71,9 +71,9 @@ func testAnalyzerBuilder(platformAPI string) func(t *testing.T, when spec.G, it 
 			p, err := platform.NewPlatform(platformAPI)
 			h.AssertNil(t, err)
 			analyzer = &lifecycle.Analyzer{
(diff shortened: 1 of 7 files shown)


analyzer_test.go
```diff
@@ -194,6 +194,21 @@ func testAnalyzerBuilder(platformAPI string) func(t *testing.T, when spec.G, it
 					h.AssertEq(t, md.Metadata, platform.LayersMetadata{})
 				})
 			})
+
+			when("run image is provided", func() {
+				it.Before(func() {
+					analyzer.RunImage = image
```
jabrown85, line 200: I was surprised to see the same `s0m3D1g3sT` value, but I guess using `image` as `RunImage` is convenient.

jabrown85 (approved):

natalieparellano:
Does anyone know why `build` and `test-linux` are expected? These jobs don't exist anymore (on `main` or the release branch).

Issue #635, opened 2021-06-14
Move logic out of cmd/analyzer


jabrown85:
### Description
We would like to push as much validation and processing out of the `cmd/analyzer.go` as possible. The `cmds` are hard to test and keep growing because there is already validation logic present.

### Proposed solution
Instead of validating inside of `cmd/analyzer.go`, we should put as much validation as possible into `analyzer.go`. This probably means introducing and using interfaces that wrap file, env, and docker/registry access. We should write unit tests to cover and possibly remove validation specific acceptance tests.

### Additional context
- We need to keep `creator` as well as library authors in mind here a bit. Having `analyzedMD, err := analyzer.Analyze()` makes sense for platforms building on top of `lifecycle`, like buildkit. So we may wish to introduce new methods that validate the image(s) that can optionally be executed by those platforms and always executed by `cmd/analyzer.go`.

PR #721, opened 2021-09-20
Refactor packages

natalieparellano:
Opening this PR as a place to get feedback...

Today, our `cmd`, `platform`, and `lifecycle` packages are pretty intertwined:
before

I think this is contributing to our "multiple apis" problem.

Here is the package structure after:

lifecycle:
lifecycle-after

platform:
platform-after

cmd:
cmd-after

Pulling apart these packages will allow us to do things like have the platform configure the lifecycle (the way the spec works) vs. having the lifecycle reach back into the platform to determine its behavior.


natalieparellano:
See 3bee890 for an example of how we could refactor the phases to inject configuration.

natalieparellano:
I made a few changes and now the diff is not quite so horrible. 

Please see 27382ac for an example of how the platform could inject its desired behavior for the analyze phase.

cmd/lifecycle/platform/v06/platform.go
```diff
@@ -0,0 +1,27 @@
...
+		previousPlatform: previousPlatform,
+	}
+}
+
+func (p *v06Platform) API() string {
+	return p.api.String()
+}
+
+func (p *v06Platform) AnalyzeOperations() []lifecycle.AnalyzeOperation {
+	return p.previousPlatform.AnalyzeOperations()
```
jabrown85, line 26: I like the way this reads
```suggestion
	return []lifecycle.AnalyzeOperation{v05.AnalyzeOperations()...}
```
I also like the idea that we could interject a new step if we make the operations public.

```suggestion
	return []lifecycle.AnalyzeOperation{v05.ReadPreviousImage, DoSomethingNewForThisVersion(), v05.RestoreLayerMetadata()}
```

natalieparellano: Changed in 7899b43

cmd/lifecycle/platform/platform.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/buildpacks/lifecycle/cmd/lifecycle/platform/pre06"
+	v06 "github.com/buildpacks/lifecycle/cmd/lifecycle/platform/v06"
+	v07 "github.com/buildpacks/lifecycle/cmd/lifecycle/platform/v07"
+)
+
+var platform03 = pre06.NewPlatform("0.3")
+var platform04 = pre06.NewPlatform("0.4")
+var platform05 = pre06.NewPlatform("0.5")
+var platform06 = v06.NewPlatform(platform05)
+var platform07 = v07.NewPlatform(platform06)
```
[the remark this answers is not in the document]
natalieparellano: Changed in 7899b43

cmd/lifecycle/platform/common/analyze.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/pkg/errors"
+
+	"github.com/buildpacks/lifecycle"
+	"github.com/buildpacks/lifecycle/platform"
+)
+
+func ReadPreviousImage(a *lifecycle.Analyzer, analyzedMD *platform.AnalyzedMetadata) error {
+	if a.Image == nil { // Image is optional in Platform API >= 0.7
```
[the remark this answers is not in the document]
natalieparellano: Changed in 7899b43

analyzer_test.go
```diff
@@ -99,13 +104,14 @@ func testAnalyzerBuilder(platformAPI string) func(t *testing.T, when spec.G, it
...
-				expectedAppMetadata   platform.LayersMetadata
-				expectedCacheMetadata platform.CacheMetadata
+				expectedAppMetadata   dataformat.LayersMetadata
+				expectedCacheMetadata dataformat.CacheMetadata
 				ref                   *testmock.MockReference
 			)
 
 			expectRestoresLayerMetadataIfSupported := func() {
 				if api.MustParse(analyzer.Platform.API()).LessThan("0.7") {
+					ops = append(ops, common.RestoreLayerMetadata)
```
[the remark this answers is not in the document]
natalieparellano: Changed in 819a998

cmd/launcher/platform/pre06/exit.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/buildpacks/lifecycle/cmd/launcher/platform/common"
+)
+
+var exitCodes = map[common.LifecycleExitError]int{
+	// launch phase errors: 700-799
+	common.LaunchError: 702, // LaunchError indicates generic launch error
+}
+
+func (p *pre06Platform) CodeFor(errType common.LifecycleExitError) int {
+	return CodeFor(errType)
```
yaelharel, line 14: Why do we need to call this function and can't we just move lines 18-21 here?

analyzer.go
```diff
@@ -24,52 +22,22 @@ type Analyzer struct {
...
-	return platform.AnalyzedMetadata{
-		Image:    imageID,
-		Metadata: appMeta,
-	}, nil
-}
-
-func (a *Analyzer) restoresLayerMetadata() bool {
-	return api.MustParse(a.Platform.API()).LessThan("0.7")
+	return *analyzedMD, nil
 }
```
yaelharel, line 38: This is great!!!

cmd/launcher/platform/v06/exit.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/buildpacks/lifecycle/cmd/launcher/platform/common"
+)
+
+var exitCodes = map[common.LifecycleExitError]int{
+	// launch phase errors: 80-89
+	common.LaunchError: 82, // LaunchError indicates generic launch error
+}
+
+func (p *v06Platform) CodeFor(errType common.LifecycleExitError) int {
+	return CodeFor(errType)
```
yaelharel, line 14: Same comment as https://github.com/buildpacks/lifecycle/pull/721/files#r714220458.

cmd/lifecycle/platform/platform.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/buildpacks/lifecycle/cmd/lifecycle/platform/common"
+	"github.com/buildpacks/lifecycle/cmd/lifecycle/platform/pre06"
+	v06 "github.com/buildpacks/lifecycle/cmd/lifecycle/platform/v06"
+	v07 "github.com/buildpacks/lifecycle/cmd/lifecycle/platform/v07"
+)
+
+func NewPlatform(apiStr string) (common.Platform, error) {
+	switch apiStr {
+	case "0.3", "0.4", "0.5":
+		return pre06.NewPlatform(apiStr), nil
```
yaelharel, line 15: Same comment as https://github.com/buildpacks/lifecycle/pull/721/files#r714253425.

platform/files.go
```diff
@@ -141,14 +141,14 @@ func (p BuildPlan) Find(bpID string) buildpack.Plan {
...
-		if !containsEntry(metRequires, planEntry) {
+		if !ContainsEntry(metRequires, planEntry) {
 			out = append(out, planEntry)
 		}
 	}
 	return BuildPlan{Entries: out}
 }
 
-func containsEntry(metRequires []string, entry BuildPlanEntry) bool {
+func ContainsEntry(metRequires []string, entry BuildPlanEntry) bool {
```
yaelharel, line 151: Why does it need to be public?

exporter.go
```diff
@@ -14,12 +14,20 @@ import (
...
 	"github.com/buildpacks/lifecycle/platform"
 )
 
+const (
+	EnvLayersDir         = "CNB_LAYERS_DIR"
+	EnvAppDir            = "CNB_APP_DIR"
+	EnvPlatformAPI       = "CNB_PLATFORM_API"
+	EnvDeprecationMode   = "CNB_DEPRECATION_MODE"
+	EnvProcessType       = "CNB_PROCESS_TYPE"
+	DeprecationModeQuiet = "quiet"
```
yaelharel, line 28: Why are these constants public?

cmd/launcher/platform/platform.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/buildpacks/lifecycle/cmd/launcher/platform/common"
+	"github.com/buildpacks/lifecycle/cmd/launcher/platform/pre06"
+	v06 "github.com/buildpacks/lifecycle/cmd/launcher/platform/v06"
+	v07 "github.com/buildpacks/lifecycle/cmd/launcher/platform/v07"
+)
+
+func NewPlatform(apiStr string) (common.Platform, error) {
+	switch apiStr {
+	case "0.3", "0.4", "0.5":
+		return pre06.NewPlatform(apiStr), nil
```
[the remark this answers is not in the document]
jabrown85: or maybe `vLegacy`. I agree the name `pre06` is kind of odd. Or maybe something silly like defining modules for each version that point to the v05.
```
v03 = v05.NewPlatform("0.3")
v04 = v05.NewPlatform("0.4")
```

I kind of wonder if instead of having `v06` have code like this:

```
func (p *v06Platform) AnalyzeOperations() []lifecycle.AnalyzeOperation {
	return []lifecycle.AnalyzeOperation{pre06.ReadPreviousImage, pre06.RestoreLayerMetadata}
}
```

We could always re-define the steps to make it obvious they are used as-is or modified
```
ReadPreviousImage = WithOptionalImage(pre06.ReadPreviousImage)
RestoreLayerMetadata = pre06.RestoreLayerMetadata

func (p *v06Platform) AnalyzeOperations() []lifecycle.AnalyzeOperation {
	return []lifecycle.AnalyzeOperation{ReadPreviousImage,RestoreLayerMetadata}
}
```

exporter.go
natalieparellano: No reason except that I copied them from the `cmd` package, and they are public in that package even though they are only used in `cmd`.

platform/files.go
natalieparellano: It doesn't... this is a holdover from an intermediate refactor. I'll change it back.

cmd/launcher/platform/pre06/exit.go
natalieparellano: I wanted to keep the same pattern that we're using here: https://github.com/buildpacks/lifecycle/blob/819a998c4ab12ff5ebca07cb95f23c6ee6f35ac3/cmd/lifecycle/platform/v06/analyze.go#L9 where a later platform could just call the function directly instead of relying on a "previous platform".

cmd/launcher/platform/v06/exit.go
natalieparellano: https://github.com/buildpacks/lifecycle/pull/721#discussion_r714275626

cmd/launcher/platform/platform.go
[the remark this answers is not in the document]
natalieparellano: It kind of bothers me having the "ops" function definitions in the `platform` package. I feel that the `lifecycle` package should sort of stand on its own, and someone should be able to import it in creating their own lifecycle binary (with the platform of their choosing).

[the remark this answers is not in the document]
natalieparellano: @yaelharel @jabrown85 please see 682f577

WRT the naming, I agree it's weird... maybe it's a bit better with 682f577 since we're not reaching back to `pre06` for any behavior. If we were to completely refactor all of our lifecycle code to use this pattern, we would end up creating `v03`, `v04`, and `v05` packages to get the differentiated behavior.

exporter.go
yaelharel: so maybe it's better to make them private?

natalieparellano: So actually after typing https://github.com/buildpacks/lifecycle/pull/721#discussion_r714285383 I realized that having these variables exported makes the `lifecycle` package more useful as a library that others might import. The `CNB_` variables are part of the spec so it makes sense to have them as part of the package interface.

cmd/launcher/platform/platform.go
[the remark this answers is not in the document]
yaelharel: I'm wondering if https://github.com/buildpacks/lifecycle/commit/682f5770e256fafdc67424f231046d6792e77151 is better than what you did before, because when, for example, platform API 15 will be released, some of the functions won't be relevant to most of the versions.
But even if we keep the last change as is, I think it's better to rename the package `pre06`.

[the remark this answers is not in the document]
jabrown85: 682f577 looks like a decent step forward to me. All the variations being next to each other is nice. I like that each version is standing on their own. It means deleting a version should not affect any other version.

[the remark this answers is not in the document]
natalieparellano: @yaelharel I agree with you, that was the reason I put it in `platform` to begin with... but after seeing how confusing it is to reference previous platforms, this seems like the lesser of two evils!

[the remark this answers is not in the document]
natalieparellano: Also hopefully one day we will deprecate and then remove support for some platforms!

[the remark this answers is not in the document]
natalieparellano: Please see 70d06dc - `v3through5` seems not bad...

aemengo (commented):
I'm worried about some awkwardness being introduced here. Most notably re: `analyzer.go`, I'm worried that:

- The `Analyze() (platform.AnalyzedMetadata, error)` has become too simple. For anything meaningful to happen, it needs `ops` to be passed in - `ops` which now exist in `cmd/lifecycle/platform/*/analyze.go`. Because of this delegation, the `Analyzer` now seems to suffer from the [Lazy Class](https://github.com/lee-dohm/code-smells#lazy-class) Code smell. Or, at the very least, _Lazy Method (I made this up)_. Seeing the `analyzer_test.go` have to specify the `ops` to test against emphasizes the point - when it should be an implementation detail.
- Paradoxically the `cmd/lifecycle/platform/*/analyze.go` files, which hold the important operations, have to delegate _back_ to the `analyzer.go` file. It's not the exact case, but this dependency reminds me of the [Feature Envy](https://github.com/lee-dohm/code-smells#feature-envy) Code Smell.
- I like this `type AnalyzeOperation` definition and that it's specified by the `Platform` object. It allows for [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). But a feature of composition is having a _clean interface_. Seeing the `Analyzer` call these operations, and then **pass a reference to itself** inside the operation does not feel like a clean interface. It's like the `Analyzer` and `Platform` are fighting for ownership.
- Lastly, like I've mentioned before, the analyzer feels a little too dynamic now. Passing functions (_between classes_) can be useful when you don't know the correct logic until later in runtime. But in this case, the logic hinges on the `platformAPI`. And we know this pretty early on. 

I wonder if the solution here is if the `Platform` object has these phase methods (`Analyze()`, `Detect()`, `etc`) and something else (_perhaps `cmd/lifecycle/*.go`_) is responsible for calling them. Because `Platform` already logically broken up, I'd imagine some issues mentioned above going away.

Some, let's say, more subjective points that I had:
- I really dislike this package name `v03through05`. I think @jabrown85 suggested `package legacy`. What's wrong with that? Let's prioritize code that's easily understood, not code that tries too hard to teach you.
- I've never been a fan of this `previousPlatform` pattern. I'm glad to see it going away. Sometimes there's really nothing wrong with duplicating code. There's plenty of [blog posts](https://assortment.io/posts/reasons-against-the-dry-principle#reasons-not-to-use-the-principle) on considering WET (_write everything twice_) code to avoid the indirection that DRY code can introduce.

Regardless, thanks for making this PR @natalieparellano! There are good ideas in here and the code base looks much more manageable. Please forgive my long musings. I love this stuff 😅

layer_metadata_restorer.go
```diff
@@ -189,7 +189,7 @@ func (la *DefaultLayerMetadataRestorer) writeLayerMetadata(layerSHAStore LayerSH
 	return layerSHAStore.add(buildpackID, metadata.SHA, layer)
 }
 
-func retrieveCacheMetadata(cache Cache, logger Logger) (platform.CacheMetadata, error) {
+func RetrieveCacheMetadata(cache Cache, logger Logger) (platform.CacheMetadata, error) {
```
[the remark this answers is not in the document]
natalieparellano: Ah, that's another one left over from a previous refactor. I'll change it.

cmd/launcher/platform/platform.go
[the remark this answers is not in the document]
natalieparellano: Per @aemengo - 

> I really dislike this package name v03through05. I think @jabrown85 suggested package legacy. What's wrong with that? Let's prioritize code that's easily understood, not code that tries too hard to teach you.

I'm also good with `legacy`.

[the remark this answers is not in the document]
yaelharel: Does it make sense that we won't need to make a new package in one of the next versions? I guess not, but if there is a chance that how would we call the package that supports the two versions?

If there is no chance that such thing will happen, I'm totally fine with `legacy`, but if there is, maybe it's better to rename it to `v03Tov05` or `v03To05`.

exporter.go
yaelharel: I don't have a strong opinion about it, but generally, as long as we don't see a specific usage for it right now, I think it's better to make them private. We can always change them to be public if needed in the future.

cmd/launcher/platform/platform.go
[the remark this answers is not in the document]
natalieparellano: I think eventually every platform version will have its own package. The only reason we don't have `v03`, `v04`, and `v05` yet is because we didn't refactor the code that pertains to those packages to use the `Platform` interface. If we always use this interface going forward (when adding support for new platforms), we won't have another platform package that supports more than one platform.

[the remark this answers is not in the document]
natalieparellano: See 4ad354f - went with `legacy`

cmd/launcher/platform/pre06/exit.go
natalieparellano: Since it seems we're all in agreement that getting rid of "previous platform" is a good step, I'll mark this as resolved for now.

yaelharel: Following your last changes, the code in `analyze.go` was changed. Do you think it's worth changing the `CodeFor` code as well?

yaelharel: Sorry, I saw your comment only after posting mine. What do you think?

natalieparellano: I wouldn't move `CodeFor` into `analyzer.go` because it doesn't pertain to the analyzer... it feels fine to leave it here, unless you have a better place for it.

natalieparellano:
> Because of this delegation, the Analyzer now seems to suffer from the Lazy Class Code smell.

I agree... though it's worth noting (because I forgot) that we have issues such as [this one](https://github.com/buildpacks/lifecycle/issues/635) to move logic from cmd -> lifecycle that might make `analyzer.Analyze` more useful.

natalieparellano:
> I wonder if the solution here is if the Platform object has these phase methods (Analyze(), Detect(), etc) and something else (perhaps cmd/lifecycle/*.go) is responsible for calling them.

I've been thinking... in some sense, everything that isn't in the `buildpack` package IS the platform. What you're describing I think would lead to most things in `lifecycle` moving to `platform`, but with duplication for each platform version. Maybe this is what we want... will need to think about it more.

cmd/launcher/platform/pre06/exit.go
yaelharel: I didn't think about moving it to `analyzer.go`. 
Since `CodeFor` is exactly the same for all versions, and the only thing that changes is the `exitCodes` map, maybe we can put the function in a common place and pass the map to it.

exporter.go
natalieparellano: Actually on removing the entanglement between the `cmd` and `platform` packages, these constants no longer need to be copied! Please see caaa415

cmd/launcher/platform/pre06/exit.go
natalieparellano: I'm not sure I follow :/ Could you give an example of how you would like the code to read?

cmd/launcher/platform/legacy/platform.go
```diff
@@ -1,8 +1,8 @@
-package pre06
+package legacy
 
 import (
 	"github.com/buildpacks/lifecycle/api"
-	"github.com/buildpacks/lifecycle/platform/common"
+	"github.com/buildpacks/lifecycle/cmd/launcher/platform/common"
 )
 
 type pre06Platform struct {
```
yaelharel, line 8: Do you want to rename it to `legacy` as well?

cmd/launcher/platform/pre06/exit.go
yaelharel: In `cmd/launcher/platform/common/exit.go`:

```
func CodeFor(errType common.LifecycleExitError, exitCodes map[common.LifecycleExitError]int) int {
	if code, ok := exitCodes[errType]; ok {
		return code
	}
	return cmd.CodeFailed
}
```

Then in `cmd/launcher/platform/[legacy&v06&v07]/exit.go`:
```
func (p *XXXPlatform) CodeFor(errType common.LifecycleExitError) int {
	return common.CodeFor(errType, exitCodes)
}
```

I think that in this way you're not copying the code for every version and it's also not weird to have the `CodeFor` twice in some of the files.

natalieparellano: I realized that since every platform (beside the legacy one) is using the same codes, we could just move that function into `common` for now. Please see e1c3bd8.

natalieparellano: On thinking about this more, what you proposed makes the most sense I think. Please see b6d9f87 - there is more duplication this way, but the code is more readable.

natalieparellano:
Closing this PR as it's a bit dated and we're not so sure of the direction (the uncontroversial bits already merged in #723).

PR #722, opened 2021-09-23
Fix umask race

### LINKED ISSUE #719: `archive.Extract` Race
### Summary

The implementation of `archive.Extract` is not thread-safe, but is called from goroutine code, causing non-deterministic behavior. Specifically, `archive.Extract` [clears and then resets the global `umask` value](https://github.com/buildpacks/lifecycle/blob/b8e2767db739fc96c3bf20b06f2a26adc76dc4cc/archive/extract.go#L20-L22). This value cannot be modified safely from within a goroutine.

This was discovered while we were debugging a [report](https://github.com/paketo-buildpacks/go-build/issues/230) in the Paketo `go-dist` buildpack. On subsequent builds, the buildpack appeared to produce a layer whose content changed when there were no changes to any of the inputs that might have caused a new layer to be created. Instead, when comparing layers between builds, we noticed that the permissions of the layer directories themselves were different. After some head scratching and tracing of our own code, we became convinced that this must be an issue with the lifecycle. That is when we discovered the `archive.Extract` code that [mutates the global `umask` value](https://github.com/buildpacks/lifecycle/blob/b8e2767db739fc96c3bf20b06f2a26adc76dc4cc/archive/extract.go#L20-L22) and the [concurrent code](https://github.com/buildpacks/lifecycle/blob/b8e2767db739fc96c3bf20b06f2a26adc76dc4cc/restorer.go#L91-L93) that invokes `archive.Extract`.

We've created a simplified reproduction program to isolate the issue.

---

##### Steps

1. Execute the [reproduction program](https://gist.github.com/ryanmoran/44b9e8fe1bbd6b59a86a2626ec1275a7#file-main-go) using `go run main.go --order serial` and witness that the `umask` values remain unchanged from beginning to end.
   ```
   ‣ go run main.go --order serial
   umask before: 22
   umask after: 22
   ```
2. Execute the [reproduction program](https://gist.github.com/ryanmoran/44b9e8fe1bbd6b59a86a2626ec1275a7#file-main-go) using `go run main.go --order parallel` and witness that the `umask` value is mutated (somewhat non-deterministically).
   ```
   ‣ go run main.go --order parallel
   umask before: 22
   umask after: 0
   ```

##### Current behavior

The lifecycle mutated the global value of `umask` leading to non-deterministic permissions when creating directories.

##### Expected

The lifecycle should set the value of `umask` back to its original state after modification in a thread-safe manner.

---

##### lifecycle version

`0.11.4`

##### platform version(s)

```
‣ pack report
Pack:
  Version:  0.21.0+git-a9c294a.build-2814
  OS/Arch:  darwin/amd64

Default Lifecycle Version:  0.11.3

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6

Config:

  [[trusted-builders]]
    name = "[REDACTED]"
```

```
‣ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.0-rc.3)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 5
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.10.47-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 11.7GiB
 Name: docker-desktop
 ID: T4XN:HJ2I:QI4X:GFX6:PW7J:XG36:TL4W:YQYY:XLMG:KA2U:7AZU:LZNA
 Docker Root Dir: /var/lib/docker
 Debug Mode: true
  File Descriptors: 43
  Goroutines: 45
  System Time: 2021-09-20T18:03:35.907829Z
  EventsListeners: 3
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: true
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

natalieparellano:
Fixes #719 

### DIFF
--- a/archive/extract.go
+++ b/archive/extract.go
@@ -15,7 +15,8 @@ type PathMode struct {
 	Mode os.FileMode
 }
 
-// Extract reads all entries from TarReader and extracts them to the filesystem
+// Extract reads all entries from TarReader and extracts them to the filesystem.
+// Note that this function is not thread-safe on Unix as it mutates the global umask.
 func Extract(tr TarReader) error {
 	// Avoid umask from changing the file permissions in the tar file.
 	umask := SetUmask(0)
--- a/archive/extract.go
+++ b/archive/extract.go
@@ -18,8 +18,8 @@ type PathMode struct {
 // Extract reads all entries from TarReader and extracts them to the filesystem
 func Extract(tr TarReader) error {
 	// Avoid umask from changing the file permissions in the tar file.
-	umask := setUmask(0)
-	defer setUmask(umask)
+	umask := SetUmask(0)
+	defer SetUmask(umask)
 
 	buf := make([]byte, 32*32*1024)
 	dirsFound := make(map[string]bool)
--- a/archive/extract.go
+++ b/archive/extract.go
@@ -16,11 +16,10 @@ type PathMode struct {
 }
 
 // Extract reads all entries from TarReader and extracts them to the filesystem.
-// Note that this function is not thread-safe on Unix as it mutates the global umask.
 func Extract(tr TarReader) error {
 	// Avoid umask from changing the file permissions in the tar file.
-	umask := SetUmask(0)
-	defer SetUmask(umask)
+	umask := setUmask(0)
+	defer setUmask(umask)
 
 	buf := make([]byte, 32*32*1024)
 	dirsFound := make(map[string]bool)
--- a/archive/extract.go
+++ b/archive/extract.go
@@ -10,11 +10,12 @@ import (
 	"github.com/pkg/errors"
 )
 
-var umask int
+var systemUmask int
 
 func init() {
-	umask = setUmask(0)
-	setUmask(umask)
+	// Avoid umask from changing the file permissions in the tar file.
+	systemUmask = setUmask(0)
+	_ = setUmask(systemUmask)
 }
 
 type PathMode struct {
@@ -24,10 +25,6 @@ type PathMode struct {
 
 // Extract reads all entries from TarReader and extracts them to the filesystem
 func Extract(tr TarReader) error {
-	// Avoid umask from changing the file permissions in the tar file.
-	umask := setUmask(0)
-	defer setUmask(umask)
-
 	buf := make([]byte, 32*32*1024)
 	dirsFound := make(map[string]bool)
 
@@ -61,7 +58,7 @@ func Extract(tr TarReader) error {
 			dirPath := filepath.Dir(hdr.Name)
 			if !dirsFound[dirPath] {
 				if _, err := os.Stat(dirPath); os.IsNotExist(err) {
-					if err := os.MkdirAll(dirPath, applyUmask(os.ModePerm, umask)); err != nil {
+					if err := os.MkdirAll(dirPath, applyUmask(os.ModePerm, systemUmask)); err != nil {
 						return errors.Wrapf(err, "failed to create parent dir %q for file %q", dirPath, hdr.Name)
 					}
 					dirsFound[dirPath] = true
--- a/archive/extract.go
+++ b/archive/extract.go
@@ -10,12 +10,19 @@ import (
 	"github.com/pkg/errors"
 )
 
+var umask int
+
+func init() {
+	umask = setUmask(0)
+	setUmask(umask)
+}
+
 type PathMode struct {
 	Path string
 	Mode os.FileMode
 }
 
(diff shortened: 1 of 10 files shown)


natalieparellano:
There might be better ways of solving this, but this seemed like the "simplest". I'm curious to know what others think.

archive/extract.go
```diff
@@ -15,11 +15,12 @@ type PathMode struct {
 	Mode os.FileMode
 }
 
-// Extract reads all entries from TarReader and extracts them to the filesystem
+// Extract reads all entries from TarReader and extracts them to the filesystem.
+// Note that this function is not thread-safe on Unix as it mutates the global umask.
 func Extract(tr TarReader) error {
```
aemengo, line 20:
```suggestion
var umask int

func init() {
	umask = setUmask(0)
	setUmask(umask)
}

// Extract reads all entries from TarReader and extracts them to the filesystem
func Extract(tr TarReader) error {
```

Unfortunately, I'm still able to reproduce @ryanmoran's case in this PR. What do you think of the following?

restorer.go
```diff
@@ -40,6 +41,8 @@ func (r *Restorer) Restore(cache Cache) error {
 	}
 
 	var g errgroup.Group
+	umask := archive.SetUmask(0)
+	defer archive.SetUmask(umask)
```
aemengo, line 45: I'm not quite sure what these lines will prevent. Ultimately, the same `archive.Extract` code will be called.

archive/extract.go
natalieparellano: I'm a bit wary to set the umask for any program that might import this package (for the entirety of the program's execution). Are you using the linked [reproduction program](https://gist.github.com/ryanmoran/44b9e8fe1bbd6b59a86a2626ec1275a7#file-main-go) to verify? If so I think it makes sense that you're still seeing `umask after: 0` in the output. [These lines](https://github.com/buildpacks/lifecycle/pull/722/files#diff-98f352cc6ef347aa505fe6d5f3e01c3a47ee0b5d1a91d1e918343284c893b51eR44-R45) are a bit like [this line](https://gist.github.com/ryanmoran/44b9e8fe1bbd6b59a86a2626ec1275a7#file-main-go-L24) in the reproduction program, setting the umask back to whatever it was before `archive.Extract` was called.

restorer.go
natalieparellano: Replied [here](https://github.com/buildpacks/lifecycle/pull/722/files#r715649087)

archive/extract.go
aemengo: I think the problem is that we're trying to solve is that we want a `fetchUmask()`, so we can use it for the Extract() operation.

But it looks like the only way to _fetch_ is to `setUmask()` to get the old value. And then set it again.

The problem with our `defer` approach is that the Extract() can take a long time. Thus leaving the "gap" (_when the throwaway value is set_) to be too long. Which wouldn't be good for both race conditions, and in-between syscalls to `open()` or `mkdir()`

natalieparellano:
@aemengo thank you for helping me understand this code! I think what you proposed makes sense.

natalieparellano:
Now we consistently get:
```
$  go run repro/repro.go --order parallel
umask before: 22
umask after: 22
```

aemengo (approved):

jabrown85 (approved):

PR #724, opened 2021-09-24
Move the layer metadata restorer and the layer sha store to a new package

### LINKED ISSUE #629: Move layer metadata restorer & layer sha store into internal package
### Description
These services are currently in the `lifecycle` package, but they are really implementation details and shouldn't be part of the public api. Let's move them into an internal package.

yaelharel:
Fixes #629


cmd/lifecycle/analyzer.go
```diff
@@ -204,7 +205,7 @@ func (a *analyzeCmd) Exec() error {
 		return err
 	}
 
-	if err := lifecycle.WriteTOML(a.analyzedPath, analyzedMD); err != nil {
+	if err := buildpack.WriteTOML(a.analyzedPath, analyzedMD); err != nil {
```
natalieparellano, line 208: Why did this function move to the buildpack package? I don't think we want the `main` package to know about buildpacks. It should only know about the platform...

layermetadata/layer_metadata_restorer.go
```diff
@@ -1,7 +1,8 @@
-package lifecycle
+package layermetadata
```
natalieparellano, line 1: Can this be an internal package?

layermetadata/layer_metadata_restorer.go
```diff
@@ -11,8 +12,19 @@ import (
...
+	SetMetadata(metadata platform.CacheMetadata) error
+	RetrieveMetadata() (platform.CacheMetadata, error)
+	AddLayerFile(tarPath string, sha string) error
+	ReuseLayer(sha string) error
+	RetrieveLayer(sha string) (io.ReadCloser, error)
+	Commit() error
+}
+
+//go:generate mockgen -package testmock -destination testmock/layer_metadata_restorer.go github.com/buildpacks/lifecycle/layermetadata MetaRestorer
+type MetaRestorer interface {
```
yaelharel, line 27: I'm not sure that the names of the package and this interface are the best. But keeping this interface name as `LayerMetadataRestorer` inside a `layermetadata` package seems weird because whoever use it, will need to use: `layermetadata.LayerMetadataRestorer`.
It's not a good idea to call the interface `Restorer` for example because `testmock.Restorer` isn't a good name either.
Name suggestions are more than welcome 😊

cmd/lifecycle/analyzer.go
yaelharel: The same `WriteTOML` function existed in both the `lifecycle` and `buildpack` packages. Since we import the `buildpack` package anyway in most of the files where we used to call `WriteTOML` under the `lifecycle` package, I thought it might be useful to just use the `buildpack` package.
As you can see, I've made this change in a separate commit so it'll be easier to make changes to it if needed.

natalieparellano: I still think it's weird that `main` should depend on the `buildpack` package. `WriteTOML` is a very simple function, maybe we could just duplicate it.

yaelharel: I understand the weirdness but I really don't like duplicating code in such case.
I found this issue because the new package that I created is also using this function as well and in order to prevent circular imports, we can either use the buildpack implementation or duplicate it again (for the third time).
What do you think about creating a helper package that all three packages will be able to use?

natalieparellano: I think it's okay for `layermetadata` to use `buildpack.WriteTOML` (because it needs to import `buildpack` anyway). But I don't think we need to remove `lifecycle.WriteTOML`. That is two copies of the function, which seems okay. Then `main` can use `lifecycle.WriteTOML`, and avoid importing `buildpack`.

layermetadata/layer_metadata_restorer.go
natalieparellano: We should also move layer_metadata_restorer_test.go

cmd/lifecycle/analyzer.go
yaelharel: As you can see in [this commit](https://github.com/buildpacks/lifecycle/pull/724/commits/b08737e7b89539676e9ad709915d643e22d109b6), all of the `main`, except the `rebaser` already import `buildpack` (even without calling the `WriteTOML` function). This is why I thought there is no problem with using the `buildpack` package also for this purpose.

natalieparellano: You are right! I guess it just reads weird, since `buildpack.WriteTOML` implies that a buildpack is doing something, whereas this action has nothing to do with buildpacks.

yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/724/commits/ced3b85ee7ae01a12682461f0fc549a70cb09c48.
What do you think about putting it in a new package?

utils/utils.go
```diff
@@ -1,4 +1,4 @@
-package lifecycle
+package utils
```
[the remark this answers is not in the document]
yaelharel: I like your idea. I don't like putting everything together in `utils` as well.
I'll let you know after I'll make some changes.
Thanks!

cmd/lifecycle/analyzer.go
yaelharel: I'm going to mark this conversation as resolved. Let's continue the discussion [here](https://github.com/buildpacks/lifecycle/pull/724#discussion_r717648603).

utils/utils.go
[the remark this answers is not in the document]
yaelharel: Please let me know what you think about the new packages in my last commit: https://github.com/buildpacks/lifecycle/pull/724/commits/3917722d377eb5c719ff6bd97ef5c155e8a7191c.
@jabrown85 @natalieparellano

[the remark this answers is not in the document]
natalieparellano: I have to say that I'm still a bit wary of creating all these tiny packages. I did a little digging to try to substantiate this feeling, and found this [post](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common):

> My recommendation to improve the name of utils or helpers packages is to analyse where they are imported and move the relevant functions into the calling package. Even if this results in some code duplication this is preferable to introducing an import dependency between two packages. In the case where utility functions are used in many places, prefer multiple packages, each focused on a single aspect with a correspondingly descriptive name.

We are doing the "multiple focused packages" part, but all of the utility functions that have been copied are used only in the `lifecycle` package - with the exception of `WriteTOML`. I think I'd still rather duplicate this function (it's very simple, and only 11 lines of code) than introduce all of these new packages with no clear need to do so.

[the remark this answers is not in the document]
yaelharel: I don't like code duplication unless the function is almost an inline function, and even then, you really need to think whether it's worth it.

Talking about `WriteTOML`, if at some point we'll find a bug in this function or we'll decide to change it a bit, having to do it in 3 different (or even 2) places will increase the probability of new bugs because we might forget to change it everywhere or we might copy it incorrectly.

Regarding the the package`imageutils` - it is being used by 3 other packages: `lifecycle`, `cache` and `main` so I would leave at least the `DecodeLabel` function there.

The package `str` is only being used by the `lifecycle` package so I'm ok with moving its functions to somewhere there.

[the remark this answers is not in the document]
jabrown85: I removed a line at the end of my comment that was originally like "if we don't like small packages I rather have some duplication". I think there is a balance to be struck here for sure. I think I lean toward small packages but I don't have very strong feelings.

layermetadata/layer_metadata_restorer.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/724/commits/35dce2dac769d44a8f367a3558346f95cc19981b.

utils/utils.go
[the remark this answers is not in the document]
natalieparellano: > Regarding the the packageimageutils - it is being used by 3 other packages: lifecycle, cache and main so I would leave at least the DecodeLabel function there.

Fair enough. But let's not make `imageutils` (too similar to `imgutil` lol) - I noticed we already have an `image` package, so maybe it could go there.

> The package str is only being used by the lifecycle package so I'm ok with moving its functions to somewhere there.

Let's just keep it in utils.go then.

> I don't like code duplication unless the function is almost an inline function, and even then, you really need to think whether it's worth it.

I think we'll just have to agree to disagree :) Can we please keep both `lifecycle.WriteTOML` and `buildpack.WriteTOML` for now - we know that we have a lot of work to do re: reading and writing TOML and perhaps new packages will emerge from that effort. If you want you could write a test that `lifecycle.WriteTOML` and `buildpack.WriteTOML` produce the same output.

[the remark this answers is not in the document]
yaelharel: I moved the imageutils files to the image package as part of the following commit:
https://github.com/buildpacks/lifecycle/pull/724/commits/6559c5c07ee41e11c6b8f553949bd43b9b93b325

Regarding `WriteTOML`:
You want to define the function in both the `lifecycle` and the `buildpack` packages, right?
But what about `layer_metadata_restorer.go`? It can't use the function that will be define in the `lifecycle` package because it will cause circular imports. Moreover, as you wrote in the past about the `lifecycle` package, it doesn't make sense that the `layermetadata` package will use the `buildpack` package for that function.
So what do you want to do? Define it again for the third time?
@natalieparellano @jabrown85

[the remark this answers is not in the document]
natalieparellano: I think in the end, we’ll probably have two packages with toml utilities - one for platform, and one for buildpack (since the way we are parsing & interpreting files will differ by each api). Maybe it makes sense to explore what that will look like as part of this PR?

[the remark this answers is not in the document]
yaelharel: I'm not sure I understand what is the problem that there will be a common package with useful functions.
I don't think that `WriteTOML` will depend on an API version.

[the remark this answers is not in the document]
yaelharel: Following the discussion in the [slack](https://buildpacks.slack.com/archives/CSFE49VGR/p1633451965101500), please see https://github.com/buildpacks/lifecycle/pull/724/commits/3caa11617113824dcdd2aa082db932963c887932.

buildpack/descriptor_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/buildpacks/lifecycle/buildpack"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+)
+
+func TestUtils(t *testing.T) {
+	spec.Run(t, "Descriptor", testDescriptor, spec.Report(report.Terminal{}))
+}
+
+func testDescriptor(t *testing.T, when spec.G, it spec.S) {
+	when(".ReadOrder", func() {
```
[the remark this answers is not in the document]
yaelharel: Of course. I deleted it by mistake. Thank you!

internal/layermetadata/layer_metadata_restorer.go
```diff
@@ -23,8 +36,8 @@ type DefaultLayerMetadataRestorer struct {
 }
 
 type LayerSHAStore interface {
-	add(buildpackID, sha string, layer *bpLayer) error
-	get(buildpackID string, layer bpLayer) (string, error)
+	add(buildpackID, sha string, layer *BpLayer) error
```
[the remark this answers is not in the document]
jabrown85: I guess because it is in an `internal` package anyway. I like the symmetry - I don't have a strong preference.

image/utils_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/buildpacks/lifecycle/internal/encoding"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+)
+
+func TestUtils(t *testing.T) {
+	spec.Run(t, "Utils", testUtils, spec.Report(report.Terminal{}))
+}
+
+func testUtils(t *testing.T, when spec.G, it spec.S) {
+	when(".WriteTOML", func() {
```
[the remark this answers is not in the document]
jabrown85: This was mostly a comment about `WriteTOML` being tested in the `image` package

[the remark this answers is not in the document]
yaelharel: Ohhh...please see https://github.com/buildpacks/lifecycle/pull/724/commits/4ba0204f81cec26afb944b869db90790323ca9d1.
Thanks!

natalieparellano:
It looks good! I made a few suggestions around naming and organization.

buildpack/layermetadata/encoderdecoder.go
```diff
@@ -0,0 +1,129 @@
...
+type EncoderDecoder interface {
+	IsSupported(buildpackAPI string) bool
+	Encode(file *os.File, lmf File) error
+	Decode(path string) (File, string, error)
+}
+
+func supportedEncoderDecoders() []EncoderDecoder {
+	return []EncoderDecoder{
+		&DefaultEncoderDecoder{},
+		&LegacyEncoderDecoder{},
```
aemengo, line 22:
```suggestion
		&defaultEncoderDecoder{},
		&legacyEncoderDecoder{},
```

This is definitely a nitpick but most of the methods, and even the interface, is not used outside of this package. Having _only_ `EncodeFile(), DecodeFile()` be given the public scope, and be moved to the top of the file, makes the file easier to parse. And better expresses developer intent.

buildpack/files.go
```diff
@@ -197,52 +193,14 @@ func containsName(unmet []Unmet, name string) bool {
...
-
-	for _, decoder := range decoders {
-		if decoder.IsSupported(buildpackAPI) {
-			return decoder.Decode(path)
-		}
-	}
-	return layertypes.LayerMetadataFile{}, "", errors.New("couldn't find a decoder")
+type LayerMetadata struct {
+	SHA string `json:"sha" toml:"sha"`
+	layermetadata.File
```
aemengo, line 205: Now that we have the `buildpack/layermetadata/file.go`, all the structs seem like they belong together. It's just a bit awkward this logic spread out.

aemengo (approved):

buildpack/layermetadata/encoderdecoder.go
natalieparellano: Please see 418562fc

buildpack/files.go
natalieparellano: @aemengo your comment made me wonder why we need the `layermetadata` package to begin with. WDYT of 63a91a52 ?

aemengo: I like it. 👍🏾

natalieparellano: Fixed in 1656b02e

jabrown85 (approved):
This turned out to be quite the change! It looks good to me, but it was a lot of files to review 😅

PR #727, opened 2021-09-28
Ensure umask is unset when extracting archive

natalieparellano:
Thanks @yaelharel for prodding me to look into this more closely

https://github.com/buildpacks/lifecycle/pull/203 added logic to update file permissions in case the system umask was applied.
https://github.com/buildpacks/lifecycle/pull/246 unset the umask before the extract to avoid having to do this extra step (thus saving time), however it made the function not thread-safe as discovered in #719.

This PR ensures that
* the umask is always unset when running the extract
* the system umask is applied to directories
* extracts running in parallel will not contaminate each other

@sclevine it looks like you have some of the history here, it would be great if you could take a look!

cc @aemengo @jabrown85


archive/extract_test.go
```diff
@@ -129,6 +129,9 @@ func testExtract(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("extracts a tar file", func() {
+			old := archive.SetUmask(0)
```
jabrown85, line 132: Is there a test we can write to not regress on expected behavior? I'm not entirely sure what happens if we leave off these 2 lines, for instance. Will this test fail and will it be clear why it failed? Should we be asserting the value of `archive.SetUmask(0)` to be `old` at the end of these tests?

restorer.go
```diff
@@ -39,6 +40,9 @@ func (r *Restorer) Restore(cache Cache) error {
 		}
 	}
 
+	old := archive.SetUmask(0)
```
jabrown85, line 43: Do the tests fail here if these are omitted?

aemengo:
To recap our thoughts. The tar archive has the correct file permissions. Thus the options being considered are:
1. We set the `umask = 0` for the entire duration of `Extract()`, _and create every file with the `FileMode` recorded in the tar archive._
2. We fetch the umask very quickly, _and create every file with the `FileMode` recorded in the tar archive._

And option 1 seems to lead us into race conditions.

🙂

restorer.go
aemengo: Forgive me, I'm still not sure why these lines are necessary, if `Extract()` is behaving correctly  - and avoiding race conditions. 😅

natalieparellano: If we do #722 only then we will need to bring back these lines from #203 (and potentially make the extract slower): https://github.com/buildpacks/lifecycle/pull/203/files#diff-64f1e499a23830d0c26ee246125d18431a7e36ad3249c5df70eb2982effc68afR150-R153

natalieparellano: @jabrown85 yes they fail... rather than unset the umask during the extract, I made it the responsibility of the calling function, otherwise there is no way that this function could be threadsafe (unless we explicitly set the permissions like in https://github.com/buildpacks/lifecycle/pull/203/files#diff-64f1e499a23830d0c26ee246125d18431a7e36ad3249c5df70eb2982effc68afR150-R153 )

archive/extract_test.go
natalieparellano: If we leave off these two lines then the test will fail with `umask should be unset by the calling function`.

jabrown85: Can we write a test for that so we don't accidentally remove that error behavior in some distant future?

natalieparellano: Makes sense, added in c0ed744

archive/extract.go
```diff
@@ -25,6 +25,12 @@ type PathMode struct {
 
 // Extract reads all entries from TarReader and extracts them to the filesystem
 func Extract(tr TarReader) error {
+	currentUmask := SetUmask(0)
```
sclevine, line 28: This affects reentrancy of `Extract`. Consumers may expect to call this method in parallel with other invocations.

natalieparellano: @sclevine updated in 1a3669d - what do you think?

natalieparellano: Sorry, it's really cf27827 through 1a3669d...

jabrown85:
Can we, instead of trying to make `Extract` thread-safe handle `umask`, put the `umask` reset logic in restorer?

[This](https://github.com/buildpacks/lifecycle/blob/8c2edb2a85f5aaa1f46babcaf300fe3a84c85505/restorer.go#L90-L92) is the bit that causes the problem, if I'm understanding this correctly.

Could we make `(r *Restorer) Restore(cache Cache) error` have the init like we used to have?

```
func init() {
	systemUmask = setUmask(0)
	_ = setUmask(systemUmask)
}

func (r *Restorer) Restore(cache Cache) error {
	umask = setUmask(0)      
        defer setUmask(systemUmask)
        ... some time later ...
        go layers.Extract(rc, "", umask) // add umask as param to Extract
}
```

We don't expect `Restorer` to be used concurrently and we know `Extract` is used concurrently so it makes sense to me to move the unsafe bits up a level.

Or, if we really don't want to pass `umask` from restorer down to archive, we could have `archive.Extract` only execute `umask(0)` if `/proc/self/status` `umask` is _not_ already zero and always _return_ the previous `umask` _if it had to set it) so the caller can do something like this so `Extract` is safe for reentry.

```
var firstUmask

go func(){
   umask, err := layers.Extract(rc, "")
   if umask != -1 {
    firstUmask = umask
  }
}

// after all are done
setUmask(prevUmask)
```

natalieparellano:
> put the umask reset logic in restorer

The latest incarnation of this PR does this. `archive.Extract` doesn't do anything except verify that the umask is already 0. Maybe we could take the verification out, and add a docs comment, but that would make the function buggy for anyone who didn't read carefully, which seems bad.

> Could we make (r *Restorer) Restore(cache Cache) error have the init like we used to have?

What would the variable set in the init be used for?

> we could have archive.Extract only execute umask(0) if /proc/self/status umask is not already zero and always return the previous umask if it had to set it

This is interesting. I had some variation of this before, but I was erroring out if the umask was not already 0 (and trying to "set it back" within `archive.Extract`, which is unsafe). Let me give it a try.

natalieparellano:
@jabrown85 one bit that I think makes what you proposed hard is that the original umask is also used within `archive.Extract`, here: https://github.com/buildpacks/lifecycle/blob/8c2edb2a85f5aaa1f46babcaf300fe3a84c85505/archive/extract.go#L57

natalieparellano:
After conversation with @jabrown85, we decided:
* Getting the system umask in an init function is potentially too early (a program could alter it later on)
* We don't need to "check the current umask by unsetting it" inside `archive.Extract` if we document that the umask must be unset prior to calling the function
* We don't entirely trust our homegrown `GetUmask()`

The latest commits reflect this thinking. It's a bit gross to change the signature of `archive.Extract` and `layers.Extract`, but it's safer this way (we think).

archive/extract_test.go
```diff
@@ -142,19 +145,25 @@ func testExtract(t *testing.T, when spec.G, it spec.S) {
...
 			file, err := os.Create(filepath.Join(tmpDir, "root"))
 			h.AssertNil(t, err)
 			h.AssertNil(t, file.Close())
 
-			h.AssertError(t, archive.Extract(tr), "failed to create directory")
+			h.AssertError(t, archive.Extract(tr, 0), "failed to create directory")
 		})
 
 		it("doesn't alter permissions of existing folders", func() {
+			oldUmask := archive.SetUmask(0)
```
jabrown85, line 159: Could we have a test that asserts the permissions on the directory created using `oldUmask`?

natalieparellano: I think in theory that is what this test is supposed to show. However it doesn't seem to make any difference what value is passed in...

natalieparellano: Added a test!

aemengo (approved):
Very elegant!

jabrown85 (approved):

natalieparellano:
The arm worker... 😭

archive/extract.go
```diff
@@ -10,29 +10,23 @@ import (
...
-
 type PathMode struct {
 	Path string
 	Mode os.FileMode
 }
 
-// Extract reads all entries from TarReader and extracts them to the filesystem
-func Extract(tr TarReader) error {
+// Extract reads all entries from TarReader and extracts them to the filesystem.
+// The umask must be unset before calling this function on unix, to ensure that files have the correct file mode.
```
yaelharel, line 19: Do we want to add that we can call `SetUmask` after this function to return the original value?

archive/extract_test.go
```diff
@@ -167,3 +178,12 @@ func testExtract(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 }
+
+func testPathPerms(t *testing.T, parentDir, path string, expectedMode os.FileMode) {
```
yaelharel, line 182: Since we're calling this function only in one use case, maybe we can put it inside the for loop. I'm not sure what will be clearer for the user.

yaelharel (approved):
Added a few small comments but I don't have strong opinion on making those changes.

archive/extract.go
natalieparellano:
```suggestion
// The umask must be unset before calling this function on unix, to ensure that files have the correct file mode. SetUmask can be used to set and unset the umask.
```

layers/extract.go
```diff
@@ -11,9 +11,15 @@ import (
 // Extract extracts entries from r to the dest directory
 // Contents of r should be an OCI layer.
 // If dest is an empty string files with be extracted to `/` or `c:\` on unix and windows filesystems respectively.
-func Extract(r io.Reader, dest string) error {
+// The umask must be unset before calling this function on unix, to ensure that files have the correct file mode.
```
natalieparellano, line 14:
```suggestion
// The umask must be unset before calling this function on unix, to ensure that files have the correct file mode. SetUmask can be used to set and unset the umask.
```

archive/extract_test.go
natalieparellano: Aye, I called this function in two places before and forgot to change it back. I would update but I'm having problems with my IDE... let's just leave it out of laziness.

PR #729, opened 2021-09-28
Use the correct tag when signing the sbom

natalieparellano:
Also there is no need to parse the digest from `crane tag` because it does not change.
This will make the code less brittle.


natalieparellano:
Tested on fork: https://github.com/natalieparellano/lifecycle/actions/runs/1283484914

jabrown85 (approved):

PR #730, opened 2021-09-30
Set Umask as part of archive.Extract

yaelharel:
A different solution for https://github.com/buildpacks/lifecycle/pull/727.


yaelharel:
@aemengo, we would love to get your review! Thanks!

archive/extract.go
```diff
@@ -15,10 +16,35 @@ type PathMode struct {
 	Mode os.FileMode
 }
 
+var (
+	umaskLock      sync.Mutex
+	extractCounter int
+	originalUmask  int
+)
```
yaelharel, line 23: We can put these three variables inside a private struct if anyone thinks it'll be clearer.
Or we can even move the variables and the related functions (`setUmaskIfNeeded` and `unsetUmaskIfNeeded`) to an internal package.

Personally, I don't think it's necessary but it's an option.

natalieparellano:
I think this is the best solution. This way we don't force the complexity of setting the umask on the caller. @yaelharel thank you for thinking of it!

archive/extract.go
```diff
@@ -15,10 +16,35 @@ type PathMode struct {
...
+
+func setUmaskIfNeeded() {
+	umaskLock.Lock()
+	extractCounter++
+	if extractCounter == 1 {
+		originalUmask = setUmask(0)
+	}
+	umaskLock.Unlock()
```
jabrown85, line 31: I would feel safer with `defer umaskLock.Unlock()` on line 27. Mostly to guard against an unexpected `panic` when setting `umask`.

archive/extract.go
```diff
@@ -15,10 +16,35 @@ type PathMode struct {
...
+}
+
+func unsetUmaskIfNeeded() {
+	umaskLock.Lock()
+	extractCounter--
+	if extractCounter == 0 {
+		_ = setUmask(originalUmask)
+	}
+	umaskLock.Unlock()
```
jabrown85, line 40: Same comment as the other. A `defer` feels appropriate here to me.

archive/extract_test.go
```diff
@@ -12,49 +12,29 @@ import (
...
 
 	"github.com/buildpacks/lifecycle/archive"
 	h "github.com/buildpacks/lifecycle/testhelpers"
 )
 
+var originalUmask int
+
 func TestArchiveExtract(t *testing.T) {
 	rand.Seed(time.Now().UTC().UnixNano())
+	originalUmask = h.GetUmask(t)
```
jabrown85, line 25: I personally like this style for tests. 

```
originalUmask = setUmask(0)
_ = setUmask(originalUmask)
```

jabrown85 (commented):
I like it - it seems simpler than I initially figured it would end up. The complexity being trapped inside of Extract is _probably_ the right move from an API perspective.

archive/extract_test.go
yaelharel: Yeah, we thought about it. @natalieparellano, it's your call. I really don't mind. Maybe what @jabrown85 suggested is clearer.

archive/extract.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/730/commits/901a5d6146ffee65c3eb3a9c460df2446af4f6ec.

archive/extract.go
yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/730/commits/901a5d6146ffee65c3eb3a9c460df2446af4f6ec.

jabrown85 (approved):

archive/extract_test.go
natalieparellano: That works for me!

natalieparellano: This would allow us to delete our `GetUmask()` helpers.

yaelharel: Thinking about it again, I'm not sure if it's possible to get rid of the `GetUmask` functions because of [this line](https://github.com/buildpacks/lifecycle/pull/730/files#diff-b12673f7f2d9439d536db38a5a66d22c553089b91aaa34501d19f34e4743edb3R94).
If we'll try to replace that line with:
```
originalUmask = setUmask(0)
_ = setUmask(originalUmask)
```
Then we might have problems with other tests that run in parallel.

And if we don't get rid of this function, I don't think it's worth changing [the suggested line](https://github.com/buildpacks/lifecycle/pull/730/files#diff-b12673f7f2d9439d536db38a5a66d22c553089b91aaa34501d19f34e4743edb3R25).

@jabrown85 @natalieparellano, what do you think?

natalieparellano: Ah, you're right!

yaelharel: I think we should just leave it as is unless any of you (or someone else) has a better solution.

yaelharel: After talking with @jabrown85, we decided to keep the code as is.

PR #741, opened 2021-10-05
Cacher when caching a layer should also cache sbom as a separate layer

natalieparellano:


natalieparellano:
It might make sense to create a `structured-sbom` branch for the lifecycle and throw all of the PRs for https://github.com/buildpacks/lifecycle/issues?q=is%3Aissue+is%3Aopen+label%3Aepic%2Fsbom on there since it will be hard to test the feature end-to-end without all of them.

PR #760, opened 2021-11-10
Don't run tests when we change markdown files

natalieparellano:


natalieparellano:
Actually this might have implications if we try to release and the last commit only changed a markdown file (then there wouldn't be a build to attach to the release). Thinking...

jabrown85:
To me, this feels like an optimization we don't really need. If making markdown only changes doesn't happen regularly, I think letting tests run is a good way to make sure flaky tests are found out.

PR #764, opened 2021-11-12
Error when previous bom is specified on buildpack API 0.7

aemengo:
if you are on buildpack API version 0.7 and output old style bom information, return error 🛑 
if you are on buildpack API version <0.7 and output new style SBoM information, show warning ⚠️ 

Screen Shot 2021-11-15 at 3 11 08 PM

cc: @dmikusa-pivotal

### DIFF
--- a/buildpack/bomfile.go
+++ b/buildpack/bomfile.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/buildpacks/lifecycle/api"
 	"github.com/buildpacks/lifecycle/buildpack/layertypes"
 )
 
@@ -90,7 +91,7 @@ func validateMediaTypes(bp GroupBuildpack, bomfiles []BOMFile, sbomMediaTypes []
 	return nil
 }
 
-func processBOMFiles(layersDir string, bp GroupBuildpack, pathToLayerMetadataFile map[string]layertypes.LayerMetadataFile, sbomMediaTypes []string) ([]BOMFile, error) {
+func (b *Descriptor) processBOMFiles(layersDir string, bp GroupBuildpack, pathToLayerMetadataFile map[string]layertypes.LayerMetadataFile, logger Logger) ([]BOMFile, error) {
 	var (
 		layerGlob = filepath.Join(layersDir, "*.sbom.*.json")
 		files     []BOMFile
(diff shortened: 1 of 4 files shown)


samj1912 (approved):

jabrown85:
I like the strictness of a buildpack author having to choose a single formation. In practical terms, how hard would it be for a buildpack author to adopt the format if they were previously unaware of cyclone/other tooling? Do we have a migration guide for those authors?

dmikusa-pivotal:
>In practical terms, how hard would it be for a buildpack author to adopt the format if they were previously unaware of cyclone/other tooling? Do we have a migration guide for those authors?

Generally speaking, I think you'd be using a tool that supports your language/runtime to generate this information. A lot of folks are using `syft` which is quick and supports a number of language/runtimes. There are other tools too.

As far as a migration guide, do we have a guide or doc for migrating between buildpack API versions? I could certainly see having notes in a guide like that about the BOM format changes.

dmikusa-pivotal (approved):

jabrown85:
> As far as a migration guide, do we have a guide or doc for migrating between buildpack API versions? I could certainly see having notes in a guide like that about the BOM format changes.

We do have migration guides [here](https://github.com/buildpacks/docs/tree/main/content/docs/reference/spec/migration).

dmikusa-pivotal:
Perfect. If someone isn't already working on it, I can try to write up something for that.

aemengo:
@dmikusa-pivotal It's the first I've been made aware of these. We'd appreciate the help 🙇🏾

buildpack/bomfile.go
```diff
@@ -101,6 +102,14 @@ func processBOMFiles(layersDir string, bp GroupBuildpack, pathToLayerMetadataFil
 		return nil, err
 	}
 
+	if api.MustParse(b.API).LessThan("0.7") {
+		if len(matches) != 0 {
+			logger.Warnf("the following SBoM files will not be respected in this buildpack api version [%s]", strings.Join(matches, ", "))
```
natalieparellano, line 107:
```suggestion
			logger.Warnf("the following SBoM files will be ignored for buildpack api version < 0.7 [%s]", strings.Join(matches, ", "))
```

buildpack/build_test.go
```diff
@@ -332,11 +318,11 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 						filepath.Join(layersDir, buildpackID, fmt.Sprintf("%s.toml", layerName)))
 
 					br, err := bpTOML.Build(buildpack.Plan{}, config, mockEnv)
-					if err != nil {
-						t.Fatalf("Unexpected error:\n%s\n", err)
-					}
+					h.AssertNil(t, err)
 
 					h.AssertEq(t, len(br.BOMFiles), 0)
+					expected := "the following SBoM files will not be respected in this buildpack api version"
```
natalieparellano, line 324:
```suggestion
					expected := "the following SBoM files will be ignored for buildpack api version < 0.7"
```

natalieparellano (approved):

buildpack/bomfile.go
aemengo: _Fixed in latest commit._

buildpack/build_test.go
aemengo: _Fixed in latest commit._

jabrown85 (approved):

PR #771, opened 2021-11-29
Remove separate packages for platform interface implementation

natalieparellano:
Despite having advocated the current approach, I now regret it. Having to duplicate the entire previous platform when a new version is supported is not optimal. E.g. for platform 0.8, the only difference from 0.7 is that standardized sBOM is now supported, but we had to copy the entire directory and essentially do a find and replace to get everything to compile.

It would be awesome if we could still move logic into the `platform` package (e.g. [here](https://github.com/buildpacks/lifecycle/blob/5744dae9b0e748ccc2daf160e26a0d3dfd5e9de6/exporter.go#L99), this could be a `SupportsSBOM` field or method on `Platform`). But the proposed organization would allow us to do this at our own pace, and reduce a fair amount of duplication.


natalieparellano:
Also makes the code slightly more efficient by having the `Platform` hold on to a `*api.Version` instead of a string.

builder_test.go
```diff
@@ -813,7 +808,9 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 
 		when("platform api < 0.4", func() {
 			it.Before(func() {
-				builder.PlatformAPI = api.MustParse("0.3")
+				var err error
+				builder.Platform = platform.NewPlatform("0.3")
+				h.AssertNil(t, err)
```
natalieparellano, line 813:
```suggestion
				builder.Platform = platform.NewPlatform("0.3")
```

builder_test.go
```diff
@@ -860,7 +857,9 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 
 		when("platform api < 0.6", func() {
 			it.Before(func() {
-				builder.PlatformAPI = api.MustParse("0.5")
+				var err error
+				builder.Platform = platform.NewPlatform("0.5")
+				h.AssertNil(t, err)
```
natalieparellano, line 862:
```suggestion
				builder.Platform = platform.NewPlatform("0.5")
```

jabrown85 (approved):

aemengo (approved):

PR #773, opened 2021-11-30
Fix bug where sbom is not restored for launch=true cache=false layers

natalieparellano:
Because the layer directory does not exist in the build container, the lifecycle
did not bother to look for any sbom files.

This changes `eachDir` to `eachLayer`. Usages of the return value of `eachLayer`
will not fail if the layer directory does not exist.


aemengo (approved):
More tautological. I like it! 👍🏾 

_Verified working on my machine._

analyzer.go
```diff
@@ -51,7 +51,7 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 
 		if api.MustParse(a.Platform.API()).AtLeast("0.8") {
 			if appMeta.BOM != nil && appMeta.BOM.SHA != "" {
-				if err := a.restorePreviousLayer(appMeta.BOM.SHA); err != nil {
+				if err := a.SBOMRestorer.RestoreFromPrevious(a.PreviousImage, appMeta.BOM.SHA); err != nil {
```
natalieparellano, line 54:
```suggestion
				if err := a.SBOMRestorer.RestoreFromImage(a.PreviousImage, appMeta.BOM.SHA); err != nil {
```
This might be better

natalieparellano:
@aemengo do you mind taking another look? The addition of the acceptance test uncovered an additional bug wherein launch=false cache=true sbom files are NOT restored when using a cache directory. The reason for this is that the `-cache-dir` input to the analyzer is ignored on platform API >= 0.7.

To fix this (without having to re-introduce the flag) I moved the logic that restores the cache sbom layer into the restorer. It  felt a bit weird to have the logic for layer restoration spread out across the analyzer and the restorer, so I created a `layer.SBOMRestorer` to consolidate it. LMK what you think.

cc @samj1912

natalieparellano:
😕  there is a permissions issue with the mounted `/layers` volume in the acceptance test. Will push up a fix...

buildpack/build.go
```diff
@@ -208,29 +211,29 @@ func (b *Descriptor) setupEnv(pathToLayerMetadataFile map[string]LayerMetadataFi
...
+	files, err := ioutil.ReadDir(bpLayersDir)
 	if os.IsNotExist(err) {
 		return map[string]LayerMetadataFile{}, nil
 	} else if err != nil {
 		return map[string]LayerMetadataFile{}, err
 	}
-	pathToLayerMetadataFile := map[string]LayerMetadataFile{}
+	bpLayers := map[string]LayerMetadataFile{}
 	for _, f := range files {
-		if !f.IsDir() {
```
jabrown85, line 220: This guard is probably still useful. It’s unlikely to happen but there could be a dir with that suffix.

internal/layer/sbom_restorer.go
```diff
@@ -0,0 +1,131 @@
...
+	layersDir string
+	logger    Logger
+}
+
+func (r *DefaultSBOMRestorer) RestoreFromPrevious(image imgutil.Image, layerDigest string) error {
+	// Sanity check to prevent panic.
+	if image == nil {
+		return errors.Errorf("restoring layer: previous image not found for %q", layerDigest)
+	}
+	r.logger.Debugf("Retrieving previous image layer for %q", layerDigest)
```
jabrown85, line 49: I might suggest changing this so we can tell it from the other restore method.
```suggestion
	r.logger.Debugf("Retrieving previous image layer for sbom layer %q", layerDigest)
```

internal/layer/sbom_restorer.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return layers.Extract(rc, "")
+}
+
+func (r *DefaultSBOMRestorer) RestoreFromCache(cache Cache, layerDigest string) error {
+	// Sanity check to prevent panic.
+	if cache == nil {
+		return errors.New("restoring layer: cache not provided")
+	}
+	r.logger.Debugf("Retrieving data for %q", layerDigest)
```
jabrown85, line 65:
```suggestion
	r.logger.Debugf("Retrieving data for sbom layer %q", layerDigest)
```

jabrown85 (approved):
Couple of nits but this looks like it does what we want.

buildpack/build.go
natalieparellano: Please see fbbb2e66

internal/layer/sbom_restorer.go
natalieparellano: Added in fbbb2e66

internal/layer/sbom_restorer.go
natalieparellano: Added in fbbb2e66

restorer.go
```diff
@@ -101,12 +96,20 @@ func (r *Restorer) Restore(cache Cache) error {
...
-		return errors.Wrap(err, "restoring data")
+	if api.MustParse(r.Platform.API()).AtLeast("0.8") {
+		g.Go(func() error {
+			if cacheMeta.BOM.SHA == "" {
+				return nil
+			}
+			if err := r.SBOMRestorer.RestoreFromCache(cache, cacheMeta.BOM.SHA); err != nil {
+				return err
+			}
+			return r.SBOMRestorer.RestoreToBuildpackLayers(r.Buildpacks)
```
aemengo, line 107: The benefit of introducing an abstraction is that it can _abstract away_ the orchestration logic (_arguably, even the platform API logic_).

aemengo (approved):
Approved, but I have some misgivings.

I don't like how the "restore sBOM" logic is now split between "analyze" and "restore". We originally wrote the code to restore everything in the "restore" phase. And opted against it to avoid introducing a `--docker` (or something else) flag.

Between introducing the flag and having the logic split up, I would have rather introduced before. Like I've said before, out-of-band, this introduces tribal knowledge regarding how the lifecycle works. Another engineer could introduce a bug later because of not having this context.

I don't disagree enough to hold the PR; just voicing my opinion - in writing. Also,
_Verified working on my machine_ 👍🏾

natalieparellano:
@aemengo thank you for your comment, and I have to say that I wholeheartedly agree with you. Upon further reflection we actually have a third bug (see https://github.com/buildpacks/lifecycle/issues/775) that is fixed in dd3796b8 but only by spreading some logic further to the build phase. We should explore making a spec change that would allow all the logic to be consolidated in the restore phase.

PR #777, opened 2021-12-03
Bumps imgutil and also updates indirect dependencies

natalieparellano:
The lifecycle and imgutil will be on the same version of ggcr.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,14 +4,12 @@ require (
 	github.com/BurntSushi/toml v0.4.1
 	github.com/apex/log v1.9.0
 	github.com/buildpacks/imgutil v0.0.0-20211203200417-76206845baac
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -3,19 +3,20 @@ module github.com/buildpacks/lifecycle
 require (
 	github.com/BurntSushi/toml v0.4.1
 	github.com/apex/log v1.9.0
-	github.com/buildpacks/imgutil v0.0.0-20211129173834-372fdd2f7ce6
+	github.com/buildpacks/imgutil v0.0.0-20211203200417-76206845baac
+	github.com/coreos/etcd v3.3.13+incompatible // indirect
```
jabrown85, line 7: This makes me think we missed a `go tidy`

jabrown85 (approved):
Looks fine to me, but I think we could tidy the deps to make it even better

PR #779, opened 2021-12-03
Guard singleton connection

matejvasek:
Right now the connection is used in serial manner so it works just fine.
But if in we used parallel operation in future there would be broken IO,
and the error message doesn't say much.

```
===> EXPORTING
Reusing layers from image with id '946b1c1f946db00a91f90a485dc7e2200de3985dd06b503ed6286f5c54196839'
ERROR: failed to  image: verifying image "946b1c1f946db00a91f90a485dc7e2200de3985dd06b503ed6286f5c54196839": error during connect: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.40/images/946b1c1f946db00a91f90a485dc7e2200de3985dd06b503ed6286f5c54196839/json": net/http: HTTP/1.x transport connection broken: malformed HTTP status code "Red"
ERROR: failed to build: executing lifecycle: failed with status code: 1
```

### DIFF
--- a/priv/docker.go
+++ b/priv/docker.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"sync/atomic"
 
 	"github.com/docker/docker/client"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


natalieparellano:
Looks good to me! @matejvasek thanks for suggesting this.

jabrown85 (approved):

natalieparellano (approved):

PR #782, opened 2021-12-08
Organize analyze inputs

natalieparellano:
cmd/lifecycle/analyzer.go is hard to make changes to, for a few reasons:

* there are several nested structs: `analyzeCmd`, `analyzeArgs`, and `platform06` and it's not obvious where to add new fields
* a lot of branching logic in `analyzeCmd.Args()`

This PR tries to address these problems by:
* Removing the `platform06` struct and moving its fields to the other structs, but with the prefix `legacy`
* ~Making clear where fields belong by renaming `analyzeArgs` -> `analyzeInputsForCreator`~
* Cleaning up a few fields that creator was providing without needing to
* Moving some logic from `analyzeCmd.Args()` into helper functions

If we like this pattern I could make similar changes for other phases if needed.

### DIFF
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -23,88 +23,86 @@ import (
 )
 
 type analyzeCmd struct {
-	//flags: inputs
-	analyzeArgs
-	stackPath string
-	uid, gid  int
-
-	//flags: paths to write data
-	analyzedPath string
+	analyzeInputsForCreator
+	additionalTags  cmd.StringSlice
+	analyzedPath    string
+	cacheImageRef   string
+	legacyCacheDir  string
+	legacyGroupPath string
+	outputImageRef  string
+	stackPath       string
+	uid, gid        int
 }
 
-type analyzeArgs struct {
-	cacheImageRef    string
+// Inputs needed when run by creator
+type analyzeInputsForCreator struct {
 	layersDir        string
-	outputImageRef   string
 	previousImageRef string
 	runImageRef      string
+	legacySkipLayers bool
 	useDaemon        bool
 
-	additionalTags cmd.StringSlice
-	docker         client.CommonAPIClient // construct if necessary before dropping privileges
-	keychain       authn.Keychain
-	platform       Platform
-	platform06     analyzeArgsPlatform06
-}
-
-type analyzeArgsPlatform06 struct {
-	cacheDir   string // not needed when run by creator
-	groupPath  string // not needed when run by creator
-	skipLayers bool
-	cache      lifecycle.Cache
-	group      buildpack.Group
+	docker      client.CommonAPIClient // construct if necessary before dropping privileges
+	keychain    authn.Keychain         // construct if necessary before dropping privileges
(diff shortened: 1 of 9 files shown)


cmd/lifecycle/analyzer.go
```diff
@@ -274,39 +253,61 @@ func (a *analyzeCmd) supportsRunImage() bool {
 	return a.platformAPIVersionGreaterThan06()
 }
 
-func (a *analyzeCmd) validateRunImageInput() error {
-	if !a.supportsRunImage() && a.runImageRef != "" {
-		return errors.New("-run-image is unsupported")
```
jabrown85, line 279: Did this check disappear on purpose?

cmd/lifecycle/creator.go
```diff
@@ -167,13 +167,10 @@ func (c *createCmd) Exec() error {
 	)
 	if c.platform.API().AtLeast("0.7") {
 		cmd.DefaultLogger.Phase("ANALYZING")
-		analyzedMD, err = analyzeArgs{
-			additionalTags:   c.additionalTags,
```
jabrown85, line 171: I guess we can get away without `additinalTags` and `outputImageRef` because creator does validation on its own?

cmd/lifecycle/analyzer.go
```diff
@@ -23,116 +23,95 @@ import (
...
+	legacyGroupPath string
+	outputImageRef  string
+	stackPath       string
+	uid, gid        int
 }
 
-type analyzeArgs struct {
-	cacheImageRef    string
+// Inputs needed when run by creator
+type analyzeInputsForCreator struct {
```
natalieparellano, line 38: I wonder if this rename is truly needed. It would be annoying to rename the struct for all the other phases. Maybe we just enforce the meaning with a comment (this is already mostly true for the other phases).

cmd/lifecycle/analyzer.go
natalieparellano: I realized that this is already covered by `cmd.FlagRunImage(&a.runImageRef)` being scoped to the `if` block in `DefineFlags()` (the error message is `flag provided but not defined: -run-image`). Maybe we could add an acceptance test for it. It's probably also worth adding a comment for `DefineFlags()` because just glancing at the function (without digging into the internals of `FlagSet`) it just looks like it's reading values into memory when it's also defining which flags are valid.

cmd/lifecycle/creator.go
natalieparellano: Though it could be worth consolidating the validation into an additional package or service that could be unit tested. That could help further our goal of moving things out of the `cmd` package.

cmd/lifecycle/analyzer.go
natalieparellano: Added comments in d1988a83

cmd/lifecycle/analyzer.go
natalieparellano: Put back in d1988a83

jabrown85 (approved):

PR #783, opened 2021-12-08
Don't strip lifecycle of debug symbols when $DEBUG=true is passed

aemengo:


aemengo:
I'm closing this PR. Although very useful for debugging, I couldn't make the entire user journey consumer-friendly.

PR #784, opened 2021-12-13
Fix slow performance for creator when restoring sbom layer from previ…

natalieparellano:
…ous image in daemon

For more context see https://github.com/buildpacks/spec/issues/276

### DIFF
--- a/acceptance/creator_test.go
+++ b/acceptance/creator_test.go
@@ -143,16 +143,17 @@ func testCreatorFunc(platformAPI string) func(t *testing.T, when spec.G, it spec
 
 		when("sbom", func() {
 			var (
-				container1 string
-				container2 string
-				container3 string
-				container4 string
-				dirBuild1  string
-				dirBuild2  string
-				dirCache   string
-				dirRun1    string
-				dirRun2    string
-				imageName  string
+				container1     string
+				container2     string
+				container3     string
+				container4     string
+				dirBuild1      string
+				dirBuild2      string
+				dirCache       string
+				dirLaunchCache string
+				dirRun1        string
+				dirRun2        string
+				imageName      string
 			)
 
 			it.Before(func() {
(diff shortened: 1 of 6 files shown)


acceptance/creator_test.go
```diff
@@ -260,9 +264,16 @@ func testCreatorFunc(platformAPI string) func(t *testing.T, when spec.G, it spec
...
+								"--volume", dirLaunchCache+":"+ctrPath("/launch-cache"),
 							)...),
 							h.WithArgs(createArgs...),
 						)
+						// check that launch cache was used
+						duration := time.Now().Sub(startTime)
+						t.Logf("Build duration: %s", duration)
+						if duration > 3*time.Second {
+							t.Fatalf("Expected second build to complete in less than 3 seconds; took %s", duration)
+						}
```
natalieparellano, line 276: I'm a bit wary of this test, as it might introduce flakes. But this seemed like the most straightforward way to test that the change does what we want. Open to suggestions...

natalieparellano:
Current lifecycle, slow re-build:
```
$  time pack build test-current -B paketobuildpacks/builder:base
base: Pulling from paketobuildpacks/builder
Digest: sha256:d4225fb4b5bceb5702fadc06f63e53ed1599150505e88ac8ad22efb3c527845b
Status: Image is up to date for paketobuildpacks/builder:base
base-cnb: Pulling from paketobuildpacks/run
Digest: sha256:f9e78e309fe5af1d376d3ce4593e3d25260604e4a62430dad7ec65b11289cf89
Status: Image is up to date for paketobuildpacks/run:base-cnb
===> ANALYZING
...
Successfully built image test-current

real    0m22.270s
user    0m0.073s
sys     0m0.061s
```

Fixed lifecycle, faster re-build with log line:
```
d7ec65b11289cf89
Status: Image is up to date for paketobuildpacks/run:base-cnb
===> ANALYZING
Restoring data for sbom from previous image
...
Successfully built image test-fix

real    0m13.925s
user    0m0.081s
sys     0m0.073s
```

acceptance/creator_test.go
jabrown85: Yeah, it could become a problem but I'm OK with it as a general idea to make sure there are no large regressions.

jabrown85 (approved):

edmorley:
@natalieparellano Hi! Does this performance issue all builds, or only those where people have opted into SBOM? (At the moment all of our buildpacks are using buildpack API 0.6 or older, so aren't generating SBOMs; trying to work out if we need this fix soon, or can wait a bit.)

PR #789, opened 2022-01-06
Set process working directory from launch.toml

### LINKED ISSUE #557: [RFC 0081] Process Specific Working Directories
[RFC 0081](https://github.com/buildpacks/rfcs/blob/main/text/0081-process-specific-working-directory.md)
https://github.com/buildpacks/rfcs/pull/144/files

mboldt:
Resolves #557.

Chatted with @natalieparellano about this before the holidays as this being a small step towards [removing shell processes](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md).

### DIFF
--- a/launch/bash.go
+++ b/launch/bash.go
@@ -2,6 +2,7 @@ package launch
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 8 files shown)


mboldt:
Changed to draft, as I realized that I have overlooked Windows cmd.exe execution.

launch/bash.go
```diff
@@ -31,6 +32,9 @@ func (b *BashShell) Launch(proc ShellProcess) error {
 		bashCommand = bashCommandWithTokens(len(proc.Args) + 1)
 	}
 	launcher += bashCommand
+	if err := os.Chdir(proc.WorkingDirectory); err != nil {
```
natalieparellano, line 35: This would also change the working directory for profile.d scripts. I think instead we'll want to throw a `cd workdir` right about here: https://github.com/buildpacks/lifecycle/blob/6f56a153034317a88f41109b76e026b93134eb89/launch/bash.go#L33

natalieparellano:
We'll need to add logic to the build phase to find the process working directory in the output launch.toml and write it to /layers/config/metadata.toml. We'll also want to inspect the buildpack api and only respect the working directory if the buildpack api supports it.

mboldt:
Thank you, @natalieparellano ! I'm going to close this for now and take another pass at it with those comments in mind.

PR #795, opened 2022-01-19
Read and store `io.buildpacks.id` in analyzer phase

jabrown85:
Implements #742

TODO:
- [ ] We need spec to land (See https://github.com/buildpacks/spec/issues/258)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .idea
+.tool-versions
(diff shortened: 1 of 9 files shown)


jabrown85:
@natalieparellano I left this in draft since we don't have a spec PR to pull requirements from. Happy to pick this back up when there are comments on the PR and/or spec guidelines that change behavior. 😄

PR #797, opened 2022-01-20
WIP: export kaniko layers on run image

jabrown85:
Loop through run-image extensions and execute them one by one

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -93,9 +93,10 @@ func (e *Exporter) Export(opts ExportOptions) (platform.ExportReport, error) {
 	}
 
 	// extender layers
-	if err := e.addExtenderLayers(opts, &meta); err != nil {
-		return platform.ExportReport{}, err
-	}
+	// the new base is currently being referenced via run-image
+	// if err := e.addExtenderLayers(opts, &meta); err != nil {
+	// 	return platform.ExportReport{}, err
+	// }
 
 	// buildpack-provided layers
 	if err := e.addBuildpackLayers(opts, &meta); err != nil {
--- a/exporter.go
+++ b/exporter.go
@@ -292,42 +292,59 @@ func (e *Exporter) addAppLayers(opts ExportOptions, slices []layers.Slice, meta 
 
 func (e *Exporter) addExtenderLayers(opts ExportOptions, meta *platform.LayersMetadata) error {
 	e.Logger.Info("Adding extender layers")
-	manifestPath := "/layers-run/kaniko/manifest.json"
-	var manifest []struct {
-		Layers []string `json:"Layers"`
-	}
 
-	f, err := os.Open(manifestPath)
+	path := filepath.Join("/layers-run/kaniko")
+	dirs, err := ioutil.ReadDir(path)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return nil
 		}
-
 		return err
 	}
 
-	defer f.Close()
+	for _, dir := range dirs {
+		if !dir.IsDir() {
+			continue
+		}
 
-	contents, _ := ioutil.ReadAll(f)
-	e.Logger.Infof("kaniko manifest: %s", string(contents))
-	f.Seek(0, 0)
+		e.Logger.Infof("Adding extender layers for %s", dir.Name())
+		manifestPath := filepath.Join(path, dir.Name(), "manifest.json")
+		var manifest []struct {
+			Layers []string `json:"Layers"`
+		}
 
-	err = json.NewDecoder(f).Decode(&manifest)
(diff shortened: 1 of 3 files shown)


test-extender.sh
```diff
@@ -13,14 +17,28 @@ FROM cnbs/sample-builder:bionic
...
 cd $SAMPLES_REPO_PATH
 
 rm -rf $SAMPLES_REPO_PATH/kaniko
 mkdir -p $SAMPLES_REPO_PATH/kaniko
 rm -rf $SAMPLES_REPO_PATH/layers/kaniko
 mkdir -p $SAMPLES_REPO_PATH/layers/kaniko
+rm -rf $SAMPLES_REPO_PATH/kaniko-run
+mkdir -p $SAMPLES_REPO_PATH/kaniko-run
+rm -rf $SAMPLES_REPO_PATH/layers-run/kaniko
+mkdir -p $SAMPLES_REPO_PATH/layers-run/kaniko
```
natalieparellano, line 41: I guess a next step in this effort could be to try to make these directories useful for caching purposes (though alternative approaches to caching such as spinning up a registry process were also discussed).

test-extender.sh
```diff
@@ -54,6 +77,37 @@ docker run \
...
   -v $PWD/cnb/buildpacks/:/cnb/buildpacks \
   -u root \
-  test-builder \
-  /cnb/lifecycle/extender kaniko build ubuntu:bionic
+  $REGISTRY_HOST/test-builder \
+  /cnb/lifecycle/extender kaniko build "$REGISTRY_HOST/test-builder"
+  #              args:    <kaniko|buildah> <build|run> <base-image>
+
+echo ">>>>>>>>>> Running extend on run image..."
+ # TODO: we should probably not have to mount /cnb/ext? Can we copy the static Dockerfiles to layers?
```
natalieparellano, line 85: That sounds like a good idea to me. The lifecycle currently just leaves the Dockerfiles where the extensions wrote them, but copying them sounds like a logical next step.

extender/main.go
```diff
@@ -108,53 +169,13 @@ func doKaniko(kind, baseimage string) {
...
+		// TODO: next base image is the one we just built, we should use digest instead
+		current_base_image = b.Destination
 	}
 
+	// run buildpacks now
 	if kind == "build" {
-		// We're in "build" mode, untar layers to root filesystem: /
-		for _, layerFile := range layerFiles {
-			workingDirectory := "/"
-			tarPath := "/layers/kaniko/" + layerFile
```
cmoulliard, line 150: We should maybe avoid to hard code `kaniko` but pass it as parameter as it could be possible that another solution could be supported in the future (buildah, ...)

extender/main.go
```diff
@@ -108,53 +169,13 @@ func doKaniko(kind, baseimage string) {
...
-
-			err = untar(tarPath, workingDirectory)
-			if err != nil {
-				panic(err)
-			}
-		}
-
 		// Run the build for buildpacks with lowered privileges.
 		// We must assume that this extender is run as root.
 		cmd := exec.Command("/cnb/lifecycle/builder", "-app", "/workspace", "-log-level", "debug")
```
cmoulliard, line 181: Do you plan to log an error message for the end user to be informed that a `root privileged are needed` ?

Remark: Why do you say that  `Run the build for buildpacks with lowered privileges.` as comment ?

exporter.go
```diff
@@ -91,6 +92,12 @@ func (e *Exporter) Export(opts ExportOptions) (platform.ExportReport, error) {
...
 	}
 
+	// extender layers
+	// the new base is currently being referenced via run-image
+	// if err := e.addExtenderLayers(opts, &meta); err != nil {
+	// 	return platform.ExportReport{}, err
+	// }
+
 	// buildpack-provided layers
 	if err := e.addBuildpackLayers(opts, &meta); err != nil {
```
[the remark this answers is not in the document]
jabrown85: This function already existed and only handles buildpack layers. I actually could remove the exporter changes and `addExtenderLayers` entirely. But I didn't do so yet depending on how we plan on moving forward. My plan was to assume a new run image reference came out of the run extension and was already pushed to the remote registry. We could alternatively have the run extensions save a run image as OCI Layout/Tarball and then have exporter build with those layers (or even push up the run image to the remote registry).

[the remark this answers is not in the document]
cmoulliard: Add then a TODO to review later or better document what this step of the code is doing ?

natalieparellano:
@jabrown85 I'm going to close this as all the work is captured in https://github.com/buildpacks/lifecycle/pull/802

PR #794, opened 2022-01-18
Warn when legacy bom is specified on buildpack API 0.7

aemengo:
as opposed to throwing an error. This allows for non-breaking platform migration from legacy bom table to sbom format

This commit partially reverts 880a801db2d4bfbb39671a66f7aadd96c0231e37

PR #798, opened 2022-01-24
If there are is no sbom layer in the cache, still copy sbom files for launch

natalieparellano:
...that were restored from the previous image.

Found this bug while verifying #794


natalieparellano:
cc @aemengo

jabrown85 (approved):

PR #800, opened 2022-01-27
Improve performance when restoring launch sboms from daemon

natalieparellano:
See spec PR: https://github.com/buildpacks/spec/pull/278/files

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -49,12 +49,9 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 			appMeta = platform.LayersMetadata{}
 		}
 
-		if a.Platform.API().AtLeast("0.8") {
-			if appMeta.BOM != nil && appMeta.BOM.SHA != "" {
-				a.Logger.Infof("Restoring data for sbom from previous image")
-				if err := a.SBOMRestorer.RestoreFromPrevious(a.PreviousImage, appMeta.BOM.SHA); err != nil {
-					return platform.AnalyzedMetadata{}, errors.Wrap(err, "retrieving launch sBOM layer")
-				}
+		if bomSHA(appMeta) != "" {
+			if err := a.SBOMRestorer.RestoreFromPrevious(a.PreviousImage, bomSHA(appMeta)); err != nil {
+				return platform.AnalyzedMetadata{}, errors.Wrap(err, "retrieving launch sBOM layer")
 			}
 		}
 	} else {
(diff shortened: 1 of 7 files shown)


internal/layer/sbom_restorer.go
```diff
@@ -29,10 +30,19 @@ type Cache interface {
 	RetrieveLayer(sha string) (io.ReadCloser, error)
 }
 
-func NewSBOMRestorer(layersDir string, logger Logger) SBOMRestorer {
+type SBOMRestorerOpts struct {
+	LayersDir string
+	Logger    Logger
+	Nop       bool
+}
```
natalieparellano, line 37: This struct makes it a bit clearer when we're calling the constructor what each option is for.

analyzer.go
```diff
@@ -49,12 +49,9 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 			appMeta = platform.LayersMetadata{}
 		}
 
-		if a.Platform.API().AtLeast("0.8") {
```
natalieparellano, line 52: Because of https://github.com/buildpacks/lifecycle/pull/800/files#diff-f664818a668bae8688d032e037ff762c9a7537beec1916963461e36dd1c40dc3R40 the analyzer can be a bit more ignorant of platform API logic, and just be an "orchestrator".

natalieparellano:
@aemengo @jabrown85 LMK what you think of this pattern. If we move logic that could differ by platform into services, then the switch on platform API logic happens in the constructor.

analyzer.go
```diff
@@ -49,12 +49,9 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
...
-			if appMeta.BOM != nil && appMeta.BOM.SHA != "" {
-				a.Logger.Infof("Restoring data for sbom from previous image")
-				if err := a.SBOMRestorer.RestoreFromPrevious(a.PreviousImage, appMeta.BOM.SHA); err != nil {
-					return platform.AnalyzedMetadata{}, errors.Wrap(err, "retrieving launch sBOM layer")
-				}
+		if bomSHA(appMeta) != "" {
```
aemengo, line 52: Forgive me for [repeating myself](https://github.com/buildpacks/lifecycle/pull/773#discussion_r761418935) but might as well move all the `if`, if there's a brand new object.

aemengo (approved):

analyzer.go
natalieparellano: I started to implement your suggestion with gusto, but now I'm more on the fence about it. Removing the `if` would mean giving `SBOMRestorer` knowledge of `platform.LayersMetadata` (and probably also `platform.CacheMetadata`) which it doesn't really need. It would be nice to be able to change the schema of these labels and leave the restorer's behavior unaffected.

aemengo: That’s totally sound rationale. I can’t argue with that 🙂

natalieparellano: I think I confused myself... we could move the empty check inside the service (we'd still pass it `bomSHA(appMeta)`). 

> To answer your original question: I like this “Null Object” pattern a lot 👍🏾

I'm so happy!

natalieparellano: > we could move the empty check inside the service

Please see 9ad2d509

jabrown85 (approved):

PR #786, opened 2021-12-15
Dockerfiles: run /bin/detect and /bin/build for extensions

natalieparellano:
This PR updates the lifecycle to:
* Run /bin/detect for extensions as part of regular detect
* Run /bin/build for extensions as a new invocation of build

Note that this PR does NOT tackle actually applying Dockerfiles to the build and/or run image - for the work being done there, see #709 

### Testing

**Update** for ease of use, please use the `test-extender.sh` script. Example invocation: `LIFECYCLE_REPO_PATH=~/workspace/lifecycle SAMPLES_REPO_PATH=~/workspace/samples/ ./test-extender.sh`

The following assumes that github.com/buildpacks/samples is checked out to the `dockerfiles-poc` branch
```
cd $LIFECYCLE_REPO_PATH

docker image rm test-builder --force

make clean build-linux-amd64

cd out/linux-amd64

cat << EOF > Dockerfile
FROM cnbs/sample-builder:bionic
COPY ./lifecycle /cnb/lifecycle
EOF

docker build -t test-builder .

cd $SAMPLES_REPO_PATH

docker run \
  -v $PWD/workspace/:/workspace \
  -v $PWD/layers/:/layers \
  -v $PWD/platform/:/platform \
  -v $PWD/cnb/ext/:/cnb/ext \
  test-builder \
  /cnb/lifecycle/detector -order /layers/order.toml -log-level debug

docker run \
  -v $PWD/workspace/:/workspace \
  -v $PWD/layers/:/layers \
  -v $PWD/platform/:/platform \
  -v $PWD/cnb/ext/:/cnb/ext \
  test-builder \
  /cnb/lifecycle/builder -use-extensions -log-level debug
```

You should see:
* Detect passes, `./layers/group.toml` is created with `samples/curl` and `samples/rebasable`; empty `./layers/plan.toml` is created
* Build passes, `./layers/config/metadata.toml` is created with the "buildpacks" (really extensions) that participated in the build and a list of Dockerfiles that should be applied by the build and/or run extender

PR #802, opened 2022-02-15
Dockerfiles POC

natalieparellano:
This PR updates https://github.com/buildpacks/lifecycle/pull/786 and integrates the work done in https://github.com/buildpacks/lifecycle/pull/797. It also pulls in code from https://github.com/redhat-buildpacks/poc/tree/main/kaniko/code.

In theory, using this branch it is possible to perform a build that uses extensions to extend the build and/or run image and get an app image that is based off the extended run image.

To see it in action, run test-extender.sh (use `REGISTRY_HOST` to configure the registry where the extended images should be saved).

Changes of note:
* extender/spec.md for proposed platform <-> extender interface
* new cmd/extender package
* new extender package
  * the `DockerfileApplier` interface (this could allow platforms to supply their own implementation, using `extender` as a library to provide CNB logic) 
* changes to detector.go and builder.go
* changes to buildpack package:
  * buildpack/build.go
  * buildpack/detect.go
  * buildpack/dockerfile.go
  * buildpack/extension.go

Significant work still to be done:
* kaniko implementation:
  * Enable and verify caching
  * Enable "ignore paths" provided by the end user
  * When extending the build image, don't push and pull the image between Dockerfile builds (we can just parse the commands in the Dockerfile)
  * Decide if we need an in-process registry
* Decide what build.Dockerfiles should be considered "valid" (e.g., they must start with `ARG #{base_image}`)
* Set the correct value of `io.buildpacks.rebasable`
* Run `genpkgs`
* Exclude `genpkgs` from the final image
* ...much more


extender/testdata/cnb/buildpacks/samples_use_curl/0.0.1/buildpack.toml
```diff
@@ -0,0 +1,9 @@
+api = "0.7"
+
+[buildpack]
+id = "samples/use_curl"
+version = "0.0.1"
+name = "Buildpack using curl"
+
+[[stacks]]
+id = "*"
```
jabrown85, line 9: nit:newline

```suggestion
id = "*"

```

extender/testdata/cnb/ext/samples_curl/0.0.1/bin/build
```diff
@@ -0,0 +1,41 @@
...
+output_dir=$1
+plan_path=$3
+
+# Create Dockerfile
+cat << EOF > ${output_dir}/Dockerfile
+ARG base_image
+FROM \${base_image}
+
+# TODO: why wasn't this needed before?
+USER root
```
jabrown85, line 16: I believe this was not needed before because we were executing the instructions on the builder which was running as root.

This raises a good question/concern. If, for run extensions, we apply `Dockerfile` after `Dockerfile` and the last `Dockerfile` leaves the `USER root` as the last instruction...is that ok? That feels wrong as one of the pillars of CNB is that the resulting image should _not_ be root IMO. It gets a bit muddier when you think about full run image replacements as well. Does the final image need to be using the same `CNB_USER_ID` as the original run image reference to make things happy/correct?

I'm sure @sclevine has some thoughts on this...

jabrown85 (approved):
This looks like a great start to me!

BarDweller:
Ran into an issue where if the Dockerfile being processed attempts to pull any new images, the pull failed because kaniko had a null/empty default platform, resulting in errors like.. 

```
time="2022-02-17T18:34:55Z" level=info msg="Retrieving image manifest redhat/ubi8-minimal:latest"
time="2022-02-17T18:34:55Z" level=info msg="Retrieving image redhat/ubi8-minimal:latest from registry index.docker.io"
time="2022-02-17T18:34:55Z" level=error msg="Error parsing the serverURL" error="docker-credential-ecr-login can only be used with Amazon Elastic Container Registry." serverURL=index.docker.io
ERROR: no child with platform  in index redhat/ubi8-minimal:latest
```

(Note the empty platform name in the final ERROR output)

Did some digging, found kaniko treats the CustomPlatform option differently to all it's others, and essentially requires it to be set to the default platform if the value is not set by a user. 

Have created a quick PR for the branch of this PR to fix that.. https://github.com/buildpacks/lifecycle/pull/804

natalieparellano:
Some thoughts on next steps...

While the POC seems to work, there are a number of features missing. Some are procedural / cosmetic, but there are a few harder problems to solve:
* isolating registry credentials from Dockerfiles
* running syft on the extended image (see related [comment](https://github.com/buildpacks/spec/pull/298#issuecomment-1099314512))
* validations on the extended image (is it rebasable, did the user change, etc.)

For isolating registry credentials, the platform would need to pull all the base images from the registry before starting the extender container. The base images could be provided to kaniko in either a local registry, or in a [cache directory](https://github.com/GoogleContainerTools/kaniko#--cache-dir) - but the latter seems not performant as kaniko expects to find the layers in a single tar, which would make it difficult to re-use layers across images. We could see about PR'ing kaniko the ability to lookup layers from a blob store.

Another problem is outputs - kaniko can output to either a registry or to an oci layout directory. The former might be more convenient as it is readily chain-able as an input to the next Dockerfile. But there are other headaches associated with creating a local registry (such as networking) that would need to be thought through.

The following diagram illustrates how this might work. It assumes we are using a local registry but the idea is the same if we try to go the other route (changing how `--cache-dir` behaves in kaniko, making in chain-able). The other discussions currently underway in the project ([1](https://github.com/buildpacks/rfcs/pull/201), [2](https://github.com/buildpacks/rfcs/pull/203)) might influence which route we decide to take. cc @jjbustamante @jromero @samj1912  

Screen Shot 2022-04-14 at 12 12 41 PM

natalieparellano:
Also just to add - https://github.com/buildpacks/spec/pull/298#issuecomment-1099212660 proposes a phased implementation that would let us get something out that would deliver value while we work out some of these thornier issues.

natalieparellano:
Closing as I think this POC has served its purpose. For further details on the "real" implementation please follow https://github.com/buildpacks/rfcs/issues/224

PR #805, opened 2022-02-23
Move logic out of cmd/lifecycle/analyzer.go

natalieparellano:
Resolves https://github.com/buildpacks/lifecycle/issues/635. By moving logic out of `main`, we can simplify our acceptance tests. It should also make the logic easier to follow (especially with respect to platform api branching). 

~~This still needs a few unit tests (marked with `TODO`) but I am marking it ready so as to receive feedback on the general direction.~~

If we like this, I will follow up with another PR to remove a bunch of tests from acceptance/analyzer_test.go.

I am hoping we could use this as a "template" for updating the other phases.


cmd/lifecycle/analyzer.go
```diff
@@ -3,139 +3,75 @@ package main
...
-	if a.platformAPIVersionGreaterThan06() {
-		cmd.FlagPreviousImage(&a.previousImageRef)
-		cmd.FlagRunImage(&a.runImageRef)
-		cmd.FlagStackPath(&a.stackPath)
-		cmd.FlagTags(&a.additionalTags)
-	} else {
-		cmd.FlagCacheDir(&a.legacyCacheDir)
-		cmd.FlagGroupPath(&a.legacyGroupPath)
-		cmd.FlagSkipLayers(&a.skipLayers)
+	switch {
```
jabrown85, line 28: I love the switch, keeping track of what was being registered was super annoying before.

cmd/lifecycle/platform/analyze_inputs.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/buildpacks/lifecycle"
+	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/image"
+	"github.com/buildpacks/lifecycle/internal/str"
+)
+
+// AnalyzeInputs holds the values of command-line flags and args.
+// Fields are the cumulative total of inputs across all supported platform APIs.
+type AnalyzeInputs struct { // TODO: make "to AnalyzerOpts" method? Do these structs need to be different?
```
jabrown85, line 16: I wonder if it might read better to do an options pattern for these type of structs. Just thinking out loud, it might not make sense.

```go
inputs, err := NewAnalyzeInputs(platformAPI,  logger, 
  WithCacheImage(),
  WithLayersDir(),
  ...
)
```

jabrown85 (commented):
I didn't do a full review of every line or anything, but I 👍  the direction.

natalieparellano:
@jabrown85 I think this is ready for another review. My only question at this point - what should be the home of `github.com/buildpacks/lifecycle/cmd/lifecycle/platform`? I was thinking `github.com/buildpacks/lifecycle/platform`, but I don't want to increase the dependencies of the launcher, which currently uses this package. What about `github.com/buildpacks/lifecycle/platform/build` and `github.com/buildpacks/lifecycle/platform/launch`?

jabrown85 (commented):
This looks pretty good to me! It is a lot of changes. The tests are happy and I think it reads and feels better.

As far as the final package location goes, I think I like `github.com/buildpacks/lifecycle/platform` and `github.com/buildpacks/lifecycle/platform/launch` for the launcher specific deps. I think the naming is good for the one we use most of time and having a specific `/launch` package for the launch-specific things feels good. I am also happy to discuss alternatives on naming

natalieparellano:
@jabrown85 I took a look at this today... the problem with putting everything that's currently in `github.com/buildpacks/lifecycle/cmd/lifecycle/platform` in `github.com/buildpacks/lifecycle/platform` is that it creates an import cycle:

```
package command-line-arguments
        imports github.com/buildpacks/lifecycle
        imports github.com/buildpacks/lifecycle/internal/layer
        imports github.com/buildpacks/lifecycle/platform    <---- things like platform.LayersMetadata
        imports github.com/buildpacks/lifecycle: import cycle not allowed
```

It also feels weird putting things like `AnalyzeInputsResolver` and `LayersMetadata` in the same package, because they have totally different consumers. 

What about `github.com/buildpacks/lifecycle/cmd/lifecycle/platform/cmd` and `github.com/buildpacks/lifecycle/cmd/lifecycle/platform/launch/cmd`? Or something else?

jabrown85:
> What about `github.com/buildpacks/lifecycle/cmd/lifecycle/platform/cmd` and `github.com/buildpacks/lifecycle/cmd/lifecycle/platform/launch/cmd`? Or something else?

Would `github.com/buildpacks/lifecycle/cli` and `github.com/buildpacks/lifecycle/cli/launch` or `github.com/buildpacks/lifecycle/cli/platform`/`github.com/buildpacks/lifecycle/cli/launch` work?

If not, I'm also ok with `github.com/buildpacks/lifecycle/cmd/lifecycle/platform/cmd`, but don't really like the duplicate `cmd` in the same import. It feels bleh.

natalieparellano:
@jabrown85 I think this is ready for another round...

platform/inputs/analyze.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return images
+}
+
+type AnalyzeResolver struct {
+	PlatformAPI *api.Version
+}
+
+// Resolve accepts AnalyzeInputs with flags filled in, and args.
+// It returns AnalyzeInputs with default values filled in, or an error if the provided inputs are not valid.
```
natalieparellano, line 59:
```suggestion
// Resolve accepts Analyze with flags filled in, and args.
// It returns Analyze with default values filled in, or an error if the provided inputs are not valid.
```
```suggestion
// Resolve accepts AnalyzeInputs with flags filled in, and args.
// It returns AnalyzeInputs with default values filled in, or an error if the provided inputs are not valid.
```

natalieparellano, line 59:
```suggestion
// Resolve accepts Analyze with flags filled in, and args.
// It returns Analyze with default values filled in, or an error if the provided inputs are not valid.
```
```suggestion
// Resolve accepts AnalyzeInputs with flags filled in, and args.
// It returns AnalyzeInputs with default values filled in, or an error if the provided inputs are not valid.
```

natalieparellano: IDK about this struct name

natalieparellano: IDK about this struct name

platform/inputs/analyze.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/buildpacks/lifecycle"
+	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/buildpack"
+	"github.com/buildpacks/lifecycle/image"
+	"github.com/buildpacks/lifecycle/internal/str"
+)
+
+// Analyze holds the values of command-line flags and args.
+// Fields are the cumulative total of inputs across all supported platform APIs.
+type Analyze struct {
```
natalieparellano, line 17: For simplicity and ease of comprehension, maybe this file and analyzer_factory.go should be one file. The idea is basically to create a pipeline from user-provided flags and args to a `lifecycle.Analyzer`.

platform/inputs/analyze.go
mboldt: I like how `Inputs` alludes to the spec.

natalieparellano:
I'm going to update this PR with some feedback from @mboldt !

natalieparellano:
Just pushed some updates, I will leave some comments shortly describing the substance and intent of the changes.

analyzer.go
```diff
@@ -15,41 +16,186 @@ type Platform interface {
...
 
+type AnalyzerFactory struct {
+	platformAPI     *api.Version
+	cacheHandler    CacheHandler
+	configHandler   ConfigHandler
+	imageHandler    ImageHandler
+	registryHandler RegistryHandler
+}
+
+func NewAnalyzerFactory(platformAPI *api.Version, cacheHandler CacheHandler, configHandler ConfigHandler, imageHandler ImageHandler, registryHandler RegistryHandler) *AnalyzerFactory {
```
natalieparellano, line 27: It seemed like the constructor for a `lifecycle.Analyzer` belonged in the `lifecycle` package.

analyzer.go
```diff
@@ -15,41 +16,186 @@ type Platform interface {
...
+	if err := f.setPrevious(analyzer, previousImageRef, launchCacheDir); err != nil {
+		return nil, err
+	}
+	if err := f.setRun(analyzer, runImageRef); err != nil {
+		return nil, err
+	}
+	return analyzer, nil
+}
+
+func (f *AnalyzerFactory) ensureRegistryAccess(
```
natalieparellano, line 105: I got rid of the "ops" pattern - it seemed to make things more complicated.

analyzer_test.go
```diff
@@ -22,29 +22,325 @@ import (
...
+					fakeImageHandler,
+					fakeRegistryHandler,
+				)
+			})
+
+			it("configures the analyzer", func() {
+				previousImage := fakes.NewImage("some-previous-image-ref", "", nil)
+				runImage := fakes.NewImage("some-run-image-ref", "", nil)
+
+				t.Log("ensures registry access")
```
natalieparellano, line 82: The tests read better now IMO.

cmd/lifecycle/main.go
```diff
@@ -76,6 +82,90 @@ func subcommand(platform Platform) {
 	}
 }
 
+// handlers
```
natalieparellano, line 85: These seem fine to live in the `main` package, because they are very hard to unit test and will be covered by acceptance.

internal/layer/metadata_restorer.go
```diff
@@ -12,23 +12,23 @@ import (
...
 }
 
 func NewMetadataRestorer(logger Logger, layersDir string, skipLayers bool) MetadataRestorer {
 	return &DefaultMetadataRestorer{
-		logger:     logger,
-		layersDir:  layersDir,
-		skipLayers: skipLayers,
+		Logger:     logger,
+		LayersDir:  layersDir,
+		SkipLayers: skipLayers,
```
natalieparellano, line 24: These could maybe be made private

internal/layer/sbom_restorer.go
```diff
@@ -41,14 +41,14 @@ func NewSBOMRestorer(opts SBOMRestorerOpts, platformAPI *api.Version) SBOMRestor
 		return &NopSBOMRestorer{}
 	}
 	return &DefaultSBOMRestorer{
-		layersDir: opts.LayersDir,
-		logger:    opts.Logger,
+		LayersDir: opts.LayersDir,
+		Logger:    opts.Logger,
```
natalieparellano, line 45: These could maybe be made private

natalieparellano: Ah - I made them public so they could be tested here: https://github.com/buildpacks/lifecycle/blob/8b9e787a0028978da28976fc6d63a4de82bc9dc0/analyzer_test.go#L104-L105

internal/layer/metadata_restorer.go
natalieparellano: I made them public so they could be tested here: https://github.com/buildpacks/lifecycle/blob/8b9e787a0028978da28976fc6d63a4de82bc9dc0/analyzer_test.go#L278-L280

acceptance/phase_test.go
```diff
@@ -144,7 +145,7 @@ func (d *targetDaemon) createFixtures(t *testing.T) {
 	var fixtures daemonImageFixtures
 
 	appMeta := minifyMetadata(t, filepath.Join("testdata", "app_image_metadata.json"), platform.LayersMetadata{})
-	cacheMeta := minifyMetadata(t, filepath.Join("testdata", "cache_image_metadata.json"), platform.CacheMetadata{})
+	cacheMeta := minifyMetadata(t, filepath.Join("testdata", "cache_image_metadata.json"), cache.Metadata{})
```
natalieparellano, line 148: I originally moved this struct to try and avoid an import cycle. I think now it could move back to the platform package, but maybe it's better here (image and volume caches have `SetMetadata(metadata Metadata)` and `RetrieveMetadata()`).

analyzer.go
```diff
@@ -15,41 +16,182 @@ type Platform interface {
...
 	SBOMRestorer  layer.SBOMRestorer
 
 	// Platform API < 0.7
 	Buildpacks            []buildpack.GroupBuildpack
 	Cache                 Cache
 	LayerMetadataRestorer layer.MetadataRestorer
+	RestoresLayerMetadata bool
+}
+
+func (f *AnalyzerFactory) NewAnalyzer(
```
natalieparellano, line 50: In the past, adding new inputs to the analyzer/restorer/etc. was a source of bugs because we would forget to update the creator. Accepting the inputs as a list, while ugly, can avoid this problem.

cmd/lifecycle/main.go
```diff
@@ -76,6 +82,90 @@ func subcommand(platform Platform) {
...
+	return h.docker != nil
+}
+
+type DefaultConfigHandler struct{}
+
+func NewConfigHandler() *DefaultConfigHandler {
+	return &DefaultConfigHandler{}
+}
+
+func (h *DefaultConfigHandler) ReadGroup(path string) ([]buildpack.GroupBuildpack, error) {
```
natalieparellano, line 158: I wanted this to be in the lifecycle package, but its dependence on `verifyBuildpackApis` (and needing to return the correct error code) made things a bit complicated.

platform/platform.go
```diff
@@ -4,24 +4,34 @@ import (
...
-	}
-	return &platform
 }
 
+// API returns the platform API.
 func (p *Platform) API() *api.Version {
 	return p.api
 }
+
+// InputsResolver resolves inputs for each of the lifecycle phases.
```
natalieparellano, line 29: Hopefully, eventually! (for now just analyze...)

platform/utils.go
```diff
@@ -0,0 +1,56 @@
+package platform
```
natalieparellano, line 1: These functions should eventually go somewhere else... but I think their new homes will be created when we do additional cleanup as part of https://github.com/buildpacks/rfcs/pull/217

acceptance/phase_test.go
natalieparellano: Moved it back to make the diff smaller...

mboldt (approved):
Chatted with @natalieparellano - looks good!

cmd/launcher/platform/platform_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func testPlatform(platformAPI *api.Version) func(t *testing.T, when testspec.G, it testspec.S) {
+	return func(t *testing.T, when testspec.G, it testspec.S) {
+		when("#NewPlatform", func() {
+			when("platform api >= 0.6", func() {
+				it.Before(func() {
+					h.SkipIf(t, platformAPI.LessThan("0.6"), "")
+				})
+
+				it("configures the platformr", func() {
```
mboldt: typo

cmd/lifecycle/platform/platform_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+func testPlatform(platformAPI *api.Version) func(t *testing.T, when testspec.G, it testspec.S) {
+	return func(t *testing.T, when testspec.G, it testspec.S) {
+		when("#NewPlatform", func() {
+			when("platform api >= 0.6", func() {
+				it.Before(func() {
+					h.SkipIf(t, platformAPI.LessThan("0.6"), "")
+				})
+
+				it("configures the platformr", func() {
```
mboldt: typo

PR #812, opened 2022-03-02
Fix: order.toml has an `order-ext` table

natalieparellano:


detector.go
```diff
@@ -115,7 +116,11 @@ func (d *Detector) detectGroup(group buildpack.Group, done []buildpack.GroupBuil
 			buildable buildpack.Buildable
 			err       error
 		)
-		if groupBuildable.Extension {
+		if len(groupBuildable.OrderExt) > 0 {
```
jabrown85, line 119: Could we make `d.ExtensionStore.Lookup` return a buildable that represents a metabuildpack given a known `groupBuildable.ID` or similar?  That way we can fall into the existing `bpDesc.IsMetaBuildpack` below?

natalieparellano: That would probably be better than the current implementation :)

Part of the pain IMO is the fact that `GroupBuildable` (what was formerly `GroupBuildpack`) is used everywhere throughout the code, for different reasons - representing the contents of order.toml, representing the contents of group.toml (with a slightly different schema), "printing" the identifier of a buildpack in a label (e.g., to associate a bom entry with a buildpack). I started to look into what might make this code a bit nicer, but it seemed to be a lot of work :/ We might look at it for the "real" implementation.

That aside, I agree with your suggestion. That and  https://github.com/buildpacks/lifecycle/pull/802#discussion_r807386907 would make the store interface more useful.

jabrown85 (approved):

PR #813, opened 2022-03-02
Cache image for kaniko is configurable

natalieparellano:
Piggy backing off the work in https://github.com/buildpacks/lifecycle/pull/810


cmd/extender/extender.go
```diff
@@ -42,7 +43,8 @@ type extendCmd struct {
 
 func (e *extendCmd) DefineFlags() {
 	cmd.FlagAppDir(&e.appDir)
-	cmd.FlagCacheDir(&e.cacheDir)
+	cmd.FlagCacheDir(&e.cacheDir) // TODO: will this ever be used by kaniko? Should we remove it?
```
jabrown85, line 46: Only if we want platforms to dump build/run image bases as tarballs so kaniko doesn't have to pull them in during extension runs. Could be useful as a readonly volume if the platform allows only certain base images.

extender/kaniko/dockerfile_applier.go
```diff
@@ -104,9 +87,9 @@ func (a *DockerfileApplier) ApplyRun(dockerfiles []extender.Dockerfile, baseImag
 			CustomPlatform:  platforms.DefaultString(),
 
 			Cache:     true,
-			CacheRepo: cacheRef,
+			CacheRepo: a.cacheImageRef,
 			CacheOptions: config.CacheOptions{
-				CacheTTL: 14 * (24 * time.Hour),
+				CacheTTL: 14 * (24 * time.Hour), // TODO: should this be configurable?
```
jabrown85, line 92: Probably? Platforms can already dump the cache image itself, but this configuration seems useful to me.

jabrown85 (approved):

PR #834, opened 2022-03-21
Strip optional parameter before checking media type

### LINKED ISSUE #828: SBOM format media types are too restrictive
### Summary

The code to check if a valid media type is set for a buildpack restricts those types to exact matches of the following strings:

* `application/vnd.cyclonedx+json`
* `application/spdx+json`
* `application/vnd.syft+json`

Unfortunately, media types are allowed to have supplemental data included in their media-type as parameters. [RFC 4855 Section 3](https://www.rfc-editor.org/rfc/rfc4855.html#section-3) outlines the format of a media type as the following:

```
type "/" subtype  *(";" parameter)
```

In our cases, we are covering the `type` and `subtype` fields, but not allowing these media types to be declared with optional `parameter` fields.

In the cases of these SBOM formats, I would like to  be able to specify the spec version formats to be output for each type. For example:

```toml
sbom-formats = [ "application/vnd.cyclonedx+json;version=1.3" ]
```

Doing so would allow buildpack authors to be more explicit about what formats they will generate as SBOM artifacts.

---

##### Steps

1. Create a buildpack that includes an `sbom-formats = [ "application/vnd.cyclonedx+json;version=1.3" ]` specification, and generates a `cyclonedx.json` SBOM output
2. Run `pack build` to build an application
3. See that the lifecycle emits the following error: `failed to build: sbom type 'application/vnd.cyclonedx+json' not declared for buildpack: 'test-sbom-buildpack@0.0.1'`

##### Current behavior

The build fails because the lifecycle says that I haven't declared the SBOM format I am generating.

##### Expected

The lifecycle should be a bit less restrictive in how it allows these formats to be specified. Instead of an exact string match, it could maybe parse the media type and ensure that the `type` and `subtype` match allowed values for the specification.

---

##### lifecycle version

`0.13.4`

##### platform version(s)

```
$ pack report
Pack:
  Version:  0.24.0+git-79a40b7.build-3148
  OS/Arch:  darwin/amd64

Default Lifecycle Version:  0.13.3

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8

Config:
```

```
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.3)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 8
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.76-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 19.55GiB
 Name: docker-desktop
 ID: HLUZ:MXT7:CRIC:GFEY:HKU2:34LY:GKYJ:4HTU:ZD32:WYWM:XCOU:PU47
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 No Proxy: hubproxy.docker.internal
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

natalieparellano:
Fixes #828

### DIFF
--- a/buildpack/bomfile.go
+++ b/buildpack/bomfile.go
@@ -65,10 +65,11 @@ func (b *BOMFile) mediaType() string {
 	}
 }
 
-func validateMediaTypes(bp GroupBuildpack, bomfiles []BOMFile, sbomMediaTypes []string) error {
-	contains := func(vs []string, t string) bool {
-		for _, v := range vs {
-			if v == t {
+func validateMediaTypes(bp GroupBuildpack, bomfiles []BOMFile, declaredTypes []string) error {
+	contains := func(declaredTypes []string, foundType string) bool {
+		for _, declaredType := range declaredTypes {
+			parts := strings.Split(declaredType, ";")
+			if foundType == parts[0] {
 				return true
 			}
 		}
@@ -76,13 +77,13 @@ func validateMediaTypes(bp GroupBuildpack, bomfiles []BOMFile, sbomMediaTypes []
 	}
 
 	for _, bomFile := range bomfiles {
-		mediaType := bomFile.mediaType()
-		switch mediaType {
+		foundType := bomFile.mediaType()
+		switch foundType {
 		case mediaTypeUnsupported:
 			return errors.Errorf("unsupported SBOM format: '%s'", bomFile.Path)
 		default:
-			if !contains(sbomMediaTypes, mediaType) {
-				return errors.Errorf("SBOM type '%s' not declared for buildpack: '%s'", mediaType, bp.String())
+			if !contains(declaredTypes, foundType) {
+				return errors.Errorf("SBOM type '%s' not declared for buildpack: '%s'", foundType, bp.String())
 			}
 		}
 	}
--- a/buildpack/bomfile.go
+++ b/buildpack/bomfile.go
@@ -1,6 +1,8 @@
 package buildpack
 
 import (
+	"fmt"
+	"mime"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


buildpack/bomfile.go
```diff
@@ -65,24 +65,25 @@ func (b *BOMFile) mediaType() string {
 	}
 }
 
-func validateMediaTypes(bp GroupBuildpack, bomfiles []BOMFile, sbomMediaTypes []string) error {
-	contains := func(vs []string, t string) bool {
-		for _, v := range vs {
-			if v == t {
+func validateMediaTypes(bp GroupBuildpack, bomfiles []BOMFile, declaredTypes []string) error {
+	contains := func(declaredTypes []string, foundType string) bool {
```
samj1912, line 69: Should this instead use https://pkg.go.dev/mime#ParseMediaType?

natalieparellano: I guess so... for now we don't really need to inspect the media type with much granularity, but it could be useful in the future.

jabrown85 (approved):

buildpack/bomfile.go
jabrown85: I think when we need the ability to understand the media type that library looks good to me, but I don't see a need for it right now. Unless I'm missing something.

natalieparellano: Thinking on this more... maybe we should go ahead and use `mime` - it's in the standard library, and if we ever need to understand the types more granularly it will already be there i.e., we won't forget about this package and try to do our own homegrown thing. Added in 946c7064

samj1912 (approved):

PR #848, opened 2022-04-11
Make runLaunch public

samj1912:
This exposes the launcher's entrypoint to be public. This would allow other projects to easily wrap the launcher to inject their own custom logic. The only public interface it exposes is that there should be a RunLauncher function that takes nil arguments and returns an error.

This would greatly help simplify the logic we have for the custom launcher to deal with https://github.com/buildpacks/rfcs/issues/210

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -1,88 +1,10 @@
 package main
 
 import (
-	"os"
-	"path/filepath"
-	"strings"
-
-	"github.com/BurntSushi/toml"
-	"github.com/heroku/color"
-
-	"github.com/buildpacks/lifecycle/api"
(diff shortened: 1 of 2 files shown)


cmd/launcher/main.go
```diff
@@ -16,10 +16,10 @@ import (
 )
 
 func main() {
-	cmd.Exit(runLaunch())
+	cmd.Exit(RunLaunch())
 }
 
-func runLaunch() error {
+func RunLaunch() error {
```
natalieparellano, line 22: Relaying contents from Slack conversation - making this public won't help, since it's still in `main`. @samj1912 to update.

samj1912:
@natalieparellano I believe we should be all good now

natalieparellano (approved):

PR #854, opened 2022-05-04
Use identity token when provided, instead of ignoring it

natalieparellano:
Fixes 651

We were waiting for an Azure environment to test with, but the fix is so small and straightforward that I'm not sure it's worth it.


auth/env_keychain.go
```diff
@@ -156,6 +156,10 @@ func authConfigToHeader(config *authn.AuthConfig) (string, error) {
 		return fmt.Sprintf("Basic %s", encoded), nil
 	}
 
+	if config.IdentityToken != "" {
+		return fmt.Sprintf("Bearer %s", config.IdentityToken), nil
```
ekcasey, line 160: Looking at the [ggcr implementation](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/bearer.go#L124) I fear it is more complicated than this. We can't simply replace the registry token with an identity token but instead need to do an oauth flow where we use the identity token to get a refresh token.

Luckily I think the solution is to do less rather than to do more. When we store the resolved auth we [convert the AuthConfig to a http header](https://github.com/buildpacks/lifecycle/blob/main/auth/env_keychain.go#L121) and then [convert it back to an AuthConfig](https://github.com/buildpacks/lifecycle/blob/main/auth/env_keychain.go#L60). We could just not do that and return the original auth config - letting ggcr handle the complexity

I am can't remember why do this. It's either
1. historical reasons, we adapted existing code to changes in the ggcr interface to fix complation failures without taking advantage of improvements
2. some of this weirdness exists to support the `CNB_REGISTRY_AUTH` weirdness

In either case we should be able to make this improvement. We should just check what if any of the functionality pack uses and ensure it has what it needs.

natalieparellano: Thanks @ekcasey for the explanation. 

AFAIK pack doesn't care about the format of `CNB_REGISTRY_AUTH` (as long as it is something the lifecycle can understand). It seems like we are using the headers solely to "stringify" the `AuthConfig`, since we never return them when resolving the keychain.

WDYT about fe91721a - it's a bit hacky, but I'm hesitant to do a full-blown refactor of this interface lest we run into compatibility issues between pack <-> lifecycle.

auth/env_keychain.go
```diff
@@ -19,55 +19,72 @@ const EnvRegistryAuth = "CNB_REGISTRY_AUTH"
...
 		envKeychain,
-		InMemoryKeychain(authn.DefaultKeychain, images...),
+		NewResolvedKeychain(authn.DefaultKeychain, images...),
 	), nil
 }
 
-// ResolvedKeychain is an implementation of authn.Keychain that stores credentials in memory.
-type ResolvedKeychain struct {
-	Auths map[string]string
-}
```
natalieparellano, line 36: Moved

auth/env_keychain.go
```diff
@@ -19,55 +19,72 @@ const EnvRegistryAuth = "CNB_REGISTRY_AUTH"
...
 		return nil, errors.Wrap(err, "reading auth env var")
 	}
-	return &ResolvedKeychain{Auths: authHeaders}, nil
+	return &EnvKeychain{AuthHeaders: authHeaders}, nil
 }
 
-// InMemoryKeychain resolves credentials for the given images from the given keychain and returns a new keychain
-// that stores the pre-resolved credentials in memory and returns them on demand. This is useful in cases where the
-// backing credential store may become inaccessible in the the future.
-func InMemoryKeychain(keychain authn.Keychain, images ...string) authn.Keychain {
```
natalieparellano, line 51: Renamed NewResolvedKeychain and moved

auth/env_keychain.go
```diff
@@ -19,55 +19,72 @@ const EnvRegistryAuth = "CNB_REGISTRY_AUTH"
...
 	return authn.Anonymous, nil
 }
 
+var (
+	basicAuthRegExp     = regexp.MustCompile("(?i)^basic (.*)$")
+	bearerAuthRegExp    = regexp.MustCompile("(?i)^bearer (.*)$")
+	identityTokenRegExp = regexp.MustCompile("(?i)^x-identity (.*)$")
+)
+
+func authHeaderToConfig(header string) (*authn.AuthConfig, error) {
```
natalieparellano, line 66: Moved

auth/env_keychain.go
```diff
@@ -76,6 +93,47 @@ func (p *providedAuth) Authorization() (*authn.AuthConfig, error) {
 	return p.config, nil
 }
 
+// NewResolvedKeychain resolves credentials for the given images from the given keychain and returns a new keychain
+// that stores the pre-resolved credentials in memory and returns them on demand. This is useful in cases where the
+// backing credential store may become inaccessible in the the future.
+func NewResolvedKeychain(keychain authn.Keychain, images ...string) authn.Keychain {
```
natalieparellano, line 99: Formerly InMemoryKeychain

auth/env_keychain.go
```diff
@@ -76,6 +93,47 @@ func (p *providedAuth) Authorization() (*authn.AuthConfig, error) {
...
+			continue
+		}
+		if authenticator == authn.Anonymous {
+			continue
+		}
+		authConfig, err := authenticator.Authorization()
+		if err != nil {
+			continue
+		}
+		registryAuths[reference.Context().Registry.Name()] = authConfig
```
natalieparellano, line 119: No more round tripping

auth/env_keychain.go
```diff
@@ -100,47 +158,42 @@ func ReadEnvVar(envVar string) (map[string]string, error) {
...
 		registryAuths[reference.Context().Registry.Name()] = header
 	}
-
 	return registryAuths
 }
 
-// BuildEnvVar creates the contents to use for authentication environment variable.
-//
-// Complementary to `ReadEnvVar`.
-func BuildEnvVar(keychain authn.Keychain, images ...string) (string, error) {
```
natalieparellano, line 134: Moved

auth/env_keychain.go
```diff
@@ -100,47 +158,42 @@ func ReadEnvVar(envVar string) (map[string]string, error) {
 	return authMap, nil
 }
 
-func buildAuthMap(keychain authn.Keychain, images ...string) map[string]string {
-	registryAuths := map[string]string{}
+// BuildEnvVar creates the contents to use for authentication environment variable.
+//
+// Complementary to `ReadEnvVar`.
+func BuildEnvVar(keychain authn.Keychain, images ...string) (string, error) {
```
natalieparellano, line 164: Moved

auth/env_keychain.go
```diff
@@ -156,28 +209,11 @@ func authConfigToHeader(config *authn.AuthConfig) (string, error) {
...
-			Auth: matches[0][1],
-		}, nil
-	}
-
-	if matches := bearerAuthRegExp.FindAllStringSubmatch(header, -1); len(matches) != 0 {
-		return &authn.AuthConfig{
-			RegistryToken: matches[0][1],
-		}, nil
+	if config.IdentityToken != "" {
+		return fmt.Sprintf("X-Identity %s", config.IdentityToken), nil
```
natalieparellano, line 213: This is a bit icky. If we are given an identity token we need some place to put it, but there is no proper header for it.

auth/env_keychain_test.go
```diff
@@ -60,25 +65,127 @@ func testEnvKeychain(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertEq(t, keychain, &auth.EnvKeychain{
+					AuthHeaders: map[string]string{},
 				})
 			})
 		})
 	})
 
-	when("InMemoryKeychain", func() {
+	when("EnvKeychain", func() {
+		when("#Resolve", func() {
```
natalieparellano, line 87: `EnvKeychain` and `ResolvedKeychain` tested separately because they are now distinct.

auth/env_keychain.go
natalieparellano: 276e3dc4 removes the authConfig -> header -> authConfig round tripping when we are given a proper keychain (vs. an "env" one).

auth/env_keychain.go
ekcasey: > This is a bit icky. If we are given an identity token we need some place to put it, but there is no proper header for it.

I had that question reading through, but quickly realized why this hack is necessary given the current interface for passing registry auth details from pack to the lifecycle. I think a comment explaining the weirdness might be appropriate to explain what is happening to the uninitiated and to our future selves.

ekcasey (approved):
Looks great!

As we discussed before it would be great if we could replace the `$CNB_REGISTRY_AUTH` interface with something more sane, especially given the hack required to make this work with identity tokens. But removing the config->header->translation feels like a big step in the right direction.

Maybe we should rename this file `keychain.go` given we are working with more keychains than just the `EnvKeychain` now?

Issue #460, opened 2020-11-10
[RFC 0066] Experimental Features


ekcasey:
[RFC 0066](https://github.com/buildpacks/rfcs/blob/main/text/0066-lifecycle-prelease-version-and-experimental-section.md)
buildpacks/rfcs#115
buildpacks/spec#141

Setup framework to handle experimental modes.

natalieparellano:
Closed in https://github.com/buildpacks/lifecycle/pull/860

Issue #459, opened 2020-11-10
[RFC TBD] Prelease APIs


ekcasey:
API parsing and validation should handle prerelease APIS.

https://github.com/buildpacks/spec/issues/161
https://github.com/buildpacks/rfcs/pull/115

natalieparellano:
Support for `CNB_EXPERIMENTAL_MODE` was added in https://github.com/buildpacks/lifecycle/pull/860, but support for APIs with prerelease versions is yet to be added. We think this will be more valuable as we approach 1.0

PR #860, opened 2022-05-24
Dockerfiles phase 1 (detect changes)

natalieparellano:
Generate changes to follow

~Merging this PR should be blocked on #459 and #460~ (we can update the platform api version in #https://github.com/buildpacks/lifecycle/pull/861)

The diff is quite huge due to the `GroupBuildpack` -> `GroupElement` rename. The most interesting files are probably:
* cmd/lifecycle/detector.go
* detector.go
* detector_test.go
* buildpack/detect.go
* buildpack.detect_test.go

I based this branch off of `cmd-package` (#805) before it was merged, so there are several commits that don't actually relate to the detector, but they don't show up in the diff now that those commits are on `main`. We squash on merge so it shouldn't matter.


analyzer.go
```diff
@@ -12,10 +12,6 @@ import (
 	"github.com/buildpacks/lifecycle/platform"
 )
 
-type Platform interface {
```
natalieparellano, line 15: Moved to builder.go since the analyzer no longer depends on it

api/apis.go
```diff
@@ -8,8 +8,8 @@ import (
 )
 
 var (
-	Platform  = newApisMustParse([]string{"0.3", "0.4", "0.5", "0.6", "0.7", "0.8", "0.9"}, nil)
-	Buildpack = newApisMustParse([]string{"0.2", "0.3", "0.4", "0.5", "0.6", "0.7", "0.8"}, nil)
+	Platform  = newApisMustParse([]string{"0.3", "0.4", "0.5", "0.6", "0.7", "0.8", "0.9", "0.10"}, nil)
+	Buildpack = newApisMustParse([]string{"0.2", "0.3", "0.4", "0.5", "0.6", "0.7", "0.8", "0.9"}, nil)
```
natalieparellano, line 12: TODO: change to pre-release api

builder_test.go
```diff
@@ -34,14 +34,14 @@ func TestBuilder(t *testing.T) {
 }
 
 //go:generate mockgen -package testmock -destination testmock/env.go github.com/buildpacks/lifecycle BuildEnv
-//go:generate mockgen -package testmock -destination testmock/buildpack_store.go github.com/buildpacks/lifecycle BuildpackStore
-//go:generate mockgen -package testmock -destination testmock/buildpack.go github.com/buildpacks/lifecycle Buildpack
+//go:generate mockgen -package testmock -destination testmock/dir_store.go github.com/buildpacks/lifecycle DirStore
+//go:generate mockgen -package testmock -destination testmock/build_module.go github.com/buildpacks/lifecycle/buildpack BuildModule
```
natalieparellano, line 38: "Build module" seemed the best way of describing something that can be a buildpack or an extension. Open to other suggestions.

buildpack/descriptor.go
```diff
@@ -59,31 +112,55 @@ func (bg Group) Append(group ...Group) Group {
 	return bg
 }
 
-// A GroupBuildpack represents a buildpack referenced in a buildpack.toml's [[order.group]].
-// It may be a regular buildpack, or a meta buildpack.
-type GroupBuildpack struct {
-	API      string `toml:"api,omitempty" json:"-"`
+// A GroupElement represents a buildpack referenced in a buildpack.toml's [[order.group]] OR
+// a buildpack or extension in order.toml OR a buildpack or extension in group.toml.
+type GroupElement struct {
```
natalieparellano, line 117: The use of a single struct to represent these 3 subtly different representations of a buildpack and/or extension seems sub-optimal, but refactoring it would introduce a lot more changes to this already large PR.

cmd/lifecycle/analyzer.go
```diff
@@ -108,7 +107,7 @@ func (a *analyzeCmd) Exec() error {
 	factory := lifecycle.NewAnalyzerFactory(
 		a.platform.API(),
 		NewCacheHandler(a.keychain),
-		NewConfigHandler(),
+		lifecycle.NewConfigHandler(&cmd.APIVerifier{}),
```
natalieparellano, line 110: Moved the config handler to the lifecycle package so that it could be unit tested

cmd/lifecycle/main.go
```diff
@@ -232,6 +202,15 @@ func verifyReadWriteAccess(imageRef string, keychain authn.Keychain) error {
 
 // helpers
 
+func appendNotEmpty(slice []string, elems ...string) []string {
```
natalieparellano, line 205: Alphabetized

handlers_test.go
```diff
@@ -0,0 +1,273 @@
...
+	h "github.com/buildpacks/lifecycle/testhelpers"
+	"github.com/buildpacks/lifecycle/testmock"
+)
+
+func TestHandlers(t *testing.T) {
+	spec.Run(t, "Handlers", testHandlers, spec.Report(report.Terminal{}))
+}
+
+func testHandlers(t *testing.T, when spec.G, it spec.S) {
+	when("#ReadGroup", func() {
```
natalieparellano, line 25: I think eventually other phases will have their own config handler, and we won't need the static method.

platform/analyze_inputs_test.go
```diff
@@ -116,35 +132,15 @@ func testAnalyzeInputs(platformAPI string) func(t *testing.T, when spec.G, it sp
 					})
 				})
 			})
-
-			when("layers path is provided", func() {
-				it("uses the group path at the layers path and writes analyzed.toml at the layers path", func() {
```
natalieparellano, line 121: Moved this to outer scope since it is applicable there also.

platform/files_test.go
```diff
@@ -12,141 +12,145 @@ import (
...
+							`"processes":null}`,
+					); s != "" {
+						t.Fatalf("Unexpected JSON:\n%s\n", s)
+					}
+				})
 			})
 		})
 	})
 
-	when("MarshalJSON", func() {
```
natalieparellano, line 79: Nested under metadata.toml

platform/files_test.go
```diff
@@ -12,141 +12,145 @@ import (
...
 )
 
-func TestStackMetadata(t *testing.T) {
-	spec.Run(t, "Test StackMetadata", testMetadata)
+func TestFiles(t *testing.T) {
+	spec.Run(t, "Files", testFiles)
 }
 
-func testMetadata(t *testing.T, when spec.G, it spec.S) {
-	when("BestRunImageMirror", func() {
```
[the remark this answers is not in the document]
natalieparellano: I'm sorry for this distracting diff. I can't remember why I needed to change this file. I think I undid the changes that were relevant to this PR and kept the re-organization changes.

buildpack/detect_test.go
```diff
@@ -21,302 +22,357 @@ import (
 )
 
 func TestDetect(t *testing.T) {
-	spec.Run(t, "Detect", testDetect, spec.Report(report.Terminal{}))
+	bpPath := filepath.Join("testdata", "by-id", "A", "v1", "buildpack.toml")
+	extPath := filepath.Join("testdata", "by-id", "extA", "v1", "extension.toml")
+	for _, path := range []string{bpPath, extPath} {
+		spec.Run(t, fmt.Sprintf("Detect-%s", filepath.Base(path)), testDetect(path), spec.Report(report.Terminal{}))
+	}
```
natalieparellano, line 29: This will run the test suite once for a buildpack and once for an extension. The diff is a bit hard to decipher but may be easier in an IDE.

detector.go
```diff
@@ -95,46 +128,66 @@ func (d *Detector) detectOrder(order buildpack.Order, done, next []buildpack.Gro
...
-		if hasID(done, groupBp.ID) {
+func (d *Detector) detectGroup(group buildpack.Group, done []buildpack.GroupElement, wg *sync.WaitGroup) ([]buildpack.GroupElement, []platform.BuildPlanEntry, error) {
+	for i, groupEl := range group.Group {
+		// Continue if element has already been processed.
+		if hasID(done, groupEl.ID) {
 			continue
 		}
 
-		bp, err := d.Store.Lookup(groupBp.ID, groupBp.Version)
+		// Resolve order if element is the order for extensions.
```
natalieparellano, line 138: Having a single "special" group element to represent the order for extensions seems like the cleanest way to do it. I also experimented with having a special ID that, when provided to the store, would return a descriptor for an extensions order (as suggested [here](https://github.com/buildpacks/lifecycle/pull/812#discussion_r818160288)) but that turned out to be messier IMO.

detector_test.go
```diff
@@ -622,22 +997,21 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
 		it("should fail if the group is empty", func() {
-			_, _, err := resolver.Resolve([]buildpack.GroupBuildpack{}, &sync.Map{})
+			_, _, err := resolver.Resolve([]buildpack.GroupElement{}, &sync.Map{})
 			if err != lifecycle.ErrFailedDetection {
 				t.Fatalf("Unexpected error:\n%s\n", err)
 			}
 
 			if s := cmp.Diff(h.AllLogs(logHandler),
 				"======== Results ========\n"+
-					"Resolving plan... (try #1)\n"+
```
natalieparellano, line 632: We don't need to try to resolve the plan to know that there are no buildpacks in the group.

buildpack/descriptor.go
```diff
@@ -59,31 +112,55 @@ func (bg Group) Append(group ...Group) Group {
...
+	Extension bool `toml:"extension,omitempty" json:"-"`
+
+	// Fields that are in order.toml only
+
+	// Optional specifies that the buildpack or extension is optional. Extensions are always optional.
+	Optional bool `toml:"optional,omitempty" json:"optional,omitempty"`
+
+	// Fields that are never written
+
+	// OrderExt holds the order for extensions during the detect phase. \
```
natalieparellano, line 141:
```suggestion
	// OrderExt holds the order for extensions during the detect phase.
```

builder_test.go
mboldt: Thought about this a little...best I could come up with was `DetectorBuilder` (similar in spirit to the standard library `ReadCloser`, for example).
Otherwise, maybe something like `GeneralizedBuildpack`.
Thought about `Builder`, but that's a different thing in our world.

Not strongly convinced any of that is better than `BuildModule`, but if you like one go for it.

detector_test.go
```diff
@@ -585,30 +781,221 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
...
 					if err, ok := err.(*buildpack.Error); !ok || err.Type != buildpack.ErrTypeFailedDetection {
 						t.Fatalf("Unexpected error:\n%s\n", err)
 					}
 				})
 			})
 		})
+
+		when("there are extensions", func() {
+			it("selects the first passing group", func() {
+				// This test doesn't use gomock.InOrder() because each call to Detect() happens in a go func.
```
mboldt, line 809: Oof, these tests need a lot of gymnastics to set up. Probably doesn't need to be part of this, but maybe worth an issue to revisit to make them more manageable? (Or maybe there is enough inherent complexity that this is the way?)

internal/lifecycle/utils.go
```diff
@@ -0,0 +1,29 @@
+package lifecycle
+
+import (
+	"github.com/buildpacks/lifecycle/buildpack"
+)
+
+func PrependExtensions(orderBp buildpack.Order, orderExt buildpack.Order) buildpack.Order {
```
mboldt, line 7: The `utils` file name threw up a flag for me. As far as GitHub could see, this function is only used in detector.go...maybe just move it there? Or if it is worth splitting into its own file, perhaps a more descriptive name?

mboldt (commented):
Overall looks good! Left some thoughts and questions (about stuff you've probably thought about a lot more than me, so take with a grain of salt ;-) ).

cmd/api_verifier.go
```diff
@@ -0,0 +1,130 @@
...
+
+type APIVerifier struct{}
+
+func (v *APIVerifier) VerifyBuildpackAPIForBuildpack(name, requested string) error {
+	return VerifyBuildpackAPI(Buildpack, name, requested)
+}
+
+func (v *APIVerifier) VerifyBuildpackAPIForExtension(name, requested string) error {
+	return VerifyBuildpackAPI(Extension, name, requested)
+}
```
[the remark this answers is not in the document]
natalieparellano: @mboldt I would love to make this better! Where do you think the logic should be consolidated / moved to?

buildpack/descriptor.go
```diff
@@ -15,14 +40,56 @@ func (b *Descriptor) ConfigFile() *Descriptor {
...
+	case b.IsBuildpack():
+		groupEl.ID = b.Buildpack.ID
+		groupEl.Version = b.Buildpack.Version
+		groupEl.Homepage = b.Buildpack.Homepage
+	case b.IsExtension():
+		groupEl.ID = b.Extension.ID
+		groupEl.Version = b.Extension.Version
+		groupEl.Homepage = b.Extension.Homepage
+		groupEl.Extension = true
+		groupEl.Optional = true
```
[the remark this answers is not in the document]
natalieparellano: You are right, it's a bit weird. TBH at the point where `ToGroupElement()` is used, we don't really care if the element is optional or not - this feeds through to the output of "plan resolution" which is passed from the detect phase to the build phase, and the build phase doesn't have a concept of optional / not optional for buildpacks, it just runs what it is given. We could just not set this for extensions, but since extensions are always optional it doesn't hurt. I think what would ultimately make this better is to pull apart `GroupElement` and have a separate struct for each of the scenarios it currently covers (see https://github.com/buildpacks/lifecycle/pull/860#discussion_r880889983) but that is a pretty big refactor to tackle along with all of these other changes 😐

internal/lifecycle/utils.go
natalieparellano: Makes sense - I'll move it.

builder_test.go
natalieparellano: I think another one that came up was `Detectable` but honestly I like `BuildModule` the best. I don't know why 🤔

cmd/lifecycle/creator.go
```diff
@@ -192,48 +198,55 @@ func (c *createCmd) Exec() error {
 			cmd.DefaultLogger,
 		)
 		if err != nil {
-			return errors.Wrap(err, "initializing analyzer")
+			return cmd.FailErr(err, "initialize analyzer")
```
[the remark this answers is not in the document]
natalieparellano: Yeah, `cmd.FailErr` is going to prepend `failed to ` to the provided "action". I think as it was before, that `initializing analyzer` error was never going to be wrapped with anything, and I was trying to keep the top-level error messages consistent (i.e., `main` should always return an `cmd.FailErr`).

cmd/api_verifier.go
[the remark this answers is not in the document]
natalieparellano: Please see also 3a0218d7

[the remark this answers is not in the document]
natalieparellano: Removed `ToGroupElement()` weirdness in bc0a6460.

buildpack/descriptor.go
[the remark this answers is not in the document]
natalieparellano: Removed `ToGroupElement()` weirdness in bc0a6460.

detector_test.go
natalieparellano: Made an issue: https://github.com/buildpacks/lifecycle/issues/866

mboldt (approved):

natalieparellano:
Thanks so much for your feedback @mboldt !

PR #867, opened 2022-06-09
Refactor: move logger interface to own package

natalieparellano:
We use the logger everywhere and it's getting annoying to copy the interface so many times. It also makes it harder for interfaces that take a logger to be in a separate package from their implementation, as we might create an import cycle.


mboldt (approved):
Looks like a good move to me!

Beyond scope for this, but curious if we've considered just using apex directly? Like the advice from that Ultimate Service workshop we attended a while back.

PR #887, opened 2022-07-26
Add GCP credential provider

natalieparellano:
This was originally added in #540 but removed in #560. The GCP credential provider from
github.com/vdemeester/k8s-pkg-credentialprovider/gcp would loop forever when the metadata
server could not be reached (see https://github.com/kubernetes/kubernetes/issues/86245).

However it appears that the GGCR re-implementation in github.com/google/go-containerregistry/pkg/v1/google
doesn't suffer from this problem, and therefore support for GCP might be safe to add back.

See also:
* https://github.com/google/go-containerregistry/pull/1234#issuecomment-1007450066
* https://github.com/vdemeester/k8s-pkg-credentialprovider/issues/18#issuecomment-1084249349

I ran the experiment described here: https://github.com/buildpacks/lifecycle/pull/560#issuecomment-811464020 on this branch and didn't observe any issues. However it would be really nice to test this in a real environment.


go.mod
```diff
@@ -1,30 +1,31 @@
 module github.com/buildpacks/lifecycle
 
 require (
-	github.com/BurntSushi/toml v1.1.0
+	github.com/BurntSushi/toml v1.2.0
```
natalieparellano, line 4: In an effort to fix clashing dependencies, I updated them all, and this is breaking the build. Will downgrade...

PR #888, opened 2022-08-01
PoC: parse both old and new Buildpack API launch.toml

mboldt:
Not pretty or complete, but proves out the approach.


natalieparellano:
Superseded by https://github.com/buildpacks/lifecycle/pull/889. Thanks for the help @mboldt!

Issue #322, opened 2020-06-24
Process-specific env vars


ekcasey:
Implement https://github.com/buildpacks/rfcs/pull/83

[RFC Text](https://github.com/buildpacks/rfcs/blob/main/text/0042-process-specific-env.md)
Spec issue: https://github.com/buildpacks/spec/issues/90

PR #900, opened 2022-08-30
Allow storing multiple commands instead of a single string

jabrown85:
This is a first step to implementing #322. This PR is updating our internal structs to allow for a slice of commands while keeping the external API and behavior the same. A future PR will implement handling multiple commands depending on the API version and changing the behavior of the launcher and metadat file output changes.

### DIFF
--- a/buildpack/build_test.go
+++ b/buildpack/build_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/apex/log/handlers/memory"
 	"github.com/golang/mock/gomock"
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
@@ -33,6 +34,10 @@ func TestBuild(t *testing.T) {
 	for _, kind := range []string{buildpack.KindBuildpack, buildpack.KindExtension} {
 		spec.Run(t, "unit-build/"+kind, testBuild(kind), spec.Report(report.Terminal{}))
 	}
+}
+
+var processCmpOpts = []cmp.Option{
+	cmpopts.IgnoreFields(launch.Process{}, "RawCommandValue"),
 }
 
 func testBuild(kind string) func(t *testing.T, when spec.G, it spec.S) {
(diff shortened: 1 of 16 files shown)


platform/files.go
```diff
@@ -85,17 +86,88 @@ type BuildMetadata struct {
 	PlatformAPI                 *api.Version             `toml:"-" json:"-"`
 }
 
+// DecodeBuildMetadataTOML reads a metadata.toml file
+func DecodeBuildMetadataTOML(path string, platformAPI *api.Version, buildmd *BuildMetadata) error {
```
natalieparellano, line 90: Could we please add a test for this function? Also this logic seems duplicated across `launch.DecodeLaunchMetadataTOML` and `buildpack.DecodeLaunchTOML` - maybe we could move the common logic to an internal package? 

The `launch` and `buildpack` functions seem not used outside the package, maybe they could be made private (this would avoid the need to test them too as they are already tested implicitly in the other units).

launch/launch.go
```diff
@@ -31,6 +39,64 @@ type Metadata struct {
...
+				return err
+			}
+			launchmd.Processes[i].Command = command
+		}
+	}
+
+	return nil
+}
+
+func (m Metadata) Matches(x interface{}) bool {
```
natalieparellano, line 75: What's the code path where this is used? Is it for tests? A comment here could be helpful for posterity...

natalieparellano (approved):
@jabrown85 apologies for the slow review! I added a couple comments but overall this LGTM

buildpack/build_test.go
```diff
@@ -35,6 +36,10 @@ func TestBuild(t *testing.T) {
 	}
 }
 
+var processCmpOpts = []cmp.Option{
+	cmpopts.IgnoreFields(launch.Process{}, "RawCommandValue"),
+}
```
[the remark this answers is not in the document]
jabrown85: [317dbef](https://github.com/buildpacks/lifecycle/pull/900/commits/317dbefebfbb2e0699edbc685f3275125c1b0aa1)

platform/files.go
jabrown85: [01eddfa](https://github.com/buildpacks/lifecycle/pull/900/commits/01eddfa4d00876c9d23fc61217601446c1596a9e)

jabrown85:
@natalieparellano did you want to take another look or 👍 to merge?

launch/launch.go
```diff
@@ -31,6 +39,74 @@ type Metadata struct {
...
+	if err = DecodeProcesses(launchmd.Processes, platformAPI, md); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func DecodeProcesses(processes []Process, platformAPI *api.Version, md toml.MetaData) error {
+	// decode the process.commands, which will differ based on platform API
+	commandsAreStrings := platformAPI.LessThan("0.11")
```
natalieparellano, line 59:
```suggestion
	commandsAreStrings := platformAPI.LessThan("0.10")
```

launch/decode_test.go
```diff
@@ -0,0 +1,94 @@
...
+			tmpDir     string
+			apiVersion *api.Version
+		)
+
+		it.Before(func() {
+			var err error
+			tmpDir, err = ioutil.TempDir("", "test-decode-metadata-toml")
+			h.AssertNil(t, err)
+
+			apiVersion = api.MustParse("0.11")
```
natalieparellano, line 36:
```suggestion
			apiVersion = api.MustParse("0.10")
```

natalieparellano:
@jabrown85 apologies for the slow reply! I made two suggestions for the platform api boundary (this will go out in 0.10, not 0.11) but otherwise this looks good to me!

jabrown85:
I need to iterate on this some more after the version changes

PR #902, opened 2022-09-06
fixes from testing.

BarDweller:
Fixes for..
- detecting build.Dockerfiles (generate.go readOutputFilesExt) {missing logic to look for build.Dockerfileand add to br.Dockerfiles with appropriate DockerfileKind}
- detecting run.Dockerfiles during scanning for updated runImage (generator.go checkNewRunImage) {missing use of launch.Escape for extID}
- filtering the buildplan based on capabilities extensions provides (generator.go now returns filteredBuildPlan as part of GenerateResult, and detector.go uses that plan when supplied to update plan.toml as part of Exec)

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -82,6 +82,8 @@ func (f *GeneratorFactory) setExtensions(generator *Generator, extensions []buil
 
 type GenerateResult struct {
 	RunImage string
+	Plan     platform.BuildPlan
+	UsePlan  bool
 }
 
 func (g *Generator) Generate() (GenerateResult, error) {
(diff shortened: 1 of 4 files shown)


generator.go
```diff
@@ -82,6 +82,8 @@ func (f *GeneratorFactory) setExtensions(generator *Generator, extensions []buil
 
 type GenerateResult struct {
 	RunImage string
+	Plan     platform.BuildPlan
+	UsePlan  bool
```
natalieparellano, line 86: Is `UsePlan` ever false when there is no error?

Issue #905, opened 2022-09-09
CVEs reported in launcher


abelsromero:
### Summary
We have obtained a report with multiple CVEs
```
	+----------------+----------+------+---------+---------+--------------------------+-----------+------------+------------+----------------------------------------------------+
	|  [1m     CVE       [0m |  [1mSEVERITY [0m |  [1mCVSS [0m |  [1mPACKAGE [0m |  [1mVERSION [0m |  [1m         STATUS          [0m |  [1mPUBLISHED [0m |  [1mDISCOVERED [0m |  [1mGRACE DAYS [0m |  [1m                   DESCRIPTION                     [0m |
	+----------------+----------+------+---------+---------+--------------------------+-----------+------------+------------+----------------------------------------------------+
	|  CVE-2022-30580 [0m |  [91;1mhigh [0m     |  [0m7.80 [0m |  [0mgo [0m      |  [0m1.17.10 [0m |  [0mfixed in 1.18.3, 1.17.11 [0m |  [0m28 days [0m   |  [0m< 1 hour [0m   |  [0m1 [0m          |  [0mCode injection in Cmd.Start in os/exec before [0m      |
	|                |          |      |         |         | 28 days ago              |           |            |            | Go 1.17.11 and Go 1.18.3 allows execution of any   |
	|                |          |      |         |         |                          |           |            |            | binaries in the working directory named either     |
	|                |          |      |         |         |                          |           |            |            | \"..com\...                                        |
	+----------------+----------+------+---------+---------+--------------------------+-----------+------------+------------+----------------------------------------------------+
	|  CVE-2022-32189 [0m |  [91;1mhigh [0m     |  [0m7.50 [0m |  [0mgo [0m      |  [0m1.17.10 [0m |  [0mfixed in 1.18.5, 1.17.13 [0m |  [0m28 days [0m   |  [0m< 1 hour [0m   |  [0m1 [0m          |  [0mA too-short encoded message can cause a panic in [0m   |
	|                |          |      |         |         | 28 days ago              |           |            |            | Float.GobDecode and Rat GobDecode in math/big in   |
	|                |          |      |         |         |                          |           |            |            | Go before 1.17.13 and 1.18.5, potentially allowing |
	|                |          |      |         |         |                          |           |            |            | a...                                               |
	+----------------+----------+------+---------+---------+--------------------------+-----------+------------+------------+----------------------------------------------------+
[...]

PR #910, opened 2022-09-13
Add GitHub Actions workflow to update go when a new version is available

natalieparellano:
Together with release/0.14.2, resolves https://github.com/buildpacks/lifecycle/issues/905.

Lifted from https://github.com/buildpacks/libcnb/blob/main/.github/workflows/pb-update-go.yml.

### DIFF
--- a/.github/workflows/check-latest-release.yml
+++ b/.github/workflows/check-latest-release.yml
@@ -15,6 +15,8 @@ jobs:
             - uses: actions/checkout@v3
             - name: Read go versions
               id: read-go
+              env:
+                GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
                 #!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


jabrown85 (approved):

natalieparellano:
Putting this is draft until it can be tested on a fork

natalieparellano:
@jabrown85 is it possible you could take another look? I think this is closer to what we need. 

Also interesting to note that `actions/setup-go` may lag a patch or two behind (it seems recently updated to 1.18.5 and 1.18.6 is the latest as of 9/6). We may try other go installation methods if we find that we need to be on the absolute latest (the `syft` scanner should help with that).

natalieparellano:
Tested on fork, didn't get to the scanning part but hopefully is okay: https://github.com/natalieparellano/lifecycle/actions/runs/3069587335/jobs/4958413910

jabrown85 (approved):

PR #903, opened 2022-09-08
Use the last part of the path when trying to find matching process.

jabrown85:
The process was comparing the base path minus the extension, which was blocking types with a `.` in them from ever matching.

Fixes #895

Signed-off-by: Jesse Brown <jabrown85@gmail.com>

PR #912, opened 2022-09-13
Fix process type parsing

natalieparellano:
Also bumps imgutil


jabrown85 (approved):

natalieparellano:
Cherry-picked from #903

PR #918, opened 2022-09-29
Dockerfile validation

BarDweller:
Validation by simple dockerfile parser.. 

Validation added as additional step after generation before dockerfiles are copied. Build fails if validation fails.

Currently instruction validation is MUST, to weaken to SHOULD, modify VerifyBuildDockerfile/VerifyRunDockerfile appropriately to just print the err from the verify(Build|Run)Verbs invocation, rather than returning it. 

All dockerfile logic moved to dockerfile.go, uses small visitor to inspect each instruction (skipping comments, and directives, and multiline continues).. assumes 1st arg for instruction is on same line with instruction, will most likely be true for the only args we care about (ARG and FROM). 

Original checkNewRunImage dockerfile logic patched to use new parser to be tolerant of comments / empty lines in dockerfile

### DIFF
--- a/buildpack/dockerfile.go
+++ b/buildpack/dockerfile.go
@@ -1,10 +1,12 @@
 package buildpack
 
 import (
-	"bufio"
+	"bytes"
 	"fmt"
-	"os"
-	"strings"
+	"io/ioutil"
+
+	"github.com/moby/buildkit/frontend/dockerfile/instructions"
+	"github.com/moby/buildkit/frontend/dockerfile/parser"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


buildpack/dockerfile.go
```diff
@@ -23,3 +30,171 @@ type ExtendArg struct {
...
+			continue
+		}
+
+		//does this line start a continuation block?
+		if !inContinueBlock {
+			checkLine = true
+		} else {
+			checkLine = false
+		}
+		if strings.HasSuffix("\\", line) {
```
natalieparellano, line 63: Should we also trim spaces here?

BarDweller: I don't think so. 
The continuation char must be the last char on the line, if there are spaces after it, it stops being a valid continuation char.. 
So here we have to test against `line` rather than `trimmedLine` because we care about the '\' chars original placement.

buildpack/dockerfile.go
```diff
@@ -23,3 +30,171 @@ type ExtendArg struct {
 	Name  string `toml:"name"`
 	Value string `toml:"value"`
 }
+
+type InstructionCallback func(verb string, arg string, lineno int, state map[string]string) (string, error)
+
+func walkDockerInstructions(filename string, handleInstruction InstructionCallback) (string, error) {
```
jabrown85, line 36: Did we look at using an existing library? I know kaniko has a [dockerfile package](https://github.com/GoogleContainerTools/kaniko/blob/239d16cd1c6301c997d511ea41b19cc0ab30621f/pkg/dockerfile/dockerfile.go#L212) that parses Dockerfiles or dockerfile commands?

It might not make sense given the requirements, but I figured I should ask.

BarDweller: I didn't, but mostly under the belief it might be better to avoid dragging in a whole new dep for parsing dockerfiles.. that much said, @natalieparellano has pointed me at the buildkit parser that's in use by kaniko, and I'll have a look tomorrow at if I can't integrate that.. although the usage would be contained within the `walkDockerInstructions` method. 
Although it's a simple syntax, there's still decent benefits to be gained from using a lib with more testing etc =)

BarDweller: I tried some experiments.. but I don't think we're going to use it.. (the kaniko/moby buildkit based dockerfile parsing library)

launcher with my parser - 2.2Mb
launcher with mobi buildkit parser fn welded to my visitor pattern - 5.7Mb

launcher with mobi buildkit parser fn with validation logic using mobi buildkit model - 5.7Mb
launcher using kaniko parser fn with validation logic using mobi buildkit model - 17Mb

(Kaniko parser fn was to try to avoid a bit of boilerplate around invoking the mobi buildkit parser.. definitely not worth the extra 11Mb!!)

The lifecycle build has a test to make sure the launcher comes in under 3Mb, and using the mobi buildkit parser/model is adding 3.5mb alone.

BarDweller: Update: launcher shouldn't have had that package in it =) @natalieparellano found the problem.. looking good so far =)

buildpack/dockerfile.go
BarDweller: Resolved, moved to mobi/buildkit parsing.

buildpack/dockerfile.go
BarDweller: Resolved, moved to mobi / buildkit parsing =)

jabrown85 (approved):

natalieparellano:
Thank you @BarDweller this is great!

PR #920, opened 2022-09-29
Allow storing multiple commands instead of a single string

jabrown85:
Re-introducing 5fa84a4c38e6f49e09be31f0e77df3e0f97e59e6 with fixes to the panic that resulted in the revert.

This is a first step to implementing https://github.com/buildpacks/lifecycle/issues/322. This PR is updating our internal structs to allow for a slice of commands while keeping the external API and behavior the same. A future PR will implement handling multiple commands depending on the API version and changing the behavior of the launcher and metadat file output changes.

I did also run the `pack` acceptance tests pointing at the linux lifecycle and it was 👍 

https://github.com/buildpacks/lifecycle/pull/900


platform/files.go
```diff
@@ -85,17 +86,67 @@ type BuildMetadata struct {
...
 func (md *BuildMetadata) MarshalJSON() ([]byte, error) {
+	type BuildMetadataProcessSerializer struct {
+		Type             string   `toml:"type" json:"type"`
+		Command          string   `toml:"command" json:"command"`
+		Args             []string `toml:"args" json:"args"`
+		Direct           bool     `toml:"direct" json:"direct"`
+		Default          bool     `toml:"default,omitempty" json:"default,omitempty"`
+		BuildpackID      string   `toml:"buildpack-id" json:"buildpackID"`
+		WorkingDirectory string   `toml:"working-dir,omitempty" json:"working-dir,omitempty"`
+	}
```
natalieparellano, line 111: I think if we gave `launch.Process` a `MarshalJSON` method this logic could go away, no?

launch/launch.go
```diff
@@ -31,6 +115,44 @@ type Metadata struct {
 	Buildpacks []Buildpack `toml:"buildpacks" json:"buildpacks"`
 }
 
+// DecodeLaunchMetadataTOML reads a launch.toml file
+func DecodeLaunchMetadataTOML(path string, launchmd *Metadata) error {
```
natalieparellano, line 127: Do we need this custom function anymore actually?

platform/files.go
jabrown85: Done! Good call. This might end up biting us a little bit when we need to serialize differently based on the platform API. We may need to pass that down as part of the `launch.Process` or re-read from env (if that is safe to do) during serialization.

Perhaps all our file structs would be better off if they always had `PlatformAPI string `toml: "-", json: "-"` so we can more easily implement `MarshalJSON` and `MarshalText`. I think the dynamic `buildpack.BOMEntry` in this same area could have also maybe been implemented similar to `launch.Process`

launch/shell.go
```diff
@@ -34,10 +35,14 @@ func (l *Launcher) launchWithShell(self string, proc Process) error {
 	if err != nil {
 		return err
 	}
+	command := ""
+	if len(proc.Command) > 0 {
```
natalieparellano, line 39: For now, this is always false, right?

jabrown85: I believe it is always true. You must have at least a single command (and right now only 1 command).

launch/launch_test.go
```diff
@@ -0,0 +1,104 @@
...
+	h "github.com/buildpacks/lifecycle/testhelpers"
+)
+
+func TestLaunch(t *testing.T) {
+	spec.Run(t, "Launch", testLaunch, spec.Sequential(), spec.Report(report.Terminal{}))
+}
+
+func testLaunch(t *testing.T, when spec.G, it spec.S) {
+	when("Process", func() {
+		when("MarshalText", func() {
```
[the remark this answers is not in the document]
jabrown85: I wondered that as well at first. The exporter tests already check the metadata toml in the image label `io.buildpacks.build.metadata` and the tests pass. Is there somewhere else you think we should double check?

launch/launch.go
```diff
@@ -1,21 +1,113 @@
...
+// UnmarshalTOML implements the toml Unmarshaler interface to allow us more control when reading a Process from toml.
+func (p *Process) UnmarshalTOML(data interface{}) error {
+	var tomlString string
+	switch v := data.(type) {
+	case string:
+		tomlString = v
+	case map[string]interface{}:
+		// turn back into a string
+		bytes, _ := encoding.MarshalTOML(v)
+		tomlString = string(bytes)
```
natalieparellano, line 76: Is this block necessary? I thought `pProcess` allows us to avoid double looping through this function. Or at least, when I comment it out, everything still passes...

launch/launch_test.go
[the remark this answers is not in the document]
natalieparellano: I saw that too... I guess we're in the clear for now but this seems like something that could bite us unexpectedly in the future. Is it possible to switch this back to `MarshalTOML`, and handle the newlines somehow?

launch/shell.go
natalieparellano: Ah yes, my bad. I meant it's always of length 1.

launch/launch.go
jabrown85: Commenting out the `case map[string]interface{}` path leads to an error.
```
=== RUN   TestDecodeMetadataTOML/DecodeMetadataTOML/decoding_metadata.toml/string_commands/decodes_string_commands_into_command_array

    /Users/jesse.brown/dev/buildpacks/lifecycle/launch/decode_test.go:83: Expected nil: could not cast data to string
```

natalieparellano: Hmm... that's weird. I guess I don't understand what's going on here :(

jabrown85: I think this changes depending on the path it took to get here. If you were to serialize _just_ a `launch.Process` back and forth it would always be a string. If you have a collection of these in a parent struct it ends up with a string for the parent, then children get a map of the that section to process.

launch/launch.go
jabrown85: Removed

launch/launch.go
natalieparellano: Ahh that makes sense... thanks for the explanation. Could we add a comment for posterity?

launch/launch_test.go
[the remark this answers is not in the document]
jabrown85: Hmm. I don't see a straight forward way to do this. The encoding change we need is [here](https://github.com/BurntSushi/toml/blob/360c9e349664b57aac916a5c41aac8702b179c09/encode.go#L319-L321). But it doesn't seem to be exposed via any public members.

The `MarshalTOML` interface usage seems to be intended for a single custom "primitive" type with a single value. I think our usage here is a bit odd as we turn a single `launch.Process` into an exploded toml/json.

If we wanted to, I think we could try something like

```
#old
Command          []string       `toml:"-" json:"-"` // ignored
Args             []string `toml:"args" json:"args"`

#new
Command          launch.CommandValue       `toml:"command" json:"command"`
Args             launch.ArgValue `toml:"args" json:"args"`
```

`launch.CommandValue` and `launch.ArgValue` would be the type we implement `MarshalText` on. This feels like a more supported and expected path, but right now we do this translation bit in the marshaller methods. The `launch.ArgValue` wouldn't have access to the `launch.Command` value in order to grab the extra command args.

```
		Command:          p.Command[0],
		Args:             append(p.Command[1:], p.Args[0:]...),
```

I could still give this a shot and see what I can come up with if you think it is worth the effort. Let me know!

platform/files.go
```diff
@@ -85,10 +86,29 @@ type BuildMetadata struct {
...
+
 func (md *BuildMetadata) MarshalJSON() ([]byte, error) {
 	if md.PlatformAPI == nil || md.PlatformAPI.LessThan("0.9") {
-		return json.Marshal(*md)
+		type BuildMetadataSerializer BuildMetadata // prevent infinite recursion when serializing
+		return json.Marshal(&struct {
+			*BuildMetadataSerializer
+		}{
+			BuildMetadataSerializer: (*BuildMetadataSerializer)(md),
+		})
```
natalieparellano, line 109: Did we intend to change this logic?

launch/launch.go
```diff
@@ -1,21 +1,114 @@
...
 	"path/filepath"
 	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/buildpacks/lifecycle/internal/encoding"
+
+	"github.com/BurntSushi/toml"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
```
natalieparellano, line 16:
```suggestion
	"github.com/BurntSushi/toml"
	"github.com/google/go-cmp/cmp"
	"github.com/google/go-cmp/cmp/cmpopts"
	"github.com/pkg/errors"
	
	"github.com/buildpacks/lifecycle/internal/encoding"
```

platform/files.go
jabrown85: Nope! Thanks.

launch/launch_test.go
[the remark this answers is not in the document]
natalieparellano: I think for now because we have a `MarshalJSON` method on `Process` we don't have to worry that `MarshalText` will ever be called when serializing JSON because `MarshalJSON` has precedence :) So we can call this done 🎉

natalieparellano (approved):
I think we need to satisfy the linter but otherwise this looks good to me. Thanks for slogging through this thorny issue @jabrown85 :)

PR #926, opened 2022-10-13
Remove <image> as an argument to the extender

natalieparellano:
Favors passing the reference via analyzed.toml

To reflect this commit in the platform spec PR: https://github.com/buildpacks/spec/pull/320/commits/f0a3ef244ab34f467f3530117f2aec6cf18b7bf1

### DIFF
--- a/extender.go
+++ b/extender.go
@@ -83,7 +83,9 @@ func (f *ExtenderFactory) setImageRef(extender *Extender, path string) error {
 	if err != nil {
 		return err
 	}
-	extender.ImageRef = analyzedMD.BuildImage.Reference
+	if analyzedMD.BuildImage != nil {
+		extender.ImageRef = analyzedMD.BuildImage.Reference
+	}
 	return nil
 }
 
--- a/extender.go
+++ b/extender.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/BurntSushi/toml"
+	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/uuid"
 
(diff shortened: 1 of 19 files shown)


extender.go
```diff
@@ -66,12 +69,24 @@ func (f *ExtenderFactory) NewExtender(
...
 	}
 	return extender, nil
 }
 
+func (f *ExtenderFactory) setImageRef(extender *Extender, path string) error {
+	analyzedMD, err := f.configHandler.ReadAnalyzed(path)
+	if err != nil {
+		return err
+	}
+	extender.ImageRef = analyzedMD.BuildImage.Reference
```
jabrown85, line 86: Do we need guards around any part of `BuildImage.Reference` being `nil`/`empty`?

platform/files.go
```diff
@@ -21,6 +21,7 @@ type AnalyzedMetadata struct {
 	PreviousImage *ImageIdentifier `toml:"image"`
 	Metadata      LayersMetadata   `toml:"metadata"`
 	RunImage      *ImageIdentifier `toml:"run-image,omitempty"`
+	BuildImage    *ImageIdentifier `toml:"build-image,omitempty"`
```
jabrown85, line 24: In the spec PR for this phase, I don't see the analyzed.toml data format updated to include `build-image`

extender.go
natalieparellano: Thank you @jabrown85! Added in 35c052e2

platform/files.go
natalieparellano: Updated spec PR!

jabrown85 (approved):

PR #933, opened 2022-10-19
Bump imgutil and docker

natalieparellano:


buildpack/dockerfile.go
```diff
@@ -17,7 +17,7 @@ const (
 	DockerfileKindRun   = "run"
 )
 
-var permittedCommandsBuild = []string{"from", "add", "arg", "copy", "env", "label", "run", "shell", "user", "workdir"}
+var permittedCommandsBuild = []string{"FROM", "ADD", "ARG", "COPY", "ENV", "LABEL", "RUN", "SHELL", "USER", "WORKDIR"}
```
natalieparellano, line 20: Docker uses a newer version of buildkit - I tried upgrading, but it breaks kaniko. However this change (which was needed when buildkit was upgraded) seems nice to have anyway.

go.mod
```diff
@@ -124,8 +124,8 @@ go 1.18
 
 replace github.com/containerd/containerd => github.com/containerd/containerd v1.6.6
 
-// These match the docker/docker's dependencies configured in:
-// https://github.com/moby/moby/blob/v20.10.18/vendor.conf
+// This matches GoogleContainerTools/kaniko dependencies configured in:
+// https://github.com/GoogleContainerTools/kaniko/blob/f9aaa9fca7bf4077778ed527c1a8a6e09e60c53c/go.mod
```
natalieparellano, line 128: We are more locked to kaniko (vs docker)

jabrown85 (approved):

PR #936, opened 2022-10-21
docs: fix coverage badge primarily

desenna:
Signed-off-by: Michael de Senna <desenna@gmail.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,12 +29,13 @@ jobs:
         env:
           TEST_COVERAGE: 1
         run: make test
-      - name: Upload Coverage
-        uses: codecov/codecov-action@v1.0.5
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
         with:
           file: ./out/tests/coverage-unit.txt
           flags: unit,os_linux
           fail_ci_if_error: true
+          verbose: true
   test-linux-arm64:
     runs-on: linux-arm64
     steps:
@@ -88,7 +89,7 @@ jobs:
         run: |
           make test
       - name: Prepare Codecov
-        uses: crazy-max/ghaction-chocolatey@v1
+        uses: crazy-max/ghaction-chocolatey@v2
         with:
           args: install codecov -y
       - name: Run Codecov
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://github.com/buildpacks/lifecycle/workflows/build/badge.svg)](https://github.com/buildpacks/lifecycle/actions)
 [![GoDoc](https://godoc.org/github.com/buildpacks/lifecycle?status.svg)](https://godoc.org/github.com/buildpacks/lifecycle)
-[![codecov](https://codecov.io/gh/buildpacks/pack/branch/main/graph/badge.svg)](https://codecov.io/gh/buildpacks/pack)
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build Status](https://github.com/buildpacks/lifecycle/workflows/build/badge.svg)](https://github.com/buildpacks/lifecycle/actions)
 [![GoDoc](https://godoc.org/github.com/buildpacks/lifecycle?status.svg)](https://godoc.org/github.com/buildpacks/lifecycle)
-[![codecov](https://codecov.io/gh/buildpacks/pack/branch/main/graph/badge.svg)](https://codecov.io/gh/buildpacks/pack)
+[![codecov](https://codecov.io/gh/buildpacks/lifecycle/branch/main/graph/badge.svg)](https://codecov.io/gh/buildpacks/lifecycle/tree/main)
```
desenna, line 5: This is to correct the badge. I think though that if someone who has admin permissions to codecov.io could set the default branch to main, as in buildpacks/pack, we could probably drop tree/main.

.github/workflows/build.yml
```diff
@@ -87,8 +88,15 @@ jobs:
...
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./out/tests/coverage-unit.txt
+          flags: unit,os_windows
+          fail_ci_if_error: true
+          verbose: true
       - name: Prepare Codecov
-        uses: crazy-max/ghaction-chocolatey@v1
+        uses: crazy-max/ghaction-chocolatey@v2
```
desenna, line 99: bump to v2 (mirroring what's in buildpacks/pack)

.github/workflows/build.yml
```diff
@@ -29,12 +29,13 @@ jobs:
         env:
           TEST_COVERAGE: 1
         run: make test
-      - name: Upload Coverage
-        uses: codecov/codecov-action@v1.0.5
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
```
desenna, line 33: bump to v3 (mirroring what's in buildpacks/pack)

.github/workflows/build.yml
```diff
@@ -29,12 +29,13 @@ jobs:
         env:
           TEST_COVERAGE: 1
         run: make test
-      - name: Upload Coverage
-        uses: codecov/codecov-action@v1.0.5
+      - name: Upload coverage to Codecov
```
desenna, line 32: This language is consistent with that of buildpacks/pack

.github/workflows/build.yml
desenna: TBH I found the split here between linux and windows a bit confusing or I guess I don't really understand it. I'm wondering if only windows installs codecov, both run tests (maybe the linux one is ignored bc not installed?), and only the linux section uploads the coverage-unit.txt to hosted codecov..?

Or is perhaps something implicit happening here that I'm missing?

natalieparellano: I'm a bit fuzzy on this myself TBH... I do see output from Codecov in the linux workflows, but I'm not sure why it doesn't need to be installed explicitly.

natalieparellano (approved):

natalieparellano:
Thanks for the PR @desenna!!!

PR #944, opened 2022-10-27
Generate lifecycle sbom

omocquais-p:
#958

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,36 @@ $(BUILD_DIR)/darwin-amd64/lifecycle/launcher:
 	$(GOENV) $(GOBUILD) -o $(OUT_DIR)/launcher -a ./cmd/launcher
 	test $$(du -m $(OUT_DIR)/launcher|cut -f 1) -le 4
 
+generate-sbom: run-syft-windows run-syft-linux-amd64 run-syft-linux-arm64
+
+run-syft-windows: install-syft
+run-syft-windows: export GOOS:=windows
+run-syft-windows: export GOARCH:=amd64
+run-syft-windows:
+	@echo "> Running syft..."
+	syft $(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.exe -o json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.syft.json -o spdx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.spdx.json -o cyclonedx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.cdx.json
+	syft $(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.exe -o json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.syft.json -o spdx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.spdx.json -o cyclonedx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.cdx.json
+
+run-syft-linux-amd64: install-syft
+run-syft-linux-amd64: export GOOS:=linux
+run-syft-linux-amd64: export GOARCH:=amd64
+run-syft-linux-amd64:
+	@echo "> Running syft..."
+	syft $(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle -o json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.syft.json -o spdx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.spdx.json -o cyclonedx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.cdx.json
+	syft $(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher -o json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.syft.json -o spdx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.spdx.json -o cyclonedx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.cdx.json
+
+run-syft-linux-arm64: install-syft
+run-syft-linux-arm64: export GOOS:=linux
+run-syft-linux-arm64: export GOARCH:=arm64
+run-syft-linux-arm64:
+	@echo "> Running syft..."
+	syft $(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle -o json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.syft.json -o spdx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.spdx.json -o cyclonedx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/lifecycle.sbom.cdx.json
+	syft $(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher -o json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.syft.json -o spdx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.spdx.json -o cyclonedx-json=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle/launcher.sbom.cdx.json
+
+install-syft:
+	@echo "> Installing syft..."
+	curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
+
 install-goimports:
 	@echo "> Installing goimports..."
 	$(GOCMD) install golang.org/x/tools/cmd/goimports@v0.1.2
@@ -270,7 +300,7 @@ clean:
 	@echo "> Cleaning workspace..."
 	rm -rf $(BUILD_DIR)
 
-package: package-linux-amd64 package-linux-arm64 package-windows-amd64
+package:  generate-sbom package-linux-amd64 package-linux-arm64 package-windows-amd64
 
 package-linux-amd64: GOOS:=linux
 package-linux-amd64: GOARCH:=amd64
--- a/acceptance/phase_test.go
+++ b/acceptance/phase_test.go
@@ -10,12 +10,15 @@ import (
 	"os/exec"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -213,6 +214,28 @@ $(BUILD_DIR)/darwin-amd64/lifecycle/launcher:
...
+copy-sboms: export GOARCH:=amd64
+copy-sboms: OUT_DIR:=$(BUILD_DIR)/$(GOOS)-$(GOARCH)/lifecycle
+copy-sboms:
+	@echo "> Copying SBOMs"
+	cp $(OUT_DIR)/*.json $(PWD)/out/linux-arm64/lifecycle
+	cp $(OUT_DIR)/*.json $(PWD)/out/windows-amd64/lifecycle
+
+build-sbom-generator-image:
+	@echo "> Building docker image for sbom generator"
+	DOCKER_BUILDKIT=1 docker build --progress=plain --file tools/sboms/Dockerfile --tag $(SBOM_GENERATOR_IMAGE) --cache-from=$(SBOM_GENERATOR_IMAGE) .
```
natalieparellano, line 229: We might inspect the current OS and only run this in a container if we're on Darwin

natalieparellano, line 229:
```suggestion
	@echo "> Building docker image for sbom generator"
	DOCKER_BUILDKIT=1 docker build --progress=plain --file tools/sboms/Dockerfile --tag $(SBOM_GENERATOR_IMAGE) .
```
I could be mistaken but I don't think this is needed if we're also building to `SBOM_GENERATOR_IMAGE`

exporter.go
```diff
@@ -157,6 +166,53 @@ func (e *Exporter) Export(opts ExportOptions) (platform.ExportReport, error) {
...
+	for _, component := range components {
+		if component == "lifecycle" {
+			targetDir = targetBuildDir
+		} else if component == "launcher" {
+			targetDir = targetLaunchDir
+		}
+		for _, extension := range extensions {
+			sbomFilename := component + extension
+
+			srcSbomAbsPath := "/cnb/lifecycle/" + sbomFilename
```
natalieparellano, line 194: We should use `filepath.Join` and on Windows, we'll need to add the root directory (we do something similar here: https://github.com/buildpacks/lifecycle/blob/5a4df34080c097528eca957576ae19e477942398/platform/defaults_windows.go#L5)

exporter_test.go
```diff
@@ -1074,6 +1074,21 @@ version = "4.5.6"
 				h.AssertNil(t, err)
 				h.AssertContains(t, fakeAppImage.SavedNames(), append(opts.AdditionalNames, fakeAppImage.Name())...)
 			})
+
+			it("should display that the SBOM is missing", func() {
```
natalieparellano, line 1078: Do we also need a test that when the SBOM is not missing, it is copied?

exporter.go
```diff
@@ -157,6 +166,53 @@ func (e *Exporter) Export(opts ExportOptions) (platform.ExportReport, error) {
...
+
+			_, err := os.Stat(srcSbomAbsPath)
+			if !errors.Is(err, os.ErrNotExist) {
+				destSbomFilename := strings.Replace(sbomFilename, component+".", "", 1)
```
natalieparellano, line 198: Would "sbom" + extension be simpler?

tools/sboms/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY tools/sboms/main.go ./
+COPY tools/sboms/generator/sbom.go ./generator/sbom.go
+
+RUN CGO_ENABLED=0 go build -o /sbom-generator
+
+FROM alpine:latest
+WORKDIR /app/
+COPY --from=builder /sbom-generator ./
+
+ENTRYPOINT ["/app/sbom-generator", "-path", "/lifecycle"]
```
natalieparellano, line 20: nit: missing newline

tools/sboms/generator/sbom.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+const (
+	CycloneDXFormat = "application/vnd.cyclonedx+json"
+	SPDXFormat      = "application/spdx+json"
+	SyftFormat      = "application/vnd.syft+json"
+)
+
+func GenerateSBOM(path string) []error {
+	fmt.Printf("\nGenerate SBOM from this path: %v", path)
```
natalieparellano, line 23:
```suggestion
	log.Printf("Generate SBOM from this path: %v", path)
```

tools/sboms/generator/sbom_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			h.AssertStringContains(t, string(spdxContent), "github.com/pkg/errors")
+			h.AssertStringContains(t, string(spdxContent), "spdxVersion")
+
+			h.AssertStringContains(t, string(syftContent), "github.com/pkg/errors")
+			h.AssertStringContains(t, string(syftContent), "https://raw.githubusercontent.com/anchore/syft/")
+		})
+	})
+
+	when("when given a directory as parameter", func() {
```
natalieparellano, line 72:
```suggestion
	when("given a directory as parameter", func() {
```

tools/sboms/generator/sbom_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+		it("returns an array with one error", func() {
+			errors := GenerateSBOM(tmpDir)
+
+			h.AssertEq(t, len(errors), 1)
+			h.AssertNotNil(t, errors[0])
+		})
+	})
+
+	when("when given an absolute path to a binary file with an executable extension .exe", func() {
```
natalieparellano, line 91:
```suggestion
	when("given an absolute path to a binary file with an executable extension .exe", func() {
```

tools/sboms/generator/sbom_test.go
```diff
@@ -0,0 +1,122 @@
...
+		it("returns the filename without the extension", func() {
+			h.AssertEq(t, GenerateFilename("/windows-amd64/lifecycle/lifecycle.exe"), "lifecycle")
+		})
+	})
+
+	when("when given an absolute path to a binary file without an extension", func() {
```
natalieparellano, line 97:
```suggestion
	when("given an absolute path to a binary file without an extension", func() {
```

tools/sboms/main.go
```diff
@@ -0,0 +1,20 @@
...
+	path string
+)
+
+func main() {
+	flag.StringVar(&path, "path", "", "path to a binary file")
+	flag.Parse()
+
+	generator.GenerateSBOM(path)
+
+	fmt.Printf("\nSBOMs have been generated\n")
```
natalieparellano, line 19:
```suggestion
	log.Printf("SBOMs have been generated")
```

tools/sboms/generator/sbom_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			tmpDir, err := os.MkdirTemp("", "sbom-test")
+			h.AssertNil(t, err)
+
+			tmpBinaryDir, err := os.MkdirTemp("", "sbom-test-binary")
+			h.AssertNil(t, err)
+
+			copyFile(t, tmpDir, "go.mod.template", "go.mod")
+			copyFile(t, tmpDir, "go.sum.template", "go.sum")
+			copyFile(t, tmpDir, "simple-app.go.template", "simple-app.go")
```
natalieparellano, line 38: Interesting use of `.template` ...I haven't seen it before, do we need it? It's common to call the directory with fixtures `testdata`

natalieparellano (commented):
Amazing work @omocquais-p! It's really exciting to see this come to fruition. I added a couple comments, but overall this is looking to be in good shape.

tools/sboms/generator/sbom_test.go
omocquais-p: ok. I have rewritten this code with the use of a new _testdata_ directory.

exporter_test.go
omocquais-p: In this test (exporter_test.go), the SBOM is missing because the SBOMs have not yet been generated. 

The alternative test case has been implemented in the acceptance test ( acceptance/exporter_test.go ) where fake SBOMs have been copied ( copyFakeSboms(t) ) in the out directory.
We have to copy the fake SBOMs before this call: _exportTest.Start(t, updateAnalyzedTOMLFixturesWithRegRepoName)_ because there is a call to _h.MakeAndCopyLifecycle()_ where all the files available in the _out_ directory are copied in new empty destination directory (_containerBinaryDir_).
Then the docker image is built (_h.DockerBuild_) from the _out_ directory which contains _binary and SBOM_s. The SBOMs are successfully packaged in the /cnb/lifecycle/ folder and we check the output "Copying launcher SBOM (sbom.syft.json)"


So with these two tests, we are covering the both test cases (SBOM is _available_ and SBOM is _missing_).

exporter.go
omocquais-p: destSbomFilename := strings.Replace(sbomFilename, component+".", "", 1)

==> no, because we want to change the name of the file: lifecycle.sbom.cdx.json ==> sbom.cdx.json and the value of the _extension_ variable is ".sbom.cdx.json"

Makefile
omocquais-p: ok. I'll remove the parameter.

tools/sboms/Dockerfile
omocquais-p: ok. I'll add the missing newline

tools/sboms/generator/sbom.go
omocquais-p: ok. I'll change the code.

exporter.go
omocquais-p: ok. I have pushed a new commit to fix this problem:
https://github.com/omocquais-p/lifecycle/commit/f230fdeffa5a40b7e0b0bc1784af7562248f33db

acceptance/phase_test.go
```diff
@@ -363,3 +367,27 @@ func withoutDaemonFixtures(phaseTest *PhaseTest) {
...
+	if goos == "darwin" {
+		goos = "linux"
+	}
+	buildLifecycleDir, err := filepath.Abs(filepath.Join("..", "out", fmt.Sprintf("%s-%s", goos, runtime.GOARCH), "lifecycle"))
+	if err != nil {
+		t.Log("Fail to locate lifecycle directory")
+	}
+
+	extensions := [3]string{".sbom.cdx.json", ".sbom.spdx.json", ".sbom.syft.json"}
+	components := [2]string{"lifecycle", "launcher"}
```
jjbustamante, line 384: I can see this is used in several places and also seems to be constant. Is it a valid suggestion to externalize it?

jjbustamante (changes requested):

acceptance/phase_test.go
jjbustamante: Maybe something like: 

```=golang
func SBOMExtensions() []string {
   return [3]string{".sbom.cdx.json", ".sbom.spdx.json", ".sbom.syft.json"}
}

func SBOMComponents() []string {
   return [2]string{"lifecycle", "launcher"}
}
```

exporter.go
```diff
@@ -157,6 +167,53 @@ func (e *Exporter) Export(opts ExportOptions) (platform.ExportReport, error) {
...
+
+				err := fsutil.Copy(srcSbomAbsPath, destSbomAbsPath)
+
+				if err != nil {
+					return errors.Wrapf(err, "Fail to copy "+sbomFilename)
+				}
+
+				e.Logger.Infof("Copying " + component + " SBOM (" + destSbomFilename + ")")
+			} else {
+				e.Logger.Warn(sbomFilename + " is missing")
```
natalieparellano, line 210: Discussed offline - we can probably do the check earlier and not create the directory if there are no SBOM files.

jjbustamante (approved):

natalieparellano (approved):
@omocquais-p great work, and thank you for the contribution! 

Note: IMO Codecov is being too picky here, as the constants are already implicitly tested elsewhere. I don't think we need to add any more tests.

PR #946, opened 2022-11-03
Fix regression in metadata.toml rendering

jabrown85:
Moved the dynamic process serialization to a new field type and off of process. The TOML that was previously generated was not expected and the TOML library we use doesn't seem to have a way to render the table the way we expect. Making a new type for the single field `command` allows us to render the TOML the way we want.

Fixes: buildpacks/lifecycle#945
Signed-off-by: Jesse Brown <jabrown85@gmail.com>

PR #950, opened 2022-11-08
Fix special character escaping when marshaling launch process

natalieparellano:
Following up on https://github.com/buildpacks/lifecycle/pull/946 and https://github.com/buildpacks/lifecycle/pull/947, this handles special characters and adds an `UnmarshalJSON` method to `Command` for library consumers.

### DIFF
--- a/platform/files.go
+++ b/platform/files.go
@@ -99,8 +99,8 @@ func DecodeBuildMetadataTOML(path string, platformAPI *api.Version, buildmd *Bui
 	// this will allow us to re-encode the metadata.toml file with
 	// the current platform API
 	buildmd.PlatformAPI = platformAPI
-	for i := range buildmd.Processes {
-		buildmd.Processes[i] = buildmd.Processes[i].WithPlatformAPI(platformAPI)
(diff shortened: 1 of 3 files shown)


platform/files.go
```diff
@@ -99,8 +99,8 @@ func DecodeBuildMetadataTOML(path string, platformAPI *api.Version, buildmd *Bui
 	// this will allow us to re-encode the metadata.toml file with
 	// the current platform API
 	buildmd.PlatformAPI = platformAPI
-	for i := range buildmd.Processes {
-		buildmd.Processes[i] = buildmd.Processes[i].WithPlatformAPI(platformAPI)
+	for i, process := range buildmd.Processes {
+		buildmd.Processes[i] = process.WithPlatformAPI(platformAPI)
```
natalieparellano, line 103: I can't remember why I did this to be honest

jabrown85 (approved):
This looks great @natalieparellano !

platform/files.go
jabrown85: Maybe a linter? It reads better

PR #899, opened 2022-08-29
Allow setting environment variables using the build image

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

PR #962, opened 2022-11-22
(updated with latest main) Allow setting environment variables using the build image

natalieparellano:
Supersedes https://github.com/buildpacks/lifecycle/pull/899


samj1912 (approved):
Thanks for this!

PR #968, opened 2022-11-30
Fix pack acceptance with 0.28.0-rc1

natalieparellano:
We need to configure git to use line feeds on Windows, otherwise
  the test fixtures and hence the layer diff IDs will be different from the test expectations.


natalieparellano:
Tested on fork: https://github.com/natalieparellano/lifecycle/actions/runs/3587585926/jobs/6038186159

samj1912 (approved):

PR #972, opened 2022-12-07
Fix: rebase does not require stack.toml

natalieparellano:
cc @jabrown85


natalieparellano:
Confirmed (in Slack) this fixes the issue

PR #973, opened 2022-12-07
Updates go to version 1.19

natalieparellano:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up go
         uses: actions/setup-go@v3
         with:
-          go-version: '1.18'
+          go-version: '1.19'
       - name: Install jq
         run: |
           mkdir -p deps/bin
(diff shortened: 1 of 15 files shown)


.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Set up go
         uses: actions/setup-go@v3
         with:
-          go-version: '1.18'
+          go-version: '1.19'
```
jabrown85, line 21: Do we have a reason why we don't want to migrate to `go-version-file: 'go.mod'`?

jabrown85 (approved):
Looks good, with a question

.github/workflows/build.yml
natalieparellano: No! Let's do it!

natalieparellano: Updated in 7cf318e3

jabrown85 (approved):

PR #975, opened 2022-12-09
Improve detector log output

### LINKED ISSUE #967: Include full lifecycle log output on failed builds to improve debugging
### Description

This feature request is in reference to a problem kpack users encounter when trying to debug problems using lifecycle log output.  In the event that no buildpacks detect as part of a kpack build, the full lifecycle logs are not printed, which makes it difficult to troubleshoot the root of the problem.    

### Proposed solution
Print full lifecycle logs. In order to avoid overwhelming the user, perhaps this verbose log output could only be surfaced on failed detection for the detect phase.

The kpack project had a couple of ideas such as attempting detection again or on a build failure try again with verbose logging. Due to how kpack builds are usually not user triggered (rather automatically triggered to match the user's desired state), it is difficult/clunky to mimic the `pack` CLI's verbose flag command. It would instead be preferable to address this at the lifecycle level.

natalieparellano:
When detect fails, print output as info level

Resolves #967


log/multi_logger.go
```diff
@@ -0,0 +1,88 @@
...
+	"io"
+
+	"github.com/apex/log"
+)
+
+var _ Logger = &MultiLogger{}
+
+type MultiLogger struct {
+	defaultLogger *DefaultLogger
+	infoLogger    *DefaultLogger
```
natalieparellano, line 13: Maybe there are better names than "default" and "info". I thought of "verbose" instead of "info", but wasn't sure.

acceptance/detector_test.go
```diff
@@ -106,7 +106,7 @@ func testDetector(t *testing.T, when spec.G, it spec.S) {
 				"docker",
 				"run",
 				"--rm",
-				"--env", "CNB_ORDER_PATH=/cnb/orders/empty_order.toml",
```
natalieparellano, line 109: Empty order produces no output, so it's not a good test

natalieparellano:
@jabrown85 it would be awesome to get your input here (as the requestor of #283) and also because I'm a bit unsure of some of the choices I made. The implementation ended up being a tad more complicated than I initially expected. Feedback is very welcome!

detector.go
```diff
@@ -56,27 +58,30 @@ type Detector struct {
...
 		Executor:       &buildpack.DefaultDetectExecutor{},
 		Logger:         logger,
 		PlatformDir:    platformDir,
-		Resolver:       &DefaultResolver{Logger: logger},
 		Runs:           &sync.Map{},
 	}
 	if err := f.setOrder(detector, orderPath, logger); err != nil {
 		return nil, err
 	}
+	if err := f.setResolverAndMultiLogger(detector, logger); err != nil {
```
jabrown85, line 82: What if we used the in-memory logger and then re-log using the passed in logger?

```golang
// setup in-memory logger for detect so we can dynamically show logs after execution
memLogger := memory.New()

// use that in our Detector
Resolver: &DefaultResolver{Logger: memLogger}
```

Then later on we always copy the entries from the memory logger to the default logger, but can skip log lines that don't match the level.

```golang
	for _, e := range memLogger.Entries {
           if detectErr != null {
             defaultLogger.HandleLog(e)
           } else if e.Level >= Info { efaultLogger.HandleLog(e) }
        }
        return group, plan, detectErr
```

natalieparellano: Hmm... but would that not result in some log messages being printed out of order?

jabrown85: Would it? Is `Entries` order not guaranteed?

natalieparellano: Ah I think I get it now... let me make a few changes

natalieparellano: @jabrown85 could you take another look? It's much simpler now... thanks :)

natalieparellano:
Oof, the linux ARM runner needs to be re-created

detector.go
```diff
@@ -58,30 +58,30 @@ type Detector struct {
...
 	detector := &Detector{
 		AppDir:         appDir,
 		BuildConfigDir: buildConfigDir,
 		DirStore:       f.dirStore,
 		Executor:       &buildpack.DefaultDetectExecutor{},
 		Logger:         logger,
 		PlatformDir:    platformDir,
+		Resolver:       NewDefaultDetectResolver(&apexlog.Logger{Handler: memHandler}),
 		Runs:           &sync.Map{},
+		MemHandler:     memHandler,
```
jabrown85, line 80: I think we could straight up hide `MemHandler` and let it get built and stored in `Detector`. It isn't something a caller needs to send in. More of an implementation detail IMO.

natalieparellano: You're right... updated in 04799076

jabrown85 (approved):
CI issue appears transient - I like the changes!

PR #981, opened 2023-01-04
renames CNB_ANALYZE_SKIP_LAYERS to CNB_SKIP_LAYERS 

joe-kimmel-vmw:
so name matches spec and resolves FIXME


jabrown85:
Looks like you need to fix DCO (See failed GitHub check)

natalieparellano:
Related Slack conversation: https://cloud-native.slack.com/archives/C033DV9EBDF/p1668721420968799

Though there's low probability anyone is using this (or else we would have heard about it by now) it may be safer to guard this on a platform API version bump. I put this on the agenda for tomorrow's WG (1/5/23) to discuss.

natalieparellano:
Per 1/5/23 we decided that just renaming the variable is fine as the current usage isn't documented anywhere.

natalieparellano (approved):

PR #984, opened 2023-01-06
Launcher sboms copy only sbom extensions

joe-kimmel-vmw:


exporter_test.go
```diff
@@ -1507,11 +1507,11 @@ version = "4.5.6"
 					opts.LauncherConfig.SBOMDir = filepath.Join(opts.LayersDir, "some-launcher-sbom-dir")
 				})
 
-				it("copies everything in the launcher SBOM directory", func() {
+				it("copies sboms from the launcher SBOM directory", func() {
 					_, err := exporter.Export(opts)
 					h.AssertNil(t, err)
 
-					h.AssertPathExists(t, filepath.Join(opts.LayersDir, "sbom", "launch", "buildpacksio_lifecycle", "launcher", "some-sbom-file"))
+					h.AssertPathExists(t, filepath.Join(opts.LayersDir, "sbom", "launch", "buildpacksio_lifecycle", "launcher", "some-sbom-file.sbom.spdx.json"))
```
natalieparellano, line 1514: Could we add some other (unwanted) file to the fixture dir and assert that the path does not exist?

natalieparellano (approved):
Looks great! Thanks for the PR @joe-kimmel-vmw!

Issue #983, opened 2023-01-05
[RFC 0115] Add `-previous-image` to rebase


jabrown85:
### Description
Add `-previous-image` as an optional input to rebase command.

### Proposed solution
Add `-previous-image` as an optional input to rebase command. When `-previous-image` is provided, use the provided image reference for the previous image and _require_ a `-tag` input. The `-previous-image` will _not_ be a push target for a successful rebase.

More details can be found in the **[RFC](https://github.com/buildpacks/rfcs/pull/262).**

### Additional context
RFC: https://github.com/buildpacks/rfcs/pull/262

jabrown85:
Question for you @natalieparellano - could/should we tie this to a platform version? Since this is a new optional argument, I suppose we don't have to.

joeybrown-sf:
[RFC](https://github.com/buildpacks/rfcs/blob/main/text/0115-rebase-immutable-image-ref.md)

PR #985, opened 2023-01-06
allow passing previous image in rebase cmd

joeybrown-sf:
This allows a rebase by a image digest reference.

Resolves [#983](https://github.com/buildpacks/lifecycle/issues/983)

Here is an example command & output:
```
➜ lifecycle rebase -previous-image localhost:5003/cnb/rebase/demo@sha256:916a9e100569ee521b86d03b8499b9b93d7d256d6e838868ae720295f2ea2f76 localhost:5003/cnb/rebase/demo:v1 localhost:5003/cnb/rebase/demo:rebased
Saving localhost:5003/cnb/rebase/demo:v1...
*** Images (sha256:07c07bba660126053df38987bbb6098784b9348226ac689ded17f588c6708ed5):
      localhost:5003/cnb/rebase/demo:v1
      localhost:5003/cnb/rebase/demo:rebased
```

### DIFF
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -102,7 +102,7 @@ func (r *rebaseCmd) Exec() error {
 		Logger:      cmd.DefaultLogger,
 		PlatformAPI: r.PlatformAPI,
 	}
-	report, err := rebaser.Rebase(r.appImage, newBaseImage, r.AdditionalTags, r.OutputImageRef)
+	report, err := rebaser.Rebase(r.appImage, newBaseImage, r.OutputImageRef, r.AdditionalTags)
 	if err != nil {
 		return cmd.FailErrCode(err, r.CodeFor(platform.RebaseError), "rebase")
 	}
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -102,7 +102,7 @@ func (r *rebaseCmd) Exec() error {
 		Logger:      cmd.DefaultLogger,
 		PlatformAPI: r.PlatformAPI,
 	}
-	report, err := rebaser.Rebase(r.appImage, newBaseImage, r.AdditionalTags)
+	report, err := rebaser.Rebase(r.appImage, newBaseImage, r.AdditionalTags, r.OutputImageRef)
 	if err != nil {
 		return cmd.FailErrCode(err, r.CodeFor(platform.RebaseError), "rebase")
 	}
@@ -129,7 +129,7 @@ func (r *rebaseCmd) setAppImage() error {
 
 	if r.UseDaemon {
 		r.appImage, err = local.NewImage(
-			r.OutputImageRef,
+			targetImageRef,
 			r.docker,
 			local.FromBaseImage(targetImageRef),
 		)
@@ -140,7 +140,7 @@ func (r *rebaseCmd) setAppImage() error {
 			return err
 		}
 		r.appImage, err = remote.NewImage(
-			r.OutputImageRef,
+			targetImageRef,
 			keychain,
 			remote.FromBaseImage(targetImageRef),
 		)
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -40,6 +40,10 @@ func (r *rebaseCmd) DefineFlags() {
 	cli.FlagPreviousImage(&r.PreviousImageRef)
 
 	cli.DeprecatedFlagRunImage(&r.DeprecatedRunImageRef)
+
+	if r.PlatformAPI.AtLeast("0.11") {
+		cli.FlagPreviousImage(&r.PreviousImageRef)
+	}
 }
 
 // Args validates arguments and flags, and fills in default values.
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -37,6 +37,7 @@ func (r *rebaseCmd) DefineFlags() {
 	cli.FlagRunImage(&r.RunImageRef)
 	cli.FlagUID(&r.UID)
 	cli.FlagUseDaemon(&r.UseDaemon)
+	cli.FlagPreviousImage(&r.PreviousImageRef)
 
 	cli.DeprecatedFlagRunImage(&r.DeprecatedRunImageRef)
 }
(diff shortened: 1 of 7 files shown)


cmd/lifecycle/rebaser.go
```diff
@@ -101,7 +102,7 @@ func (r *rebaseCmd) Exec() error {
 		Logger:      cmd.DefaultLogger,
 		PlatformAPI: r.PlatformAPI,
 	}
-	report, err := rebaser.Rebase(r.appImage, newBaseImage, r.AdditionalTags)
+	report, err := rebaser.Rebase(r.appImage, newBaseImage, r.AdditionalTags, r.OutputImageRef)
```
jabrown85, line 105: What's the thought behind not executing `r.appImage.Rename(outputImageRef)` at this level?

I worry a bit about the `r.appImage` getting `Save`'d before a rename. It is supposed to be untouched if given as the `-previous-image`.

rebaser_test.go
```diff
@@ -70,33 +70,33 @@ func testRebaser(t *testing.T, when spec.G, it spec.S) {
 	when("#Rebase", func() {
 		when("app image and run image exist", func() {
```
jabrown85, line 71: If we keep the input paramter to `Rebase`, I think we should add a test that sends in a different value for the destination and assert we did not change the given `appImage`.

cmd/lifecycle/rebaser.go
```diff
@@ -37,6 +37,7 @@ func (r *rebaseCmd) DefineFlags() {
 	cli.FlagRunImage(&r.RunImageRef)
 	cli.FlagUID(&r.UID)
 	cli.FlagUseDaemon(&r.UseDaemon)
+	cli.FlagPreviousImage(&r.PreviousImageRef)
```
jabrown85, line 40: @natalieparellano did you have thoughts on if we should guard this on a platform API version or not? It isn't a breaking change, so I'm inclined to ship it without a guard...but that's me 😄

cmd/lifecycle/rebaser.go
joeybrown-sf: It's not here because some of the work inside rebaser.Rebase gets meta data from the `appImage`. 

Specifically, when we call [image.DecodeLabel](https://github.com/buildpacks/lifecycle/blob/50646839c80cc899193a8ed587ad59a20ec3f15a/rebaser.go#L30) inside that method, that in turn calls [image.Found](https://github.com/buildpacks/imgutil/blob/main/image.go#L64).

Here's the description for image.Found:

> Found tells whether the image exists in the repository by `Name()`.

So if we're solving this by using rename, we cannot rename until after [image.DecodeLabel](https://github.com/buildpacks/lifecycle/blob/50646839c80cc899193a8ed587ad59a20ec3f15a/rebaser.go#L30).

rebaser_test.go
joeybrown-sf: I don't quite follow. Would that be a different test than [this one](https://github.com/buildpacks/lifecycle/blob/50646839c80cc899193a8ed587ad59a20ec3f15a/rebaser_test.go#L357-L363)?

cmd/lifecycle/rebaser.go
joeybrown-sf: Maybe we create a new method that is something like `image.DecodeLabelNoValidation`?

jabrown85: That makes sense.

It feels odd that we have to pass in `AdditionalTags` and `OutputImageRef` and that we call `Save` on `appImage`, which today has `OutputImageRef` as the name already.

I wonder if we could make the signature more like 

```golang
func (r *Rebaser) Rebase(previousImage imgutil.Image, newBaseImage imgutil.Image, exportTargets []string) (RebaseReport, error) { {}
```

We could `Rename(exportTargets[0])` and then `Save(exportTargets[1:]...)`.

I'm still not in love the `Rename` flow - I would like to investigate your first approach before fully committing to this. In theory using the `remote.FromBaseImage(previousImage)` with an `outputImageRef` seems possible.

rebaser_test.go
jabrown85: That test is checking `fakeAppImage.SavedNames()` against `additionalNames, outputImageRef` but it isn't checking that the `previous-image` is NOT touched

cmd/lifecycle/rebaser.go
natalieparellano: Aye... I'm inclined to guard it just because that's what we do elsewhere and having two different patterns might be confusing. That said, maybe it makes sense to revisit our current strategy as the strictness we exercise around forwards-compatibility might be too extreme.

jabrown85: I do think that is a discussion worth having. Anything to ease the process for newcomers is a good thing IMO. But we can discuss that later.

What platform version would you want to use? I can make a PR to the spec to document the new parameter and behavior.

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/buildpacks/lifecycle
 
+replace github.com/buildpacks/imgutil => github.com/joeybrown-sf/imgutil v0.0.0-20230109163644-1c47df232fe2
```
jabrown85, line 3: We'll need to get the PR to `imgutil` first

jabrown85 (changes requested):

joeybrown-sf:
Depends on updates from [imgutil PR 167](https://github.com/buildpacks/imgutil/pull/167).

rebaser_test.go
```diff
@@ -333,25 +343,35 @@ func testRebaser(t *testing.T, when spec.G, it spec.S) {
...
 				h.AssertNil(t, fakeAppImage.SetLabel(platform.StackIDLabel, ""))
 				h.AssertNil(t, fakeNewBaseImage.SetLabel(platform.StackIDLabel, "io.buildpacks.stacks.cflinuxfs3"))
 
-				_, err := rebaser.Rebase(fakeAppImage, fakeNewBaseImage, additionalNames)
+				_, err := rebaser.Rebase(fakeAppImage, fakeNewBaseImage, fakeAppImage.Name(), additionalNames)
 				h.AssertError(t, err, "stack not defined on app image")
 			})
 		})
+
+		when("outputImageRef is different than app image name", func() {
```
jabrown85, line 367: nit
```suggestion
		when("outputImageRef is different than workingImage name", func() {
```

jabrown85 (approved):
Looks good to me!

PR #994, opened 2023-01-31
Stack removal: detect phase and analyze.toml support Targets

joe-kimmel-vmw:
fixes: #743 

- removes stacks from buildpack descriptors and detection logic,  
- shims from stack to target where relevant.
- changes and simplifies some pointer-struct to just strings,  which does change the analyze.toml spec. As far as I can tell this is victimless change, but probably there's a corresponding PR in the specs repo I haven't made yet?


natalieparellano (commented):
@joe-kimmel-vmw it's awesome to see this taking shape. I left a few comments for now

buildpack/bp_descriptor.go
```diff
@@ -61,6 +90,23 @@ func (d *BpDescriptor) String() string {
 	return d.Buildpack.Name + " " + d.Buildpack.Version
 }
 
+// Equals compares the two structs but they have to be really entirely identical, note this may be to strict IRL? how to tell?
+func (t *TargetMetadata) Equals(o *TargetMetadata) bool {
```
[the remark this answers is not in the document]
joe-kimmel-vmw: "satisfies" ?

cmd/lifecycle/creator.go
```diff
@@ -144,7 +144,7 @@ func (c *createCmd) Exec() error {
 			lifecycle.NewConfigHandler(),
 			dirStore,
 		)
-		detector, err := detectorFactory.NewDetector(c.AppDir, c.BuildConfigDir, c.OrderPath, c.PlatformDir, cmd.DefaultLogger)
+		detector, err := detectorFactory.NewDetector(c.AppDir, c.BuildConfigDir, c.OrderPath, c.PlatformDir, c.AnalyzedPath, cmd.DefaultLogger)
```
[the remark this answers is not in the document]
joe-kimmel-vmw: just for clarity --  "alphabetized but the logger is still last"?  or "alphabetized including the logger"?  I'm not immediately spotting another place where we do this. 

what if we make ... a new struct ... for making our other struct?  😬  e.g. 
```
struct DetectorInputs {  // i'm probably pretending Go uses C syntax here but let's be honest, if there was a sword stuck in a stone C would be the only language capable of pulling it out to become the one true compiler of England
  AppDir string
  BuildConfigDir string
  OrderPath string
  PlatformDir string
  AnalyzedPath string
  Logger PaulBunyan // whatever the right type is, idk :) 
}
```
then we could pass one of those structs in,  and basically it's just a hack to make Go give us keyword arguments , so the function call would look like:

`detectorFactory.NewDetector(DetectorInputs{AppDir: c.Appdir, BuildConfigDir: c.BuildConfigDir, ...})` 

OR honestly,  the only time we _don't_ already have a struct that we're just copying all the info out of is the unit tests,  so we could even make some convenience shims:
`detectorFactory.NewFromCreator(creator)` and `detectorFactory.NewFromDetectCmd(d)` .... oh or maybe those would be a little roundabout to avoid import cycles, but still.  overall I vaguely like the idea of using structs to replace long args lists (i.e. i miss python's kwargs) and/or having specialized functions to construct the new object from the object that already has all the data we want (i.e. i miss C++'s function overloading),  but as I'm calling out I'm also definitely guilty of wanting Go to be more like its cousins and less like itself,  so maybe there's some good arguments against such thoughts.

platform/files.go
```diff
@@ -83,8 +99,9 @@ type LayerMetadata struct {
 }
 
 type RunImageMetadata struct {
-	TopLayer  string `json:"topLayer" toml:"top-layer"`
-	Reference string `json:"reference" toml:"reference"`
+	TopLayer  string                   `json:"topLayer" toml:"top-layer"`
+	Reference string                   `json:"reference" toml:"reference"`
+	Target    buildpack.TargetMetadata `json:"target" toml:"target"`
```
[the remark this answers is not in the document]
joe-kimmel-vmw: Thanks --  this is mostly addressed in [this commit](https://github.com/buildpacks/lifecycle/pull/994/commits/7314fa9df829ddc8dcd2f19bc802dcd5a6f62dd9) but i want to leave this as unresolved for a while longer since I'm expecting some further refinements to what I did :)

cmd/lifecycle/creator.go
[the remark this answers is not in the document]
joe-kimmel-vmw: ok thx --  i think using `platform.LifecycleInputs` wouldn't give us a way to indicate clearly which fields are optional in the larger struct.   I also think since this only gets called from a 2 places (plus unit tests) it feels like over-engineering to do any of the heavier things i outlined above.   So that leaves alphabetizing...  but fwiw i had to sing the alphabet song to figure out which comes first: "analyzedPath" or "appDir",  so apologies if i still got it wrong  :)

buildpack/testdata/buildpack/by-id/B/v1/bin/build
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+set -euo pipefail
```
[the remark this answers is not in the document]
joe-kimmel-vmw: I agree with you and debated symlinks initially, and have been debating them some more since you left this comment :) . I'm still of the opinion that symlinks are a little like regular expressions in that I've usually seen them solve a problem short-term by creating a different problem that happens later. It might even just be the confusion someone has when they think they're only editing the file for a single test-case but then all the other ones start to fail?   

but idk, it's very true that at this time, these tests look like i'm trying to maximize my LOC metrics 😁 - they're identical, redundant, _and_ the contents don't even really matter for the purposes of the tests.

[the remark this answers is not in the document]
joe-kimmel-vmw: on further consideration, it looks like those files are entirely unnecessary, so i just deleted them :)

platform/export_inputs.go
```diff
@@ -97,7 +97,7 @@ func FillExportRunImage(i *LifecycleInputs, logger log.Logger) error {
 			if err != nil {
 				return err
 			}
-			if analyzedMD.RunImage == nil || analyzedMD.RunImage.Reference == "" {
```
natalieparellano, line 100: Isn't it still theoretically possible that this could be nil?

platform/files.go
```diff
@@ -87,6 +87,11 @@ type PreviousImageRunImageMetadata struct {
 	Reference string `json:"reference" toml:"reference"`
 }
 
+type RunImage struct {
+	Reference string                   `toml:"reference"`
+	Target    buildpack.TargetMetadata `json:"target" toml:"target"`
```
natalieparellano, line 92: `buildpack.TargetMetadata` is subtly different from `RunImage.Target` in that there can be multiple distributions:

```
type TargetMetadata struct {
	Os            string                 `json:"os" toml:"os"`
	Arch          string                 `json:"arch" toml:"arch"`
	ArchVariant   string                 `json:"arch-variant" toml:"arch-variant"`
	Distributions []DistributionMetadata `json:"distributions" toml:"distributions"` <- a single image can only have one of these
}
```

We might want to make a `platform.TargetMetadata` that looks like:

```
type TargetMetadata struct {
	Os            string                 `json:"os" toml:"os"`
	Arch          string                 `json:"arch" toml:"arch"`
	ArchVariant   string                 `json:"arch-variant" toml:"arch-variant"`
	Distribution  buildpack.DistributionMetadata `json:"distribution" toml:"distribution"`
}
```

I think this could make `TargetMetadata.Satisfies` simpler as we'll only have to check that "buildpack.Distributions" contains "platform.Distribution" instead of list equality.

detector.go
```diff
@@ -86,6 +87,14 @@ func (f *DetectorFactory) NewDetector(appDir, buildConfigDir, orderPath, platfor
 	if err := f.setOrder(detector, orderPath, logger); err != nil {
 		return nil, err
 	}
+
+	var amd platform.AnalyzedMetadata
```
jabrown85, line 91: Does this need to be guarded by Platform API version? Is there always an analyzed toml available to read on every version?

detector.go
```diff
@@ -195,32 +209,55 @@ func (d *Detector) detectGroup(group buildpack.Group, done []buildpack.GroupElem
...
+			bpDescriptor, err = d.DirStore.LookupBp(groupEl.ID, groupEl.Version)
 			if err != nil {
 				return nil, nil, err
 			}
 
+			targetMatch := false
+			if d.AnalyzeMD.RunImage.Target.IsWildcard() {
+				targetMatch = true
+			} else {
+				for _, target := range bpDescriptor.Targets {
```
jabrown85, line 228: Is the idea that buildpacks won't have targets unless they are opting into this validation? Does this validation need to be guarded at a platform API level?

platform/files.go
```diff
@@ -31,7 +40,7 @@ type ImageIdentifier struct {
 	Reference string `toml:"reference"`
 }
 
-func ReadAnalyzed(analyzedPath string, logger log.Logger) (AnalyzedMetadata, error) {
+func ReadAnalyzed(analyzedPath string, logger log.Logger) (AnalyzedMetadata, error) { // TODO should we be updating the platform api value here when we read it in? (would anyone ever call different phases of the same run with different api versions??)
```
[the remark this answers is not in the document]
joe-kimmel-vmw: thanks!

detector.go
joe-kimmel-vmw: hmm good point.  I was writing thinking that they'd all be empty/wildcard, but i guess there's a case where one of them would get the bionic stack and the other wouldn't, so then it wouldn't satisfy even if somebody was trying to run with an older API which would be a new bug/regression from user perspective,  so anyways you're right. API guard is the right call.  thanks!

detector.go
joe-kimmel-vmw: thanks - i think this is resolved by passing in the struct

detector.go
```diff
@@ -195,32 +209,55 @@ func (d *Detector) detectGroup(group buildpack.Group, done []buildpack.GroupElem
...
 		)
 		if groupEl.Kind() == buildpack.KindBuildpack {
-			descriptor, err = d.DirStore.LookupBp(groupEl.ID, groupEl.Version)
+			bpDescriptor, err = d.DirStore.LookupBp(groupEl.ID, groupEl.Version)
 			if err != nil {
 				return nil, nil, err
 			}
 
+			targetMatch := false
+			if d.AnalyzeMD.RunImage.Target.IsWildcard() {
```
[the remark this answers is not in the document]
joe-kimmel-vmw: hmm. i think just semantic readability -- confessionally, in my heart i'm a bit of a ruby/python programmer who wishes that each line of code could be an (admittedly kind of awkward) english sentence, so this appeals to me.  But I see your point that i've pulled out a function with a name for some logic that i'm really only using in one place.  Feel free to say something about "stylistically historically in this codebase we prefer... " and I can  just inline it, but re-reading it, it still passes my own personal threshold at this moment  :)

platform/files.go
```diff
@@ -20,10 +20,10 @@ import (
 // analyzed.toml
 
 type AnalyzedMetadata struct {
-	PreviousImage *ImageIdentifier `toml:"image"`
-	Metadata      LayersMetadata   `toml:"metadata"`
-	RunImage      *ImageIdentifier `toml:"run-image,omitempty"`
-	BuildImage    *ImageIdentifier `toml:"build-image,omitempty"`
+	PreviousImageRef string           `toml:"image-ref"`
```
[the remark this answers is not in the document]
joe-kimmel-vmw: i think this is resolved with the more recent changes and the file is now backwards compatible.

platform/files.go
joe-kimmel-vmw: thanks. i think you're right, and that let me to update a few other things,  so it's now `IsSatisfiedBy` b/c the method needs to know about both the TargetMetaData types (in platform and in buildpack) so it needs to be in the 'higher' package up the dependency graph...

platform/export_inputs.go
joe-kimmel-vmw: no i don't think so -- it's not a pointer anymore, so it should always be either the "zero" values of that struct, or the filled-in values of that struct.

detector.go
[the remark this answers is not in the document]
jabrown85: I was thinking about it from a external API builder. If we think a platform user may have the need to run this code, say in a custom `prepare` type of step, it would make it so they wouldn't have to know to check for wildcard explicitly. But maybe this file/validation isn't likely to be consumed externally anyway.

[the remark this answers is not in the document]
joe-kimmel-vmw: ok yeah, i gotchu --  moved into the detector and lowercased.

env/build.go
```diff
@@ -5,8 +5,9 @@ import (
 	"strings"
 )
 
+// BuildEnvIncludelist are env vars that are passed-through where the builder would supply them and then lifecycle would pass it through to the buildpack
```
natalieparellano, line 8:
```suggestion
// BuildEnvIncludelist are env vars that, if set in the lifecycle's execution environment - either in a builder or by the platform, are passed-through to buildpack executables
```

buildpack/bp_descriptor_test.go
```diff
@@ -29,5 +29,76 @@ func testBpDescriptor(t *testing.T, when spec.G, it spec.S) {
...
+			h.AssertEq(t, descriptor.WithAPI, "0.12")
+			h.AssertEq(t, descriptor.Buildpack.ID, "B")
+			h.AssertEq(t, descriptor.Buildpack.Name, "Buildpack B")
+			h.AssertEq(t, descriptor.Buildpack.Version, "v1")
+			h.AssertEq(t, descriptor.Buildpack.Homepage, "Buildpack B Homepage")
+			h.AssertEq(t, descriptor.Buildpack.SBOM, []string{"application/vnd.cyclonedx+json"})
+			// specific behaviors for this test
+			h.AssertEq(t, descriptor.Stacks[0].ID, "io.buildpacks.stacks.bionic")
+			h.AssertEq(t, len(descriptor.Targets), 1)
+			h.AssertEq(t, descriptor.Targets[0].Arch, "x86_64")
```
natalieparellano, line 82:
```suggestion
			h.AssertEq(t, descriptor.Targets[0].Arch, "amd64")
```
(this was wrong in the spec PR)

buildpack/bp_descriptor.go
```diff
@@ -6,13 +6,38 @@ import (
...
+	Stacks      []StackMetadata  `tome:"stacks"` // just for backwards compat so we can check if it's the bionic stack, which we translate to a target
+
+}
+
+type StackMetadata struct {
+	ID string `toml:"id"`
+}
+
+type TargetPartial struct {
+	Os          string `json:"os" toml:"os"`
```
natalieparellano, line 28:
```suggestion
	OS          string `json:"os" toml:"os"`
```

buildpack/bp_descriptor.go
```diff
@@ -38,6 +63,19 @@ func ReadBpDescriptor(path string) (*BpDescriptor, error) {
 	if descriptor.WithRootDir, err = filepath.Abs(filepath.Dir(path)); err != nil {
 		return &BpDescriptor{}, err
 	}
+
+	apiVersion, err := api.NewVersion(descriptor.WithAPI)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(descriptor.Targets) == 0 && apiVersion.AtLeast("0.12") {
```
natalieparellano, line 72:
```suggestion
	if len(descriptor.Targets) == 0 {
```
I apologize, the lifecycle issue and spec PR were not clear about this - we actually want to infer the target data for all (and especially older) Buildpack API; this is part of our [backwards compatibility strategy](https://github.com/buildpacks/spec/pull/336#issuecomment-1454099934). The only time we want to guard on API versions is when we write data.

platform/files.go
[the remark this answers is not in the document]
natalieparellano: Even though the information is preserved, this TOML annotation will render the file as:
```
image-ref: "previous-image-ref"
```
 instead of 
```
[image]
reference = "previous-image-ref"
```
as required by the platform spec [here](https://github.com/buildpacks/spec/blob/main/platform.md#analyzedtoml-toml).

natalieparellano (commented):
Adding a few comments for now, still reviewing

platform/files_test.go
```diff
@@ -154,4 +154,80 @@ func testFiles(t *testing.T, when spec.G, it spec.S) {
 			})
 		})
 	})
+	when("analyzed.toml", func() {
+		when("it is old it stays old", func() {
+			it("serializes and deserializes", func() {
```
natalieparellano, line 159: Though it's painful I think we do need to compare to a hard-coded string as the style of this test is hiding the regression mentioned here: https://github.com/buildpacks/lifecycle/pull/994#discussion_r1126643263

cmd/lifecycle/creator.go
```diff
@@ -172,7 +172,8 @@ func (c *createCmd) Exec() error {
 			lifecycle.NewConfigHandler(),
 			dirStore,
 		)
-		detector, err := detectorFactory.NewDetector(c.AppDir, c.BuildConfigDir, c.OrderPath, c.PlatformDir, cmd.DefaultLogger)
+		// QUESTION:  does everyone else think it's correct to pass in an empty/default analyzed.MD here since in this old API version the detect runs before the analyze?
```
natalieparellano, line 175: Seems okay to me... this whole "factory.NewPhase" thing is crying out for a refactor (now that we have `LifecycleInputs` for all phases why not pass the inputs in as you suggested earlier instead of having N arguments... then maybe the creator could introduce some overrides for structs that we don't actually write out in a single-phase invocation - analyzed.toml, group.toml, plan.toml ...but let's get this PR through first).

buildpack/bp_descriptor_test.go
```diff
@@ -29,5 +29,76 @@ func testBpDescriptor(t *testing.T, when spec.G, it spec.S) {
...
+			h.AssertEq(t, descriptor.Targets[0].Os, "linux")
+			h.AssertEq(t, descriptor.Targets[0].Distributions[0].Name, "ubuntu")
+			h.AssertEq(t, descriptor.Targets[0].Distributions[0].Version, "18.04")
+		})
+
+		it("does not translate non-special stack values", func() {
```
natalieparellano, line 88: We can assume linux/amd64 if ./bin/build is present and windows/amd64 if ./bin/build.exe or ./bin/build.bat is present

platform/files.go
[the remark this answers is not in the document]
joe-kimmel-vmw: I think i'm [shimming it appropriately for older apis](https://github.com/buildpacks/lifecycle/pull/994/files#diff-631f47fe759684bf2670a0a309f32e09dbcf15a1e470097b7e9ee15f0fda3d92R93) ... can we change the spec?  😁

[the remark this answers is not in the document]
natalieparellano: oic... sorry, I missed that!

platform/files_test.go
natalieparellano: Update: actually there is no regression, though testing the string would be the proof!

platform/files.go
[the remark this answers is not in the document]
natalieparellano: We can, for newer platforms, but we'd need to include it in the [spec PR](https://github.com/buildpacks/spec/pull/335) with a justification for why it needs to be different. Personally I don't think our hatred of pointers is sufficient to change something that people are already used to.

buildpack/bp_descriptor.go
natalieparellano: (I think this is more meticulously correct but I can't remember where I read this)

joe-kimmel-vmw: quick grep suggests it's more consistent.  thanks!

buildpack/bp_descriptor_test.go
joe-kimmel-vmw: this one might be worth a followup chat -- do we do that kind of inference anywhere else already?

platform/files_test.go
joe-kimmel-vmw: (we said i should test the string)

buildpack/bp_descriptor.go
```diff
@@ -38,6 +62,32 @@ func ReadBpDescriptor(path string) (*BpDescriptor, error) {
...
+			binFiles, err := os.ReadDir(binDir)
+			if err != nil {
+				return &BpDescriptor{}, err
+			}
+			for _, bf := range binFiles {
+				fname := bf.Name()
+				if fname == "build.exe" || fname == "build.bat" {
+					descriptor.Targets = append(descriptor.Targets, TargetMetadata{TargetPartial: TargetPartial{OS: "windows", Arch: "amd64"}})
+				}
+				if fname == "build" { // TODO: Question: do you think we should check whether its executable bit is set? the spec just does it by name...
```
natalieparellano, line 85:
```suggestion
				if fname == "build" {
```
Name seems okay for our purposes here

platform/files.go
```diff
@@ -20,6 +21,14 @@ import (
 // analyzed.toml
 
 type AnalyzedMetadata struct {
+	PreviousImageRef string           `toml:"image-ref"`
```
natalieparellano, line 24: Weren't we going to put this back to `ImageIdentifier`?

buildpack/bp_descriptor.go
natalieparellano: Resolved in 4c260baf

platform/files.go
natalieparellano: Resolved in 4c260baf

natalieparellano (approved):
Looks good to me! Thanks for all your hard work @joe-kimmel-vmw

PR #1009, opened 2023-02-13
Make a single constructor for lifecycle inputs

natalieparellano:
- The logic to update the default path for TOML files was repeated across phases
- In general it is safe to provide default values for inputs that might not be relevant to the current phase, as these will be ignored when constructing a new service for the phase; e.g., platform.LifecycleInputs.OrderPath will be ignored when constructing a lifecycle.Exporter
- As more inputs are shared across phases (e.g., analyzed.toml is now an input to the detect phase), duplicating the logic for providing default values is becoming more cumbersome


platform/detect_inputs.go
```diff
@@ -7,59 +7,5 @@ import (
...
-	case platformAPI.AtLeast("0.5"):
-		inputs = defaultDetectInputs05()
-	default:
-		inputs = defaultDetectInputs03To04()
-	}
-	inputs.PlatformAPI = platformAPI
-	return inputs
-}
-
-func defaultDetectInputs() LifecycleInputs {
```
natalieparellano, line 27: Just updated detect for now, but we can remove the individual constructors for each phase if we like this pattern

platform/lifecycle_inputs.go
```diff
@@ -55,6 +55,82 @@ type LifecycleInputs struct {
 	KanikoCacheTTL        time.Duration
 }
 
+func NewLifecycleInputs(platformAPI *api.Version) LifecycleInputs {
+	inputs := LifecycleInputs{
+		// Operator config
```
natalieparellano, line 60: Mirrored the organization in platform/defaults.go

jabrown85 (approved):
I like it!

platform/lifecycle_inputs.go
```diff
@@ -100,207 +171,6 @@ func notIn(list []string, str string) bool {
...
 
-var (
-	ErrOutputImageRequired           = "image argument is required"
-	ErrRunImageRequiredWhenNoStackMD = "-run-image is required when there is no stack metadata available"
-	ErrRunImageRequiredWhenNoRunMD   = "-run-image is required when there is no run metadata available"
-	ErrSupplyOnlyOneRunImage         = "supply only one of -run-image or (deprecated) -image"
-	ErrRunImageUnsupported           = "-run-image is unsupported"
-	ErrImageUnsupported              = "-image is unsupported"
-	MsgIgnoringLaunchCache           = "Ignoring -launch-cache, only intended for use with -daemon"
-)
```
natalieparellano, line 111: Moved to other file for readability

natalieparellano:
Marking as ready, but @jjbustamante we can wait for the OCI layout PR to land first to avoid extra work on your side :)

natalieparellano:
Look at all that deleted code!

jabrown85 (approved):

natalieparellano:
I brought this up to date with the latest main and made a couple other changes, chiefly requiring the desired value of layers directory to be read before a new "platform" can be instantiated. 

The old pattern was to provide a "placeholder" layers directory and then after-the-fact go back and update all the placeholder paths with the correct value now that the layers directory is known. This had the downside that when adding any new inputs that defaulted to a child of the layers directory (such as the newly added "generated" dir which defaults to `<layers>/generated`) we would need to remember to add it as a path to update later and failing to do this resulted in confusing errors.

platform/lifecycle_inputs.go
```diff
@@ -102,307 +185,6 @@ func notIn(list []string, str string) bool {
...
-	ErrOutputImageRequired           = "image argument is required"
-	ErrRunImageRequiredWhenNoStackMD = "-run-image is required when there is no stack metadata available"
-	ErrRunImageRequiredWhenNoRunMD   = "-run-image is required when there is no run metadata available"
-	ErrSupplyOnlyOneRunImage         = "supply only one of -run-image or (deprecated) -image"
-	ErrRunImageUnsupported           = "-run-image is unsupported"
-	ErrImageUnsupported              = "-image is unsupported"
-	ErrMultipleTargetsUnsupported    = "exporting to multiple targets is unsupported"
-	ErrLayoutDirRequired             = "defining a layout directory is required when OCI Layout feature is enabled. Use -layout-dir flag or CNB_LAYOUT_DIR environment variable"
-	MsgIgnoringLaunchCache           = "Ignoring -launch-cache, only intended for use with -daemon"
-)
```
natalieparellano, line 115: Moved to resolve_inputs.go

platform/resolve_analyze_inputs_test.go
```diff
@@ -161,48 +161,13 @@ func testAnalyzeInputs(platformAPI string) func(t *testing.T, when spec.G, it sp
...
-						err := platform.ResolveInputs(platform.Analyze, &inputs, logger)
+						err := platform.ResolveInputs(platform.Analyze, inputs, logger)
 						h.AssertNil(t, err)
 						h.AssertNoLogEntry(t, logHandler, `no stack metadata found at path ''`)
 						h.AssertNoLogEntry(t, logHandler, `Previous image with name "" not found`)
 					})
 				})
 			})
 		})
-
```
natalieparellano, line 172: These are now tested in lifecycle_inputs_test.go

platform/resolve_create_inputs_test.go
```diff
@@ -159,45 +159,13 @@ func testCreateInputs(platformAPI string) func(t *testing.T, when spec.G, it spe
...
-						err := platform.ResolveInputs(platform.Create, &inputs, logger)
+						err := platform.ResolveInputs(platform.Create, inputs, logger)
 						h.AssertNil(t, err)
 						h.AssertNoLogEntry(t, logHandler, `no stack metadata found at path ''`)
 						h.AssertNoLogEntry(t, logHandler, `Previous image with name "" not found`)
 					})
 				})
 			})
 		})
-
```
natalieparellano, line 170: These are now tested in lifecycle_inputs_test.go

cmd/lifecycle/main.go
```diff
@@ -18,31 +18,41 @@ import (
 )
 
 func main() {
-	switch strings.TrimSuffix(filepath.Base(os.Args[0]), filepath.Ext(os.Args[0])) {
+	phase := strings.TrimSuffix(filepath.Base(os.Args[0]), filepath.Ext(os.Args[0]))
```
joe-kimmel-vmw, line 21: fwiw I'm a fan of explicitly naming this string `phase` instead of leaving it anonymous 👏

cmd/lifecycle/main.go
```diff
@@ -18,31 +18,41 @@ import (
...
+	if phase == "rebaser" {
+		cli.Run(&rebaseCmd{Platform: platform.NewPlatformFor(platformAPIWithExitOnError(), "<layers>")}, false)
```
joe-kimmel-vmw, line 23: it's less clear why this one case needs to be above the switch though?  it seems like this is the only one that doesn't call `cli.FlagLayersDir` , along with the Exit / ExitWithVersion cases?  it looked like even everything under `subcommand` calls that -- were you trying to factor it out so we only have one line that says `cli.FlagLayersDir(&layersDir)` ?

joe-kimmel-vmw (commented):
sorry i feel like i'm leaving comments on your sidequest and not the main change,  feel free to ignore me

jabrown85 (approved):
I love this type of work!

cmd/lifecycle/main.go
natalieparellano: Yeah, when you call `cli.FlagLayersDir(&layersDir)` that registers the flag as valid, and it isn't for the rebaser. Alternatively we could do something like
```
if phase == "rebaser" {
		cli.Run(&rebaseCmd{Platform: platform.NewPlatformFor(platformAPIWithExitOnError(), "<layers>")}, false) 
} else {
		cli.FlagLayersDir(&layersDir) // assumes all phases that follow (including invalid phases) respect -layers
}
```
Would that be clearer? I'm open to other suggestions too

joe-kimmel-vmw: if it really is valid for every single other thing (which it seemed like?) then your if/else technology above would be a good solution.   otherwise we'll just have to live with it being _almost_ everywhere.  nbd

natalieparellano: Sounds good. Let's do if/else, it's clearer and less duplicative. I'll make the change

PR #1010, opened 2023-02-14
WIP: Send dispatch event to buildpacks/samples on lifecycle release

### LINKED ISSUE #801: Add automation to re-generate sample builders when lifecycle is published
### Description
This will ensure the sample builders are kept up-to-date with the lifecycle.

### Proposed solution
[Slack thread](https://buildpacks.slack.com/archives/CST4A3ECV/p1638807846018600?thread_ts=1638800094.018200&cid=CST4A3ECV)

> Can I propose that the [#implementation](https://buildpacks.slack.com/archives/CSFE49VGR) team add automation to have this done similar to how we do with pack releases?
See [here](https://github.com/buildpacks/samples/blob/aa79de59c4bae24e32f15fda467d02ae9cd94b01/.github/workflows/build-and-deploy.yml#L10-L12) and [here](https://github.com/buildpacks/pack/blob/main/.github/workflows/delivery-release-dispatch.yml).
Additionally, we’d need to update the `builder.toml`[ files](https://github.com/search?q=repo%3Abuildpacks%2Fsamples+filename%3Abuilder.toml&type=Code&ref=advsearch&l=&l=) to set an explicit version of the lifecycle instead of relying on pack’s default since they don’t typically correlate.

glokta1:
Adds workflow to send dispatch event to `buildpacks/samples` on lifecycle releases. This will be used as a trigger on the receiving end to rebuild the samples and update their `builder.toml` files with the latest lifecycle version.

Lifecycle version is sent in dispatch payload.

Resolves #801 

**Question:** Am I correct in thinking the activity type of `release` should be `released` and not `published`? I took it from the `post-release.yml` file but I presume we want to send the dispatch event only for (stable) releases, not pre-releases or draft releases?

### DIFF
--- a/.github/workflows/release-dispatch.yml
+++ b/.github/workflows/release-dispatch.yml
@@ -15,3 +15,4 @@ jobs:
           token: ${{ secrets.PLATFORM_GITHUB_TOKEN }}
           event-type: lifecycle-release
           repository: buildpacks/samples
+          client-payload: '{"version": "${{ github.event.release.tag_name }}"}'
--- a/.github/workflows/release-dispatch.yml
+++ b/.github/workflows/release-dispatch.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Repository Dispatch
         uses: peter-evans/repository-dispatch@v2
         with:
-          token: ${{ secrets.PLATFORM_GITHUB_TOKEN }}
+          token: ${{ secrets.IMPLEMENTATION_GITHUB_TOKEN }}
           event-type: lifecycle-release
           repository: buildpacks/samples
           client-payload: '{"version": "${{ github.event.release.tag_name }}"}'
--- a/.github/workflows/release-dispatch.yml
+++ b/.github/workflows/release-dispatch.yml
@@ -3,7 +3,7 @@ name: release-dispatch
 on:
   release:
     types:
-      - published
+      - released
 
 jobs:
   send-release-dispatch:
--- a/.github/workflows/release-dispatch.yml
+++ b/.github/workflows/release-dispatch.yml
@@ -1,0 +1,17 @@
+name: release-dispatch
+
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  send-release-dispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.PLATFORM_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-dispatch.yml
```diff
@@ -0,0 +1,18 @@
+name: release-dispatch
+
+on:
+  release:
+    types:
+      - published
```
natalieparellano, line 6: Great catch! I do think we want this to be `released`

.github/workflows/release-dispatch.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  send-release-dispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.PLATFORM_GITHUB_TOKEN }}
```
natalieparellano, line 15:
```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}
```

natalieparellano:
Thanks for the PR @glokta1! I'll test this out on my fork...

.github/workflows/release-dispatch.yml
natalieparellano: Oof, I thought the GITHUB_TOKEN would have enough permissions, but maybe not. Investigating...

.github/workflows/release-dispatch.yml
natalieparellano, line 6:
```suggestion
      - released
```

natalieparellano (commented):
@glokta1 thanks for this! I added a couple of suggestions.

natalieparellano (approved):

natalieparellano:
Looks good to me! @jabrown85 WDYT?

jabrown85 (approved):

natalieparellano:
@glokta1 apologies for letting this PR sit here - I think we were waiting for https://github.com/buildpacks/samples/pull/138 to react to the dispatch before merging them together. It looks like https://github.com/buildpacks/samples/pull/138 registers the event but doesn't actually update the builder.toml files or published images. Let us know if you intend on picking that back up, otherwise we might just close both PRs for now.

glokta1:
Hey @natalieparellano, I should apologize for the lack of communication. I encountered some difficulties in setting up the test environment while writing the workflow and left it at the time. I will pick this back up in a 2-3 days' time :)

natalieparellano:
No worries, thanks a lot for your effort @glokta1 ❤️

PR #1013, opened 2023-02-17
Preserves test output for easier grepping

joe-kimmel-vmw:
- unit and acceptances tests are tee'd to a file in /out (which is already .gitignore-d)
- this makes it easier to grep for failures after the test run, if you're into that kind of thing
- rhymes with a previous change in pack: https://github.com/buildpacks/pack/pull/1585


jjbustamante (approved):
LGTM!

Issue #997, opened 2023-02-02
[RFC #0105] Restorer changes for run image extension (Dockerfiles phase 3)


natalieparellano:
Spec PR: https://github.com/buildpacks/spec/pull/347/files

The `restorer`:

```
  - MUST resolve `run-image.reference` to a digest in `<analyzed>` if not present
  - MUST populate `run-image.target` data in `<analyzed>` if not present
  - **If** `<analyzed>` has `run-image.extend = true`, the lifecycle:
    - MUST download from the registry and save in OCI layout format the `run-image` in `<analyzed>` to `<kaniko-dir>/cache`
```

natalieparellano:
We may only need the image manifest and config (like for build image extension). Will confirm as part of #998

natalieparellano:
Edit: actually I don't think we need `-generated`, as analyzed.toml has all the info we need.

PR #1014, opened 2023-02-17
Restorer changes for run image extension

natalieparellano:
Fixes https://github.com/buildpacks/lifecycle/issues/997


natalieparellano:
Still needs to add target data; blocking on https://github.com/buildpacks/lifecycle/pull/994

acceptance/extender_test.go
```diff
@@ -90,9 +90,10 @@ func testExtenderFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				buildImageDigest := buildImageHash.String()
 				baseCacheDir := filepath.Join(kanikoDir, "cache", "base")
 				h.AssertNil(t, os.MkdirAll(baseCacheDir, 0755))
-				layoutPath, err := selective.Write(filepath.Join(baseCacheDir, buildImageDigest), empty.Index)
+				var sparseImage imgutil.Image
+				sparseImage, err = sparse.NewImage(filepath.Join(baseCacheDir, digest.DigestStr()), remoteImage)
```
natalieparellano, line 94:
```suggestion
				sparseImage, err = sparse.NewImage(filepath.Join(baseCacheDir, buildImageDigest), remoteImage)
```

natalieparellano:
This will be easier to review as a diff on top of `runext/detect-996` ...I'm working on getting that one up-to-date with `main`, then I'll repoint. For now, the changes shown include commits from both branches.

natalieparellano:
Still need to bring this up-to-date with main (hence the huge diff)

acceptance/exporter_test.go
```diff
@@ -371,29 +371,38 @@ func assertImageOSAndArchAndCreatedAt(t *testing.T, imageName string, phaseTest
...
-	analyzedMD = assertAnalyzedMetadata(t, placeHolderPath)
-	analyzedMD.RunImage = &platform.RunImage{Reference: phaseTest.targetRegistry.fixtures.ReadOnlyRunImage}
-	encoding.WriteTOML(strings.TrimSuffix(placeHolderPath, ".placeholder"), analyzedMD)
-
-	placeHolderPath = filepath.Join("testdata", "exporter", "container", "layers", "layout-analyzed.toml.placeholder")
-	analyzedMD = assertAnalyzedMetadata(t, placeHolderPath)
-	// Values from image acceptance/testdata/exporter/container/layout-repo in OCI layout format
-	analyzedMD.RunImage = &platform.RunImage{
-		Reference: "/layout-repo/index.docker.io/library/busybox/latest@sha256:445c45cc89fdeb64b915b77f042e74ab580559b8d0d5ef6950be1c0265834c33",
+	regPlaceholders := []string{
```
natalieparellano, line 374: This helper function exists because we expect the run image in analyzed.toml to contain the registry IP and port, which aren't known until we start the test

buildpack/bp_descriptor.go
```diff
@@ -31,10 +31,10 @@ type TargetPartial struct {
 
 type TargetMetadata struct {
 	TargetPartial
-	Distributions []DistributionMetadata `json:"distributions" toml:"distributions"`
+	Distributions []OSDistribution `json:"distributions" toml:"distributions"`
```
natalieparellano, line 34: Preferred the slightly more descriptive name

cmd/lifecycle/restorer.go
```diff
@@ -111,64 +139,84 @@ func (r *restoreCmd) Exec() error {
...
-		return fmt.Errorf("failed to read image: %w", err)
+	return digest.DigestStr() == ""
+}
+
+func (r *restoreCmd) supportsBuildImageExtension() bool {
+	return r.PlatformAPI.AtLeast("0.10")
+}
+
+func (r *restoreCmd) supportsRunImageExtension() bool {
+	return r.PlatformAPI.AtLeast("0.12") && !r.UseDaemon && !r.UseLayout
```
natalieparellano, line 171: We might be able to relax this requirement later

platform/files.go
```diff
@@ -50,14 +50,9 @@ type ImageIdentifier struct {
...
 
 type RunImage struct {
-	Reference string          `toml:"reference"`
-	Extend    bool            `toml:"extend,omitempty"`
-	Target    *TargetMetadata `json:"target,omitempty" toml:"target,omitempty"`
-}
-
-type TargetMetadata struct {
-	buildpack.TargetPartial
-	Distribution *buildpack.DistributionMetadata `json:"distribution,omitempty" toml:"distribution,omitempty"`
```
natalieparellano, line 60: Moved to run_image.go, but if this is the right place for it I can put it back

jabrown85 (approved):

cmd/lifecycle/restorer.go
```diff
@@ -90,10 +79,52 @@ func (r *restoreCmd) Privileges() error {
...
+		} else if needsUpdating(analyzedMD.RunImage) {
+			cmd.DefaultLogger.Debugf("Updating analyzed metadata...")
+			runImage, digest, err := newRemoteImage(analyzedMD.RunImage.Reference, r.keychain)
+			if err != nil {
+				return cmd.FailErr(err, "reading run image")
+			}
+			targetData, err := platform.ReadTargetData(runImage)
+			if err != nil {
+				return cmd.FailErr(err, "reading target data from run image")
+			}
```
natalieparellano, line 116: I overlooked the daemon/layout case here. We need https://github.com/buildpacks/imgutil/pull/184 to land, then we could use that instead to get target data for all export formats.

cmd/lifecycle/restorer.go
```diff
@@ -111,64 +142,84 @@ func (r *restoreCmd) Exec() error {
...
+	if imageRef == "" {
+		return nil, name.Digest{}, nil
+	}
+	baseCacheDir := filepath.Join(kanikoDir, "cache", "base")
+	if err := os.MkdirAll(baseCacheDir, 0755); err != nil {
+		return nil, name.Digest{}, fmt.Errorf("failed to create cache directory: %w", err)
 	}
-	buildImageHash, err := remoteImage.Digest()
+	// get remote image
+	remoteImage, digest, err := newRemoteImage(imageRef, r.keychain)
```
natalieparellano, line 186: If exporting to layout format we could look for the run image in the layout directory, but since the pack implementation currently requires the builder to exist in a registry (even when exporting to a daemon) we could keep this requirement and drop it whenever https://github.com/buildpacks/pack/issues/1623 is completed.

cmd/lifecycle/restorer.go
```diff
@@ -90,10 +79,52 @@ func (r *restoreCmd) Privileges() error {
 }
 
 func (r *restoreCmd) Exec() error {
-	if r.supportsBuildImageExtension() {
-		if err := r.pullBuilderImageIfNeeded(); err != nil {
-			return cmd.FailErr(err, "read builder image")
+	var analyzedMD platform.AnalyzedMetadata
+	if _, err := toml.DecodeFile(r.AnalyzedPath, &analyzedMD); err == nil {
```
joe-kimmel-vmw, line 83: here can you use the `platform.ReadAnalyzed` func?

platform/run_image.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+const (
+	TargetLabel                = "io.buildpacks.target"
+	OSDistributionNameLabel    = "io.buildpacks.distribution.name"
+	OSDistributionVersionLabel = "io.buildpacks.distribution.version"
+)
+
+type TargetMetadata struct {
+	ID           string          `toml:"id"`
```
joe-kimmel-vmw, line 17: heh -- i also just added this ID Field. but i think here you got rid of that "partial" struct i was playing with... pros and cons, but now it's decoupled from what happens in the buildpack side, which imo was the only value of that partial (to preserve & codify the coupling).  I thought it was correct to keep them coupled (and add the ID field to both structs) but if we're not doing that then it's probably correct to delete the TargetPartial struct entirely from bp_descriptor.go.

platform/run_image.go
```diff
@@ -0,0 +1,78 @@
...
+	OS           string          `toml:"os"`
+	Arch         string          `toml:"arch"`
+	ArchVariant  string          `toml:"variant"`
+	Distribution *OSDistribution `toml:"distribution"`
+}
+
+type OSDistribution struct {
```
joe-kimmel-vmw, line 24: here too --  what's your guidance on de/coupling with the bp_descriptor.go file?    My gut still says that they're reading & writing the same things, so it's better to have them using the same structs.  But I also think it's somewhere between possible and likely that I'm missing something..

cmd/lifecycle/restorer.go
```diff
@@ -90,10 +79,52 @@ func (r *restoreCmd) Privileges() error {
...
+			targetData, err := platform.ReadTargetData(runImage)
+			if err != nil {
+				return cmd.FailErr(err, "reading target data from run image")
+			}
+			analyzedMD.RunImage = &platform.RunImage{
+				Reference:      digest.String(),
+				Extend:         true,
+				TargetMetadata: &targetData,
+			}
+		} else if needsUpdating(analyzedMD.RunImage) {
```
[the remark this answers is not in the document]
natalieparellano: Updated in 0f2305aa

cmd/lifecycle/restorer.go
natalieparellano: Updated in 5adcca54

platform/run_image.go
natalieparellano: In this case the platform spec and the buildpack spec happen to match each other (and happily so), but there isn't any technical reason that they are tied together. By contrast where for example `platform.BuildMetadata` depends on `buildpack.Label` that schema is directly coming from the buildpack spec and how buildpacks define labels. At least, that is my rationalization.

platform/run_image.go
natalieparellano: I removed it just because it's simpler, but if we find we need it, we can put it back.

natalieparellano:
Let's block this on https://github.com/buildpacks/lifecycle/pull/1030 as the changes from that PR will be useful here

cmd/lifecycle/exporter.go
```diff
@@ -64,12 +64,12 @@ func (e *exportCmd) DefineFlags() {
 	cli.FlagProcessType(&e.DefaultProcessType)
 	cli.FlagProjectMetadataPath(&e.ProjectMetadataPath)
 	cli.FlagReportPath(&e.ReportPath)
-	cli.FlagRunImage(&e.RunImageRef)
+	cli.FlagRunImage(&e.RunImageRef) // Note: this flag isn't valid on Platform 0.7 and later, and any provided value is ignored
```
natalieparellano, line 67:
```suggestion
	cli.FlagRunImage(&e.RunImageRef) // Note: this flag isn't valid on Platform 0.7 and later
```
(randomly found when reviewing the code) Actually, I don't think we even ignore the provided value :/ we should probably change this, especially when we formally deprecate platforms older than 0.7

cmd/lifecycle/exporter.go
```diff
@@ -64,12 +64,12 @@ func (e *exportCmd) DefineFlags() {
...
 	cli.FlagStackPath(&e.StackPath)
 	cli.FlagUID(&e.UID)
 	cli.FlagUseDaemon(&e.UseDaemon)
 
-	cli.DeprecatedFlagRunImage(&e.DeprecatedRunImageRef)
+	cli.DeprecatedFlagRunImage(&e.DeprecatedRunImageRef) // Note: this flag isn't valid on Platform 0.7 and later, and any provided value is ignored
```
natalieparellano, line 72:
```suggestion
	cli.DeprecatedFlagRunImage(&e.DeprecatedRunImageRef) // Note: this flag isn't valid on Platform 0.7 and later
```

platform/files.go
```diff
@@ -89,40 +102,6 @@ func (t *TargetMetadata) IsSatisfiedBy(o *buildpack.TargetMetadata) bool {
 	return true
 }
 
-func GetTargetFromImage(image imgutil.Image) (*TargetMetadata, error) {
```
natalieparellano, line 92: Moved to run_image.go

natalieparellano:
@jabrown85 @joe-kimmel-vmw I think this is ready for one more look

cmd/lifecycle/exporter.go
joe-kimmel-vmw: wait what does that mean,  that it's not valid but "we don't even ignore the provided value" -- what do we do with it?

PR #1019, opened 2023-02-21
Add Layer Metadata test as living doc

AidanDelaney:
Layer Metadata is straightforward, adding a test to act as a living document.

### DIFF
--- a/buildpack/layermetadata_test.go
+++ b/buildpack/layermetadata_test.go
@@ -1,0 +1,44 @@
+package buildpack_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	"github.com/buildpacks/lifecycle/buildpack"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+)
+
+func TestLayerMetadata(t *testing.T) {
+	spec.Run(t, "unit-layermetadata", testLayerMetadata, spec.Report(report.Terminal{}))
+}
+
+func testLayerMetadata(t *testing.T, when spec.G, it spec.S) {
+	when("#LayerMetadata", func() {
+		var (
+			metadataFile *os.File
+		)
+		it.Before(func() {
+			var err error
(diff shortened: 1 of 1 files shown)


buildpack/layermetadata_test.go
```diff
@@ -0,0 +1,47 @@
...
+		)
+		it.Before(func() {
+			var err error
+			metadataFile, err = os.CreateTemp("", "test")
+			h.AssertNil(t, err)
+		})
+		it.After(func() {
+			os.Remove(metadataFile.Name())
+		})
+		it("decode file correctly", func() {
```
natalieparellano, line 34:
```suggestion
		it("decodes file correctly", func() {
```

buildpack/layermetadata_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/buildpacks/lifecycle/buildpack"
+	h "github.com/buildpacks/lifecycle/testhelpers"
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+	"github.com/tj/assert"
+)
+
+func TestLayerMetadata(t *testing.T) {
+	if runtime.GOOS != "windows" {
```
natalieparellano, line 16: Why no Windows?

natalieparellano (commented):
Sounds good to me! FWIW this is implicitly tested [here](https://github.com/buildpacks/lifecycle/blob/3be3585644a67d44c034f1a823b3ae26790a3485/buildpack/build_test.go#L357-L392) but it's always nice to test a smaller unit.

buildpack/layermetadata_test.go
AidanDelaney: Oh. Mostly because I didn't have a Windows machine on the train with me :)  Have changed this.

buildpack/layermetadata_test.go
AidanDelaney: Thanks.  Made this change.

natalieparellano (approved):

PR #1022, opened 2023-02-24
Extender changes for run image extension

### LINKED ISSUE #998: [RFC #0105] Extender changes for run image extension (Dockerfiles phase 3)
Spec PR: https://github.com/buildpacks/spec/pull/347/files

For newer platform API (`0.12` and above), the `extender` accepts new `-extended` and `-kind` flags. `-extended` is the location of a directory to output extended run image layers. `-kind` is the kind of image to extend (`build` (default) or `run`).

When extending the run image, the `extender` should find the image-to-extend's manifest & config in `<kaniko-dir>/cache` (the restorer pulls these down).

After extending the image, the `extender`:

```
  - **If** any `run.Dockerfile` set the label `io.buildpacks.rebasable` to `false` or left the label unset:
    - SHALL set the label `io.buildpacks.rebasable` to `false` on the extended run image
  - SHALL copy the manifest and config for the extended run image, along with any new layers, to `<extended>`
```

The extender should provide `user_id` and `group_id` build args to each Dockerfile (build or run).

natalieparellano:
Fixes #998 

Branching off other feature branch for now, will repoint when that is merged


platform/lifecycle_inputs.go
```diff
@@ -71,11 +73,17 @@ const PlaceholderLayers = "<layers>"
 func NewLifecycleInputs(platformAPI *api.Version) *LifecycleInputs {
 	// FIXME: api compatibility should be validated here
 
+	var skipLayers bool
+	if boolEnv(EnvSkipLayers) || boolEnv(EnvSkipRestore) {
+		skipLayers = true
+	}
```
natalieparellano, line 79: Unrelated bug

internal/selective/write.go
```diff
@@ -59,7 +59,7 @@ func (l Path) writeImage(img v1.Image) error {
 	if err != nil {
 		return err
 	}
-	if err := l.WriteBlob(cfgName, io.NopCloser(bytes.NewReader(cfgBlob))); err != nil {
```
natalieparellano, line 62: Not sure why this stood out to me

acceptance/extender_test.go
```diff
@@ -55,6 +57,9 @@ func TestExtender(t *testing.T) {
 	extendRegFixtures = extendTest.targetRegistry.fixtures
 
 	for _, platformAPI := range api.Platform.Supported {
+		if platformAPI.LessThan("0.10") {
+			continue
+		}
```
natalieparellano, line 62: Speeds the tests up a little

extender.go
```diff
@@ -1,40 +1,47 @@
...
+	CacheTTL          time.Duration            // a platform input
+	DockerfileApplier DockerfileApplier        // uses kaniko, BuildKit, or other to apply the provided Dockerfile to the provided image
+	Extensions        []buildpack.GroupElement // extensions are ordered from group.toml
 }
 
 //go:generate mockgen -package testmock -destination testmock/dockerfile_applier.go github.com/buildpacks/lifecycle DockerfileApplier
 type DockerfileApplier interface {
-	Apply(workspace string, digest string, dockerfiles []extend.Dockerfile, options extend.Options) error
+	ImageFor(reference string) (v1.Image, error)
+	Apply(dockerfile extend.Dockerfile, toBaseImage v1.Image, withBuildOptions extend.Options, logger log.Logger) (v1.Image, error)
```
natalieparellano, line 43: Pretty significant refactor, now instead of taking a list of Dockerfiles the applier takes a single Dockerfile. This allows us to consolidate all the CNB-spec logic in the `lifecycle` package without it bleeding into `internal/extend/kaniko`.

extender.go
```diff
@@ -113,38 +121,272 @@ func (f *ExtenderFactory) verifyAPIs(groupExt []buildpack.GroupElement, logger l
...
+}
+
+func (e *Extender) saveSelective(image v1.Image, origTopLayerHash string) error {
+	// save sparse image (manifest and config)
+	imageHash, err := image.Digest()
+	if err != nil {
+		return fmt.Errorf("getting image hash: %w", err)
+	}
+	toPath := filepath.Join(e.ExtendedDir, "run", imageHash.String())
+	layoutPath, err := selective.Write(toPath, empty.Index) // FIXME: this should use the imgutil layout/sparse package instead, but for some reason sparse.NewImage().Save() fails when the provided base image is already sparse
```
natalieparellano, line 213: This was a bummer to discover. When we update imgutil we can remove some logic (e.g., `topLayer`) from this file.

internal/extend/kaniko/dockerfile_applier.go
```diff
@@ -17,25 +21,65 @@ const (
...
+	if err != nil {
+		return nil, fmt.Errorf("getting image from hash '%s': %w", hash.String(), err)
+	}
+	return v1Image, nil
 }
 
-func NewDockerfileApplier(logger log.Logger) *DockerfileApplier {
-	return &DockerfileApplier{
-		logger: logger,
+func (a *DockerfileApplier) Cleanup() error {
```
natalieparellano, line 61: This could cause problems when images are extended in parallel (if the kaniko directory is re-used, which it might be). It may be safer to move this logic to the exporter.

natalieparellano:
@jabrown85 @joe-kimmel-vmw this one is finally ready for review :)

extender.go
```diff
@@ -113,38 +121,272 @@ func (f *ExtenderFactory) verifyAPIs(groupExt []buildpack.GroupElement, logger l
...
+	logger.Debugf("Extending base image for %s: %s", kind, e.ImageRef)
+	dockerfiles, err := e.dockerfilesFor(kind, logger)
 	if err != nil {
-		return fmt.Errorf("failed to get digest for reference %s: %w", e.ImageRef, err)
+		return nil, fmt.Errorf("getting %s dockerfiles: %w", kind, err)
+	}
+
+	var (
+		configFile *v1.ConfigFile
+		rebasable  = true // for now, we don't require the initial base image to have io.buildpacks.rebasable=true
```
joe-kimmel-vmw, line 288: that's gonna be a tough one right? bc if we don't start that way, then technically we'd break backwards compatibility to ever require it?   is the issue that basically nobody's gonna decorate their images like that,  so we kind of just say "the absence of rebasable=false label is true"?  

I guess i'm questioning whether the comment should imply that in the future we'll really have a default of rebaseable false,  given that we're starting with a default of rebaseable true.   are we skirting a spec change by saying "yeah yeah, we'll do it later"?? 😆

extender.go
```diff
@@ -113,38 +121,272 @@ func (f *ExtenderFactory) verifyAPIs(groupExt []buildpack.GroupElement, logger l
...
+
+		// apply Dockerfile
+
+		if baseImage, err = e.DockerfileApplier.Apply(
+			dockerfile,
+			baseImage,
+			buildOptions,
+			logger,
+		); err != nil {
+			return nil, fmt.Errorf("applying dockerfile to image: %w", err)
```
joe-kimmel-vmw, line 311: if/when we return partway through this loop what does that mean for the user's docker images?  are they going to be partially extended and in an inconsistent or unusable state?  is there a way to ... "roll back the transaction"?   or maybe indicate to the user "our condolences for the loss of your docker image. you have a backup, right?"   Or is the UX cool, here?

platform/files.go
```diff
@@ -76,7 +76,7 @@ type OSDistribution struct {
 	Version string `json:"version" toml:"version"`
 }
 
-// Satisfies treats optional fields (ArchVariant and Distributions) as wildcards if empty, returns true if
+// IsSatisfiedBy treats optional fields (ArchVariant and Distributions) as wildcards if empty, returns true if
```
joe-kimmel-vmw, line 79: thanks

joe-kimmel-vmw (commented):
I left you a couple questions but idk, i'm not actually an approver anyhow 😆 feel free to ignore me,  or tell me why i'm wrong / what i'm missing

acceptance/extender_test.go
```diff
@@ -148,10 +167,103 @@ func testExtenderFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
...
+					t.Log("second build extends the build image by pulling from the cache directory")
+					secondOutput := h.DockerRunWithCombinedOutput(t,
+						extendImage,
+						h.WithFlags(extendFlags...),
+						h.WithArgs(extendArgs...),
+					)
+					h.AssertStringDoesNotContain(t, secondOutput, "Did not find cache key, pulling remote image...")
+					h.AssertStringDoesNotContain(t, secondOutput, "Error while retrieving image from cache: oci")
+					h.AssertStringDoesNotContain(t, secondOutput, "ca-certificates")                                                             // shows that first cache layer was used
+					h.AssertStringDoesNotContain(t, secondOutput, "No cached layer found for cmd RUN apt-get update && apt-get install -y tree") // shows that second cache layer was used
```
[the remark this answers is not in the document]
natalieparellano: You are so right.

For line 3, `ca-certificates`, the positive assertion actually does exist [here](https://github.com/buildpacks/lifecycle/blob/21f9560b590a98205aa4f251812c01e39351c58b/acceptance/extender_test.go#L151). For line 4, `No cached layer found...`, I'll add it ina43a606a. For these lines, we could maybe add some sort of time-based assertion (since using the cache will be faster) and also I **think** they are also covered by [this line](https://github.com/buildpacks/lifecycle/blob/21f9560b590a98205aa4f251812c01e39351c58b/acceptance/extender_test.go#L214) since generating a new layer will cause the image SHA to be different.

For lines 1-2, there's probably no good option, and maybe we don't even need these tests anymore. In development I was trying to make sure that I used a particular code path but the end result of either was the same. For now I added constants in 007da926.

extender.go
natalieparellano: Fair point - I'l remove "for now" because I don't think we'll ever actually make this a requirement.

extender.go
natalieparellano: Great question! It should be okay - we're using the filesystem as our export target. Specifically, we're writing to the kaniko dir as a "scratch" workspace and only copying the image layers to `<extended>/run/<image>` after all Dockerfiles have been applied.

platform/files.go
natalieparellano: You can thank my IDE :)

acceptance/extender_test.go
```diff
@@ -39,6 +39,11 @@ var (
 	extendTest           *PhaseTest
 )
 
+const (
+	msgPullingRemoteImage          = "Did not find cache key, pulling remote image"
+	msgErrRetrievingImageFromCache = "Error while retrieving image from cache: oci"
```
joe-kimmel-vmw, line 44: i see --   ideally imo these constants would be used in the (production) code that generates the logs, and then consumed from the tests,  so that it's clear that those log lines are relied on someplace else outside the immediate domain,  and the tests would update effortlessly with the constants.

natalieparellano: Ohhhh I see... those log lines are emitted by kaniko :/

PR #1024, opened 2023-02-28
Verify read access while selecting best `run-image` mirror

pbusko:
Instead of simply picking the first run-image, the lifecycle will select the first **accessible** reference, if there's no match by registry.

See https://github.com/buildpacks/spec/pull/357

### DIFF
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -47,8 +47,8 @@ func TestAnalyzer(t *testing.T) {
 	analyzeDaemonFixtures = analyzeTest.targetDaemon.fixtures
 	analyzeRegFixtures = analyzeTest.targetRegistry.fixtures
 
-	for _, platformAPI := range []string{"0.12"} {
-		spec.Run(t, "acceptance-analyzer/"+platformAPI, testAnalyzerFunc(platformAPI), spec.Parallel(), spec.Report(report.Terminal{}))
+	for _, platformAPI := range api.Platform.Supported {
+		spec.Run(t, "acceptance-analyzer/"+platformAPI.String(), testAnalyzerFunc(platformAPI.String()), spec.Parallel(), spec.Report(report.Terminal{}))
 	}
 }
 
--- a/acceptance/analyzer_test.go
+++ b/acceptance/analyzer_test.go
@@ -47,8 +47,8 @@ func TestAnalyzer(t *testing.T) {
 	analyzeDaemonFixtures = analyzeTest.targetDaemon.fixtures
 	analyzeRegFixtures = analyzeTest.targetRegistry.fixtures
 
-	for _, platformAPI := range api.Platform.Supported {
-		spec.Run(t, "acceptance-analyzer/"+platformAPI.String(), testAnalyzerFunc(platformAPI.String()), spec.Parallel(), spec.Report(report.Terminal{}))
+	for _, platformAPI := range []string{"0.12"} {
+		spec.Run(t, "acceptance-analyzer/"+platformAPI, testAnalyzerFunc(platformAPI), spec.Parallel(), spec.Report(report.Terminal{}))
 	}
 }
 
@@ -271,7 +271,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				output, err := cmd.CombinedOutput()
 
 				h.AssertNotNil(t, err)
-				expected := "ensure registry read access to some-run-image-from-run-toml"
+				expected := "failed to find accessible run image"
 				h.AssertStringContains(t, string(output), expected)
 			})
 		})
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -67,6 +67,7 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
 		return cmd.FailErrCode(err, cmd.CodeForInvalidArgs, "parse arguments")
 	}
 	a.LifecycleInputs.OutputImageRef = args[0]
+	a.LifecycleInputs.AccessChecker = &platform.RemoteImageStrategy{}
 	if err := platform.ResolveInputs(platform.Analyze, a.LifecycleInputs, cmd.DefaultLogger); err != nil {
 		return cmd.FailErrCode(err, cmd.CodeForInvalidArgs, "resolve inputs")
 	}
--- a/cmd/lifecycle/analyzer.go
+++ b/cmd/lifecycle/analyzer.go
@@ -67,7 +67,6 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
 		return cmd.FailErrCode(err, cmd.CodeForInvalidArgs, "parse arguments")
 	}
 	a.LifecycleInputs.OutputImageRef = args[0]
-	a.LifecycleInputs.AccessChecker = &platform.RemoteImageStrategy{}
 	if err := platform.ResolveInputs(platform.Analyze, a.LifecycleInputs, cmd.DefaultLogger); err != nil {
 		return cmd.FailErrCode(err, cmd.CodeForInvalidArgs, "resolve inputs")
 	}
--- a/platform/files.go
+++ b/platform/files.go
@@ -10,10 +10,14 @@ import (
 	"github.com/buildpacks/lifecycle/internal/fsutil"
 
 	"github.com/BurntSushi/toml"
+	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/pkg/errors"
 
+	"github.com/buildpacks/imgutil/remote"
+
 	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/auth"
 	"github.com/buildpacks/lifecycle/buildpack"
 	"github.com/buildpacks/lifecycle/internal/encoding"
 	"github.com/buildpacks/lifecycle/launch"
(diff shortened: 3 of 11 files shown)


platform/files.go
```diff
@@ -279,38 +283,79 @@ type RunImageMetadata struct {
...
+	for _, image := range runImageMirrors {
+		ok, err := accessChecker.CheckReadAccess(image, keychain)
+		if err != nil {
+			return "", err
+		}
 
-func byRegistry(reg string, imgs []string) (string, error) {
-	if len(imgs) < 1 {
-		return "", errors.New("no images provided to search")
+		if ok {
```
jabrown85, line 328: Is there a world where we need to gate this new behavior behind a Platform API check?

platform/files.go
```diff
@@ -279,38 +283,79 @@ type RunImageMetadata struct {
...
+	for _, repo := range repos {
+		ref, err := name.ParseReference(repo, name.WeakValidation)
 		if err != nil {
 			continue
 		}
 		if reg == ref.Context().RegistryStr() {
-			return img, nil
+			ok, err := accessChecker.CheckReadAccess(repo, keychain)
+			if err != nil {
+				return ""
```
jabrown85, line 349: If an error occurs at `accessChecker.CheckReadAccess` - how will we be able to see the failure and get the information to the user?

c0d1ngm0nk3y: Currently - not at all. The next working mirror would be taken and if no run image could be found a generic error would be risen. The problem is that there might be several errors if there are several mirrors. Or even some fail and others works.

platform/files.go
c0d1ngm0nk3y: I don't think so. The only changed behaviour would be that a run image was chosen that was not readable (e.g. some registry, but different repo) and in this case another run image might now work. I would rather see it as improvement/bugfix. The check would make sure that the user receives an error or am I missing something? 

WDYT?

jabrown85: I think I'm ok with this. @natalieparellano may feel differently though 😄

natalieparellano: I agree, this should be safe to back-port to previous Platform APIs

platform/files.go
natalieparellano: It might make sense to log a warning here

platform/files.go
```diff
@@ -279,38 +283,79 @@ type RunImageMetadata struct {
 	Mirrors []string `toml:"mirrors" json:"mirrors,omitempty"`
 }
 
-func (rm *RunImageMetadata) BestRunImageMirror(registry string) (string, error) {
+//go:generate mockgen -package testmock -destination ../testmock/access_checker.go github.com/buildpacks/lifecycle/platform AccessChecker
+type AccessChecker interface {
+	CheckReadAccess(repo string, keychain authn.Keychain) (bool, error)
```
natalieparellano, line 288: Could we maybe extend `[ImageHandler](https://github.com/buildpacks/lifecycle/blob/b6af8b55150542b20f896bfff8ed8fd47a0e1bc5/image/handler.go)` to include this? This was added recently, if you pull in the latest main you should see it (apologies in advance for the merge conflict)

cmd/lifecycle/analyzer.go
```diff
@@ -66,6 +66,7 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
 		return cmd.FailErrCode(err, cmd.CodeForInvalidArgs, "parse arguments")
 	}
 	a.LifecycleInputs.OutputImageRef = args[0]
+	a.LifecycleInputs.AccessChecker = &platform.AccessCheckerImpl{}
```
natalieparellano, line 69: If we do https://github.com/buildpacks/lifecycle/pull/1024/files?show-viewed-files=true&file-filters%5B%5D=#r1124696772 we could pass the image handler in when we call `ResolveInputs()` rather than having the access checker live on this struct. This would require updating `LifecycleInputsOperation` to include an image handler, but that seems fine.

natalieparellano (changes requested):
@pbusko thanks for this! I think broadly this PR accomplishes what we want. I'm requesting changes to hopefully align the code with some recently introduced constructs (the image handler) and keep `platform.LifecycleInputs` scoped to only the things that a platform (operator) would provide. Feel free to offer other suggestions if you think there is a better way to do it.

platform/files.go
c0d1ngm0nk3y: > It might make sense to log a warning here

2 quick follow-up questions:

1. If a run-image matching the image's registry was found without read access, a warning sounds reasonable. Do you think we should also warn for "normal" run-mirrors? For example, for use one mirror would be located in the VPN, so this warning would be kind of normal when you are not in the VPN.
2. Is there a better way to pass around the `log.Logger` in each function call?

natalieparellano: > Do you think we should also warn for "normal" run-mirrors?

You're right, that might get noisy. Let's not warn in this case.

Re: passing around the logger, I'm not sure, we're not very consistent about this (sometimes the logger is embedded, sometimes it's just passed around). I'm happy to defer to whatever you think is best.

platform/files.go
c0d1ngm0nk3y: We tried, but the problem seems to be that creating an `ImageHandler` needs a `KeyChain` and the `Keychain` needs input from runMetadata which is not available before `resolve_inputs.go`.

cmd/lifecycle/analyzer.go
c0d1ngm0nk3y: The main reason for putting it to the struct in the first place was that it eases the testing because we can inject a mock. Our main concern was to adapt every `LifecycleInputsOperation` although only 2 are needing it.

platform/files.go
c0d1ngm0nk3y: > I'm happy to defer to whatever you think is best.

We had a closer look and the changes needed to pass around the logger seem not to be worth logging a warning in this edge case. This would be if the run image matches the registry of the image, the image can be written, but the run image can't be pulled. Right? WDYT?

platform/files.go
natalieparellano: I see... that is a bummer. The issue that I could see with where `CheckReadAccess` is called is that we'd do this check even when exporting to a daemon or layout path, and those targets will fail. So we'll want the `AccessChecker` interface to have different implementations for those targets with basically a no-op `CheckReadAccess`.

natalieparellano: Thinking more about the comment above, if we are going to check read access to images in the daemon (for local builds) then resolving the inputs must be done after the client is constructed (in `cmd.Privileges()`). This unfortunately requires the re-wiring of a bunch of stuff that I tried to illustrate in this [draft PR](https://github.com/buildpacks/lifecycle/pull/1031/files). Unfortunately I couldn't think of a better way to get the access checker than just passing it to every `LifecycleInputsOperation`. If you can think of a better way to do it, I am all for it.

c0d1ngm0nk3y: > The issue that I could see with where `CheckReadAccess` is called is that we'd do this check even when exporting to a daemon or layout path, and those targets will fail.

I have probably missed the point, but we are using `CheckReadAccess` "only" on the `run-image` and the `run-image-mirrors`. So shouldn't the export target have nothing to do with that?

c0d1ngm0nk3y: @natalieparellano The more I thought of this and tried things out, the more it felt that `AccessChecker` and `image.Handler` should not be the same thing. The `AccessHandler` is actually only used for run-images (always remote, right?)

I tried something different. Calling it `ImageStrategy` (better names are more than welcome!) since that what it basically is imho. We want just another way to verify readability of images in the unit tests.

I also got rid of the according `gomock` mocks and used a simple test implementation. Which is a huge gain in regards to readability of the unit tests.

WDYT?

Please have a look, I pushed it a a separate commit, so it would be easy to throw it away again.

platform/files.go
```diff
@@ -339,38 +343,78 @@ type RunImageForExport struct {
...
 }
 
-func (rm *RunImageForExport) BestRunImageMirror(registry string) (string, error) {
+type ImageStrategy interface {
+	CheckReadAccess(repo string, keychain authn.Keychain) (bool, error)
+}
+
+type RemoteImageStrategy struct{}
+
+var _ ImageStrategy = &RemoteImageStrategy{}
```
natalieparellano, line 352: Curious why we need this global

cmd/lifecycle/analyzer.go
```diff
@@ -67,6 +67,7 @@ func (a *analyzeCmd) Args(nargs int, args []string) error {
 		return cmd.FailErrCode(err, cmd.CodeForInvalidArgs, "parse arguments")
 	}
 	a.LifecycleInputs.OutputImageRef = args[0]
+	a.LifecycleInputs.AccessChecker = &platform.RemoteImageStrategy{}
```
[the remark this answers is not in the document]
natalieparellano: @pbusko just circling back here. Did you have any questions about the feedback or want to discuss anything further? I think this comment ^^ is the only thing truly blocking.

platform/files.go
c0d1ngm0nk3y: We don't :)

cmd/lifecycle/analyzer.go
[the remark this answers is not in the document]
c0d1ngm0nk3y: @natalieparellano Sorry for the delay. Could you have another look? We moved it to `platform.ResolveInputs` and checked for `UseLayout` and `UseDaemon`.

c0d1ngm0nk3y:
> I'll have a look at your changes soon.

@natalieparellano Did you have a chance to have a closer look? Is something still missing from your side?

natalieparellano:
@c0d1ngm0nk3y - many apologies, I missed your comment from 2 weeks ago prodding me to take another look 😞 I'll take a look today

platform/resolve_analyze_inputs_test.go
```diff
@@ -36,6 +36,7 @@ func testResolveAnalyzeInputs(platformAPI string) func(t *testing.T, when spec.G
 			inputs.OutputImageRef = "some-output-image" // satisfy validation
 			logHandler = memory.New()
 			logger = &log.Logger{Handler: logHandler}
+			inputs.UseDaemon = true //to prevent access checking of run images
```
natalieparellano, line 39:
```suggestion
			inputs.UseDaemon = true // to prevent access checking of run images
```
To fix the linter

natalieparellano:
Looks good @c0d1ngm0nk3y! I think you need to run `make lint`. The code looks solid - I'm going to follow up on the spec change at Working Group on Thursday (see [comment](https://github.com/buildpacks/spec/pull/357#issuecomment-1532096451)).

acceptance/analyzer_test.go
```diff
@@ -271,7 +271,7 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 				output, err := cmd.CombinedOutput()
 
 				h.AssertNotNil(t, err)
-				expected := "ensure registry read access to some-run-image-from-run-toml"
```
c0d1ngm0nk3y, line 274: The error the user might face changed, but  that shouldn't be an issue, right?

c0d1ngm0nk3y (commented):
The acceptance tests failed, so we had to make another small adjustment. Can you check and approve the workflow again?

acceptance/analyzer_test.go
natalieparellano: That is fine! The new error messages shows the feature is working as intended :)

acceptance/analyzer_test.go
```diff
@@ -47,8 +47,8 @@ func TestAnalyzer(t *testing.T) {
 	analyzeDaemonFixtures = analyzeTest.targetDaemon.fixtures
 	analyzeRegFixtures = analyzeTest.targetRegistry.fixtures
 
-	for _, platformAPI := range api.Platform.Supported {
-		spec.Run(t, "acceptance-analyzer/"+platformAPI.String(), testAnalyzerFunc(platformAPI.String()), spec.Parallel(), spec.Report(report.Terminal{}))
+	for _, platformAPI := range []string{"0.12"} {
```
natalieparellano, line 50:
```suggestion
	for _, platformAPI := range api.Platform.Supported {
```
Aye, I'm afraid we do need to run the analyzer tests for all supported platforms. I think we skip over older platforms in the extender acceptance tests, but that is only because the extender isn't supported on older platforms.

platform/resolve_inputs.go
```diff
@@ -178,11 +184,8 @@ func fillRunImageFromRunTOMLIfNeeded(i *LifecycleInputs, logger log.Logger) erro
 	if len(runMD.Images) == 0 {
 		return errors.New(ErrRunImageRequiredWhenNoRunMD)
 	}
-	i.RunImageRef, err = runMD.Images[0].BestRunImageMirror(targetRegistry)
-	if err != nil {
```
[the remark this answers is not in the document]
natalieparellano: Good catch!

natalieparellano:
@c0d1ngm0nk3y @pbusko apologies for the extended back and forth on this one - if you could just pull in https://github.com/buildpacks/lifecycle/pull/1024/files#r1188710899, if the tests all pass then we will merge this in.

c0d1ngm0nk3y:
@natalieparellano There was a small oversight in the test, I fixed that. Sorry. The tests should be fine now.

platform/resolve_inputs.go
```diff
@@ -20,6 +20,12 @@ var (
 )
 
 func ResolveInputs(phase LifecyclePhase, i *LifecycleInputs, logger log.Logger) error {
+	if i.UseDaemon || i.UseLayout {
```
jabrown85, line 23: Is `LocalimageStrategy` correct for Daemon run images in all cases?

I thought `pack build <blah> --publish` , for instance, would read the run image from the target registry to place the app layers onto without pulling it down.

natalieparellano: Ah yes, but in the `--publish` case `i.UseDaemon` would be false

c0d1ngm0nk3y: According to [this comment](https://github.com/buildpacks/lifecycle/pull/1024#issuecomment-1476360242)

> If the export target is the daemon, then we also expect to find the run image in a daemon. We don't allow "mixing and matching" although maybe (someday) we should.

But does `--publish` not imply that you export to a registry and not the daemon?

jabrown85: As long as `i.UseDaemon` is false I think we are good.

natalieparellano:
Green! Thanks for the PR @c0d1ngm0nk3y and @pbusko!

natalieparellano (approved):

c0d1ngm0nk3y:
@natalieparellano We would really like to consume this change. Do you know what the current plans for release `0.17.0` are?

natalieparellano:
@c0d1ngm0nk3y we are getting really close - at this point, we need the specs to be finalized and shipped before we can ship the lifecycle. You can track the progress here: https://github.com/buildpacks/lifecycle/issues/1165

At this point, I am hoping for a lifecycle 0.17.0-rc.4 early next week, and a lifecycle 0.17.0 late next week. I have a record of being overly optimistic, but I think that's achievable 🙏🏼 🤞🏼

natalieparellano:
> I am hoping for a lifecycle 0.17.0-rc.4 early next week

I missed the mark by a couple of days, but `0.17.0-rc.4` is out now. If it's possible to try it out and verify that things are still working fine that will help us get the "real" one out sooner.

PR #1030, opened 2023-03-09
Pull Os and Arch from RunImage; handle * Stacks

joe-kimmel-vmw:
for https://github.com/buildpacks/lifecycle/issues/742

### DIFF
--- a/acceptance/phase_test.go
+++ b/acceptance/phase_test.go
@@ -13,6 +13,7 @@ import (
 	"runtime"
 	"strings"
 	"testing"
+	"time"
 
 	ih "github.com/buildpacks/imgutil/testhelpers"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -398,4 +399,20 @@ func SBOMExtensions() []string {
 
 func SBOMComponents() []string {
 	return []string{"lifecycle", "launcher"}
+}
+
+func assertImageOSAndArch(t *testing.T, imageName string, phaseTest *PhaseTest) { //nolint  - these functions are in fact used, i promise
+	inspect, _, err := h.DockerCli(t).ImageInspectWithRaw(context.TODO(), imageName)
+	h.AssertNil(t, err)
+	h.AssertEq(t, inspect.Os, phaseTest.targetDaemon.os)
+	h.AssertEq(t, inspect.Architecture, phaseTest.targetDaemon.arch)
+	fmt.Println("asserted that image has OS:", phaseTest.targetDaemon.os, "and arch: ", phaseTest.targetDaemon.arch)
+}
+
+func assertImageOSAndArchAndCreatedAt(t *testing.T, imageName string, phaseTest *PhaseTest, expectedCreatedAt time.Time) { //nolint
+	inspect, _, err := h.DockerCli(t).ImageInspectWithRaw(context.TODO(), imageName)
+	h.AssertNil(t, err)
+	h.AssertEq(t, inspect.Os, phaseTest.targetDaemon.os)
+	h.AssertEq(t, inspect.Architecture, phaseTest.targetDaemon.arch)
+	h.AssertEq(t, inspect.Created, expectedCreatedAt.Format(time.RFC3339))
 }
--- a/acceptance/phase_test.go
+++ b/acceptance/phase_test.go
@@ -406,7 +406,6 @@ func assertImageOSAndArch(t *testing.T, imageName string, phaseTest *PhaseTest) 
 	h.AssertNil(t, err)
 	h.AssertEq(t, inspect.Os, phaseTest.targetDaemon.os)
 	h.AssertEq(t, inspect.Architecture, phaseTest.targetDaemon.arch)
-	fmt.Println("asserted that image has OS:", phaseTest.targetDaemon.os, "and arch: ", phaseTest.targetDaemon.arch)
 }
 
 func assertImageOSAndArchAndCreatedAt(t *testing.T, imageName string, phaseTest *PhaseTest, expectedCreatedAt time.Time) { //nolint
--- a/analyzer.go
+++ b/analyzer.go
@@ -219,9 +219,11 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 		if err != nil {
 			return platform.AnalyzedMetadata{}, errors.Wrap(err, "identifying run image")
 		}
-		atm, err = platform.GetTargetFromImage(a.RunImage)
-		if err != nil {
-			return platform.AnalyzedMetadata{}, errors.Wrap(err, "unpacking metadata from image")
+		if a.PlatformAPI.AtLeast("0.12") {
+			atm, err = platform.GetTargetFromImage(a.RunImage)
+			if err != nil {
+				return platform.AnalyzedMetadata{}, errors.Wrap(err, "unpacking metadata from image")
+			}
 		}
 	}
 
--- a/analyzer.go
+++ b/analyzer.go
@@ -272,7 +272,10 @@ func (a *Analyzer) getTargetFromImage(image imgutil.Image) (*platform.TargetMeta
 	if err != nil {
 		return &tm, err
 	}
-	// TODO: arch variant
+	tm.ArchVariant, err = image.Variant()
+	if err != nil {
+		return &tm, err
+	}
 	labels, err := image.Labels()
 	if err != nil {
 		return &tm, err
--- a/analyzer.go
+++ b/analyzer.go
@@ -213,10 +213,15 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 		appMeta = platform.LayersMetadata{}
 	}
 
+	var atm *platform.TargetMetadata
 	if a.RunImage != nil {
 		runImageIDReference, err = a.getImageIdentifier(a.RunImage)
 		if err != nil {
 			return platform.AnalyzedMetadata{}, errors.Wrap(err, "identifying run image")
+		}
+		atm, err = a.getTargetFromImage(a.RunImage)
+		if err != nil {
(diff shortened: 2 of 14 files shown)


analyzer.go
```diff
@@ -252,6 +257,41 @@ func (a *Analyzer) getImageIdentifier(image imgutil.Image) (string, error) {
 	return identifier.String(), nil
 }
 
+func (a *Analyzer) getTargetFromImage(image imgutil.Image) (*platform.TargetMetadata, error) {
```
natalieparellano, line 260: I need to use this in the restorer also, so could we move this to the platform package?

buildpack/bp_descriptor.go
```diff
@@ -24,9 +24,10 @@ type StackMetadata struct {
 }
 
 type TargetPartial struct {
-	OS          string `json:"os" toml:"os"`
 	Arch        string `json:"arch" toml:"arch"`
 	ArchVariant string `json:"arch-variant" toml:"arch-variant"`
+	ID          string `json:"id" toml:"id"`
```
natalieparellano, line 29:
```suggestion
```
Actually this isn't part of the buildpack spec - we provide the run image's target id in `CNB_TARGET_ID` but it isn't part of buildpack.toml for the purposes of determining which run images the buildpack is compatible with.

analyzer.go
```diff
@@ -213,11 +213,16 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
 		appMeta = platform.LayersMetadata{}
 	}
 
+	var atm *platform.TargetMetadata
 	if a.RunImage != nil {
 		runImageRef, err = a.getImageIdentifier(a.RunImage)
 		if err != nil {
 			return platform.AnalyzedMetadata{}, errors.Wrap(err, "identifying run image")
 		}
+		atm, err = a.getTargetFromImage(a.RunImage)
```
natalieparellano, line 222: I think we want to guard this on the platform API being at least 0.12

analyzer.go
```diff
@@ -252,6 +257,41 @@ func (a *Analyzer) getImageIdentifier(image imgutil.Image) (string, error) {
...
+	if err != nil {
+		return &tm, err
+	}
+	labels, err := image.Labels()
+	if err != nil {
+		return &tm, err
+	}
+	distName, distNameExists := labels["io.buildpacks.distribution.name"]
+	distVersion, distVersionExists := labels["io.buildpacks.distribution.version"]
+	if distNameExists || distVersionExists {
```
natalieparellano, line 285: Nice check :) I forgot about this in my restorer PR

buildpack/bp_descriptor.go
joe-kimmel-vmw: 💡

analyzer.go
joe-kimmel-vmw: ✅

acceptance/phase_test.go
```diff
@@ -399,3 +400,19 @@ func SBOMExtensions() []string {
 func SBOMComponents() []string {
 	return []string{"lifecycle", "launcher"}
 }
+
+func assertImageOSAndArch(t *testing.T, imageName string, phaseTest *PhaseTest) { //nolint  - these functions are in fact used, i promise
+	inspect, _, err := h.DockerCli(t).ImageInspectWithRaw(context.TODO(), imageName)
+	h.AssertNil(t, err)
+	h.AssertEq(t, inspect.Os, phaseTest.targetDaemon.os)
+	h.AssertEq(t, inspect.Architecture, phaseTest.targetDaemon.arch)
+	fmt.Println("asserted that image has OS:", phaseTest.targetDaemon.os, "and arch: ", phaseTest.targetDaemon.arch)
```
jabrown85, line 409: Did you mean to leave this in? I'm fine with it staying but wanted to ask.

buildpack/bp_descriptor.go
```diff
@@ -83,7 +86,7 @@ func ReadBpDescriptor(path string) (*BpDescriptor, error) {
 					descriptor.Targets = append(descriptor.Targets, TargetMetadata{TargetPartial: TargetPartial{OS: "windows", Arch: "amd64"}})
 				}
 				if fname == "build" {
-					descriptor.Targets = append(descriptor.Targets, TargetMetadata{TargetPartial: TargetPartial{OS: "linux", Arch: "amd64"}})
+					descriptor.Targets = append(descriptor.Targets, TargetMetadata{TargetPartial: TargetPartial{OS: "linux", Arch: "amd64"}}) // Question: why is the arch part of this inference -- should we leave the arch empty when we infer the OS?
```
jabrown85, line 89: That's a good question - I don't think we should set anything besides the `OS` if we can. Otherwise a buildpack that could work across arm/amd64 would not work on an arm image unless they specifically added the targets. Right?

jabrown85 (approved):

acceptance/phase_test.go
joe-kimmel-vmw: :) ooops!  good catch - yeah agree it's harmless but I'll delete it,   was leftover from some debugging...

buildpack/bp_descriptor.go
joe-kimmel-vmw: Thanks - that's exactly what i mean!   I think... ok, here's the tension. 

[our spec](https://github.com/buildpacks/spec/pull/335/files#diff-e603760990971da3f77be4bb8d77c3405098f006814fd8c054d2d15f395b8330R193) says we should abide by the OCI spec .  

the OCI spec says that arch values SHOULD conform to valid GOARCH values, 

which are:
> Choices for $GOARCH are amd64 (64-bit x86, the most mature port), 386 (32-bit x86), arm (32-bit ARM), arm64 (64-bit ARM), ppc64le (PowerPC 64-bit, little-endian), ppc64 (PowerPC 64-bit, big-endian), mips64le (MIPS 64-bit, little-endian), mips64 (MIPS 64-bit, big-endian), mipsle (MIPS 32-bit, little-endian), mips (MIPS 32-bit, big-endian), s390x (IBM System z 64-bit, big-endian), and wasm (WebAssembly 32-bit).


hmmm so i guess we're already diverging a little bit here by adding the handling of `*` stacks - honestly this might be another reasonable place to put a `*` though - in our case these could be e.g. just shell scripts, which would be expected to execute on any architecture,  right?  so it feels reasonable to keep using a wildcard,  unless that means we're going to eventually output something that some OCI registry will refuse to accept?    but the alternative would be that buildpack authors would have to maintain two identical things that differ only by specifying arm/x86?

Like we're missing a generic value for `arch` (or "generics" for the DSL describing arch) because it's a non-sequitor from the perspective of the go compiler, and we're a few levels deep into people saying "let's just use whatever is working for the layer below us" ,  right?

jabrown85: Yeah, I think I like wildcard myself. The bash/executable that runs everywhere is far more likely for our project. 🤷

joe-kimmel-vmw: cool - updated here and left a comment in the spec PR: https://github.com/buildpacks/spec/pull/335/files#r1137876951

detector.go
```diff
@@ -223,10 +232,11 @@ func (d *Detector) detectGroup(group buildpack.Group, done []buildpack.GroupElem
 
 			if d.PlatformAPI.AtLeast("0.12") {
 				targetMatch := false
-				if isWildcard(d.AnalyzeMD.RunImageTarget()) {
+				if isWildcard(d.AnalyzeMD.RunImageTarget()) || hasWildcard(bpDescriptor.Targets) {
 					targetMatch = true
 				} else {
 					for i := range bpDescriptor.Targets {
+						fmt.Println("Checking for match against descriptor:", bpDescriptor.Targets[i])
```
natalieparellano, line 239:
```suggestion
						d.Logger.Debugf("Checking for match against descriptor:", bpDescriptor.Targets[i])
```
(this could be helpful as a debug message)

detector.go
```diff
@@ -192,6 +192,15 @@ func isWildcard(t platform.TargetMetadata) bool {
 	return t.Arch == "" && t.OS == ""
 }
 
+func hasWildcard(ts []buildpack.TargetMetadata) bool {
```
natalieparellano, line 195: Is it possible to add a test for this case?

natalieparellano (commented):
This is looking good @joe-kimmel-vmw :) I added a couple comments

buildpack/bp_descriptor_test.go
```diff
@@ -118,11 +118,11 @@ func testBpDescriptor(t *testing.T, when spec.G, it spec.S) {
 			h.AssertEq(t, descriptor.Stacks[0].ID, "some.non-magic.value")
 			h.AssertEq(t, len(descriptor.Targets), 1)
 			h.AssertEq(t, len(descriptor.Targets), 1)
-			h.AssertEq(t, descriptor.Targets[0].Arch, "amd64")
+			h.AssertEq(t, descriptor.Targets[0].Arch, "*")
 			h.AssertEq(t, descriptor.Targets[0].OS, "linux")
 			h.AssertEq(t, len(descriptor.Targets[0].Distributions), 0)
 		})
-		it("detects both windows/amd64 if batch files are present and linux", func() {
+		it("detects both windows/* if batch files are present and linux", func() {
```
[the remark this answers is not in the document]
joe-kimmel-vmw: sorry, this might be a bad test-case name.  the test case has files with no extension (`build`) for linux as well as files with .bat extension (`build.bat`) for windows.  does that help?

detector.go
joe-kimmel-vmw: I was born to add a test for this case!

buildpack/bp_descriptor_test.go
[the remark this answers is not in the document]
joe-kimmel-vmw: we figured it out!

natalieparellano (approved):

PR #1031, opened 2023-03-09
WIP: make image handler an access checker

natalieparellano:
To facilitate conversation around https://github.com/buildpacks/lifecycle/pull/1024

If we want to check access to images in a daemon, we need to have the client instantiated already (it is a singleton) so resolving the inputs must be done after dropping privileges.


natalieparellano:
In general the fact that we have to make so many code changes to do a relatively simple thing is an indicator that our cmd/platform package still needs a refactor

natalieparellano:
Closing in favor of other implementation

Issue #1000, opened 2023-02-02
[RFC #0105] Rebaser changes for run image extension (Dockerfiles phase 3)


natalieparellano:
Spec PR: https://github.com/buildpacks/spec/pull/347/files

For newer platform API (`0.12` and above), the `rebaser` should require `-force` if the image has the label `io.buildpacks.rebasable` set to `false`.

Additionally:

```
- **If** `<run-image>` is not provided by the platform, the value will be [resolved](#run-image-resolution) from the contents of the `runImage` key in the `io.buildpacks.lifecycle.metdata` label on `<image>`
```

Thus instead of `stack.runImage.image` we use `runImage.image` on the label (note the schema change in the spec PR).

PR #1032, opened 2023-03-09
Rebaser changes for run image extension

jabrown85:
- [Add --force flag to rebase command](https://github.com/buildpacks/lifecycle/commit/47c62034cbb1bf67ccbaf83705609ae1b2bef36c)
- [Validate io.buildpacks.rebasable label on rebase](https://github.com/buildpacks/lifecycle/commit/bfd3344f5c99b3475496eb59ca0954b16deb3298)
- [Read rebase from io.buildpacks.lifecycle.metdata runImage](https://github.com/buildpacks/lifecycle/commit/aa42fa084e34d05798bf68f979a6c68676d00f43)

Open Questions
1. Should the `runImage` not be present, should we be failing? The spec doesn't say it allows for the fallback to the old `stack` key..but it feels aggressive to me.
2. The spec says specifically to fail is the value of rebaseable is `false`. I coded it as such, but anything other than `false` is treated as true. This may be unexpected.

Ref: https://github.com/buildpacks/lifecycle/issues/1000


natalieparellano:
> Should the runImage not be present, should we be failing? The spec doesn't say it allows for the fallback to the old stack key..but it feels aggressive to me.

I think we should fall through to the stack key. Otherwise it would be hard to rebase old apps that were created before the platform was upgraded.

jabrown85:
> > Should the runImage not be present, should we be failing? The spec doesn't say it allows for the fallback to the old stack key..but it feels aggressive to me.
> 
> I think we should fall through to the stack key. Otherwise it would be hard to rebase old apps that were created before the platform was upgraded.

I agree. Do we need a spec change then? Rebase spec doesn't say this will happen.

natalieparellano:
@jabrown85 this one looks fully updated - should we merge it?

jabrown85:
I'm happy if you are @natalieparellano! No one has reviewed it officially so go right ahead 😄

natalieparellano (approved):
Looks great!

PR #1037, opened 2023-03-17
Replace deprecated GHA ::set-output syntax

phil9909:
Use the new way of producing outputs in a github action as the old way is deprecated and will be removed (31st May 2023).

See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


jabrown85 (approved):

natalieparellano:
Thanks for the PR @phil9909!

Issue #914, opened 2022-09-22
FIXMEs in the code


natalieparellano:
### Description
There are a number of (mostly) small refactors denoted with `FIXME` in the lifecycle code. It would be awesome if we could go through and either tackle them or create an issue for them.

anushkamittal20:
Hey @natalieparellano, I would like to make my first contribution to buildpacks community with this issue. I understand this might be a great issue to help understand the project and get familiar with the codebase. I see there are 18 "FIXMEs" that can be worked on, it would be great to discuss and work on them.

natalieparellano:
Welcome @anushkamittal20! We'd be grateful to have your help here :) I'd be happy to discuss further any FIXMEs you'd like to tackle. (It wouldn't be necessary to address all of them in a single PR, in fact it may be easiest to create one PR per FIXME or related set.) We can continue the conversation here, or if you prefer more real-time support, feel free to reach out in #buildpacks-implementation in [CNCF Slack](slack.cncf.io).

anushkamittal20:
Thank you @natalieparellano! 
I was thinking we could go through all the FIXMEs and maybe document what we can do about them, with options being tackling them and creating a separate issue, and then I could pick a few to tackle here.
Would love to know what you think.
I will continue this conversation over the slack channel as you suggested!
Again thank you!

satyazzz123:
@natalieparellano I have raised a PR . please have a look . Thank you

PR #1038, opened 2023-03-19
FIXME: pass the logger and print the warning inside

anushkamittal20:
References: [issue](https://github.com/buildpacks/lifecycle/issues/914)
[slack thread](https://cloud-native.slack.com/archives/C0331B5QS02/p1678133031976489)

I have made changes to remove the message returned from `DecodeLayerMetadataFile` and the consequent changes that were required.


anushkamittal20:
@natalieparellano I started working on this issue, and I was hoping to get suggestions on how to [tackle this.](https://github.com/buildpacks/lifecycle/pull/1038/files#diff-62e865141bd023a8b34c82ff14ff46ab061b28a7a9990d85b07c3297b9924442L155)
Thank you very much!

buildpack/layermetadata.go
```diff
@@ -34,23 +35,25 @@ func EncodeLayerMetadataFile(lmf LayerMetadataFile, path, buildpackAPI string) e
...
 	}
 	defer fh.Close()
 
 	decoders := supportedEncoderDecoders()
 
 	for _, decoder := range decoders {
 		if decoder.IsSupported(buildpackAPI) {
-			return decoder.Decode(path)
+			lmf, str, err := decoder.Decode(path)
+			logger.Warnf(str)
```
natalieparellano, line 52: We can check here that the string is not empty

natalieparellano (commented):
@anushkamittal20 this looks great! Apologies for the slow reply (I've been out). You are definitely on the right track! It would be good to add a couple tests in layermetadata_test.go. You can construct a "spy" logger for this, if you search for `assertLogEntry` you'll find where we do something similar elsewhere.

buildpack/layermetadata.go
anushkamittal20: Got it!

jjbustamante:
@anushkamittal20 I think you forgot to run `make format` before pushing your code and the CI failed, just do that, run `make format` in your local environment and then push the changes.

natalieparellano:
@anushkamittal20 I **think** (though I'm not sure) that the logic you quoted should move inside `DecodeLayerMetadataFile` around where we call `decoder.Decode` in that function. We have the buildpack API at that point, so we can log or return an error depending on the version.

anushkamittal20:
@jjbustamante, I will run `make format` and push it. Thank you for letting me know!
@natalieparellano, I will try to move the logic in `DecodeLayerMetadataFile` and see if that works all right, thank you for your feedback. That was helpful!
P.S. Sorry for the delay, I will implement these changes and push them to the branch.

natalieparellano:
Looks like it needs another `make format`

natalieparellano:
@anushkamittal20 any update on this one?

anushkamittal20:
@natalieparellano wanted to update you on this.
I was facing dependency issues on my OS and wasn't able to run `make format` I am currently trying to resolve the issues or work around them. Will push the new changes ASAP.

buildpack/build.go
```diff
@@ -164,24 +164,24 @@ func runBuildCmd(d BpDescriptor, bpLayersDir, planPath string, inputs BuildInput
...
 		return eachLayer(layersDir, d.WithAPI, func(path, buildpackAPI string) (LayerMetadataFile, error) {
-			layerMetadataFile, msg, err := DecodeLayerMetadataFile(path+".toml", buildpackAPI)
+			layerMetadataFile, err := DecodeLayerMetadataFile(path+".toml", buildpackAPI, logger)
 			if err != nil {
 				return LayerMetadataFile{}, err
 			}
-			if msg != "" {
-				logger.Warn(msg)
-			}
+			// if msg != "" {
```
jjbustamante, line 171: @anushkamittal20 Could you just remove these lines?

buildpack/build.go
```diff
@@ -164,24 +164,24 @@ func runBuildCmd(d BpDescriptor, bpLayersDir, planPath string, inputs BuildInput
...
 	return eachLayer(layersDir, d.WithAPI, func(path, buildpackAPI string) (LayerMetadataFile, error) {
-		layerMetadataFile, msg, err := DecodeLayerMetadataFile(path+".toml", buildpackAPI)
+		layerMetadataFile, err := DecodeLayerMetadataFile(path+".toml", buildpackAPI, logger)
 		if err != nil {
 			return LayerMetadataFile{}, err
 		}
-		if msg != "" {
-			return LayerMetadataFile{}, errors.New(msg)
-		}
+		// if msg != "" {
```
jjbustamante, line 182: Also these one, you can just remove them

buildpack/layermetadata.go
```diff
@@ -34,23 +36,30 @@ func EncodeLayerMetadataFile(lmf LayerMetadataFile, path, buildpackAPI string) e
...
 
 	decoders := supportedEncoderDecoders()
 
 	for _, decoder := range decoders {
 		if decoder.IsSupported(buildpackAPI) {
-			return decoder.Decode(path)
+			lmf, str, err := decoder.Decode(path)
+			if str != "" {
+				logger.Warn(str)
+				if !api.MustParse(buildpackAPI).LessThan("0.6") {
```
natalieparellano, line 55:
```suggestion
				if api.MustParse(buildpackAPI).AtLeast("0.7") {
```
This reads a bit clearer

buildpack/layermetadata.go
```diff
@@ -34,23 +36,30 @@ func EncodeLayerMetadataFile(lmf LayerMetadataFile, path, buildpackAPI string) e
...
+					return LayerMetadataFile{}, errors.New(str)
+				}
```
natalieparellano, line 57:
```suggestion
				if api.MustParse(buildpackAPI).LessThan("0.6") {
					logger.Warn(str)
				} else {
					return LayerMetadataFile{}, errors.New(str)
				}
```
I think this is what we want... the test failures may be due to a test needing to be updated

buildpack/layermetadata_test.go
```diff
@@ -33,12 +45,23 @@ func testLayerMetadata(t *testing.T, when spec.G, it spec.S) {
...
 			var lmf buildpack.LayerMetadataFile
-			lmf, s, err := buildpack.DecodeLayerMetadataFile(metadataFile.Name(), "0.9")
+			lmf, err = buildpack.DecodeLayerMetadataFile(metadataFile.Name(), "0.9", logger)
 			h.AssertNil(t, err)
-			h.AssertEq(t, s, "")
 			h.AssertEq(t, lmf.Cache, true)
 			h.AssertEq(t, lmf.Build, false)
 			h.AssertEq(t, lmf.Launch, false)
 		})
+		it("metadata file in wrong format", func() {
```
natalieparellano, line 54:
```suggestion
		it("logs a warning when the metadata file has wrong format (on older apis)", func() {
```

buildpack/layermetadata_test.go
```diff
@@ -33,12 +45,23 @@ func testLayerMetadata(t *testing.T, when spec.G, it spec.S) {
...
+			err := os.WriteFile(metadataFile.Name(), []byte("cache = true"), 0400)
+			h.AssertNil(t, err)
+
+			var lmf buildpack.LayerMetadataFile
+			lmf, _ = buildpack.DecodeLayerMetadataFile(metadataFile.Name(), "0.9", logger)
```
natalieparellano, line 59:
```suggestion
			lmf, err := buildpack.DecodeLayerMetadataFile(metadataFile.Name(), "0.5", logger)
			h.AssertNil(t, err)
```

buildpack/layermetadata_test.go
natalieparellano: We can make another `it` block for `returns an error when the metadata file has wrong format` that provides a newer buildpack API on the equivalent of line 59

natalieparellano:
@anushkamittal20 thanks for taking another look at this. I think it's on the right track, we just need two different test cases for old and new buildpack API.

anushkamittal20:
Hey @natalieparellano, thank you very much for your review.
I will work on the suggested changes and get back to you!

buildpack/layermetadata_test.go
```diff
@@ -33,12 +45,34 @@ func testLayerMetadata(t *testing.T, when spec.G, it spec.S) {
...
-			lmf, s, err := buildpack.DecodeLayerMetadataFile(metadataFile.Name(), "0.9")
+			lmf, err = buildpack.DecodeLayerMetadataFile(metadataFile.Name(), "0.9", logger)
 			h.AssertNil(t, err)
-			h.AssertEq(t, s, "")
 			h.AssertEq(t, lmf.Cache, true)
 			h.AssertEq(t, lmf.Build, false)
 			h.AssertEq(t, lmf.Launch, false)
 		})
+		it("logs a warning when the metadata file has wrong format (on older apis)", func() {
+			err := os.WriteFile(metadataFile.Name(), []byte("cache = true"), 0400)
```
natalieparellano, line 55:
```suggestion
			err := os.WriteFile(metadataFile.Name(), []byte("[types]\ncache = true"), 0400)
```
I **think** this will solve it

anushkamittal20: Thank you very much @natalieparellano, this fixes the failing test. The tests look good now and the changes should be valid.

PR #1050, opened 2023-03-29
detect linux flavors by reading etc/os-release

joe-kimmel-vmw:


internal/fsutil/os_detection.go
```diff
@@ -0,0 +1,55 @@
...
+	"os"
+	"strings"
+)
+
+type OSInfo struct {
+	Name    string
+	Version string
+}
+
+type Detector interface {
```
joe-kimmel-vmw, line 13: iirc this breaks one of those golang dictates of like, "define the interface where you need it, not alongside the first struct that implements it" but on the other hand it's not like the golang core team is going to personally slap me in the face for this infraction.   lmk if we have standards i should follow...

internal/fsutil/os_detection.go
```diff
@@ -0,0 +1,55 @@
...
+func (d *Detect) HasLinuxFile() bool {
+	finfo, err := os.Stat("/etc/os-release")
+	if err != nil {
+		return false
+	}
+	return !finfo.IsDir() && finfo.Size() > 0
+}
+
+func (d *Detect) ReadLinuxFile() (string, error) {
+	nobodyLikesBytes, err := os.ReadFile("/etc/os-release")
```
jabrown85, line 31:
```suggestion
	return io.ReadAll("/etc/os-release")
```

platform/files_test.go
```diff
@@ -266,3 +273,51 @@ func testFiles(t *testing.T, when spec.G, it spec.S) {
...
+			tm := platform.TargetMetadata{}
+			d := mockDetector{contents: "this is just test contents really",
+				t:       t,
+				HasFile: true}
+			platform.PopulateTargetOSFromFileSystem(&d, &tm, logr)
+			h.AssertEq(t, "opensesame", tm.Distribution.Name)
+			h.AssertEq(t, "3.14", tm.Distribution.Version)
+
+			// now pretend we're on windows
+			d.HasFile = false
```
jabrown85, line 311: Maybe these 3 cases should be in 3 `it` blocks?

jabrown85 (approved):

internal/fsutil/os_detection.go
joe-kimmel-vmw: lol sorry- i mean to fix that variable name,  but `ReadAll` looks like the better solution all around - thanks!

joe-kimmel-vmw: hmmm  : 

> cannot use io.ReadAll("/etc/os-release") (value of type []byte) as string value in return statement

I stand by my claim that nobody likes bytes.

internal/fsutil/os_detection.go
natalieparellano: Why do we need the interface? It looks like it only has one implementation. If we wanted we could make a no-op implementation for not-linux, then `HasLinuxFile()` could go away.

platform/files_test.go
```diff
@@ -266,3 +273,59 @@ func testFiles(t *testing.T, when spec.G, it spec.S) {
...
+		Version: "3.14",
+	}
+}
+
+func testPopulateTarget(t *testing.T, when spec.G, it spec.S) {
+	when("the data is available", func() {
+		it("populates appropriately", func() {
+			logr := &log.Logger{Handler: memory.New()}
+			tm := platform.TargetMetadata{}
+			d := mockDetector{contents: "this is just test contents really",
```
natalieparellano, line 303: Oh. I see... here's the other implementation.

natalieparellano (approved):

internal/fsutil/os_detection.go
jabrown85: eh, my fault. I think `ioutil` used to have something that read it into a string. I may be confusing that with another function.

internal/fsutil/os_detection.go
```diff
@@ -0,0 +1,55 @@
...
+type Detector interface {
+	HasLinuxFile() bool
+	ReadLinuxFile() (string, error)
+	GetInfo(osReleaseContents string) OSInfo
+}
+
+type Detect struct {
+}
+
+func (d *Detect) HasLinuxFile() bool {
```
natalieparellano, line 22:
```suggestion
func (d *Detect) HasSystemdFile() bool {
```
As you've educated me on the finer points here, maybe this will help others understand why we didn't go with the suggestion mentioned [here](https://github.com/buildpacks/lifecycle/pull/1050/files#r1153496828)

PR #1053, opened 2023-03-29
wildcard arch and os support: oops we missed a spot

joe-kimmel-vmw:


jabrown85 (approved):

natalieparellano:
Sorry for the ARM troubles... I'm trying to rescue the runner

natalieparellano:
@joe-kimmel-vmw could you set the DCO to pass?

PR #1055, opened 2023-03-30
Exporter changes for run image extension

### LINKED ISSUE #999: [RFC #0105] Exporter changes for run image extension (Dockerfiles phase 3)
Spec PR: https://github.com/buildpacks/spec/pull/347/files

For newer platform API (`0.12` and above), the `exporter` accepts `-run` instead of `-stack` and a new `-extended` flag. `-extended` is the location of the directory with extended run image layers. `-run` is the location of a `run.toml` file with schema:

```
[[image]]
 image = "<image>"
 mirrors = ["<mirror1>", "<mirror2>"]
```

When constructing the app image:

```
    - **If** image extensions were used to extend the `run-image` in [`analyzed.toml`](#analyzedtoml-toml):
      - MUST be based on the image in `<extended>`
    - **Else**:
      - MUST be based on the `run-image` in [`analyzed.toml`](#analyzedtoml-toml)
```

When finding run image mirrors, the `exporter` should use the appropriate value in `run.toml`.

The io.buildpacks.lifecycle.metadata label should be serialized according to the new schema defined in the spec PR.

natalieparellano:
Fixes #999 

Also fixes some issues with other phases found during demo testing


acceptance/extender_test.go
```diff
@@ -40,8 +40,12 @@ var (
 )
 
 const (
-	msgPullingRemoteImage          = "Did not find cache key, pulling remote image"
```
natalieparellano, line 43: I don't think this message actually ever existed in kaniko, only in my local copy 🙈

acceptance/phase_test.go
```diff
@@ -415,3 +415,64 @@ func assertImageOSAndArchAndCreatedAt(t *testing.T, imageName string, phaseTest
...
+	h.AssertEq(t, len(contents) > 0, true)
+
+	var runMD platform.RunMetadata
+	_, err = toml.Decode(string(contents), &runMD)
+	h.AssertNil(t, err)
+
+	return &runMD
+}
+
+func updateTOMLFixturesWithTestRegistry(t *testing.T, phaseTest *PhaseTest) { //nolint
```
natalieparellano, line 431: The setup for acceptance is getting a little unwieldy. We might try just mounting in the `/layers` directory instead of building it into the test container (which would actually be more realistic). But that is a bigger overhaul than I'd like to do now

internal/extend/kaniko/dockerfile_applier.go
```diff
@@ -21,11 +21,22 @@ const (
 )
 
 var (
-	kanikoBaseCacheDir  = filepath.Join(kanikoDir, "cache", "base")
```
natalieparellano, line 24: This was never actually used in a successful code path

tools/test-fork.sh
```diff
@@ -28,33 +28,36 @@ if [[ $reponame != "lifecycle" ]]; then
...
 
 if [[ -z $3 ]]; then
-  echo "Remove all tests to make things faster"
+  echo "Removing all tests to make things faster"
   sed -i '' "s/make test/echo test/g" .github/workflows/*.yml
   sed -i '' "s/make acceptance/echo acceptance/g" .github/workflows/*.yml
   echo "$(sed '/pack-acceptance/,$d' .github/workflows/build.yml)" > .github/workflows/build.yml
+  echo "Removing Codecov"
+  sed -i '' "/- name: Upload coverage to Codecov/,+6d" .github/workflows/build.yml
+  sed -i '' "/- name: Prepare Codecov/,+6d" .github/workflows/build.yml
```
natalieparellano, line 59: This and the hint about `COSIGN_*` secret names is the most useful change

buildpack/dockerfile.go
```diff
@@ -32,13 +32,22 @@ const (
...
 	Kind        string
 	Path        string
-	NewBase     string
+	// WithBase if populated indicates that the Dockerfile switches the image base to the provided value.
+	// If WithBase is empty, Extend should be true, otherwise there is nothing for the Dockerfile to do.
+	// However if WithBase is populated, Extend may be true or false.
+	WithBase string
+	// Extend if true indicates that the Dockerfile contains image modifications
+	// and if false indicates that the Dockerfile only switches the image base.
+	// If Extend is false, WithBase should be empty, otherwise there is nothing for the Dockerfile to do.
```
natalieparellano, line 47:
```suggestion
	// If Extend is false, WithBase should be populated, otherwise there is nothing for the Dockerfile to do.
```

buildpack/dockerfile.go
```diff
@@ -32,13 +32,22 @@ const (
...
 
 type DockerfileInfo struct {
 	ExtensionID string
 	Kind        string
 	Path        string
-	NewBase     string
+	// WithBase if populated indicates that the Dockerfile switches the image base to the provided value.
+	// If WithBase is empty, Extend should be true, otherwise there is nothing for the Dockerfile to do.
+	// However if WithBase is populated, Extend may be true or false.
+	WithBase string
```
natalieparellano, line 44: For some reason `WithBase` seemed more clear than `NewBase`. But we can change it back if that's not true

generator.go
```diff
@@ -148,7 +148,7 @@ func (g *Generator) Generate() (GenerateResult, error) {
 	}
 
 	g.Logger.Debug("Checking for new run image")
-	base, newBaseIdx, extend := g.checkNewRunImage(dockerfiles)
+	base, extend := g.checkNewRunImage(dockerfiles)
```
natalieparellano, line 151:
```suggestion
	base, extend := g.checkRunImageAfter(dockerfiles)
```

generator.go
```diff
@@ -181,6 +182,9 @@ func (g *Generator) Generate() (GenerateResult, error) {
 }
 
 func satisfies(images []platform.RunImageForExport, imageName string) bool {
+	if imageName == "" {
+		return true
+	}
```
natalieparellano, line 187: This covers the case where we didn't switch the run image, so the original run image still satisfies the constraints. But we might be able to move this check to a place where its meaning is more obvious

generator_test.go
```diff
@@ -124,14 +124,18 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
 		stdout, stderr *bytes.Buffer
 	)
 
+	// We need to create the temp directory before `it.Before` blocks because we want the variable to be populated
+	// when we set up the table tests.
```
natalieparellano, line 128: `spec` nuances

exporter_test.go
```diff
@@ -246,7 +246,7 @@ func testExporter(t *testing.T, when spec.G, it spec.S) {
 				})
 			})
 
-			it("creates app layer on Run image", func() {
+			it("creates app layer on run image", func() {
```
natalieparellano, line 249: This PR obviously needs some more unit tests 🙃

exporter.go
```diff
@@ -233,6 +256,100 @@ func (e *Exporter) copyDefaultSBOMsForComponent(component, dstDir string) error
...
+func uncompressLayerAt(layerPath string, toArtifactsDir string) (string, error) {
+	sourceLayer, err := os.Open(layerPath)
+	if err != nil {
+		return "", err
+	}
+	zr, err := gzip.NewReader(sourceLayer)
+	if err != nil {
+		return "", err
+	}
+	tmpLayerPath := filepath.Join(toArtifactsDir, filepath.Base(layerPath)) // for now, used the compressed digest to uniquely identify the layer
```
natalieparellano, line 334:
```suggestion
	tmpLayerPath := filepath.Join(toArtifactsDir, filepath.Base(layerPath)) // for now, used the compressed digest to uniquely identify the layer as we may be writing concurrently to this directory
```

exporter.go
```diff
@@ -233,6 +256,100 @@ func (e *Exporter) copyDefaultSBOMsForComponent(component, dstDir string) error
...
+	if isLocalImage(opts.WorkingImage) {
+		localImage = true
+		if artifactsDir, err = os.MkdirTemp("", "lifecycle.exporter.layer"); err != nil {
+			return err
+		}
+	}
+	for _, l := range extendedLayers {
+		hex, err := l.DiffID()
+		if err != nil {
+			continue // failed to get the diffID because the blob doesn't exist
```
natalieparellano, line 288: We should check the err type

exporter.go
```diff
@@ -534,6 +651,20 @@ func (e *Exporter) addOrReuseLayer(image imgutil.Image, layer layers.Layer, prev
 	return layer.Digest, image.AddLayerWithDiffID(layer.TarPath, layer.Digest)
 }
 
+func (e *Exporter) addOrReuseExtensionLayer(image imgutil.Image, layer layers.Layer) (string, error) {
+	rc, err := image.GetLayer(layer.Digest)
+	if err != nil {
+		// assume we failed to get the layer because it doesn't exist
```
natalieparellano, line 657: We should check the err type

cmd/lifecycle/restorer.go
```diff
@@ -124,12 +124,12 @@ func (r *restoreCmd) Exec() error {
 			if err != nil {
 				return cmd.FailErr(err, "read target data from run image")
 			}
-			ref, err := digestReference(analyzedMD.RunImage.Reference, runImage)
+			digestRef, err := digestReference(analyzedMD.RunImage.Reference, runImage)
```
natalieparellano, line 127: `runImage.Identifier()` might do here

cmd/lifecycle/exporter.go
```diff
@@ -197,12 +208,14 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore lifecycle.Cache, an
...
 		DefaultProcessType: e.DefaultProcessType,
+		ExtendedDir:        e.ExtendedDir,
 		LauncherConfig:     launcherConfig(e.LauncherPath, e.LauncherSBOMDir),
 		LayersDir:          e.LayersDir,
 		OrigMetadata:       analyzedMD.Metadata,
 		Project:            projectMD,
 		RunImageRef:        runImageID,
-		Stack:              stackMD,
+		RunImageForExport:  runImageForExport,
+		Stack:              platform.StackMetadata{RunImage: runImageForExport}, // for backwards compat
```
natalieparellano, line 218: I'm actually not sure if we want this. This would allow newer images to run on older platforms. But maybe we should only populate this from stack.toml if it is present.

cmd/lifecycle/exporter.go
```diff
@@ -221,9 +234,29 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore lifecycle.Cache, an
 }
 
 func (e *exportCmd) initDaemonAppImage(analyzedMD platform.AnalyzedMetadata) (imgutil.Image, string, error) {
+	if isDigestRef(e.RunImageRef) {
+		// If the run image reference is a digest reference, this means extensions were used to extend the runtime base image.
+		// The restorer uses a name reference to pull the image from the registry (because the extender needs a manifest), and writes the digest reference to analyzed.toml.
+		// However, the exporter can't use a digest reference, so we convert it back into a name reference.
```
natalieparellano, line 240:
```suggestion
		// However, the exporter can't use a digest reference for local images, so we convert it back into a name reference.
```

cmd/lifecycle/exporter.go
natalieparellano: I went back and read the code and `getRunImageForExport()` gets stack information from stack.toml for older platforms, so I think this is what we want.

cmd/lifecycle/restorer.go
natalieparellano: On reading the code again, no - a sparse image identifier includes the file path

exporter.go
```diff
@@ -534,6 +658,20 @@ func (e *Exporter) addOrReuseLayer(image imgutil.Image, layer layers.Layer, prev
...
+		e.Logger.Infof("Adding layer %s\n", layer.ID)
```
jabrown85, line 665: nit: I would prefer unique log lines for extension related output
```suggestion
		e.Logger.Infof("Adding extension layer %s\n", layer.ID)
```

jabrown85 (approved):
😅 - Sure are a bunch of changes! Looks solid and this is gated behind experimental. 👍

acceptance/phase_test.go
joe-kimmel-vmw: do you want to capture that as a separate #goodfirstissue? if it's just a tests refactor.  one of us might get to it or someone looking for an opportunity to contribute might.

buildpack/dockerfile.go
joe-kimmel-vmw: https://en.wikipedia.org/wiki/Ace_of_Base

cmd/lifecycle/exporter.go
```diff
@@ -221,9 +234,29 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore lifecycle.Cache, an
 }
 
 func (e *exportCmd) initDaemonAppImage(analyzedMD platform.AnalyzedMetadata) (imgutil.Image, string, error) {
+	if isDigestRef(e.RunImageRef) {
+		// If the run image reference is a digest reference, this means extensions were used to extend the runtime base image.
```
joe-kimmel-vmw, line 238: this comment seems like it might not age well since there could be other reasons in the future to have digest references,  but it's also surprising to me that we're guaranteed to turn a digest back into a tag?   also it seems like the implementation of parseReference in GGCR will actually 'succeed' if it just turns it back into a digest?

```
// ParseReference parses the string as a reference, either by tag or digest.
func ParseReference(s string, opts ...Option) (Reference, error) {
	if t, err := NewTag(s, opts...); err == nil {
		return t, nil
	}
	if d, err := NewDigest(s, opts...); err == nil {
		return d, nil
	}
	return nil, newErrBadName("could not parse reference: " + s)
}
```

so maybe i'm not fully appreciating what we're up to here.   but i think mostly it's just the comment that seems like it's not future proof. e.g. rewording it as
`// if extensions were used to extend the image them the run image reference will be a digest reference`
fixes that fragility, b/c it means if something else comes along in the future that also has the same result, that sentence will still be true.   (whereas the current wording implies that run image extensions is the only process that could have this result)

generator.go
```diff
@@ -204,52 +205,58 @@ func (g *Generator) getGenerateInputs() buildpack.GenerateInputs {
...
+			continue
 		}
-		if dockerfiles[i].NewBase == "" {
+		if dockerfiles[i].Extend {
 			extend = true
 		}
+		if dockerfiles[i].WithBase != "" {
+			newBase = dockerfiles[i].WithBase
+			g.Logger.Debugf("Found a run.Dockerfile from extension '%s' setting run image to '%s' ", dockerfiles[i].ExtensionID, newBase)
+			ignoreNext = true
```
joe-kimmel-vmw, line 253: i think right now "ignoreNext" never gets reset, so why not exit the loop at this point?

joe-kimmel-vmw (commented):
LGTM

generator.go
```diff
@@ -204,52 +205,58 @@ func (g *Generator) getGenerateInputs() buildpack.GenerateInputs {
...
 			continue
 		}
-		if dockerfiles[i].NewBase != "" {
-			newBase = dockerfiles[i].NewBase
-			newBaseIdx = i
-			g.Logger.Debugf("Found a run.Dockerfile configuring image '%s' from extension with id '%s'", newBase, dockerfiles[i].ExtensionID)
-			break
+		if ignoreNext {
+			// If a run.Dockerfile following this one (in the build, not in the loop) switches the run image,
+			// we can ignore this run.Dockerfile as it has no effect.
```
natalieparellano, line 243:
```suggestion
			// we can ignore this run.Dockerfile as it has no effect.
			// We set Ignore to true so that when the Dockerfiles are copied to the "generated" directory,
			// we'll add the suffix `.ignore` so that the extender won't try to apply them.			
```

cmd/lifecycle/exporter.go
natalieparellano: Fair point, updated the comment in f24dc342

exporter.go
natalieparellano: Updated in 4707bcfc

exporter.go
natalieparellano: Updated (though needs improving in imgutil) in 4707bcfc

PR #1056, opened 2023-03-30
0.12 Rebase Validation and Label Sync

jabrown85:
- [Validate OS, Arch, Variant and optional labels on rebase](https://github.com/buildpacks/lifecycle/commit/bb8522457d9c59adb83a4b817cfad2b272bdbc51)
- [Copy io.buildpacks.base.* labels on rebase](https://github.com/buildpacks/lifecycle/commit/68655c8219ca980e74175a8a237739e51d5c5228)

For https://github.com/buildpacks/lifecycle/issues/744

### DIFF
--- a/platform/files.go
+++ b/platform/files.go
@@ -4,6 +4,7 @@ package platform
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 
 	"github.com/buildpacks/lifecycle/internal/fsutil"
@@ -102,6 +103,35 @@ func (t *TargetMetadata) IsSatisfiedBy(o *buildpack.TargetMetadata) bool {
 		}
 	}
 	return true
+}
+
+// IsValidRebaseTargetFor treats optional fields (ArchVariant and Distribution fields) as wildcards if empty, returns true if all populated fields match
+func (t *TargetMetadata) IsValidRebaseTargetFor(appTargetMetadata *TargetMetadata) bool {
+	if t.Arch != appTargetMetadata.Arch || t.OS != appTargetMetadata.OS {
+		return false
+	}
+	if t.ArchVariant != "" && appTargetMetadata.ArchVariant != "" && t.ArchVariant != appTargetMetadata.ArchVariant {
+		return false
+	}
+
+	if t.Distribution != nil && appTargetMetadata.Distribution != nil {
+		if t.Distribution.Name != appTargetMetadata.Distribution.Name {
+			return false
+		}
+		if t.Distribution.Version != appTargetMetadata.Distribution.Version {
+			return false
+		}
+	}
+	return true
+}
+
+func (t *TargetMetadata) String() string {
+	var distName, distVersion string
+	if t.Distribution != nil {
+		distName = t.Distribution.Name
+		distVersion = t.Distribution.Version
+	}
+	return fmt.Sprintf("OS: %s, Arch: %s, ArchVariant: %s, Distribution: (Name: %s, Version: %s)", t.OS, t.Arch, t.ArchVariant, distName, distVersion)
 }
 
 // PopulateTargetOSFromFileSystem populates the target metadata you pass in if the information is available
--- a/rebaser.go
+++ b/rebaser.go
@@ -86,7 +86,12 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 		return RebaseReport{}, errors.Wrap(err, "set app image metadata label")
 	}
 
-	hasPrefix := func(l string) bool { return strings.HasPrefix(l, "io.buildpacks.stack.") }
+	hasPrefix := func(l string) bool {
+		if r.PlatformAPI.AtLeast("0.12") {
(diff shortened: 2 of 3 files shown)


rebaser.go
```diff
@@ -137,6 +142,71 @@ func (r *Rebaser) validateRebaseable(appImg imgutil.Image) error {
...
+			}
+			newBaseArch, err := newBaseImg.Architecture()
+			if err != nil {
+				return errors.Wrap(err, "get new base image architecture")
+			}
+			if appArch != newBaseArch {
+				return fmt.Errorf("incompatible architecture: '%s' is not compatible with '%s'", newBaseArch, appArch)
+			}
+
+			appVariant, err := appImg.Variant()
```
natalieparellano, line 173: `platform.TargetMetadata.IsSatisfiedBy()` captures a couple edge cases that we might want to account for (missing variant matches any, etc.).

rebaser.go
```diff
@@ -137,6 +142,71 @@ func (r *Rebaser) validateRebaseable(appImg imgutil.Image) error {
 		if !r.Force && rebaseable == "false" {
 			return fmt.Errorf("app image is not marked as rebaseable")
 		}
+
+		// check the OS, architecture, and variant values
+		// if they are not the same, the image cannot be rebased unless the force flag is set
+		if !r.Force {
+			appOS, err := appImg.OS()
```
natalieparellano, line 149: `platform.GetTargetFromImage(image imgutil.Image)` could make this a bit simpler, maybe

rebaser.go
```diff
@@ -137,6 +142,71 @@ func (r *Rebaser) validateRebaseable(appImg imgutil.Image) error {
...
+			if err != nil {
+				return errors.Wrap(err, "get app image os")
+			}
+			newBaseOS, err := newBaseImg.OS()
+			if err != nil {
+				return errors.Wrap(err, "get new base image os")
+			}
+			if appOS != newBaseOS {
+				return fmt.Errorf("incompatible os: '%s' is not compatible with '%s'", newBaseOS, appOS)
```
joe-kimmel-vmw, line 158: nit here, and i think also applicable a few other places (even arch)  we're doing an equality check but the word we're using is "compatible" -- so i could be an ubuntu-based distro that promises 100% compatibility, but if i list my os as "mint" or whatever then the string won't equal ubuntu ...   and it's a nit b/c I'm not arguing we should be maintaining a detailed compatiblity table in the code, but I do think we should change the language to `is not equal to` so that it's more obvious to anyone confronting this error message that we're just doing the simplest thing

jabrown85: Updated error message in f4092ea5

rebaser.go
jabrown85: Thanks for that! Seems to have cleaned up things in f4092ea5

rebaser.go
jabrown85: I think I got most of that covered now in f4092ea5 - let me know if you see gaps tho

platform/files.go
```diff
@@ -102,6 +103,35 @@ func (t *TargetMetadata) IsSatisfiedBy(o *buildpack.TargetMetadata) bool {
...
+		return false
+	}
+
+	if t.Distribution != nil && appTargetMetadata.Distribution != nil {
+		if t.Distribution.Name != appTargetMetadata.Distribution.Name {
+			return false
+		}
+		if t.Distribution.Version != appTargetMetadata.Distribution.Version {
+			return false
+		}
```
natalieparellano, line 121: Should we consider "missing" as "matches any" here? I actually don't think we do that in the detector logic, but maybe we should? cc @joe-kimmel-vmw

jabrown85: Did you two get a chance to chat @natalieparellano?

Thinking through the transition from 0.11 to 0.12

Builds app on 0.11 - run image has no distro name/version
Run image gets updated with labels
Platform gets updated to 0.12
Rebase is executed
Fails due to "" on old not equal to "ubuntu/22.04" of new

Hmm. I think it would actually fail earlier as Arch != Arch or OS != OS.

Does that sound right as it is coded/spec'd? Do we want platforms to have added the labels during 0.11 including rebasing all images? Or should we potentially skip this bit of validation entirely if the target image was built with < 0.12?

Builds app on 0.11 - run image has no distro name/version
Run image gets updated with labels
Rebase image on 0.11 - OS/Arch/Distro/etc. validation is skipped
Successful Rebase
Platform gets updated to 0.12
Rebase is executed - validation occurs
Successful Rebase as everything matches from before

natalieparellano: > Or should we potentially skip this bit of validation entirely if the target image was built with < 0.12?

That makes sense to me. If the image was built on an older platform there is no way for the labels to be there.

jabrown85: Took care of that in [bd6b6c6](https://github.com/buildpacks/lifecycle/pull/1056/commits/bd6b6c6879fa90f0b23503ac198eda553e7a4727)

natalieparellano (approved):

PR #1060, opened 2023-04-05
Validate the cache image and previous image in the analyze phase

### LINKED ISSUE #1044: Reusing nonexistent layers when exporting
### Summary

It is possible for an image stored in a registry to become corrupt.  Google Container Registry is particularly susceptible because the underlying storage is provided by a customer owned Cloud Storage project. Customers can set retention policies that break images in unexpected ways (e.g. deleting a few layers but not the manifest or config).

The export phase fails when trying to re-use missing layers from a corrupt `CNB_CACHE_IMAGE` or `CNB_PREVIOUS_IMAGE`, because it doesn't validate the layers actually exist.

---

##### Steps

1. Build an image and `--publish` it to Google Container Registry using `pack`:
```
pack build \
  --builder=gcr.io/buildpacks/builder \
  --cache-image=gcr.io/my-proj/corrupt-img:cache \
  --publish \
  gcr.io/my-proj/corrupt-img:latest
``` 
2. Corrupt the image by deleting the top layer of the app image:
```
# use crane to dig out the top layer digest
LAYER=$(crane manifest gcr.io/mattrobertsontest/gotestmin:latest | jq -r ".layers | .[-1] | .digest")
# use gsutil to delete the layer from the underlying cloud storage bucket
gsutil rm gs://artifacts.my-proj.appspot.com/containers/images/$LAYER
```
3. Re-run the `pack build` from step 1 to observe a cache hit on a corrupt image

##### Current behavior

The exporter fails with an error message like:

```
[exporter] Saving gcr.io/my-proj/gotestmin:latest...
[exporter] *** Images (sha256:ddaf944ac6873ad852ce266e53bf0d792be83f094dd1e29ed27076f374b8dab9):
[exporter]       gcr.io/my-proj/gotestmin:latest - GET https://storage.googleapis.com/artifacts.my-proj.appspot.com/containers/images/sha256:5a44e4f7b58d74fe6f92dd7028075c91191128d1e2e7f39846fe061a9a98836e?access_token=REDACTED: unexpected status code 404 Not Found: <?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: artifacts.my-proj.appspot.com/containers/images/sha256:5a44e4f7b58d74fe6f92dd7028075c91191128d1e2e7f39846fe061a9a98836e</Details></Error>
[exporter] ERROR: failed to export: failed to write image to the following tags: [gcr.io/my-proj/gotestmin:latest: GET https://storage.googleapis.com/artifacts.my-proj.appspot.com/containers/images/sha256:5a44e4f7b58d74fe6f92dd7028075c91191128d1e2e7f39846fe061a9a98836e?access_token=REDACTED: unexpected status code 404 Not Found: <?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: artifacts.my-proj.appspot.com/containers/images/sha256:5a44e4f7b58d74fe6f92dd7028075c91191128d1e2e7f39846fe061a9a98836e</Details></Error>]
ERROR: failed to build: executing lifecycle. This may be the result of using an untrusted builder: failed with status code: 62
```

##### Expected

The lifecycle should not try to re-use layers that don't exist on the registry. One idea is to perform a [fast validation](https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/validate.go#L66) of the cache and previous image to ensure all the layers exist during the analyze phase. When validation fails, the image would be [treated as though it doesn't exist](https://github.com/buildpacks/lifecycle/blob/1299eb80d1781ed5d50a27ff8559026c482e61f8/image/labels.go#L11).

##### lifecycle version

0.16.0

##### platform version(s)

```
❯ pack report
Pack:
  Version:  0.23.0+git-0db2c77.build-3056
  OS/Arch:  linux/amd64

Default Lifecycle Version:  0.13.1

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8

Config:

  [[trusted-builders]]
    name = "[REDACTED]"
```

matthewrobertson:
Updates the analyzer to only reuse layers from the previous app image or cache image if they are not corrupt. This uses the new `imgutil::Valid()` that was added in https://github.com/buildpacks/imgutil/pull/190

Fixes #1044


analyzer.go
```diff
@@ -251,6 +251,10 @@ func (a *Analyzer) getImageIdentifier(image imgutil.Image) (string, error) {
 		a.Logger.Infof("Image with name %q not found", image.Name())
 		return "", nil
 	}
+	if !image.Valid() {
+		a.Logger.Infof("Ignoring image with name %q because it was corrupt", image.Name())
```
jabrown85, line 255: Is a `corrupt` image the only reason it would fail at this point?

cache/image_cache.go
```diff
@@ -58,7 +58,7 @@ func NewImageCacheFromName(name string, keychain authn.Keychain, logger log.Logg
 }
 
 func (c *ImageCache) Exists() bool {
-	return c.origImage.Found()
+	return c.origImage.Valid()
```
jabrown85, line 61: Here we are delegating `Exists()` to `Valid` but in `analyzer.go` we are explicitly doing both `.Found()` and `.Valid()`. Is that because we want specific Not Found/Corrupt messages in analzyer but it doesn't much matter for cache images?

analyzer.go
matthewrobertson: Yes, `Valid()` will return false if the image is not found, the manifest cannot be parsed or something listed in the manifest does not exist in the registry. The `!image.Found()` check on line 254, has already confirmed that manifest exists, so this condition will only occur if the something is corrupt.

cache/image_cache.go
matthewrobertson: I just wanted to preserve the log message that was already present in `getImageIdentifier`. The `.Valid()` function returns false if the image doesn't exist, so `i.Exists() && i.Valid()` will send one redundant `HEAD` request to confirm that manifest exits.

matthewrobertson:
I actually tested this end to end by creating a custom builder with my updated lifecycle and it seems like this is not sufficient to solve the issue. I can see the corrupt image is ignored during analysis but I am now getting a new error while exporting:

```
ERROR: failed to export: previous image did not have layer with diff id "sha256:7dd44d3eb001c70a8cfc002f048f29a3ee3ba335356f1b4d2f7cf8d0cc29173a"
ERROR: failed to build: executing lifecycle: failed with status code: 62
```

analyzer.go
matthewrobertson: This wasn't actually preventing the tag from being decoded.

cache/image_cache.go
matthewrobertson: Found a better way to do this that I think makes more sense and provides better logging.

analyzer.go
```diff
@@ -195,22 +195,15 @@ func (a *Analyzer) Analyze() (platform.AnalyzedMetadata, error) {
...
-			return platform.AnalyzedMetadata{}, errors.Wrap(err, "identifying previous image")
-		}
-
-		// continue even if the label cannot be decoded
-		if err = image.DecodeLabel(a.PreviousImage, platform.LayerMetadataLabel, &appMeta); err != nil {
-			appMeta = platform.LayersMetadata{}
-		}
-
-		if err = a.SBOMRestorer.RestoreFromPrevious(a.PreviousImage, bomSHA(appMeta)); err != nil {
+	if sha := bomSHA(appMeta); sha != "" {
```
jabrown85, line 203: This is going to break some tests - I know the `SBOMRestorer.RestoreFromPrevious` is mocked and expectations may need to be updated if there are cases where the `bomSHA` returns `""` in those tests.

matthewrobertson: Fixed.

matthewrobertson:
windows failures are fixed by this PR in imgutil: https://github.com/buildpacks/imgutil/pull/196

jabrown85:
@matthewrobertson looks like we still have failures in windows about corrupt images. https://github.com/buildpacks/lifecycle/actions/runs/4758563102/jobs/8456787984?pr=1060#step:6:6897

If you don't have any ideas maybe we can get someone with a windows rig to help take a look

matthewrobertson:
I still see logs like "Ignoring cache image "xxx" because it was corrupt" in test output. I will poke around to see if I can figure out why they are failing the validation.

matthewrobertson:
@jabrown85 looks like the code was doing what it was supposed to do 😄 Windows base image layers weren't being pushed to the registry because of this: https://docs.docker.com/engine/reference/commandline/dockerd/#allow-push-of-nondistributable-artifacts 

Should be fixed now 🤞

jabrown85 (approved):
Nice work! I'll let @natalieparellano weigh in as well before merge but it looks great to me and all the tests are now passing 😄

.github/workflows/build.yml
```diff
@@ -76,6 +76,7 @@ jobs:
             $config=(Get-Content C:\ProgramData\docker\config\daemon.json | ConvertFrom-json)
           }
           $config | Add-Member -Force -Name "insecure-registries" -value @("$IPAddress/32") -MemberType NoteProperty
+          $config | Add-Member -Force -Name "allow-nondistributable-artifacts" -value @("$IPAddress/32") -MemberType NoteProperty
```
natalieparellano, line 79: IIUC not having this set was causing us to push invalid (missing base layers) Windows images. Would it be possible to update the documentation [here](https://github.com/buildpacks/lifecycle/blob/main/DEVELOPMENT.md#caveats) to inform contributors on Windows or our future selves to enable this setting locally?

matthewrobertson: Done

PR #1061, opened 2023-04-05
Enable pre-releases in addition to release candidates

natalieparellano:
While "release candidate" is still a pre-release, it conveys more readiness than a "pre-release" or "preview release".
We'd like to ship pre-release lifecycle artifacts
so that users can test out new features without them being fully implemented as they would in a release candidate.


natalieparellano:
Tested on fork:

Release drafted: https://github.com/natalieparellano/lifecycle/actions/runs/4622275681/jobs/8174750115 (resulting release is a pre-release)

Release published: https://github.com/natalieparellano/lifecycle/actions/runs/4622289696 (images re-tagged as semver but not `latest`)

There is an argument to be made that we should actually be parsing the version as semver instead of string matching, but that is more cumbersome.

jabrown85 (approved):

PR #1076, opened 2023-04-28
Refactor: create new package for files in the platform spec

natalieparellano:
We've discussed this for awhile, this should make things easier to find


platform/files/analyzed.go
```diff
@@ -0,0 +1,228 @@
...
+
+	"github.com/buildpacks/lifecycle/buildpack"
+	"github.com/buildpacks/lifecycle/internal/fsutil"
+	"github.com/buildpacks/lifecycle/log"
+)
+
+// Analyzed is written by the analyzer as analyzed.toml to record information about:
+// * the previous image (if it exists),
+// * the run image,
+// * the build image (if provided).
```
natalieparellano, line 17: Improving our docs coverage :)

platform/files/analyzed.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return fmt.Sprintf("OS: %s, Arch: %s, ArchVariant: %s, Distribution: (Name: %s, Version: %s)", t.OS, t.Arch, t.ArchVariant, distName, distVersion)
+}
+
+type OSDistribution struct {
+	Name    string `json:"name" toml:"name"`
+	Version string `json:"version" toml:"version"`
+}
+
+// FIXME: the target logic in this file might be better located in a dedicated package.
```
natalieparellano, line 165: I kinda feel like this package should just be about serializing & deserializing data, leaving other CNB business logic to other packages. But I didn't want to change too much in one PR.

platform/files/group.go
```diff
@@ -0,0 +1,5 @@
+package files
+
+// group.toml is written by the detector to record the buildpacks and image extensions
+// that will be used for the build.
+// See `buildpack.Group` for further information.
```
natalieparellano, line 5: Pulling this out of the buildpack package is a whole other refactor...

joe-kimmel-vmw (approved):
LGTM - thanks!

platform/files/analyzed.go
natalieparellano: I managed to move this stuff back to the platform package

natalieparellano: I managed to move this stuff back to the platform package

platform/run_image.go
```diff
@@ -38,65 +45,85 @@ func GetRunImageForExport(inputs LifecycleInputs) (RunImageForExport, error) {
...
-	if err != nil {
-		return &tm, err
-	}
-	tm.Arch, err = image.Architecture()
-	if err != nil {
-		return &tm, err
-	}
-	tm.ArchVariant, err = image.Variant()
-	if err != nil {
-		return &tm, err
```
natalieparellano, line 68: Moved to target_data.go

natalieparellano:
After this PR, we can make sure every file has a "read" method so that the ConfigHandler interface is fully satisfied

cmd/lifecycle/rebaser.go
```diff
@@ -176,7 +177,7 @@ func (r *rebaseCmd) setAppImage() error {
 		}
 
 		// for older platforms, we find the best mirror for the run image as this point
-		r.RunImageRef, err = md.Stack.BestRunImageMirror(registry, r.LifecycleInputs.AccessChecker)
+		r.RunImageRef, err = platform.BestRunImageMirrorFor(registry, md.Stack.RunImage, r.LifecycleInputs.AccessChecker)
```
joe-kimmel-vmw, line 180: 👍

cmd/lifecycle/restorer.go
```diff
@@ -125,15 +126,15 @@ func (r *restoreCmd) Exec() error {
 			if err != nil {
 				return cmd.FailErr(err, "read run image")
 			}
-			targetData, err := platform.GetTargetFromImage(runImage)
+			targetData, err := platform.GetTargetMetadata(runImage)
```
joe-kimmel-vmw, line 129: 👍

detector.go
```diff
@@ -244,7 +245,7 @@ func (d *Detector) detectGroup(group buildpack.Group, done []buildpack.GroupElem
 				} else {
 					for i := range bpDescriptor.Targets {
 						d.Logger.Debugf("Checking for match against descriptor:", bpDescriptor.Targets[i])
-						if d.AnalyzeMD.RunImage.TargetMetadata.IsSatisfiedBy(&bpDescriptor.Targets[i]) {
+						if platform.TargetSatisfiedForBuild(*d.AnalyzeMD.RunImage.TargetMetadata, bpDescriptor.Targets[i]) {
```
joe-kimmel-vmw, line 248: mixed feelings on turning this from "function with a receiver and one arg" into "function with two args" but I think you got the name right

exporter.go
```diff
@@ -430,7 +431,7 @@ func (e *Exporter) addBuildpackLayers(opts ExportOptions, meta *platform.LayersM
 				if lmd.Cache {
 					return fmt.Errorf("layer '%s' is cache=true but has no contents", fsLayer.Identifier())
 				}
-				origLayerMetadata, ok := opts.OrigMetadata.MetadataForBuildpack(bp.ID).Layers[fsLayer.Name()]
+				origLayerMetadata, ok := opts.OrigMetadata.LayersMetadataFor(bp.ID).Layers[fsLayer.Name()]
```
joe-kimmel-vmw, line 434: 👍

platform/files/order.go
```diff
@@ -0,0 +1,5 @@
+package files
+
+// order.toml is provided by the platform to enumerate the buildpacks and image extensions
```
joe-kimmel-vmw, line 3: this file is great - is the implication that `buildpack.order` is the struct used to write "the platform's order.toml" ?   what does that mean re. the separation of the buildpack and platform apis?

platform/files/report.go
```diff
@@ -0,0 +1,21 @@
+package files
+
+import "github.com/buildpacks/lifecycle/buildpack"
+
+// Report is written by the exporter as report.toml to record information about the build.
+// It is not included in the output image, but can be saved off by the platform before the build container exits.
```
joe-kimmel-vmw, line 6: in general these comments are great and answer the primary questions I would have (what file name matches this struct, what phase writes it, how can i find it, etc).   In this case (and maybe any similar ones i skimmed past) would it be reasonable to name the flag that causes it to be output?   e.g. `but will be saved off by the platform before the build container exits if the --report flag is provided" ?

platform/run_image.go
```diff
@@ -38,65 +45,85 @@ func GetRunImageForExport(inputs LifecycleInputs) (RunImageForExport, error) {
...
+		}
 	}
-	ret = append(ret, "CNB_TARGET_DISTRO_NAME="+distName)
-	ret = append(ret, "CNB_TARGET_DISTRO_VERSION="+distVersion)
-	if tm.ID != "" {
-		ret = append(ret, "CNB_TARGET_ID="+tm.ID)
+	return ""
+}
+
+type ImageStrategy interface {
```
joe-kimmel-vmw, line 111: imo we're not getting much mileage out of this interface:  because it's for exactly one function and the receivers are  stateless, it could be a "function signature type" and we could be passing function pointers as arguments.   that would reduce boilerplate and be more readable imo,  because you would have to make new empty structs just to access the function implementation. 

Do you think it's likely that we would add a second coupled function (so that the interface/strategy would be for a family of functions instead of just a single func?)

joe-kimmel-vmw (approved):
I'm comfortable with this as-is although I did try to indicate which parts I liked the most and least. Feel free to only address any comments I made that you agree with, none of my stated opinions are very strongly held here.

platform/run_image.go
natalieparellano: SGTM. Updated in 67e81d09

platform/files/order.go
natalieparellano: This is surely a point of intersection :)

detector.go
natalieparellano: Yeah... the former signature was so much more readable. But then, it would have to live in the `files` package 😞

platform/files/report.go
natalieparellano: Sounds good! Updated in 09c47792

acceptance/analyzer_test.go
```diff
@@ -1062,12 +1062,12 @@ func testAnalyzerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 	}
 }
 
-func assertAnalyzedMetadata(t *testing.T, path string) *platform.AnalyzedMetadata {
+func assertAnalyzedMetadata(t *testing.T, path string) *files.Analyzed {
```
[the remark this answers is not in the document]
natalieparellano: I like that! I think it will become more obvious when this will be helpful if/when we create a buildpack/files package and the names collide.

PR #1087, opened 2023-05-09
Add support for other operating systems

quantumsheep:


.github/workflows/build.yml
```diff
@@ -48,6 +49,30 @@ jobs:
...
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+      - name: Test
+        uses: vmactions/freebsd-vm@v0
+        with:
+          usesh: true
+          prepare: |
+            wget https://go.dev/dl/go1.20.4.freebsd-amd64.tar.gz
+            tar -C /usr/local -xzf go1.20.4.freebsd-amd64.tar.gz
```
natalieparellano, line 67: Could we interpolate the version somehow so that we're using the same go version across all 3 OSes?

.github/workflows/build.yml
```diff
@@ -148,6 +174,10 @@ jobs:
         with:
           name: lifecycle-linux-arm64-sha256
           path: out/lifecycle-v*+linux.arm64.tgz.sha256
+      - uses: actions/upload-artifact@v2
+        with:
+          name: lifecycle-freebsd-x86-64
+          path: out/lifecycle-v*+freebsd.x86-64.tgz
```
natalieparellano, line 180: Is the intent also to make the artifact available on releases? If so we'll also need to update draft-release.yml (for attaching the artifact to the release), post-release.yml (to re-tag images from Git commit to semver), and around line 236 in this file to publish the Git-tagged image.

quantumsheep: The last commits should do it 😃

quantumsheep:
@jabrown85 Thanks! The last commit should fix the issue :)

jabrown85:
@quantumsheep looks like we need some more CI tweaks 🤞

PR #1090, opened 2023-05-12
(temporary) remove anchore scan step to allow publishing

joe-kimmel-vmw:


jabrown85:
I just saw this - happy to approve but I also think we can use an [ignore file](https://github.com/anchore/grype#specifying-matches-to-ignore)

jabrown85 (approved):

natalieparellano:
Are talking about https://github.com/advisories/GHSA-hqxw-f8mx-cpmw? If so I am agreed this can be safely ignored via .grype.yaml

PR #1092, opened 2023-05-16
ignore non-impactful CVE

joe-kimmel-vmw:
Signed-off-by: Joe Kimmel <jkimmel@vmware.com>


natalieparellano (approved):

natalieparellano:
Do we need to add this to `main` also?

jjbustamante:
I'd suggest we should do it

Issue #1057, opened 2023-03-31
Fixups for run image extension


natalieparellano:
In reviewing the work that has been done already, some gaps were identified between the desired implementation and the actual implementation. Probably more gaps will be identified in working through https://github.com/buildpacks/pack/issues/1623. This issue is meant to be a catch-all issue but each gap is likely small and could be considered a "good first issue".

* Detector: if the run image based is switched by an image extension we should verify that the new base exists in run.toml (need to verify that run.toml path is an input to the detector, and if not, update the [spec PR](https://github.com/buildpacks/spec/pull/347))
* Extender: if the final user after all Dockerfiles have been applied does not match the CNB_USER_ID, we should warn
* Extender: if the final user after all Dockerfiles have been applied is root, we should fail
* Extender: the kaniko working directory should be a temp directory inside /kaniko instead of /kaniko itself (allows build image extension and run image extension to happen simultaneously without clobbering each other, while sharing the same cache)
* ...more to add here

joe-kimmel-vmw:
> Should extensions expose and/or the lifecycle validate target data for extensions?

It seems to me like something is warranted here --  e.g. if i'm writing an extension that uses a package manager like apt, yum, apk, nix, etc., then I would want that extension to only run on the appropriate distribution of linux and I might even be interested in writing several extensions in order to support multiple distributions or package managers. But maybe that's a rare use-case?

PR #1093, opened 2023-05-16
extender: better uid checks and error if it stay as root

joe-kimmel-vmw:
should resolve two bullet points on https://github.com/buildpacks/lifecycle/issues/1057

### DIFF
--- a/extender.go
+++ b/extender.go
@@ -307,15 +307,14 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
 	}
 	workingHistory = configFile.History
 	buildOptions := e.extendOptions()
+	userID, groupID := userFrom(*configFile)
+	origUserID := userID
 	for _, dockerfile := range dockerfiles {
-		// create args
-		userID, groupID := userFrom(*configFile)
 		dockerfile.Args = append([]extend.Arg{
 			{Name: argBuildID, Value: uuid.New().String()},
 			{Name: argUserID, Value: userID},
 			{Name: argGroupID, Value: groupID},
 		}, dockerfile.Args...)
-
 		// apply Dockerfile
 		if baseImage, err = e.DockerfileApplier.Apply(
 			dockerfile,
@@ -325,8 +324,7 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
 		); err != nil {
 			return nil, fmt.Errorf("applying Dockerfile to image: %w", err)
 		}
-
-		// update rebasable, history in config
+		// update rebasable, history in config, and user/group IDs
 		configFile, err = baseImage.ConfigFile()
 		if err != nil || configFile == nil {
 			return nil, fmt.Errorf("getting image config: %w", err)
@@ -350,6 +348,17 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
 			)
 		}
 		configFile.History = workingHistory
+		prevUserID := userID
+		userID, groupID = userFrom(*configFile)
+		if userID == "0" {
+			logger.Warnf("Extension from %s changed the user ID from %d to %d; this must not be the final user ID (a following extension must reset the user).", prevUserID, userID, dockerfile.Path)
+		}
+	}
+	if userID == "0" {
+		return baseImage, fmt.Errorf("the final user ID is 0 (root); please add another extension that resets the user to non-root")
+	}
+	if userID != origUserID {
+		logger.Warnf("The original user ID was %s but the final extension left the user ID set to %s.", origUserID, userID)
 	}
 	if kind == buildpack.DockerfileKindBuild {
 		return baseImage, nil
--- a/extender_test.go
+++ b/extender_test.go
@@ -227,7 +227,7 @@ func testExtender(t *testing.T, when spec.G, it spec.S) {
 				// first dockerfile
 
 				firstConfig := &v1.ConfigFile{Config: v1.Config{
-					User: "1234:5678",
+					User: "0:5678",
 				}}
 				someFakeImage.ConfigFileReturnsOnCall(0, firstConfig, nil)
 				fakeDockerfileApplier.EXPECT().Apply(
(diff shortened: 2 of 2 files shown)


cmd/lifecycle/extender.go
```diff
@@ -60,6 +60,7 @@ func (e *extendCmd) Exec() error {
 	extender, err := extenderFactory.NewExtender(
 		e.AnalyzedPath,
 		e.AppDir,
+		fmt.Sprint(e.UID),
```
natalieparellano, line 63: This UID is what the extender is running as (or rather, what it will drop privileges to once extension is finished), which _should_ be the CNB_USER_ID from the builder image, but is not guaranteed to be. Also when extending run images the runtime user may be different from the build time user.

extender.go
```diff
@@ -327,6 +333,12 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
 			rebasable = false
 		}
 	}
+	if userID == "0" {
+		return baseImage, fmt.Errorf("extension process ended with root user; please add another extension at the end that resets the user")
+	}
+	if userID != e.CnbUID {
```
natalieparellano, line 339: Rather than `e.CnbUID` I think we want to use the first value of `userID` after it is initialized.

When extending build images (only) we _could_ warn when the CNB_USER_ID from the original image config doesn't match the original USER. `pack builder create` doesn't enforce this today, but maybe it should (see thoughts from @BarDweller [here](https://cloud-native.slack.com/archives/C033DV9EBDF/p1683750839164209?thread_ts=1683741259.856339&cid=C033DV9EBDF)).

extender.go
```diff
@@ -298,16 +301,19 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
...
 		dockerfile.Args = append([]extend.Arg{
 			{Name: argBuildID, Value: uuid.New().String()},
 			{Name: argUserID, Value: userID},
 			{Name: argGroupID, Value: groupID},
 		}, dockerfile.Args...)
 
 		// apply Dockerfile
-
+		if userID == "0" {
+			logger.Warnf("extension from %s is using root user. This must not be the final user ID (you must have other non-root extensions after this).", dockerfile.Path)
```
natalieparellano, line 315:
```suggestion
			logger.Warnf("Extension from %s changed the user ID from %d to %d; this must not be the final user ID (a following extension must reset the user).", origUserID, userID, dockerfile.Path)
```

extender.go
```diff
@@ -327,6 +333,12 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
 			rebasable = false
 		}
 	}
+	if userID == "0" {
+		return baseImage, fmt.Errorf("extension process ended with root user; please add another extension at the end that resets the user")
```
natalieparellano, line 337:
```suggestion
		return baseImage, fmt.Errorf("The final user ID is 0 (root); please add another extension that resets the user to non-root.")
```

extender.go
joe-kimmel-vmw: cool - i think this is addressed in the most recent commit but lmk

extender.go
```diff
@@ -298,16 +301,22 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
 		return nil, fmt.Errorf("getting image config: %w", err)
 	}
 	buildOptions := e.extendOptions()
+	var userID, origUserID, prevUserID, groupID string
+	prevUserID = e.CnbUID // placeholder initial value
```
natalieparellano, line 305: Is this used anywhere?

cmd/lifecycle/extender.go
natalieparellano: I think if we're always just reading from the config then we don't need the new field, right?

extender.go
```diff
@@ -302,7 +299,7 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
 	}
 	buildOptions := e.extendOptions()
 	var userID, origUserID, prevUserID, groupID string
-	prevUserID = e.CnbUID // placeholder initial value
+	prevUserID = "original user ID" // placeholder initial value
```
natalieparellano, line 302: If you move `userID, groupID = userFrom(*configFile)` out of the `for` loop you will always know the previous user ID

joe-kimmel-vmw: cool - thanks.   actually it may have been an error to have it where it was before?   it's weird because we update `baseImage` and `configFile` in the lower half of the loop but we were setting the userID in the upper half of the loop, so i think that meant it was never getting updated the last time? but idk the unit tests pass either way so maybe i'm missing something.

extender.go
```diff
@@ -298,16 +298,22 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
...
-		userID, groupID := userFrom(*configFile)
 		dockerfile.Args = append([]extend.Arg{
 			{Name: argBuildID, Value: uuid.New().String()},
 			{Name: argUserID, Value: userID},
 			{Name: argGroupID, Value: groupID},
 		}, dockerfile.Args...)
-
+		if origUserID == "" {
+			origUserID = userID
+		}
```
natalieparellano, line 312: I think we also need to move this out of the for loop 🙃

extender.go
```diff
@@ -298,16 +298,22 @@ func (e *Extender) extend(kind string, baseImage v1.Image, logger log.Logger) (v
...
 		// apply Dockerfile
-
+		if userID == "0" {
+			logger.Warnf("Extension from %s changed the user ID from %d to %d; this must not be the final user ID (a following extension must reset the user).", prevUserID, userID, dockerfile.Path)
+		}
```
natalieparellano, line 316: This should go after `userID, groupID = userFrom(*configFile)` ...I think

joe-kimmel-vmw: hmm - if you move it below that line in the loop then you miss the first userID entirely . but you're right that if it's here in the loop then you miss the last user ID - but that's the one that we'll raise an error on. 

So options include:
- (1) keep it here / where it is
pros:  you only have this check once, and every case is covered by _something_
cons: it's a little unintuitive, and you won't get a warning before your error

- (2) move a copy of this above the loop to cover the initial case, and a copy of this down in the loop to cover the updates.
pros: you always get the same warning, including before the error occurs
cons: you have slightly duplicated logic. 


IIRC this whole situation is why languages like C have "do...while" loops so that you can have your initialization happen inside the loop,  but 🤷 

I think I'll try option 2 above since it's the most consistent UX.

joe-kimmel-vmw: oh... i see. it doesn't apply on the initial case, since the initial user ID doesn't come from an extension so it can't be root!

natalieparellano:
Looks good @joe-kimmel-vmw! Is it possible you could fix the DCO? I think you need to re-sign https://github.com/buildpacks/lifecycle/pull/1093/commits/92b3363bc7ca0bb2db55078dbeb39cffaf1c11c5

natalieparellano (approved):

Issue #411, opened 2020-09-09
[RFC #0102] Image history metadata


nebhale:
Currently, when a layer is created from a `Dockerfile`, some additional metadata is added to the created image containing the directives that were run to create a particular layer:

<img width="1389" alt="image" src="https://user-images.githubusercontent.com/60754/92660108-a41fcc00-f2ae-11ea-861e-59eaecc2497d.png">

Since buildpacks don't have "directives", this metadata is blank creating what I call a "black hole" for metadata:
<img width="1389" alt="image" src="https://user-images.githubusercontent.com/60754/92660287-f82ab080-f2ae-11ea-89fa-79a0f6aca796.png">

The `lifecycle` should populate whatever metadata is necessary in order to fill in this black hole.  As a straw man, I'd suggest each layer getting something like `<buildpack name> <buildpack version> – <layer name>` (e.g. `Paketo BellSoft Liberica Buildpack 3.2.0 – jre`) but I'm not really fussed with the actual content of this metadata.

jromero:
Was looking a bit at this out of curiosity and found the the field used in the config is under `history.created_by` and it's described in the [spec](https://github.com/opencontainers/image-spec/blob/e562b04403929d582d449ae5386ff79dd7961a11/config.md#properties) as "The command which created the layer.". 

I wondered if there was any real considerations that would need to be taken into account like _something_ trying to use this to recreate said image and there by needing to be executable. The nebulous description led me to explore certain usages and my ultimate finding was that there really isn't a definitive outcome. I'm only sharing this here in case it may help someone else.

Docker, using `docker history cnbs/sample-stack-run:bionic --no-trunc`, does current provide _executable_ bits like:

```shell
/bin/sh -c mkdir -p /run/systemd && echo 'docker' > /run/systemd/container
```

as well as proper comments such as: 

```shell
/bin/sh -c #(nop)  ARG cnb_gid=1000
```

but then this when using `ARGS` which I failed to find any real docs into it's format (non-executable?): 

```shell
|2 cnb_gid=1000 cnb_uid=1000 /bin/sh -c groupadd cnb --gid ${cnb_gid} &&   useradd --uid ${cnb_uid} --gid ${cnb_gid} -m -s /bin/bash cnb
```

ddobrin:
This is a question which is being asked by virtually every client of my company after demos.
Using the dive utility is an option, however is something totally different and can't be scripted.
Thank  you, it would be very helpful to have

[...]

PR #1099, opened 2023-05-22
Add history when adding buildpack or extension layers

natalieparellano:
Fixes https://github.com/buildpacks/lifecycle/issues/411


exporter.go
```diff
@@ -672,12 +676,12 @@ func (e *Exporter) addOrReuseExtensionLayer(image imgutil.Image, layer layers.La
...
 		e.Logger.Infof("Adding extension layer %s\n", layer.ID)
 		e.Logger.Debugf("Layer '%s' SHA: %s\n", layer.ID, layer.Digest)
-		return layer.Digest, image.AddLayerWithDiffID(layer.TarPath, layer.Digest)
+		return layer.Digest, image.AddLayerWithDiffIDAndHistory(layer.TarPath, layer.Digest, layer.History)
 	}
 	_ = rc.Close() // close the layer reader
 	e.Logger.Infof("Reusing layer %s\n", layer.ID)
 	e.Logger.Debugf("Layer '%s' SHA: %s\n", layer.ID, layer.Digest)
-	return layer.Digest, image.ReuseLayer(layer.Digest)
+	return layer.Digest, image.ReuseLayer(layer.Digest) // TODO: ReuseLayerWithHistory
```
natalieparellano, line 684: We need the ability to add history when re-using a layer so that re-builds where the previous image was built on an earlier platform will still have history in the re-used layers

jabrown85:
@natalieparellano does the history from the base image make it into the final image if present?

layers/factory.go
```diff
@@ -5,10 +5,22 @@ import (
...
 )
 
+const (
+	BuildpackLayerName      = "Layer: '%s', Created by buildpack: %s"
+	ExtensionLayerName      = "Layer: '%s', Created by extension: %s"
+	LauncherConfigLayerName = "Buildpacks Launcher Config"
+	LauncherLayerName       = "Buildpacks Application Launcher"
+	ProcessTypesLayerName   = "Buildpacks Process Types"
+	SBOMLayerName           = "Software Bill-of-Materials"
+	SliceLayerName          = "Application Slice: %d"
```
jabrown85, line 21: Probably not worth it - but I would support having `Application Layer` if you only have a single slice.

layers/factory.go
```diff
@@ -5,10 +5,22 @@ import (
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+
 	"github.com/buildpacks/lifecycle/archive"
 	"github.com/buildpacks/lifecycle/log"
 )
 
+const (
+	BuildpackLayerName      = "Layer: '%s', Created by buildpack: %s"
+	ExtensionLayerName      = "Layer: '%s', Created by extension: %s"
+	LauncherConfigLayerName = "Buildpacks Launcher Config"
```
jabrown85, line 17: Should this be named a little more generic given the layer name?
```suggestion
	LauncherConfigLayerName = "Buildpacks Config"
```

jabrown85 (approved):
This is _super_ cool to see! Commented but no concern from me.

layers/factory.go
natalieparellano: I left as-is for now because https://github.com/buildpacks/rfcs/blob/main/text/0102-history.md has "launcher", but we can change it if anyone has a strong opinion

PR #1104, opened 2023-05-31
When rebasing, update stack/runImage key in lifecycle metadata label

natalieparellano:
if the provided run image does not match existing metadata.

See https://github.com/buildpacks/spec/pull/360

### DIFF
--- a/rebaser_test.go
+++ b/rebaser_test.go
@@ -1,6 +1,7 @@
 package lifecycle_test
 
 import (
+	"encoding/json"
 	"math/rand"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


rebaser_test.go
```diff
@@ -129,6 +130,117 @@ func testRebaser(t *testing.T, when spec.G, it spec.S) {
 				h.AssertEq(t, md.App, []interface{}{map[string]interface{}{"sha": "123456"}})
 			})
 
+			when("new base image has no run image metadata", func() {
```
natalieparellano, line 133:
```suggestion
			when("new base image is not found in run image metadata", func() {
```

joe-kimmel-vmw (approved):

PR #1106, opened 2023-05-31
Add more logging so we know when the lifecycle actually starts

natalieparellano:
In working through https://github.com/buildpacks/lifecycle/issues/1007 it was hard to know if the lifecycle had even started when we were stuck in a loop trying to get the auth, because the first log lines emitted by the lifecycle were emitted after the Analyzer had started.

If we had this logging in place beforehand,
we would have known the problem was somewhere in Privileges()

### DIFF
--- a/cmd/lifecycle/cli/command.go
+++ b/cmd/lifecycle/cli/command.go
@@ -24,7 +24,7 @@ type Command interface {
 	Exec() error
 }
 
-func Run(c Command, asSubcommand bool) {
+func Run(c Command, withName string, asSubcommand bool) {
 	var (
 		printVersion bool
 		logLevel     string
@@ -55,6 +55,7 @@ func Run(c Command, asSubcommand bool) {
 	if err := cmd.DefaultLogger.SetLevel(logLevel); err != nil {
 		cmd.Exit(err)
 	}
+	cmd.DefaultLogger.Debugf("Starting %s...", withName)
 
 	// Warn when CNB_PLATFORM_API is unset
 	if os.Getenv(platform.EnvPlatformAPI) == "" {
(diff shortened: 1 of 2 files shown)


cmd/lifecycle/cli/command.go
```diff
@@ -24,7 +24,7 @@ type Command interface {
 	Exec() error
 }
 
-func Run(c Command, asSubcommand bool) {
+func Run(c Command, withName string, asSubcommand bool) {
```
joe-kimmel-vmw, line 27: just a nit.- `withPhaseName` or `phaseName` would be more clear to me than `withName`, but maybe it isn't always the name of a phase?

joe-kimmel-vmw (approved):

cmd/lifecycle/cli/command.go
natalieparellano: Updated in [389cc1b](https://github.com/buildpacks/lifecycle/pull/1106/commits/389cc1be4409c9790832856eb91d462d2da29dbe)

cmd/lifecycle/cli/command.go
```diff
@@ -55,18 +55,22 @@ func Run(c Command, asSubcommand bool) {
...
 	}
+	cmd.DefaultLogger.Debugf("Starting %s...", withPhaseName)
 
 	// Warn when CNB_PLATFORM_API is unset
 	if os.Getenv(platform.EnvPlatformAPI) == "" {
 		cmd.DefaultLogger.Warnf("%s is unset; using Platform API version '%s'", platform.EnvPlatformAPI, platform.DefaultPlatformAPI)
 		cmd.DefaultLogger.Infof("%s should be set to avoid breaking changes when upgrading the lifecycle", platform.EnvPlatformAPI)
 	}
 
+	cmd.DefaultLogger.Debugf("Parsing flags...")
```
natalieparellano, line 66:
```suggestion
	cmd.DefaultLogger.Debugf("Parsing inputs...")
```

PR #1109, opened 2023-06-02
metadata label is more clearly named and always has stacks

joe-kimmel-vmw:
see: https://github.com/buildpacks/lifecycle/issues/1057

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -123,9 +123,7 @@ func (e *Exporter) Export(opts ExportOptions) (files.Report, error) {
 		meta.RunImage.Image = opts.RunImageForExport.Image
 		meta.RunImage.Mirrors = opts.RunImageForExport.Mirrors
 		// ensure we always copy the new RunImage into the old stack to preserve old behavior
-		meta.Stack = &files.Stack{}
-		meta.Stack.RunImage.Image = meta.RunImage.Image
-		meta.Stack.RunImage.Mirrors = meta.RunImage.Mirrors
+		meta.Stack = opts.Stack
 	} else { // the past was a simpler time, but in fairness the eventual stack-less future will also be simpler.
 		meta.Stack = &opts.Stack
 	}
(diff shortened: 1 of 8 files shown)


exporter.go
```diff
@@ -115,9 +115,18 @@ func (e *Exporter) Export(opts ExportOptions) (files.Report, error) {
...
+			opts.RunImageForExport.Image = opts.Stack.RunImage.Image
+			opts.RunImageForExport.Mirrors = opts.Stack.RunImage.Mirrors
+		}
 		meta.RunImage.Image = opts.RunImageForExport.Image
 		meta.RunImage.Mirrors = opts.RunImageForExport.Mirrors
-	} else {
+		// ensure we always copy the new RunImage into the old stack to preserve old behavior
+		meta.Stack = &files.Stack{}
+		meta.Stack.RunImage.Image = meta.RunImage.Image
+		meta.Stack.RunImage.Mirrors = meta.RunImage.Mirrors
```
natalieparellano, line 128:
```suggestion
		meta.Stack = opts.Stack
```
This should be equivalent because of https://github.com/buildpacks/lifecycle/blob/28084b74124ad4484bd9e67ea4c469b05b5f38fa/cmd/lifecycle/exporter.go#L219

exporter.go
```diff
@@ -115,9 +115,18 @@ func (e *Exporter) Export(opts ExportOptions) (files.Report, error) {
 	meta.RunImage.Reference = opts.RunImageRef
 
 	if e.PlatformAPI.AtLeast("0.12") {
+		// if it's new API version but only using Stack still, then normalize the model by copying stack into the RunImage
+		if opts.RunImageForExport.Image == "" && len(opts.RunImageForExport.Mirrors) == 0 {
+			opts.RunImageForExport.Image = opts.Stack.RunImage.Image
+			opts.RunImageForExport.Mirrors = opts.Stack.RunImage.Mirrors
+		}
```
natalieparellano, line 122: Technically this backwards-compat layer is happening at builder creation, but by the time we get to the lifecycle I think it makes sense to just use the information from run.toml if we're on Platform API 0.12

natalieparellano, line 122:
```suggestion
```

natalieparellano (approved):

PR #1110, opened 2023-06-05
turn on additional linter checks including godoc comments

joe-kimmel-vmw:
The goal of this commit is to enforce stricter style, including godoc comments, but only going forward.

For reference, the linters enabled in this commit can be seen running against the current repo here:
https://github.com/buildpacks/lifecycle/actions/runs/5182278788/jobs/9338882119?pr=1110

but by specifying the `new-from-rev` option in golanci.yaml config we're able to say that these should only apply in the future. This functions as a "since" commit,  meaning "the linters only apply to commits since this one."

The soft-agreement of turning this on would be something like:
- the "since" commit is allowed to be bumped forward for large refactors where fixing 10s or 100s of linter errors is an unreasonable additional burden
- new features and bugfixes that incur only a handful of linter fixes should perform those linter fixes without moving the "since" commit.

In particular the linter that we're most interested in is enforcing godocs for Exported fields going forwards, which looks like this:

`exported: exported const TargetLabel should have comment (or a comment on this block) or be unexported (revive)`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
       - name: Set up go
         uses: actions/setup-go@v3
         with:
@@ -41,6 +43,8 @@ jobs:
     runs-on: linux-arm64
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
       - name: Set up go
         uses: actions/setup-go@v3
         with:
@@ -59,6 +63,8 @@ jobs:
           git config --global core.eol lf
           git config --global core.symlinks true
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
       - name: Set up go
         uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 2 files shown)


golangci.yaml
```diff
@@ -34,9 +34,11 @@ linters-settings:
 
 
 issues:
+  exclude-use-default: false
```
joe-kimmel-vmw, line 37: golangci lint overrides the default settings of revive and other sub-linters to be less pedantic. This setting is saying "don't override the defaults" or "don't exclude the default linter settings" and that's what turns back on godoc linting (among a few others, see earlier commits in this PR where the linters fail for full set of examples in our codebase)

.github/workflows/build.yml
```diff
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
```
joe-kimmel-vmw, line 19: iirc by default we do a super-shallow clone with no history at all, but if you do fetch-depth 0 honestly i forget what it means, but it means "fetch me enough git history that i can find the previous commit referenced in the `  new-from-rev` tag"

PR #1111, opened 2023-06-05
Improved matching when selecting run image metadata

natalieparellano:
We need to parse the reference in case the registry is not specified (defaults to Docker Hub)

### DIFF
--- a/platform/run_image.go
+++ b/platform/run_image.go
@@ -32,14 +32,15 @@ func GetRunImageForExport(inputs LifecycleInputs) (files.RunImageForExport, erro
 		return files.RunImageForExport{}, err
 	}
 	if len(runMD.Images) == 0 {
-		return files.RunImageForExport{}, err
+		return files.RunImageForExport{}, nil
 	}
+	inputRef := parseMaybe(inputs.RunImageRef)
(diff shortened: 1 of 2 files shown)


platform/run_image_test.go
```diff
@@ -52,27 +52,53 @@ func testRunImage(t *testing.T, when spec.G, it spec.S) {
 			})
 
 			when("contains an image matching run image ref", func() {
-				inputs.RunImageRef = "some-run-image-from-run-toml"
+				inputs.RunImageRef = "some-run-image-from-run-toml-1"
```
natalieparellano, line 55: Added `-1` because by default we return the first image in run.toml and the test still passes

platform/run_image.go
```diff
@@ -32,14 +32,15 @@ func GetRunImageForExport(inputs LifecycleInputs) (files.RunImageForExport, erro
 		return files.RunImageForExport{}, err
 	}
 	if len(runMD.Images) == 0 {
-		return files.RunImageForExport{}, err
+		return files.RunImageForExport{}, nil
```
joe-kimmel-vmw, line 35: it's not clear to me why this line was changed - is it covered by the tests below?

Issue #1098, opened 2023-05-05
Run image in `pack inspect` is not correctly updated after rebasing


harryli0108:
### Summary

Run image in `pack inspect <img-name-1>` is not correctly updated after running `pack rebase <img-name-1> --run-image <img-name-2>`.

---

##### Steps

1. Build an image with a buildpack and call it <img-name-1>
2. Build a run image with `docker build <run-img>` (Optional Step)
3. Run `pack inspect <img-name-1> --output json` and note the content in the run image part
4. Run `pack rebase <img-name-1> --run-image <run-img>` to rebase image 1 with a new run image
5. Run `pack inspect <img-name-1> --output json` and note the content in the run image part again. 

##### Current behavior

The run image field when running pack inspect is not reflected correctly after rebase command.
##### Expected behavior

The run image name and tag should be changed after rebasing with a new run image.

harryli0108:
Hi @natalieparellano and @hernandanielg. Thank you very much for the prompted replies and the quick PR! However, I'm still running into the same issue even with rebasing using the latest API version with the lts, prerelease, and the [tar.gz created from the PR merge Github workflow](https://github.com/buildpacks/lifecycle/actions/runs/5137487679). I use the following command to rebase the image using lifecycle:
`./lifecycle/rebaser -previous-image <image-name-getting-rebased>:<old-tag> <image-name-after-rebased>:<new-tag>` with the following environment variables set:
`CNB_PLATFORM_API=0.11` (for lts) or `CNB_PLATFORM_API=0.12` (for prerelease)
`CNB_USE_DAEMONO=false`
`CNB_RUN_IMAGE=<latest-run-image-we-released>`
`CNB_REGISTRY_AUTH={"<reg-url>":"<auth-type> <token>"}`

Afterwards, I run `pack inspect <image-name-after-rebased>:<new-tag>` to check the remote image info, and the run image is still using the old one. 

Could you take a look if possible? Much appreciated!

natalieparellano:
Thanks for the report @harryli0108 - I'll take a look

natalieparellano:
@harryli0108 I am probably doing something different than you, but I'm not seeing this behavior in my testing. I'm using the daemon here, but when I do:

```
$  pack inspect hello-extensions
Inspecting image: hello-extensions

REMOTE:
(not present)

LOCAL:

Stack: io.buildpacks.samples.stacks.alpine

Base Image:
[...]

PR #1132, opened 2023-06-22
Fixes for rebase

natalieparellano:
* When deciding whether to update run image metadata, consider the platform API __of the rebaser__ vs the platform API of the previous image
* Warn whenever `--force` (that is provided) is needed to successfully complete the operation

Related discussion: https://github.com/buildpacks/lifecycle/issues/1098#issuecomment-1602849946

### DIFF
--- a/rebaser.go
+++ b/rebaser.go
@@ -148,7 +148,7 @@ func containsName(origMetadata files.LayersMetadataCompat, newBaseName string) b
 	if origMetadata.Stack == nil {
 		return false
 	}
-	return origMetadata.Stack.ToRunImageForRebase().Contains(newBaseName)
+	return origMetadata.Stack.RunImage.Contains(newBaseName)
 }
 
 func validateStackID(appImg, newBaseImage imgutil.Image) error {
--- a/rebaser.go
+++ b/rebaser.go
@@ -92,6 +92,7 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 					newBaseImage.Name(),
 					existingRunImageMD,
 				)
+				// update original metadata
 				origMetadata.RunImage.Image = newBaseImage.Name()
 				origMetadata.RunImage.Mirrors = []string{}
 				newStackMD := origMetadata.RunImage.ToStack()
--- a/rebaser.go
+++ b/rebaser.go
@@ -74,7 +74,7 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 		return RebaseReport{}, fmt.Errorf("get run image id or digest: %w", err)
 	}
 	origMetadata.RunImage.Reference = identifier.String()
-	if appPlatformAPI != "" && r.PlatformAPI.AtLeast("0.12") {
+	if r.PlatformAPI.AtLeast("0.12") {
 		// update stack and runImage if needed
 		if !containsName(origMetadata, newBaseImage.Name()) {
 			var existingRunImageMD string
--- a/rebaser.go
+++ b/rebaser.go
@@ -18,6 +18,12 @@ import (
 	"github.com/buildpacks/lifecycle/platform/files"
 )
 
+var (
+	msgAppImageNotMarkedRebasable       = "app image is not marked as rebasable"
+	msgRunImageMDNotContainsName        = "rebase app image: new base image '%s' not found in existing run image metadata: %s"
+	msgUnableToSatisfyTargetConstraints = "unable to satisfy target os/arch constraints; new run image: %s, old run image: %s"
+)
+
 type Rebaser struct {
 	Logger      log.Logger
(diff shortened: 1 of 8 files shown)


rebaser.go
```diff
@@ -50,41 +56,63 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
...
+		return RebaseReport{}, fmt.Errorf("get rebase run image top layer SHA: %w", err)
 	}
 	identifier, err := newBaseImage.Identifier()
 	if err != nil {
-		return RebaseReport{}, errors.Wrap(err, "get run image id or digest")
+		return RebaseReport{}, fmt.Errorf("get run image id or digest: %w", err)
 	}
 	origMetadata.RunImage.Reference = identifier.String()
-	if appPlatformAPI != "" && api.MustParse(appPlatformAPI).AtLeast("0.12") {
+	if appPlatformAPI != "" && r.PlatformAPI.AtLeast("0.12") {
```
natalieparellano, line 77: This is the important change

rebaser.go
```diff
@@ -50,41 +56,63 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
...
-			origMetadata.RunImage.Mirrors = []string{}
-			newStackMD := origMetadata.RunImage.ToStack()
-			origMetadata.Stack = &newStackMD
+		if !containsName(origMetadata, newBaseImage.Name()) {
+			var existingRunImageMD string
+			switch {
+			case origMetadata.RunImage.Image != "":
+				existingRunImageMD = encoding.ToJSONMaybe(origMetadata.RunImage)
+			case origMetadata.Stack != nil:
+				existingRunImageMD = encoding.ToJSONMaybe(origMetadata.Stack.RunImage)
```
natalieparellano, line 85: Fallback to `stack` for older images

rebaser.go
```diff
@@ -163,34 +201,42 @@ func validateMixins(appImg, newBaseImg imgutil.Image) error {
 }
 
 func (r *Rebaser) validateTarget(appImg imgutil.Image, newBaseImg imgutil.Image) error {
-	rebaseable, err := appImg.Label(platform.RebaseableLabel)
+	rebasable, err := appImg.Label(platform.RebasableLabel)
```
natalieparellano, line 204: According to Google, this is the more popular spelling (and matches the RFC)

rebaser.go
```diff
@@ -50,41 +56,64 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
...
+		return RebaseReport{}, fmt.Errorf("get rebase run image top layer SHA: %w", err)
 	}
 	identifier, err := newBaseImage.Identifier()
 	if err != nil {
-		return RebaseReport{}, errors.Wrap(err, "get run image id or digest")
+		return RebaseReport{}, fmt.Errorf("get run image id or digest: %w", err)
 	}
 	origMetadata.RunImage.Reference = identifier.String()
-	if appPlatformAPI != "" && api.MustParse(appPlatformAPI).AtLeast("0.12") {
+	if appPlatformAPI != "" && r.PlatformAPI.AtLeast("0.12") {
```
jabrown85, line 77: Should we still be checking for the presences of `appPlatformAPI` in order to continue?

rebaser.go
```diff
@@ -50,41 +56,64 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
...
 		// update stack and runImage if needed
-		if !origMetadata.RunImage.Contains(newBaseImage.Name()) {
-			origMetadata.RunImage.Image = newBaseImage.Name()
-			origMetadata.RunImage.Mirrors = []string{}
-			newStackMD := origMetadata.RunImage.ToStack()
-			origMetadata.Stack = &newStackMD
+		if !containsName(origMetadata, newBaseImage.Name()) {
```
jabrown85, line 79: Should there be an else case here that also updates the metadata? If, for instance, there was an app image that had something like

`mybase/blah` and mirror of `mybase-mirror/blah` and now they've changed to `mybase/blah` and `mybase-mirrornew/blah`. I think we'd want to re-write the metadata to match so that the new mirror is propagated forward.

rebaser.go
natalieparellano: Oh whoops... thank you!

natalieparellano: Updated in f29ebbb6

rebaser.go
natalieparellano: But how would the new mirror be communicated to the rebaser? We can learn about a new run image via `-run-image` or `CNB_RUN_IMAGE`, but where would the mirror information come from?

jabrown85: oh ok - I think I got confused a bit.  Ignore this 😄

jabrown85 (approved):

PR #1133, opened 2023-06-26
When checking read/write access, if there is no access, surface the e…

natalieparellano:
…rror that occurred

This can be helpful in debugging failed builds where access is expected

### DIFF
--- a/cmd/lifecycle/main.go
+++ b/cmd/lifecycle/main.go
@@ -143,7 +143,9 @@ func verifyReadAccess(imageRef string, keychain authn.Keychain) error {
 		return nil
 	}
 	img, _ := remote.NewImage(imageRef, keychain)
-	if !img.CheckReadAccess() {
+	canRead, err := img.CheckReadAccess()
+	if !canRead {
+		cmd.DefaultLogger.Debugf("Error checking read access: %s", err)
 		return errors.Errorf("ensure registry read access to %s", imageRef)
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


jabrown85 (approved):
LGTM

platform/run_image.go
```diff
@@ -78,31 +78,23 @@ func BestRunImageMirrorFor(targetRegistry string, runImageMD files.RunImageForEx
 
 	// Select the first run image we have access to
 	for _, image := range runImageMirrors {
-		ok, err := checkReadAccess(image, keychain)
-		if err != nil {
-			return "", err
-		}
-		if ok {
+		if ok, _ := checkReadAccess(image, keychain); ok {
```
natalieparellano, line 81: Ideally we'd log the error at this point, but that requires a logger, which is a bigger change

cmd/lifecycle/main.go
```diff
@@ -143,7 +143,9 @@ func verifyReadAccess(imageRef string, keychain authn.Keychain) error {
 		return nil
 	}
 	img, _ := remote.NewImage(imageRef, keychain)
-	if !img.CheckReadAccess() {
+	canRead, err := img.CheckReadAccess()
+	if !canRead {
+		cmd.DefaultLogger.Debugf("Error checking read access: %s", err)
```
cmoulliard, line 148: We should log that without using Debug as some users using Tekton or kpack could not be aware about how to enable log level !!

cmd/lifecycle/main.go
```diff
@@ -154,7 +156,9 @@ func verifyReadWriteAccess(imageRef string, keychain authn.Keychain) error {
 		return nil
 	}
 	img, _ := remote.NewImage(imageRef, keychain)
-	if !img.CheckReadWriteAccess() {
+	canReadWrite, err := img.CheckReadWriteAccess()
+	if !canReadWrite {
+		cmd.DefaultLogger.Debugf("Error checking read/write access: %s", err)
```
cmoulliard, line 161: We should log that without using Debug as some users using Tekton or kpack could not be aware about how to enable log level !!

PR #1136, opened 2023-06-29
Stop deleting cache images

### LINKED ISSUE #803: [RFC 0116] Stop deleting cache images for newer Platform API versions
### Description

AWS ECR does not support the delete image registry API used by pack. See https://github.com/aws/containers-roadmap/issues/1229

When publishing a cache image to an ECR registry, users get the following error:

```
Unable to delete previous cache image: DELETE https://xxx.dkr.ecr.us-east-1.amazonaws.com/v2/ecrrepository-ns9dbfvxadcv/manifests/sha256:a16e6b50e11222fe92b2e1adb357c351b352aed1f165254b56919dec23779a23: unexpected status code 404 Not Found: 404 page not found
```

It doesn't seem to cause any issues, but looks like a problem for end-users.

### Proposed solution

Well ideally AWS would fix their implementation, but the issue has been open for a year with no movement. Perhaps pack could detect ECR repos (maybe by an URL regex 🤷) and skip this step or perform the deletion by other means.

I just ignore the log line and move on.

### Additional context
- [ ] This feature should be documented somewhere

dlion:
With this PR the cache will just override the original image with the new image without attempting to delete it first.

* Original Issue: https://github.com/buildpacks/lifecycle/issues/803
* Readable RFC: https://github.com/buildpacks/rfcs/blob/5c1d82583b0baf1e1fbb6cd2d6eaa201aabaa276/text/0000-stop-deleting-cache-images.md
* Tracking RFC issue: https://github.com/buildpacks/rfcs/issues/268

Resolves #803 and it partially resolves https://github.com/buildpacks/rfcs/issues/268


jabrown85 (approved):
This looks great! Did we say we didn't need to guard this behind an API version @natalieparellano? I don't have a strong opinion. The delete is not in spec.

jjbustamante:
@jabrown85 @natalieparellano 

One doubt about this based on the [motivation](https://github.com/buildpacks/rfcs/blob/5c1d82583b0baf1e1fbb6cd2d6eaa201aabaa276/text/0000-stop-deleting-cache-images.md#motivation) section in the RFC.

How does it impact pack users? Do we need to advertise to users they need to delete or take care of the cache images saved in the registry? or do we need to create an issue on pack to think about this and if we want to offer some feature to manage the cache?

No related [directly](https://cloud-native.slack.com/archives/C0331B61A1Y/p1683817627443339) to this conversation on slack but just wanted to keep track of it

jabrown85:
> How does it impact pack users? Do we need to advertise to users they need to delete or take care of the cache images saved in the registry? or do we need to create an issue on pack to think about this and if we want to offer some feature to manage the cache?

I _think_ that is up to `pack` on how to handle that. If `pack` wants to continue the existing experience then it can coded directly into `pack` to handle the registry operations required. If we aren't worried about this change for `pack` users, a note in the subsequent release maybe? At least, that is my thinking. 

I would be 👍  holding this until `pack` team decides a path though as `pack` is our number one platform 😄

dlion:
@jabrown85 
I agree on let pack decides how to handle it, is there anything I should do to speed up the decision process? 
Of course I'm happy to open a PR to pack if the team decides to go in that direction

dlion:
UPDATE: I requested to talk about it in the next working group
(https://cloud-native.slack.com/archives/C0331B61A1Y/p1688459224497819)

cache/image_cache.go
```diff
@@ -111,36 +111,20 @@ func (c *ImageCache) Commit() error {
...
 
+func saveImage(c *ImageCache) error {
 	if err := c.newImage.Save(); err != nil {
 		return errors.Wrapf(err, "saving image '%s'", c.newImage.Name())
 	}
-	c.committed = true
 
-	if origImgExists {
-		// Deleting the original image is for cleanup only and should not fail the commit.
-		if err := c.DeleteOrigImage(); err != nil {
```
natalieparellano, line 124: Let's gate this on a Platform API change, so that it doesn't surprise end users.

jabrown85:
As @natalieparellano pointed out in working group - we previously decided to gate this new behavior on platform api version @dlion. Let's do that for our platform and builder operators.

cache/image_cache.go
dlion: Sure, I was thinking to put this change into the https://github.com/buildpacks/spec/blob/main/platform.md#caching section.
What do you think?

dlion:
> As @natalieparellano pointed out in working group - we previously decided to gate this new behavior on platform api version @dlion. Let's do that for our platform and builder operators.

Sure, I'll comment this PR whenever I open the spec PR so we can be in the same page

cache/image_cache.go
dlion: Okay now I got it, I've got confused by the `gate` name which usually I called it `feature flag`, sorry for the misunderstanding.
All right, I'm gonna put this behind a feature flag and update the Spec accordingly :)

cache/cache_deleter.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/buildpacks/imgutil"
+	"github.com/pkg/errors"
+
+	"github.com/buildpacks/lifecycle/log"
+)
+
+// ImageDeleter defines the methods available to delete and compare cached images
+type ImageDeleter interface {
+	DeleteImage(image imgutil.Image)
+	OldImageIsTheSameAsTheNewImage(oldImage imgutil.Image, newImage imgutil.Image) bool
```
joe-kimmel-vmw, line 14: In this interface is the "old" and "new" important or could this method be given a shorter name like `ImagesEq` or `Equals` ?

dlion:
All right, after a bit of refactoring now the PR should be ready to be reviews 😄 
What I changed:
* Now we just need to pass the ApiPlatform evaluated condition a the very first level and not pass the whole object all the way down.
* The boolean flag helps us to abstract the ApiPlatform concept from the technical implementation of the deleter allowing us in the future to activate/deactivate the deletion using other factors like for example an external conf or test it without the need of using mocks.
* The check is now done in an isolated and specialized component

Thoughts are welcomed 😃

analyzer.go
```diff
@@ -151,7 +151,7 @@ func (f *AnalyzerFactory) setBuildpacks(analyzer *Analyzer, group buildpack.Grou
 
 func (f *AnalyzerFactory) setCache(analyzer *Analyzer, imageRef string, dir string) error {
 	var err error
-	analyzer.Cache, err = f.cacheHandler.InitCache(imageRef, dir)
+	analyzer.Cache, err = f.cacheHandler.InitCache(imageRef, dir, f.platformAPI.LessThan("0.13"))
```
natalieparellano, line 154: Nice :)

cmd/lifecycle/main.go
```diff
@@ -91,13 +91,15 @@ func NewCacheHandler(keychain authn.Keychain) *DefaultCacheHandler {
 	}
 }
 
-func (ch *DefaultCacheHandler) InitCache(cacheImageRef string, cacheDir string) (lifecycle.Cache, error) {
+// InitCache it's a factory used to create either a NewImageCache or a NewVolumeCache
```
natalieparellano, line 94:
```suggestion
// InitCache is a factory used to create either a NewImageCache or a NewVolumeCache
```

cache/image_deleter_test.go
```diff
@@ -0,0 +1,119 @@
...
+			cacheDeleter := NewImageDeleter(mockLogger, true)
+
+			cacheDeleter.DeleteImage(fakeErrorImage)
+
+			h.AssertEq(t, mockLogger.Calls, 1)
+		})
+	})
+
+	when("delete functionality has been deactivated", func() {
+		it("should avoid performing deleting operations", func() {
```
natalieparellano, line 55:
```suggestion
		it("should not perform deleting operations", func() {
```
nit

natalieparellano (commented):
@dlion nice work! I made a few suggestions - LMK your thoughts :)

cache/image_deleter.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/buildpacks/lifecycle/log"
+)
+
+//go:generate mockgen -package testmock -destination ../testmock/image_deleter.go github.com/buildpacks/lifecycle/cache ImageDeleter
+
+// ImageDeleter defines the methods available to delete and compare cached images
+type ImageDeleter interface {
+	DeleteImage(image imgutil.Image)
+	ImagesEq(oldImage imgutil.Image, newImage imgutil.Image) (bool, error)
```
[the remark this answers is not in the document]
dlion: Uhm, yeah I guess we can make it available outside of the interface signature but I need to think what would be the implication to put the `ImagesEq` function inside the `DeleteImage` or a wrapper method to do both (something I thought to do as some point) as it does something different than just deleting the image. 🤔 
But to do so, I have to change how I implemented it inside the `image_cache.Commit`method since it uses both methods.
Considering that this component is quite well separated I didn't think to change the signature but now that you are pointing this out could have sense

cache/image_cache_test.go
```diff
@@ -291,74 +297,5 @@ func testImageCache(t *testing.T, when spec.G, it spec.S) {
...
-					fakeNewImage = fakes.NewImage("fake-image", "", local.IDIdentifier{ImageID: "fakeOrigImage"})
-					subject = cache.NewImageCache(fakeOriginalImage, fakeNewImage, testLogger)
-				})
-				it("should not delete original image", func() {
-					err := subject.DeleteOrigImage()
-					h.AssertNil(t, err)
-					h.AssertEq(t, fakeOriginalImage.Found(), true)
-				})
-			})
-		})
```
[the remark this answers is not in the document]
natalieparellano: Sorry, where? Can you provide the line number? I looked but I didn't see it :/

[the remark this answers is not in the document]
natalieparellano: In particular I'm thinking about `when("original and new image are the same"` `it("should not delete original image"`

cache/image_deleter.go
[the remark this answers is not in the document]
natalieparellano: I'm thinking something like
```
type ImageDeleter interface {
	DeleteImageIfExists(image imgutil.Image)
}
```
...and not exposing any of `ImagesEq` to the `cache` package

cache/image_cache_test.go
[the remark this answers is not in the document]
dlion: Oh sorry I misinterpreted your comment, you were totally right!
Before forgot to un-stash it but now with the new implementation we should be covered 😄

cache/image_deleter.go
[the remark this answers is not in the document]
dlion: Thank you for pointing this out @natalieparellano !
Indeed the ImagesEq was breaking the [Interface Segretation Principle](https://en.wikipedia.org/wiki/Interface_segregation_principle) and [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) principle.
I changed the implementation following the Dependency Inversion Principle maintaining a clear separation of responsibilities allowing more flexibility in choosing different image comparing strategies without affecting the `ImageDeleter` interface using the `ImageComparer` component.

As a future improvement we can move this component in the `imageutil` library.

Let me know what do you think 😄

natalieparellano (approved):
@dlion thanks for your work on this! It looks good to me. 

@jabrown85 did you want to have another look?

jabrown85 (approved):

natalieparellano:
@dlion are you able to fix the DCO?

dlion:
@natalieparellano there was a commit not signed-off, fixed it 😄 I think then we can merge 🎉

dlion:
@natalieparellano Codecov failed again 🫠 we should reproduce what we have done with pack (https://github.com/buildpacks/pack/pull/1809).
Can you re-run the Github Actions?

Issue #1138, opened 2023-07-03
ENV Vars are harder to discover than cmd line options


joe-kimmel-vmw:
### Description
suppose you run a command like:
```
CNB_USER_ID=1000 CNB_GROUP_ID=1000 CNB_PLATFORM_API=0.12 ../lifecycle/out/darwin-amd64/lifecycle/detector --log-level debug -layers="."  -app apps/bash-script
```

An astute maintainer might notice that there's no order.toml specified, and the user will in fact get an error message relating to order.toml.   However, unlike the way for setting the "layers" directory, the way for setting the order.toml path is not discoverable by running `detector --help`. 

### Proposed solution
- we could include the env var in the error message.

### Describe alternatives you've considered
- Is there a general way to add env vars to the cobra help output?

joe-kimmel-vmw:
hmmm it turns out just in general even the cmd line options require careful reading to discover , since there totally is one in this case.

PR #1139, opened 2023-07-03
adds CNB_ORDER_PATH env var hint to error message when order.toml fil…

joe-kimmel-vmw:
…e can't be found

see: https://github.com/buildpacks/lifecycle/issues/1138

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -1,7 +1,9 @@
 package lifecycle
 
 import (
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


jabrown85 (approved):

natalieparellano:
Do we still want this one? I think we originally thought that the order path was a special case, but then we realized that it is also a flag (like the other paths). I'm going to put this in draft for now...

PR #1140, opened 2023-07-04
Support insecure registries

dlion:
This PR Implements the [support for insecure registries](https://github.com/buildpacks/rfcs/blob/main/text/0111-support-insecure-registries.md).

## Sample command
`creator --run-image=develoment-registry.com/run-java:v16 --insecure-registry=develoment-registry.com --insecure-registry=testing-registry.com testing-registry.com/java-sample:latest`

## Others

Tracking issue: https://github.com/buildpacks/rfcs/issues/246


dlion:
FIY: This PR has been paused due of the need to have a test utility on the imgutil repo

dlion:
This is failing due of the line https://github.com/buildpacks/lifecycle/actions/runs/5669003123/job/15360828068?pr=1140#step:5:4517 which panic due of the missing initialization of the map inside the `WithRegistrySettings` method.
Fixed here: https://github.com/buildpacks/imgutil/pull/211

platform/lifecycle_inputs.go
```diff
@@ -85,11 +86,12 @@ func NewLifecycleInputs(platformAPI *api.Version) *LifecycleInputs {
...
-		LogLevel:    envOrDefault(EnvLogLevel, DefaultLogLevel),
-		PlatformAPI: platformAPI,
-		ExtendKind:  envOrDefault(EnvExtendKind, DefaultExtendKind),
-		UseDaemon:   boolEnv(EnvUseDaemon),
-		UseLayout:   boolEnv(EnvUseLayout),
+		LogLevel:         envOrDefault(EnvLogLevel, DefaultLogLevel),
+		PlatformAPI:      platformAPI,
+		ExtendKind:       envOrDefault(EnvExtendKind, DefaultExtendKind),
+		UseDaemon:        boolEnv(EnvUseDaemon),
+		InsecureRegistry: nil,
```
natalieparellano, line 93: Should this be `InsecureRegistries`?

platform/lifecycle_inputs_test.go
```diff
@@ -172,6 +175,7 @@ func testLifecycleInputs(t *testing.T, when spec.G, it spec.S) {
 				h.AssertEq(t, inputs.UID, 1234)
 				h.AssertEq(t, inputs.UseDaemon, true)
 				h.AssertEq(t, inputs.UseLayout, true)
+				h.AssertEq(t, inputs.InsecureRegistry, str.Slice(nil))
```
natalieparellano, line 178: Shouldn't this expect `some-insecure-registry`?

cmd/lifecycle/cli/flags.go
```diff
@@ -168,6 +168,11 @@ func FlagForceRebase(force *bool) {
 	flagSet.BoolVar(force, "force", *force, "execute rebase even if operation is unsafe")
 }
 
+// FlagInsecureRegistry sets insecure-registry parameter as available
+func FlagInsecureRegistry(insecureRegistry *str.Slice) {
```
natalieparellano, line 172: I think according to the RFC we also want to add this to: exporter, restorer, rebaser

image/remote_handler.go
```diff
@@ -9,18 +11,31 @@ import (
...
 	}
 
+	options := []remote.ImageOption{
+		remote.FromBaseImage(imageRef),
+	}
+
+	if len(h.insecureRegistries) > 0 {
+		for _, insecureRegistry := range h.insecureRegistries {
+			if strings.HasPrefix(imageRef, insecureRegistry) {
+				options = append(options, remote.WithRegistrySetting(insecureRegistry, true, true))
```
natalieparellano, line 30: The RFC says `The second parameter specify whether the registry is insecure. The third parameter can be always false.` ...which, it's a little confusing - why have a third parameter if it's always false? According to imgutil the third parameter disables SSL certificate validation. I wonder if we should try exposing it to the end user somehow...

cmd/lifecycle/cli/flags.go
```diff
@@ -168,6 +168,11 @@ func FlagForceRebase(force *bool) {
...
+	flagSet.Var(insecureRegistry, "insecure-registry", "to use 'insecure' registries")
```
natalieparellano, line 173:
```suggestion
	flagSet.Var(insecureRegistry, "insecure-registry", "insecure registries")
```
(for parity with `FlagTags` which works similarly)

natalieparellano (commented):
Apologies for being slow to review @dlion. I left a few comments... let's discuss how we might be able to test this out at the acceptance level.

platform/lifecycle_inputs_test.go
natalieparellano: We should probably also check that we can parse a comma separated list...

platform/lifecycle_inputs.go
dlion: Uhm, in the RFC the flag is `--insecure-registry` so I tried to kept that.
Because if I changed it, we would have something like:
```
FlagInsecureRegistries(insecureRegistries *str.Slice) {
	flagSet.Var(insecureRegistries, "insecure-registry", "to use 'insecure' registries")
}
```
But it is kind of misleading because the name and the param are about insecureRegistries but the name of the param is `insecure-registry` which can cause confusion.
For this reason I kept InsecureRegistry on the input parts and then used insecureRegistries only in the image part.
What do you think?
Should I keep everything plural or singular? Should we change the parameter name?

dlion: I'm gonna change the name and push it, so we can see how it looks like :)

image/remote_handler.go
dlion: Yeah, my thought here was that if we are using `WithRegistrySetting` it means we need to skip any SSL validation so I left that as a true.
What do you think? I agree that if we are considering of using always a specific value we should change the signature on imgutil

platform/lifecycle_inputs_test.go
dlion: Yeah, I tried to copy the UseExtendendFlags way but I got it wrong.
Thanks for pointing this out Natalie!

cmd/lifecycle/cli/flags.go
dlion: You are right, I wanted first to be sure of doing the right thing in the creator.
I pushed some changes and now I'm gonna work on the other parts starting from some tests, thanks!

dlion:
Uhm, locally the `TestCreator` is passing, same for the `amd64` on the CI so I guess it's flaky.
@natalieparellano Can you have a look on my latest changes to see if the code makes sense and I haven't forgot anything?
After your approval I will go starting investigating on the default values for the `WithRegistrySettings` function.

.gitignore
```diff
@@ -6,9 +6,18 @@
...
 acceptance/testdata/exporter/container/layers/*analyzed.toml
 acceptance/testdata/exporter/container/other_layers/*analyzed.toml
+
 acceptance/testdata/restorer/container/layers/*analyzed.toml
+
+acceptance/testdata/rebaser/container/cnb/run.toml
+acceptance/testdata/rebaser/container/layers/analyzed.toml
+acceptance/testdata/rebaser/container/layers/run-image-extended-analyzed.toml
+acceptance/testdata/rebaser/container/layers/some-analyzed.toml
+acceptance/testdata/rebaser/container/other_layers/analyzed.toml
```
natalieparellano, line 23: I think we can remove these, since the files were removed

acceptance/restorer_test.go
```diff
@@ -34,7 +34,6 @@ var (
 
 func TestRestorer(t *testing.T) {
 	h.SkipIf(t, runtime.GOOS == "windows", "Restorer acceptance tests are not yet supported on Windows")
-	h.SkipIf(t, runtime.GOARCH != "amd64", "Restorer acceptance tests are not yet supported on non-amd64")
```
natalieparellano, line 37: Woo!

image/handler_test.go
```diff
@@ -0,0 +1,68 @@
...
+	testmockauth "github.com/buildpacks/lifecycle/testmock/auth"
+)
+
+//go:generate mockgen -package testmockauth -destination ../testmock/auth/mock_keychain.go github.com/google/go-containerregistry/pkg/authn Keychain
+
+func TestHandler(t *testing.T) {
+	spec.Run(t, "ImageHandler", testHandler, spec.Sequential(), spec.Report(report.Terminal{}))
+}
+
+func testHandler(t *testing.T, when spec.G, it spec.S) {
```
natalieparellano, line 22: Nice, thanks for these :)

natalieparellano (commented):
Looking good @dlion - and you added an acceptance test for every phase 🥇 

Thanks for being so thorough and for making the code incrementally better :)

cmd/lifecycle/main.go
```diff
@@ -113,18 +113,20 @@ func (ch *DefaultCacheHandler) InitCache(cacheImageRef string, cacheDir string,
...
-func NewRegistryHandler(keychain authn.Keychain) *DefaultRegistryHandler {
+func NewRegistryHandler(keychain authn.Keychain, insecureRegistries []string) *DefaultRegistryHandler {
 	return &DefaultRegistryHandler{
-		keychain: keychain,
+		keychain:         keychain,
+		insecureRegistry: insecureRegistries,
 	}
 }
 
 func (rv *DefaultRegistryHandler) EnsureReadAccess(imageRefs ...string) error {
```
dlion, line 127: This check makes difficult to test the `insecure-registry` feature properly so for now I'm skipping it.
We are doing this check in the analyzer only (and used also in the creator of course making the acceptance test difficult to create)

dlion:
### Some update on the refactoring

I wanted to use the `RegistryHandler` in the `image.NewHandler` using the `RegistryHandler.GetInsecureRegistryOptions` to get the RegistrySetting option that we can use in the `remote.NewImage` call.

I tried to start creating a test for it but then I realised that the `WithRegistrySetting` [function implementation](https://github.com/buildpacks/imgutil/blob/main/remote/options.go#L100) returns an anonymous function that receive a parameter of type [options](https://github.com/buildpacks/imgutil/blob/main/remote/options.go#L13C6-L13C13) which is a private struct making it impossible to use/mock it from lifecycle.

Refactoring this part would require lots of time and effort, so I decided to leave it there, trying to minimize the duplication as much as possible and then open a new issue to address this problem as a future improvement.

image/registry_handler.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, imageRef := range imageRefs {
+		if err := verifyReadWriteAccess(imageRef, rv.keychain, rv.GetInsecureRegistryOptions(imageRef)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// GetInsecureRegistryOptions returns a list of WithRegistrySetting imageOptions matching the specified imageRef prefix
+func (rv *DefaultRegistryHandler) GetInsecureRegistryOptions(imageRef string) []remote.ImageOption {
```
natalieparellano, line 56: nit: since this is already a "registry" handler maybe this could just be `GetInsecureOptions`

natalieparellano (commented):
Amazing work @dlion! I left a few comments. Overall this is looking really great.

go.mod
```diff
@@ -5,7 +5,7 @@ require (
 	github.com/GoogleContainerTools/kaniko v1.13.0
 	github.com/apex/log v1.9.0
 	github.com/awslabs/amazon-ecr-credential-helper/ecr-login v0.0.0-20230522190001-adf1bafd791a
-	github.com/buildpacks/imgutil v0.0.0-20230626185301-726f02e4225c
+	github.com/buildpacks/imgutil v0.0.0-20230802154001-5c57feb120b3
```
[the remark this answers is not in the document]
dlion: Yes, I was planning to do so once you were back and the PR was merged 😄 
Thanks for reminding me!

image/registry_handler.go
dlion: I like it!

cmd/lifecycle/restorer.go
```diff
@@ -281,3 +290,16 @@ func (r *restoreCmd) restore(layerMetadata files.LayersMetadata, group buildpack
 	}
 	return nil
 }
+
+func (r *restoreCmd) getInsecureRegistryOptions(imageRef string) []remote.ImageOption {
```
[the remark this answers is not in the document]
dlion: yes, this could be a temporary solution but I'm still in favour of fixing the `options` struct visibility in the imgutil library in order to be able to use the `InsecureHandler` everywhere we need it in lifecycle.
I'll do it but I'll add a @TODO or a //FIXME as a reminder :)

image/registry_handler.go
```diff
@@ -0,0 +1,99 @@
...
+/*
+TODO: This is a temporary solution in order to get insecure registries in other components too
+TODO: Ideally we should fix the `imgutil.options` struct visibility in order to mock and test the `remote.WithRegistrySetting`
+TODO: function correctly and use the RegistryHandler everywhere it is needed.
+*/
+func GetInsecureOptions(insecureRegistries []string, imageRef string) []remote.ImageOption {
+	var opts []remote.ImageOption
+	if len(insecureRegistries) > 0 {
+		for _, insecureRegistry := range insecureRegistries {
+			if strings.HasPrefix(imageRef, insecureRegistry) {
```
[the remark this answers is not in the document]
dlion: Uhm, the insecure-registry feature disables any certificate checks and it should be used only in certain occasion (like testing and towards a registry in a secure network).

This check has been done to verify that one of the `--insecure-registry=<registry.address>` is the same as the one provided in the refImage (like `<registry.address>/app-name`)

So personally I would keep that check just to avoid any human error.
Does it make sense?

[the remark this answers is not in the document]
jabrown85: Does the `insecureRegistry` on `remote.WithRegistrySetting(insecureRegistry, true)` not limit this to the registry internally?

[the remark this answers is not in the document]
dlion: Uhm not that I'm aware of: https://github.com/buildpacks/imgutil/blob/main/remote/options.go#L101C6-L101C25

But I guess as a next improvement we can integrate this check inside the `NewImage` function if we want to

cmd/lifecycle/rebaser.go
```diff
@@ -163,10 +164,17 @@ func (r *rebaseCmd) setAppImage() error {
 		if err != nil {
 			return err
 		}
+
+		var opts = []remote.ImageOption{
+			remote.FromBaseImage(targetImageRef),
+		}
+
+		opts = append(opts, image.GetInsecureOptions(r.InsecureRegistries, targetImageRef)...)
```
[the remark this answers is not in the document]
dlion: Uhm good catch, I'll add it to the `runImageRef` as well, thanks!
For the multiple insecure registries the feature works already with multiple insecure registries, did I need to implement a different way to handle this specific situation in the rebaser? 🤔

jabrown85:
I am a bit confused. There are 2 distinct insecure paths that seemingly are tied to a single setting.

The first path is accessing http only registries, which seems to work in imgutil with
```golang
        destinationImageName := "myregistry.domain.com:6443/test/test:latest"
	remoteImage, err := remote.NewImage(destinationImageName,
		authn.DefaultKeychain,
		remote.FromBaseImage(baseImageName),		
		remote.WithRegistrySetting("myregistry.domain.com:6443", true),
	)
```

However, if `myregistry.domain.com:6443` only accepts HTTPS requests - this does _not_ work. Am I missing a way to _only_ set TLSSkipVerify? I still need to speak over HTTPS, but am OK skipping the TLS verification step for a self-signed situation.

image/registry_handler.go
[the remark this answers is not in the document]
jabrown85: I tried spinning up two http registries and couldn't find a combination of `remote` options to make it work as I expected. In this example I have 2 insecure registries and I would like to be able to use them both in creation of a single image.

Note: I'm using imgutil here and not lifecycle, as I wanted to see how the registry primitives work.

```
	baseImageName := "myregistry.domain.com:6443/test/test:latest"
	destinationImageName := "myregistry.domain.com:7443/test/test:latest"

	remoteImage, err := remote.NewImage(destinationImageName,
		authn.DefaultKeychain,
		remote.FromBaseImage(baseImageName),
		remote.WithRegistrySetting("myregistry.domain.com:6443", true),
		remote.WithRegistrySetting("myregistry.domain.com:7443", true),
	)
```

The image registry setting for the base repo is not applied. I always get

```
connect to repo store "myregistry.domain.com:6443/test/test:latest": Get "https://myregistry.domain.com:6443/v2/": http: server gave HTTP response to HTTPS client
```

I can get a single registry setting to work - but not both. Maybe this is expected and I'm doing something wrong though.

dlion:
>  I still need to speak over HTTPS, but am OK skipping the TLS verification step for a self-signed situation.
Uhm, we haven't considered that scenario, we thought that if you are using insecure-registry it means you are doing that for testing purposes or an internal network where having a self-sign certificated doesn't really matter.

Here the PR I opened to simplify those parameters into a single one trying to emulate what Google did: https://github.com/buildpacks/imgutil/pull/218

Do you think we should change this implementation giving that scenario?

I also mention @natalieparellano since we had this conversation a few weeks ago (and from that the PR)

image/registry_handler.go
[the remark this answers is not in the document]
dlion: Uhm, according to what I tested on lifecycle:
1. check the registry with https failing
2. try with http

So I guess it just use the first one and stop whenever it succeed.
Did you try to make the first failing but the second to succeed?

btw thank you for these tests! I didn't spawn up an insecure-registry with imgutil, (if you have any doc I can follow to reproduce those steps would be great!), I just relied on lifecycle acceptance tests

jabrown85:
Hmm, maybe we need to use the `remote.DefaultTransport` instead of `http.Transport` in imgutil. The transport looks to have some magic in it that may help with the http/https negotiation.

jabrown85:
Ok - I think I found part of the problem. The `doSave` [here](https://github.com/buildpacks/imgutil/blob/main/remote/save.go#L92) needs the same `if insecure { InsecureSkipVerify stuff }` logic. When you create a remote image with a base image, it hits the `InsecureSkipVerify` path on reading from that registry, but doesn't do anything with `insecure` during the save process. 

I do wonder if there are other operations that need the `InsecureSkipVerify` that we don't have yet in imgutil remote. Found/Delete/etc.

I do believe we still need to be able to have multiple `remote.WithRegistrySetting` to in `remote.NewImage`. The current pattern only lets a single registry be marked as insecure/tlsnoverify AFAICT.

image/registry_handler.go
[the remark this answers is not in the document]
jabrown85: I followed the instructions [here](https://docs.docker.com/registry/insecure/#use-self-signed-certificates) for spinning up a registry with a self-signed cert. Create the cert, update your host file, and then I spun up a few in different terminals like this.

```
docker run -it -v "$(pwd)"/certs:/certs \
  -e REGISTRY_HTTP_ADDR=0.0.0.0:443 \
  -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
  -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
  -p 6443:443 \
  registry:2

docker run -it -v "$(pwd)"/certs:/certs \
  -e REGISTRY_HTTP_ADDR=0.0.0.0:443 \
  -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
  -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
  -p 7443:443 \
  registry:2


docker run -it  -e REGISTRY_HTTP_ADDR=0.0.0.0:8080 \
  -p 8080:8080 \
  registry:2

docker run -it  -e REGISTRY_HTTP_ADDR=0.0.0.0:8080 \
  -p 9080:8080 \
  registry:2
```

This way I had a combination of HTTP/TLS registries to play around with as target/base registries. I didn't do a full test matrix or anything at this point. We can use `crane` to compare behavior as well when it makes sense. If you aren't diving back into `imgutil` yourself, let me know and I will try to make some time to make changes and add tests there soon™️ . I think we have to get `imgutil` in a good spot before we land any lifecycle changes in case the interfaces need to change.

dlion:
Amazing, thanks @jabrown85 !

I will add the transport (with the InsecureSkipVerify) to the doSave as well. 👏🏼 

Maybe you are right, we need to add the insecureSkipVerify also in operations like `found`, `valid`, and `delete`.
I'm going to open a new PR for that. 😄

image/registry_handler.go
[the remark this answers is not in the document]
dlion: Great thanks!
Indeed I'm still trying to understand how imgutil works under the hood and tested that parts wasn't really easy so your help would be really appreciated and yes I agree with you that we should have always imgutil in a good shape being a very important library for lifecycle.
If you have time to adjust/test it that would be awesome, I can learn a bit more about its internals 😄

dlion:
Here we go: https://github.com/buildpacks/imgutil/pull/219
@jabrown85 let me know if you notice something else I need to change in order to move forward this PR, thank you! 😄

go.mod
```diff
@@ -5,7 +5,7 @@ require (
 	github.com/GoogleContainerTools/kaniko v1.13.0
 	github.com/apex/log v1.9.0
 	github.com/awslabs/amazon-ecr-credential-helper/ecr-login v0.0.0-20230522190001-adf1bafd791a
-	github.com/buildpacks/imgutil v0.0.0-20230626185301-726f02e4225c
+	github.com/buildpacks/imgutil v0.0.0-20230908131741-6a790d131d36
```
natalieparellano, line 8: Can we bump the version again?

image/registry_handler.go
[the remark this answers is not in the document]
natalieparellano: Would this be fixed by https://github.com/buildpacks/imgutil/pull/220?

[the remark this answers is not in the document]
jabrown85: I'm not sure. I still think we need to always `remote.WithRegistrySetting(insecureRegistry, true)` for ALL the insecure registries all the time. The contains check doesn't make sense to me, but I may be missing something.

In the case of creating a new image in an insecure registry, but using a _different_ insecure registry as the base, could we do something like this?

```golang
appImage, err := remote.NewImage(
		e.OutputImageRef,
		e.keychain,
                image.GetInsecureOptions(r.InsecureRegistries), // sets up ALL the insecure registries that may be used
                remote.FromBaseImage(r.RunImageRef),		
	)
```

If we make it so that the caller has to pass the image into `image.GetInsecureOptions` for each image, it means we it would look like this.

```golang
appImage, err := remote.NewImage(
		e.OutputImageRef,
		e.keychain,
                image.GetInsecureOptions(r.InsecureRegistries, e.RunImageRef),
                image.GetInsecureOptions(r.InsecureRegistries, e.OutputImageRef),
		opts...,
	)
```

That seems harder to do for no benefit. I don't think there is any harm in always having the map with all the possible insecure registries..even if NONE of the images reference it. It's like having the `insecureRegistries` value in a docker config.

go.mod
dlion: Done 😄

image/registry_handler.go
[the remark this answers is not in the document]
dlion: It makes sense to me, I just modified the `GetInsecureOptions` implementation to accept all insecure-registries.
Let me know what do you think 😄

image/registry_handler.go
```diff
@@ -58,13 +56,11 @@ TODO: This is a temporary solution in order to get insecure registries in other
 TODO: Ideally we should fix the `imgutil.options` struct visibility in order to mock and test the `remote.WithRegistrySetting`
 TODO: function correctly and use the RegistryHandler everywhere it is needed.
 */
-func GetInsecureOptions(insecureRegistries []string, imageRef string) []remote.ImageOption {
+func GetInsecureOptions(insecureRegistries []string) []remote.ImageOption {
 	var opts []remote.ImageOption
 	if len(insecureRegistries) > 0 {
```
jabrown85, line 61: I think we can remove this redundant if len check. If there is nothing in there it won't loop anyway

image/registry_handler_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	h "github.com/buildpacks/lifecycle/testhelpers"
+)
+
+func TestRegistryHandler(t *testing.T) {
+	spec.Run(t, "RegistryHandler", testRegistryHandler, spec.Parallel(), spec.Report(report.Terminal{}))
+}
+func testRegistryHandler(t *testing.T, when spec.G, it spec.S) {
+	when("insecure registry", func() {
+		it("returns WithRegistrySetting options for the domains specified", func() {
```
jabrown85, line 17: Can we get a test added that returns more than one?

platform/lifecycle_inputs.go
```diff
@@ -235,6 +237,14 @@ func envOrDefault(key string, defaultVal string) string {
 	return defaultVal
 }
 
+func sliceEnv(k string) str.Slice {
+	envVal := os.Getenv(k)
+	if envVal != "" {
+		return strings.Split(strings.ReplaceAll(envVal, " ", ""), ",")
```
jabrown85, line 243: What's the idea behind replacing the spaces? Is this in one of the specs?

image/registry_handler.go
dlion: Ops you are right, fixing it 😄

platform/lifecycle_inputs.go
dlion: This is been done if I remember correctly to avoid things like `--flatten=something, something, something`.
So in this way, we can either provide this format or this one `--flatten=something,something,something`.
Is this something we need to document in the spec? 🤔 At the end of the day we are not changing the current behavior, we are just preventing other formats from being passed through and not working properly.

image/registry_handler_test.go
dlion: Sure thing!

platform/lifecycle_inputs.go
jabrown85: I don't have a problem with us helping out end users - but I think it being stuffed into `sliceEnv` is unexpected. It would break a rather innocent looking example like `--labels=red,granny smith,green` down the road. If we want to ensure the specific env doesn't have spaces, I think we should do that outside of this helper. Does that make sense?

jabrown85 (approved):
This looks pretty good to me! I'll let @natalieparellano take a peek as well.

natalieparellano (approved):

cmd/lifecycle/restorer.go
```diff
@@ -42,6 +42,7 @@ func (r *restoreCmd) DefineFlags() {
 		cli.FlagGeneratedDir(&r.GeneratedDir)
 		cli.FlagUseLayout(&r.UseLayout)
 		cli.FlagLayoutDir(&r.LayoutDir)
+		cli.FlagInsecureRegistries(&r.InsecureRegistries)
```
natalieparellano, line 45: Is it possible to move this into a `if r.PlatformAPI.AtLeast("0.13")` block?

platform/lifecycle_inputs.go
```diff
@@ -85,11 +86,12 @@ func NewLifecycleInputs(platformAPI *api.Version) *LifecycleInputs {
...
-		LogLevel:    envOrDefault(EnvLogLevel, DefaultLogLevel),
-		PlatformAPI: platformAPI,
-		ExtendKind:  envOrDefault(EnvExtendKind, DefaultExtendKind),
-		UseDaemon:   boolEnv(EnvUseDaemon),
-		UseLayout:   boolEnv(EnvUseLayout),
+		LogLevel:           envOrDefault(EnvLogLevel, DefaultLogLevel),
+		PlatformAPI:        platformAPI,
+		ExtendKind:         envOrDefault(EnvExtendKind, DefaultExtendKind),
+		UseDaemon:          boolEnv(EnvUseDaemon),
+		InsecureRegistries: sliceEnv(EnvInsecureRegistries),
```
natalieparellano, line 93: Maybe we should only read into this field if `platformAPI.AtLeast("0.13")`

cmd/lifecycle/restorer.go
dlion: I moved all the `cli.FlagInsecureRegistries(&r.InsecureRegistries)` from the 0.12 guard to the 0.13 ones and added this too, thanks 😄

platform/lifecycle_inputs.go
dlion: I wouldn't do that because right now we are not supporting the 0.13 version -the latest is the 0.12- and it would make the `lifecycle_inputs_tests` fail since it's run the test toward the latest version.
I think it can be left as is it since we have the guards in all the steps, it means that the InsecureRegistries field might be populated but never consumed.

natalieparellano:
@jjbustamante I believe so... it's worth noting that while this has been merged, it only applies for Platform APIs later than 0.12, for which we don't have anything released yet in the spec.

PR #1142, opened 2023-07-05
Add support for s390x

dilipgb:
This is the PR to enable buildpacks on the s390x platform.


natalieparellano:
Fixes https://github.com/buildpacks/lifecycle/issues/811

jabrown85 (commented):
This looks great to me!  I want @natalieparellano to take a peek as well since this is a bit different than our average PR contribution with all the IBM cloud work happening.

.github/workflows/build.yml
```diff
@@ -347,5 +440,4 @@ jobs:
           git checkout v0.30.0
           $env:LIFECYCLE_PATH="..\lifecycle-v${{ env.LIFECYCLE_VERSION }}+windows.x86-64.tgz"
           $env:LIFECYCLE_IMAGE="buildpacksio/lifecycle:${{ env.LIFECYCLE_IMAGE_TAG }}"
-          make acceptance
-
+          make acceptance
```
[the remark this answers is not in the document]
dilipgb: added back the newlines.

.github/workflows/build.yml
```diff
@@ -54,6 +54,86 @@ jobs:
...
+          do
+            check=$(ibmcloud is ins | awk '/'$ZVSI_INSTANCE_NAME'/{print $3}')
+            if [[ $check == 'failed' ]]
+            then
+              echo "Failed to run the ZVSI"
+              break
+              fi
+          done
+      - name: Install dependencies and run all tests on s390x ZVSI
+        uses: appleboy/ssh-action@v0.1.10
```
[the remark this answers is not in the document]
natalieparellano: I don't think we have any set criteria, but let me ask around to see if there are guidelines we can follow

.github/workflows/build.yml
```diff
@@ -159,6 +240,14 @@ jobs:
...
           name: lifecycle-linux-arm64-sha256
           path: out/lifecycle-v*+linux.arm64.tgz.sha256
+      - uses: actions/upload-artifact@v2
+        with:
+          name: lifecycle-linux-s390x
+          path: out/lifecycle-v*+linux.s390x.tgz
+      - uses: actions/upload-artifact@v2
+        with:
+          name: lifecycle-linux-s390x-sha256
+          path: out/lifecycle-v*+linux.s390x.tgz.sha256
```
natalieparellano, line 250: I think we'll also want to update the release workflow here: https://github.com/buildpacks/lifecycle/blob/aeb177ca468cc814ab47fac0cb878497495b5819/.github/workflows/draft-release.yml#L82

It would be helpful if we could somehow test this on a fork to ensure there aren't any other gotchas that we're failing to find. https://github.com/buildpacks/lifecycle/blob/main/tools/test-fork.sh could be helpful here (though it's a bit finicky, and may require some updating).

.github/workflows/build.yml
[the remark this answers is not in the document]
natalieparellano: I didn't find much beyond the guidance [here](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) ...but I think we could mitigate the risk by moving this job to its own workflow file. That way, any secrets not used by the workflow would be scrubbed according to [here](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#accessing-secrets). We could also [restrict the permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes) on the `GITHUB_TOKEN` for the new workflow to be read-only. 

@dilipgb would you be amenable to making these changes to improve our security posture? I'm happy to work with you on it if that would be helpful. I don't think this requires further end-to-end testing on a fork, as building and storing the artifacts for s390x can still happen within build.yml since we are using cross-compilation.

.github/workflows/build.yml
dilipgb: yes. I updated it to if (urlList.length != 12) { 
this works. With new changes pushed I need to verify this again.

.github/workflows/build.yml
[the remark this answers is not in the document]
dilipgb: @natalieparellano I seperated s390x related tests to separate workflow. Can you please take a look?

.github/workflows/test-s390x.yml
```diff
@@ -0,0 +1,92 @@
+name: test-s390x
+on:
+  push:
+    branches:
+      - main
+      - 'release/**'
+  pull_request:
+    branches:
+      - main
+      - 'release/**'
```
natalieparellano, line 10: If I understand correctly, secrets won't be exposed on pull requests from forks, so will this workflow always fail in those cases?

.github/workflows/test-s390x.yml
```diff
@@ -0,0 +1,92 @@
...
+      ZVSI_FP_NAME: bp-floating-ci-${{ github.run_id }} 
+      ZVSI_INSTANCE_NAME: bp-zvsi-ci-${{ github.run_id }} 
+      ZVSI_VPC: r038-b86410fb-6889-4094-821d-1ceb72176bc4
+      ZVSI_ZONE_NAME: ca-tor-1
+      ZVSI_PROFILE_NAME: bz2-4x16
+      ZVSI_SUBNET: 02q7-af49cf76-0471-49f4-a245-b60d63196efb
+      ZVSI_IMAGE: r038-67da7290-933e-42e4-a1c2-ff439b4b6d2c
+      ZVSI_KEY: r038-f8e35cee-0475-4315-a051-40f4d2fb6af9
+      ZVSI_RG_ID: 31e0f1b03d1b4d42bd98b6768ba5421a
+      ZVSI_SG: r038-1e644f46-8044-4043-9108-2b7dc178827f
```
natalieparellano, line 26: Is there an argument for putting any of these in secrets instead of hard coding?

.github/workflows/draft-release.yml
```diff
@@ -24,7 +24,7 @@ jobs:
             exit 1
           fi
           echo "LIFECYCLE_VERSION=$version" >> $GITHUB_ENV
-      - name: Determine download urls for linux-x86-64, linux-arm64, and windows
+      - name: Determine download urls for linux-x86-64, linux-arm64, linux-s390x, and windows
```
natalieparellano, line 27: We should also expect 10 artifacts below (we recently removed cosign key and sha, which is why the number changed)

natalieparellano (commented):
Looks good @dilipgb! I added a couple comments. @jabrown85 do you want to take another look as well?

jabrown85:
This looks good to me @natalieparellano

.github/workflows/test-s390x.yml
```diff
@@ -0,0 +1,92 @@
...
+      - main
+      - 'release/**'
+  pull_request:
+    branches:
+      - main
+      - 'release/**'
+
+jobs:
+  test-linux-s390x:
+    if: (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/release*')
```
dilipgb, line 14: This condition skip the workflow if its on the fork.

.github/workflows/test-s390x.yml
dilipgb: https://github.com/buildpacks/lifecycle/pull/1142#discussion_r1345275208
You can check the condtion on the comment in the above link. That skips the execution when running from fork.

.github/workflows/draft-release.yml
dilipgb: I changed it to 10.

natalieparellano (approved):

PR #1145, opened 2023-07-10
Require the CNB_PLATFORM_API env var to be set

natalieparellano:
Instead of defaulting to (deprecated) version 0.3

Instead of defaulting to a newer Platform API version, which would silently change behavior for platforms that do not currently set this env var

First PR in support of https://github.com/buildpacks/rfcs/blob/main/text/0110-deprecate-apis.md


jabrown85 (approved):

joe-kimmel-vmw (approved):
LGTM.  

  all the `//nolints` feel a little tragic but at a glance it looks like you can't use the `skip-dirs` directory for only one kind of linter.  
  so we could decide to just never lint our acceptance tests at all,  but I think there's no granular way of saying "my acceptance tests should be allowed to pass unsanitized strings directly to the command line but please do all the other linty checks."

PR #1146, opened 2023-07-10
Deprecate buildpack and platform APIs 

natalieparellano:
Depends on https://github.com/buildpacks/lifecycle/pull/1145

According to https://github.com/buildpacks/rfcs/blob/main/text/0110-deprecate-apis.md


natalieparellano:
TODO: update this PR to also remove behaviors defined in https://github.com/buildpacks/spec/blob/main/buildpack.md#build-plan-toml-requiresversion-key

natalieparellano:
TODO: create a corresponding issue in buildpacks/pack

buildpack/layermetadata.go
```diff
@@ -29,7 +27,7 @@ func EncodeLayerMetadataFile(lmf LayerMetadataFile, path, buildpackAPI string) e
 	encoders := supportedEncoderDecoders()
 
 	for _, encoder := range encoders {
-		if encoder.IsSupported(buildpackAPI) {
```
natalieparellano, line 32: We should bring the arg back so that this structure is useful if we ever decide to add another encoder

internal/layer/metadata_restorer.go
```diff
@@ -141,27 +141,11 @@ type SHAStore interface {
 	Get(buildpackID string, layer buildpack.Layer) (string, error)
 }
 
-func NewSHAStore(useShaFiles bool) SHAStore {
-	if useShaFiles {
-		return &fileStore{}
-	}
+// NewSHAStore TODO
```
natalieparellano, line 144: TODO

natalieparellano:
Marking this as ready for review. 

We should figure out what we should do about the "glue" we're currently providing to make newer buildpacks work with older buildpacks that are now deprecated - specifically, the top-level `version` fields in plan.toml and the legacy unstructured BOM. I'd be inclined to just rip it out and make it a hard error when top-level `version`s are provided (the alternative being to just silently ignore the version like all other unsupported fields), but we should we discuss. Whatever we decide, we can put it in another PR.

natalieparellano:
Closing in favor of smaller PRs that only deprecate one API (buildpack or platform) at a time

PR #1148, opened 2023-07-12
Remove deprecated buildpack APIs

natalieparellano:
According to https://github.com/buildpacks/rfcs/blob/main/text/0110-deprecate-apis.md


cmd/lifecycle/main.go
```diff
@@ -189,8 +190,8 @@ func verifyBuildpackApis(group buildpack.Group) error {
 	for _, bp := range group.Group {
 		if bp.API == "" {
 			// if this group was generated by this lifecycle bp.API should be set
-			// but if for some reason it isn't default to 0.2
-			bp.API = "0.2"
+			// but if for some reason it isn't default to latest
+			bp.API = api.Buildpack.Latest().String()
```
jabrown85, line 194: I wonder if we should think about failing in this case instead of defaulting.

cmd/launcher/cli/launcher.go
```diff
@@ -87,8 +87,8 @@ func verifyBuildpackAPIs(bps []launch.Buildpack) error {
 	for _, bp := range bps {
 		if bp.API == "" {
 			// If the same lifecycle is used for build and launcher we should never end up here
-			// but if for some reason we do, default to 0.2
-			bp.API = "0.2"
+			// but if for some reason we do, default to latest
+			bp.API = api.Buildpack.Latest().String()
```
jabrown85, line 91: Same question here - seems a bit weird we end up with a buildpack that doesn't have an API. Is this in the spec?

jabrown85 (approved):
So much code deleted ❤️ 😮 

Left a couple questions

cmd/lifecycle/main.go
natalieparellano: Yeah... that's probably fair. The lifecycle has ensured there is an API version here for many years now!

cmd/launcher/cli/launcher.go
natalieparellano: Yeah, buildpack API is here: https://github.com/buildpacks/spec/blob/main/platform.md#metadatatoml-toml

cmd/lifecycle/main.go
natalieparellano: Updated in fb806d31

cmd/launcher/cli/launcher.go
natalieparellano: Updated in fb806d31

PR #1152, opened 2023-07-13
Remove async go-subroutine delete keeping it sync

dlion:
Fixing the PR https://github.com/buildpacks/lifecycle/pull/1136

### DIFF
--- a/cache/image_deleter.go
+++ b/cache/image_deleter.go
@@ -35,7 +35,7 @@ func (c *ImageDeleterImpl) DeleteOrigImageIfDifferentFromNewImage(origImage, new
 		}
 
 		if !same {
-			go c.deleteImage(origImage)
+			c.deleteImage(origImage)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


jabrown85 (approved):

natalieparellano:
The linux ARM runner is timing out :/ I don't think it relates to this PR (it happens sometimes) but I'm not sure. We might try restarting the runner.

dlion:
@natalieparellano 
I just kind of reverted this commit https://github.com/buildpacks/lifecycle/pull/1136/commits/0fb6d92368af66e18e0c620c2338738e2fa82a25 so it should pass 🤔 what the arm tests are doing differently from the amd64 and the windows ones?
The implementation shouldn't have effect on that, it's odd

natalieparellano:
The ARM runner was acting up - I restarted it and all is good now. Thanks for all your hard work @dlion!

PR #1156, opened 2023-07-20
fix to go 1.20.5 to avoid any issues with recent breaking changes

joe-kimmel-vmw:


natalieparellano:
Is this still needed in light of https://github.com/buildpacks/lifecycle/pull/1158?

I could be wrong, but it seems the pack tests that are failing are using "[copy from container](https://github.com/buildpacks/pack/actions/runs/5591663269/job/15145321140?pr=1841#step:11:161)" methods that the lifecycle doesn't use.

natalieparellano:
I think we can close this one. https://github.com/moby/moby/releases/tag/v24.0.5 now exists and we should be bumping up to that rather than downgrading our go version...

PR #1157, opened 2023-07-20
upgrades setup-go to v4 which mostly sounds like it enables some cach…

joe-kimmel-vmw:
…ing by default


natalieparellano (commented):
@joe-kimmel-vmw could you also bump the version in check-latest-release.yml and post-release.yml?

natalieparellano (approved):

PR #1159, opened 2023-07-25
Fix restorer target update

natalieparellano:
When we instantiate the remote image for the purpose of checking target data, we should provide "from base image" so that we don't end up returning data for an empty image


natalieparellano:
TDD notes...

Failing test (prior to code changes): https://github.com/buildpacks/lifecycle/actions/runs/5660989254/job/15337920259#step:5:10277

Note that `03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10` is a "special" SHA for empty images - e.g., 

```
	image, _ := remote.NewImage("busybox", authn.DefaultKeychain)
	fmt.Println(image.Identifier())
```

prints `index.docker.io/library/busybox@sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10`.

Oddly, I couldn't get this to fail locally. Locally, the run image reference returned is something like `192.168.1.43:64606/some-read-only-run-image-dtakgtkcmh@sha256:d0143fdc9f75c98c264f019d1c690bb17845d4b514ce1a44555eef3794ce3980`

I'm __assuming__ this has something to do with oddities around the test registry, but I'm not sure

cmd/lifecycle/restorer.go
```diff
@@ -116,7 +116,7 @@ func (r *restoreCmd) Exec() error {
 			}
 		} else if r.supportsTargetData() && needsUpdating(analyzedMD.RunImage) {
 			cmd.DefaultLogger.Debugf("Updating run image info in analyzed metadata...")
-			remoteRunImage, err = remote.NewImage(runImageName, r.keychain)
+			remoteRunImage, err = remote.NewImage(runImageName, r.keychain, remote.FromBaseImage(runImageName))
```
joe-kimmel-vmw, line 119: I can't figure out why this would work -- maybe it has the effect of (re-)loading the image?

joe-kimmel-vmw (approved):
LGTM in the sense that unit tests don't lie, but it's a little mysterious.

acceptance/restorer_test.go
```diff
@@ -274,6 +276,8 @@ func testRestorerFunc(platformAPI string) func(t *testing.T, when spec.G, it spe
 					h.AssertStringContains(t, analyzedMD.RunImage.Reference, restoreRegFixtures.ReadOnlyRunImage+"@sha256:")
 					h.AssertEq(t, analyzedMD.RunImage.Image, restoreRegFixtures.ReadOnlyRunImage)
 					h.AssertEq(t, analyzedMD.RunImage.TargetMetadata.OS, "linux")
+					t.Log("does not return the digest for an empty image")
+					h.AssertStringDoesNotContain(t, analyzedMD.RunImage.Reference, restoreRegFixtures.ReadOnlyRunImage+"@sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10")
```
dlion, line 280: Thank you for highlight that this is a special SHA, I didn't know it.
My advice is to have it stored inside a kind of constant variable with a nice name to describe it properly and improve the readability.
From the external point of view it's like a "magic number" and should be clarified.

Perhaps something like `const  SHA_EMPTY_IMAGE = "@sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10"`
To have something like:
```
h.AssertStringDoesNotContain(t, analyzedMD.RunImage.Reference, restoreRegFixtures.ReadOnlyRunImage+SHA_EMPTY_IMAGE)
```

Or if you prefer, have something like:
```
var readOnlyEmptyRunImage = restoreRegFixtures.ReadOnlyRunImage+SHA_EMPTY_IMAGE
```
and have something
```
h.AssertStringDoesNotContain(t, analyzedMD.RunImage.Reference, readOnlyEmptyRunImage)
```

What do you think?

natalieparellano: Thanks for the suggestion @dlion ❤️ I was slow to update this one but I'll keep it in mind for next time

cmd/lifecycle/restorer.go
natalieparellano: `remote.NewImage` in `imgutil` sadly doesn't work the same as `remote.Image` in GGCR - without the base image option, all the getters on the image return data for an empty image! I sadly forgot this fact when writing this code.

natalieparellano:
> LGTM in the sense that unit tests don't lie, but it's a little mysterious.

There was something wrong with my local setup, I can see the failing tests now. All good!

Issue #1153, opened 2023-07-14
how can i get time-cost from different phase


TigerWj:
hi, I'm using /lifecycle/creator to run build  instead of pack，and i need to statistic the time-cost from differe phase，like analyse / export，so how can I do it?  it seems there are no metric log that can use

joe-kimmel-vmw:
Hi @TigerWj  -- Thanks for filing this issue as it's a very helpful signal.    @natalieparellano and I have been talking about this from a few different angles for a little while, and it's great to know that someone else is interested too!

First -- in general the creator will always be faster than running the individual phases, just because it means you're only starting one executable instead of starting separate processes for each phase.   But I think you're really asking for a high-level "profiling" overview that shows how long each phase takes?

One strategy for getting this would be to add some info level logging to the beginning and end of each phase, right? 
in that vein, you can see some recent PRs that will be part of the next lifecycle release (0.17) where we're already moving in that direction (such as [this one](https://github.com/buildpacks/lifecycle/pull/1106)

Another strategy would be to add a command-line flag that would turn on golang's profiling tool and output pprof files for each phase.  These would have a lot of information about where time is spent within each phase and would likely require a little familiarity with the codebase and/or golang profiling to work with. 

Another strategy would be to explicitly collect the length of time of each phase and output that into the logs. 

Can you help us understand your specific goals in measuring the times of the phases, and whether any or all of the above would be helpful?

TigerWj:
@joe-kimmel-vmw  thanks. I think the first strategy, add some info level logging to the beginning and end of each phase, is useful. and it is what i exactly did in local.I'am trying  use buildpack in a build sysytem，so it's important for me to make the times of differ pahses clear, without this, it is hard to make optimization. And I think it will be useful for everyone。
the second strategy, sounds a bit complicated,  i think it good for someone who need to measuring times for every action inside each pahse.

samj1912:
@joe-kimmel-vmw we would be interested in this as well. One potential option is report.toml. It would be great if lifecycle can populate phase information there.

matejvasek:
@TigerWj Do you experience any performance issues?
[...]

PR #1160, opened 2023-07-26
timestamp logs for entry/exit for all the top-level Lifecycle package…

joe-kimmel-vmw:
… functions

See discussion in https://github.com/buildpacks/lifecycle/issues/1153

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -190,6 +190,8 @@ func (f *AnalyzerFactory) setRun(analyzer *Analyzer, imageRef string) error {
 
 // Analyze fetches the layers metadata from the previous image and writes analyzed.toml.
 func (a *Analyzer) Analyze() (files.Analyzed, error) {
+	ft := log.NewFuncTimer("Analyzer", a.Logger)
+	defer ft.RecordEnd()
 	var (
 		err              error
 		appMeta          files.LayersMetadata
--- a/analyzer.go
+++ b/analyzer.go
@@ -190,7 +190,7 @@ func (f *AnalyzerFactory) setRun(analyzer *Analyzer, imageRef string) error {
 
 // Analyze fetches the layers metadata from the previous image and writes analyzed.toml.
 func (a *Analyzer) Analyze() (files.Analyzed, error) {
-	ft := log.NewFuncTimer("Analyzer", a.Logger)
+	ft := log.NewRecordStart("Analyzer", a.Logger)
 	defer ft.RecordEnd()
 	var (
 		err              error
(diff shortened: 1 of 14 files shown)


log/timelog_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+func (m mockLog) Error(msg string) {
+	m.incr("Error")
+}
+func (m mockLog) Errorf(fmt string, v ...interface{}) {
+	m.incr("Error")
+}
+
+func TestCache(t *testing.T) {
+	spec.Run(t, "Exporter", testCache, spec.Parallel(), spec.Report(report.Terminal{}))
```
dlion, line 53: The name of the func and the run descriptor should be changed to `TestTimelog` and `TImelog` or something around that 😄

log/timelog_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func testCache(t *testing.T, when spec.G, it spec.S) {
+	when("#it's like this", func() {
+		it("it's like that", func() {
```
dlion, line 58: Same thing here for the name of the function and when/it description 😄

dlion:
Thanks Joe! I added a few comments 😄

analyzer.go
```diff
@@ -190,6 +190,8 @@ func (f *AnalyzerFactory) setRun(analyzer *Analyzer, imageRef string) error {
 
 // Analyze fetches the layers metadata from the previous image and writes analyzed.toml.
 func (a *Analyzer) Analyze() (files.Analyzed, error) {
+	ft := log.NewFuncTimer("Analyzer", a.Logger)
```
jabrown85, line 193: I think the 2 step approach of creating the timer and manually deferring is good to have, but maybe a convenience method would be nice for the most common approach?
```suggestion
	defer log.MeasureDuration("Analyzer", a.Logger)
```

joe-kimmel-vmw: Thanks @jabrown85  - The `New` function is actually also writing a Start timestamp to the logs, so maybe I should call it `NewAndLog` or similar?   There is a 3-step version available if you (1) make the struct yourself, (2) call `RecordStart` and then do  ` defer RecordEnd()`

joe-kimmel-vmw: or in other words,  I'm not sure there's a one-liner that outputs both a start and end log timestamp

jabrown85: Ah, right. I think a method could be written like this

```
func MeasureDuration(msg string) {
  ft := log.NewRecordStart(msg, b.Logger)
  return ft.RecordEnd()
}

// caller
defer log.MeasureDuration("Analyzer", a.Logger)() <-- notice second function is deferred
```

joe-kimmel-vmw: indeed it can! :)

jabrown85 (approved):

PR #1166, opened 2023-07-28
Simplifies target matching logic per spec PR review

natalieparellano:
See https://github.com/buildpacks/spec/pull/369


natalieparellano:
@jabrown85 I added 135d78f696f03391395d700290be019e5c323be8 with stricter validation for rebase (if the original run image had distribution information, then the new run image should also have it). I think that is more consistent with what we have in the spec:

```
- **If** `<force>` is `true`, the following [target data](#target-data) values in the output `<image>` config MUST be derived from the new `<run-image>`:
  - `os`
  - `architecture`
  - `variant` (if specified)
  - `io.buildpacks.base.distro.name` (if specified)
  - `io.buildpacks.base.distro.version` (if specified)
- **Else** the target data above MUST match the old run image if `<force>` is `false`
```

jabrown85 (approved):

PR #1169, opened 2023-07-31
Add support for darwin-arm64 chipset

dlion:
Add build make actions to build lifecycle for MacOS M1 chipset

### DIFF
--- a/darwin.mk
+++ b/darwin.mk
@@ -1,0 +1,18 @@
+define build_darwin_lifecycle
+	@echo "> Building lifecycle for $(TARGET)..."
+	$(GOENV) $(GOBUILD) -o $(OUT_DIR)/lifecycle -a ./cmd/lifecycle
+	@echo "> Creating lifecycle symlinks for $(1)..."
+	ln -sf lifecycle $(OUT_DIR)/detector
+	ln -sf lifecycle $(OUT_DIR)/analyzer
+	ln -sf lifecycle $(OUT_DIR)/restorer
+	ln -sf lifecycle $(OUT_DIR)/builder
+	ln -sf lifecycle $(OUT_DIR)/exporter
+	ln -sf lifecycle $(OUT_DIR)/rebaser
+endef
+
+define build_darwin_launcher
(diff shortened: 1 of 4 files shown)


jabrown85 (approved):
Seems ok to me!

natalieparellano:
I'm also good with this but I wonder if we could use this as an opportunity to improve our Makefile. It has tons of duplication and is close to 400 lines long at this point :/ ...perhaps we could add some helper scripts to make things more concise?

dlion:
It sounds good to me.
I just pushed some changes:
- Created a separate darwin.mk module where to have the builder functions
- Call those functions from the main Makefile

In the future we can do the same for the other actions but I preferred to stay in the scope of this PR.
I will open a new issue for this possible refactoring

darwin.mk
```diff
@@ -0,0 +1,18 @@
+define build_darwin_lifecycle
```
natalieparellano, line 1: Is there anything darwin-specific about this script? Could it just be `build_lifecycle`?

natalieparellano: I'd lean toward a lifecycle.mk and launcher.mk in this case, since we know that we'll just be setting different env vars for different os/arch targets. But yes let's go ahead and update the other targets as a separate effort.

dlion:
Thank you for you comments @natalieparellano 
I just pushed and updated version 👍🏼

launcher.mk
```diff
@@ -0,0 +1,6 @@
+define build_darwin_launcher
```
natalieparellano, line 1:
```suggestion
define build_launcher
```

lifecycle.mk
```diff
@@ -0,0 +1,11 @@
+define build_darwin_lifecycle
```
natalieparellano, line 1:
```suggestion
define build_lifecycle
```

natalieparellano (approved):

dlion:
Is there anything I have to do in order to have this PR merged? @natalieparellano :)

PR #1182, opened 2023-08-21
Update version of pack used in acceptance tests

natalieparellano:


jabrown85 (approved):

.github/workflows/build.yml
```diff
@@ -255,7 +255,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.20.5'
+          go-version-file: 'go.mod'
```
jjbustamante, line 258: @jkutner you removed the `check-latest` configuration added by @hone at some point, I do not remember why. adding the [go-version-file](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file) will check for the latest go version if I am not wrong. Any thoughts?

natalieparellano: Looks like we are still using `check-latest` in pack: https://github.com/buildpacks/pack/blob/c38f7da1b43d5b142d1e870f3d17da93b454c64f/.github/workflows/build.yml#L184

We could add it here, but since we are just running the tests I think it's okay if the patch is a little old

PR #1184, opened 2023-08-22
Print the build environment given --verbose option

AidanDelaney:
Allow developers to see the environment that is set from any BuildConfigDir settings.

### Summary
In https://github.com/buildpacks/rfcs/blob/main/text/0109-build-config.md?plain=1#L15 we proposed platform defined environment settings.  This PR allows `lifecycle` users to see the values of the environment.

#### Release notes

When verbose output is selected, print the computed build environment.

---

### Related

Resolves #___

---


natalieparellano:
Is there any concern that we might leak sensitive information by printing this out?

natalieparellano:
@AidanDelaney any update on this one? I'm going to move it to draft for now

AidanDelaney:
Let's kill this.   It's a security risk and anyone who wants to do it can _explicity_ run `env` as an inline buildpack.

PR #1200, opened 2023-09-15
Update some actions and configure dependabot to update actions

natalieparellano:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: daily
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
   test-linux-amd64:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           fetch-depth: '0'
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -231,11 +229,11 @@ jobs:
           MANIFEST_SHA=$(docker manifest push buildpacksio/lifecycle:${LIFECYCLE_IMAGE_TAG})
           echo "MANIFEST_SHA: $MANIFEST_SHA"
 
-          COSIGN_PASSWORD=${{ secrets.COSIGN_PASSWORD }} cosign sign -r \
-            -key <(echo -n "${{ secrets.COSIGN_PRIVATE_KEY }}" | base64 --decode) \
+          COSIGN_PASSWORD=${{ secrets.COSIGN_PASSWORD }} cosign sign -r --tlog-upload=false \
```
natalieparellano, line 232: Ideally we'd update cosign to use transparency logs instead of a private key... I'll do that in a separate PR

natalieparellano:
Tested on fork: https://github.com/natalieparellano/lifecycle/actions

.github/dependabot.yml
```diff
@@ -5,3 +5,7 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: daily
```
natalieparellano, line 11: @jjbustamante we could add something like this to pack to ensure actions are kept up-to-date there

jabrown85 (approved):

PR #1201, opened 2023-09-18
Update cosign

natalieparellano:
### Summary

Updates GitHub actions workflows that use `cosign` to sign and verify lifecycle images to use [keyless signing](https://github.com/sigstore/docs/blob/main/content/en/signing/overview.md). This is the default as of `cosign` 2.x.x and is considered more secure than signing with a long-lived private key. Note that the `cosign verify` command for verifying signatures is also updated.

When this is merged, we can remove `COSIGN_PASSWORD` and `COSIGN_PRIVATE_KEY` from repo secrets. We'll also need to update our Docker Hub [page](https://hub.docker.com/r/buildpacksio/lifecycle) help text with the new information.

#### Release notes

* When signing lifecycle images in GitHub actions with `cosign`, use keyless signing

---

### Related

Resolves #___

---

### Context

Tested (with some effort) on fork: https://github.com/natalieparellano/lifecycle/actions


.github/workflows/build.yml
```diff
@@ -111,6 +111,8 @@ jobs:
       - test-linux-arm64
       - test-windows
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
natalieparellano, line 115: Needed to request an OIDC token: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings

.github/workflows/draft-release.yml
```diff
@@ -6,6 +6,8 @@ on:
 jobs:
   draft-release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
natalieparellano, line 10: I updated the repo "Actions permissions" settings to use the more restrictive "Read repository contents and package permissions" instead of the more permissive "Read and write permissions", thus this is now required.

jabrown85 (approved):

PR #1205, opened 2023-09-22
Move lifecycle package to sub-directory

natalieparellano:
### Summary

We've been wanting to do this for awhile now. This will help cleanup our root directory, so that we can more easily distinguish code from configuration.

#### Release notes

n/a


acceptance/restorer_test.go
```diff
@@ -14,9 +14,9 @@ import (
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
-	"github.com/buildpacks/lifecycle"
 	"github.com/buildpacks/lifecycle/api"
 	"github.com/buildpacks/lifecycle/cmd"
+	"github.com/buildpacks/lifecycle/lifecycle"
```
jabrown85, line 19: What do you think about moving things to a new package called `internal`? I dislike the `lifecycle/lifecycle` in the imports. Are the things under `lifecycle` meant for to be public APIs? I don't actually know our stance. I know we use types from various places like `github.com/buildpacks/lifecycle/platform`, which reads nicely.

```suggestion
	"github.com/buildpacks/lifecycle/internal"
```

natalieparellano: I believe `lifecycle.Rebaser` is imported by pack and kpack ...not sure about the other phases though

natalieparellano: What about `github.com/buildpacks/lifecycle/phase`? Or making all but the rebaser internal?

natalieparellano: Maybe it could be `github.com/buildpacks/lifecycle/internal/build` and `github.com/buildpacks/lifecycle/rebase` ...or something

jabrown85: I like `github.com/buildpacks/lifecycle/rebase` or `github.com/buildpacks/lifecycle/phase` fine!

In case we want to visualize what this looks like before deciding...

```golang
rebaser := rebase.Rebaser{}
rebaser := rebase.New(opts)

rebaser := phase.Rebaser{}
rebaser := phase.NewRebaser(opts)

// alternatives
rebaser := platform.NewRebaser(opts)
rebaser := core.NewRebaser(opts)
rebaser := buildpacksio.NewRebaser(opts)
rebaser := exec.NewRebaser(opts)
rebaser := stage.NewRebaser(opts)
```

The whole `internal` folder thing golang does can be a bit aggressive. If we want to make sure we don't accidentally block out users who are importing things we don't suspect, I think we should probably avoid nesting under `internal`.

phase/handlers.go
```diff
@@ -12,26 +12,34 @@ import (
...
+// CacheHandler wraps initialization of a cache image or cache volume.
+//
+//go:generate mockgen -package testmock -destination testmock/cache_handler.go github.com/buildpacks/lifecycle/phase CacheHandler
 type CacheHandler interface {
 	InitCache(imageRef, dir string, deletionEnabled bool) (Cache, error)
 }
 
-//go:generate mockgen -package testmock -destination testmock/dir_store.go github.com/buildpacks/lifecycle DirStore
+// DirStore is a repository of buildpacks and/or image extensions.
+// Each element should be present on disk according to the format outlined in the Buildpack Interface Specification,
```
natalieparellano, line 23:
```suggestion
// Each element should be present on disk according to the format outlined in the Platform Interface Specification,
```

jabrown85 (approved):

PR #1209, opened 2023-10-03
Use the the latest version of containerd 1.6.x instead of 1.7.x

natalieparellano:
### Summary

We have tried to be on the latest version of containerd as it has been a source of CVEs in the past. However the 1.7.x line is incompatible with older versions of k8s. This change would allow library consumers to upgrade the lifecycle without having to simultaneously upgrade their clusters.

#### Release notes

Uses the highest version of `github.com/containerd/containerd` v1.6.x to facilitate importing the lifecycle as a library in k8s

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -126,8 +126,8 @@ require (
 
 go 1.20
 
-// Ensure we only pull in the latest version of containerd
-replace github.com/containerd/containerd => github.com/containerd/containerd v1.7.6
+// Ensure we only pull in the latest version of containerd 1.6.x
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.6.24
 
 // Pin docker/docker until it is upgraded in kaniko
 replace github.com/docker/docker => github.com/docker/docker v23.0.6+incompatible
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -126,8 +126,8 @@ require (
 
 go 1.20
 
-// Ensure we only pull in the latest version of containerd
-replace github.com/containerd/containerd => github.com/containerd/containerd v1.7.6
+// Ensure we only pull in the latest version of containerd 1.6.x
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.6.24
```
jabrown85, line 130: Can we remove this in favor of `github.com/containerd/containerd v1.6.24` in the main require block? Or what does that end up downgrading?

jabrown85: I wonder if this fixes the issue then - I don't think go tool respects the replace directives when importing as a library.

natalieparellano: Ohhh... you are right 🤔

jabrown85: It looks like [kaniko](https://github.com/GoogleContainerTools/kaniko/blob/main/go.mod#L42) is probably one of the main packages forcing our hand here to a later containerd version. Oh well I suppose.

PR #1222, opened 2023-10-18
Fix usage of URL based registries

c0d1ngm0nk3y:
### Summary

When reading the keychain from the environment in order, the registry were taken exactly as specified. This resulted in some cases where the credentials were not found and required the user to specify exactly the expected format although `docker pull` worked with the same credentials.

#### Release notes

Fixes access of url based repositories for specific cases.

---

### Related

Resolves #___

---

### Context

Example that did not work:

```json
{
    "auths": {
        "https://index.docker.io/v1/": {
            "auth": "..."
        }
    }
}
```


natalieparellano (approved):
Thanks for the PR @c0d1ngm0nk3y!

PR #1227, opened 2023-10-27
Do not obscure error message

c0d1ngm0nk3y:
### Summary

When the lifecycle failed because there was no readable run image, the error was rather misleading:

`-run-image is required when there is no stack metadata available`

This was caused by hiding the underlying error.

#### Release notes

Improve error message in case run-image is not readable.

---

### Related

Resolves #___

---

### Context

When the run image is on a protected registry, the `auth` data must be provided to the lifecycle.

### DIFF
--- a/log/timelog_test.go
+++ b/log/timelog_test.go
@@ -24,28 +24,28 @@ func (m mockLog) incr(key string) {
 	}
 }
 
-func (m mockLog) Debug(msg string) {
+func (m mockLog) Debug(_ string) {
 	m.incr("Debug")
 }
-func (m mockLog) Debugf(fmt string, v ...interface{}) {
+func (m mockLog) Debugf(_ string, _ ...interface{}) {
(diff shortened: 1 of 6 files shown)


natalieparellano (approved):
Thank you @c0d1ngm0nk3y 🙏🏼

natalieparellano:
Aye, now the linter is unhappy because the spacing for our exported message vars has changed and they were uncommented. You could update the commit sha here: https://github.com/buildpacks/lifecycle/blob/367b451eb708dff1aa6e5ba179e1c88c4b9917e2/golangci.yaml#L38 if you don't feel like adding comments.

c0d1ngm0nk3y:
> Aye, now the linter is unhappy...

I have made some changes. IMHO, not the most useful linter warning though :) Please have a look.

natalieparellano (approved):

PR #1235, opened 2023-10-31
Fix post-release workflow

natalieparellano:
### Summary

Fixes post-release workflow by capturing the image digest for the linux-s390x image.

#### Release notes

Fixes post-release workflow by capturing the image digest for the linux-s390x image.


natalieparellano:
Tested on fork: https://github.com/natalieparellano/lifecycle/actions/runs/6709843378

jabrown85 (approved):

PR #1236, opened 2023-10-31
Consolidate methods that read and write platform spec'd TOML

natalieparellano:
in platform/files package.

With the exception of reading <layers>/config/metadata.toml in the launcher, as we want to avoid importing the buildpack package, and platform/files depends on it.

#### Release notes

Consolidates methods that read and write platform spec'd TOML in the platform/files package

### DIFF
--- a/platform/files/handle_toml.go
+++ b/platform/files/handle_toml.go
@@ -1,0 +1,197 @@
+// Package files contains schema and helper methods for working with lifecycle configuration files.
+package files
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/BurntSushi/toml"
+
+	"github.com/buildpacks/lifecycle/api"
+	"github.com/buildpacks/lifecycle/buildpack"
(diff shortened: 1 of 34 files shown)


platform/files/handle_toml.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/buildpacks/lifecycle/buildpack"
+	"github.com/buildpacks/lifecycle/internal/encoding"
+	"github.com/buildpacks/lifecycle/log"
+)
+
+// Handler is the default handler used to read and write lifecycle configuration files.
+var Handler = &TOMLHandler{}
+
+// TOMLHandler reads and writes lifecycle configuration files in TOML format.
+type TOMLHandler struct{}
```
natalieparellano, line 20: If we ever wanted the platform to be able to supple JSON or YAML files, we could implement an alternate handler with minimal other changes.

jabrown85 (approved):
Nice work!

PR #1271, opened 2024-01-16
Add FreeBSD Support

hut8:
### Summary

Adds support for FreeBSD

#### Release notes

`lifecycle` will now build on FreeBSD.

---

### Related

https://github.com/buildpacks/lifecycle/pull/1087 - this is rather old, had a lot of conflicts, and probably needs extensive CI tweaking. So I chose to start fresh, but used some parts of this as a starting point, so credit to @quantumsheep

Surprisingly there is no issue that mentions FreeBSD but there should be 😄 

---

### Context

I need this because the fly.io CLI depends on this package and I want to be able to run it from FreeBSD.

https://github.com/superfly/flyctl/issues/712

### DIFF
--- a/buildpack/build.go
+++ b/buildpack/build.go
@@ -83,7 +83,7 @@ func (e *DefaultBuildExecutor) Build(d BpDescriptor, inputs BuildInputs, logger 
 
 	logger.Debug("Running build command")
 	if err := runBuildCmd(d, bpLayersDir, planPath, inputs, inputs.Env); err != nil {
-		return BuildOutputs{}, fmt.Errorf("failed to run build command: %w", err)
+		return BuildOutputs{}, err
 	}
 
 	logger.Debug("Processing layers")
@@ -130,7 +130,6 @@ func runBuildCmd(d BpDescriptor, bpLayersDir, planPath string, inputs BuildInput
 		inputs.PlatformDir,
 		planPath,
 	) // #nosec G204
-	fmt.Printf("running build cmd: %v", cmd.String())
 	cmd.Dir = inputs.AppDir
 	cmd.Stdout = inputs.Out
 	cmd.Stderr = inputs.Err


natalieparellano:
@hut8 thanks for this!

> What's up with the ARM64 test failing?

That seems to be a flake. I can re-kick the CI.

> In FreeBSD, when I'm running docker, I do it in a separate virtual machine under bhyve and use TCP to connect the two.

Here is where we initialize the docker connection in the lifecycle: https://github.com/buildpacks/lifecycle/blob/3c8876ce042837896e76d2d31e8c18c48a8aa738/priv/docker.go#L16

I think it should just work if you've set `DOCKER_HOST` in the lifecycle's environment. LMK if you have already tried that. We do more complicated initialization in pack (see [here](https://github.com/buildpacks/pack/blob/3f1ecb280f2c9d3aef948152df286153ddd8bd0e/internal/build/phase_config_provider.go#L155) and [here](https://github.com/buildpacks/pack/blob/3f1ecb280f2c9d3aef948152df286153ddd8bd0e/cmd/docker_init.go#L23)) but so far we haven't needed that in the lifecycle.

patmaddox:
@hut8 Have you attempted to build flyctl with this branch? I've tried adding `replace github.com/buildpacks/lifecycle => github.com/hut8/lifecycle ecf97c6bc078` to flyctl go.mod and here's what I get:

```
$ PATH=/usr/local/go121/bin:$PATH go get
...
github.com/buildpacks/lifecycle: cannot find module providing package github.com/buildpacks/lifecycle

$ PATH=/usr/local/go121/bin:$PATH mod tidy
...
github.com/buildpacks/lifecycle: module github.com/buildpacks/lifecycle@latest found (v0.18.4, replaced by github.com/hut8/lifecycle@v0.0.0-20240117010513-ecf97c6bc078), but does not contain package github.com/buildpacks/lifecycle
```

So I'm not sure how to override this dependency with your branch to try building flyctl.

PR #1276, opened 2024-01-24
Provide context directory for extensions

modulo11:
### Summary

This PR implements the [extension layer](https://github.com/buildpacks/rfcs/blob/main/text/0000-extension-layer.md) RFC by providing one of the following context directory to the kaniko build during the extension phase:

- `$CNB_OUTPUT_DIR/context`
- `$CNB_OUTPUT_DIR/context.build` and/or `$CNB_OUTPUT_DIR/context.run`

More details and examples can be found in the RFC.

#### Release notes

Provide and allow extensions to write into a shared or specific context directory.

---

### Related

---


loewenstein:
@natalieparellano once the spec changes are in, this would be worth it a look. Or already now, if there is review feedback that requires changes.

natalieparellano:
@modulo11 @loewenstein thank you for this - apologies for being slow to review the changes. It looks like there might be a couple of tests to fix, but I'll take a deeper look tomorrow to see if I could offer more useful feedback.

buildpack/buildcontext.go
```diff
@@ -0,0 +1,57 @@
...
+		contexts = append(contexts, ContextInfo{
+			ExtensionID: d.Extension.ID,
+			Path:        sharedContextDir,
+		})
+	} else if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	for _, dir := range []string{DockerfileKindRun, DockerfileKindBuild} {
+		contextDir := filepath.Join(extOutputDir, dir)
```
natalieparellano, line 40: Will this not look for the context dir in `<output>/run` or `<output>/build`? Shouldn't it be `<output>/context.run` or `<output>/context.build`?

phase/extender.go
```diff
@@ -366,12 +366,12 @@ func (e *Extender) dockerfilesFor(kind string, logger log.Logger) ([]extend.Dock
 }
 
 func (e *Extender) dockerfileFor(kind, extID string) (*extend.Dockerfile, error) {
-	dockerfilePath := filepath.Join(e.GeneratedDir, kind, launch.EscapeID(extID), "Dockerfile")
+	dockerfilePath := filepath.Join(e.GeneratedDir, launch.EscapeID(extID), fmt.Sprintf("%s.Dockerfile", kind))
```
natalieparellano, line 369: I think we need to switch on the Platform API version here so that the lifecycle will continue to work for older platforms

phase/generator.go
```diff
@@ -153,8 +160,8 @@ func (g *Generator) getGenerateInputs() buildpack.GenerateInputs {
 
 func (g *Generator) copyDockerfiles(dockerfiles []buildpack.DockerfileInfo) error {
 	for _, dockerfile := range dockerfiles {
-		targetDir := filepath.Join(g.GeneratedDir, dockerfile.Kind, launch.EscapeID(dockerfile.ExtensionID))
-		var targetPath = filepath.Join(targetDir, "Dockerfile")
+		targetDir := filepath.Join(g.GeneratedDir, launch.EscapeID(dockerfile.ExtensionID))
+		var targetPath = filepath.Join(targetDir, filepath.Base(dockerfile.Path))
```
natalieparellano, line 164: I believe we'd also need a switch here

buildpack/generate.go
```diff
@@ -114,6 +115,7 @@ func readOutputFilesExt(d ExtDescriptor, extOutputDir string, extPlanIn Plan, lo
 	var err error
 	var dfInfo DockerfileInfo
 	var found bool
+	var contexts []ContextInfo
```
natalieparellano, line 118: When we call `runGenerateCmd` we'll also need to set `extOutputDir` correctly:

> the $CNB_OUTPUT_DIR must point to the $CNB_GENERATED_DIR/<extension-id> folder instead of a temporary directory.

phase/extender.go
```diff
@@ -387,16 +387,35 @@ func (e *Extender) dockerfileFor(kind, extID string) (*extend.Dockerfile, error)
...
+		}
+	}
+
+	return e.AppDir, nil
+}
+
+func (e *Extender) extendOptions(dockerfile extend.Dockerfile) extend.Options {
 	return extend.Options{
-		BuildContext: e.AppDir,
+		BuildContext: dockerfile.ContextDir,
```
natalieparellano, line 418: We should default to the app dir if no context dir is provided

phase/extender.go
modulo11: Which one would be the correct platform API version? `v0.13`?

natalieparellano: Either sounds OK to me

buildpack/generate.go
modulo11: It's a bit hidden, but should be addressed by https://github.com/buildpacks/lifecycle/blob/a3fc1ef8508c88c1cdc37890b9224096241ebe66/phase/generator.go#L83. WDYT?

phase/extender.go
modulo11: Defaulted in https://github.com/buildpacks/lifecycle/blob/a3fc1ef8508c88c1cdc37890b9224096241ebe66/phase/extender.go#L376.

buildpack/generate.go
natalieparellano: Sounds good :)

loewenstein:
> @modulo11 @loewenstein thank you for this - apologies for being slow to review the changes. It looks like there might be a couple of tests to fix, but I'll take a deeper look tomorrow to see if I could offer more useful feedback.

@modulo11 did we have a look for what causes the test failures?

phase/generator.go
```diff
@@ -154,7 +169,13 @@ func (g *Generator) getGenerateInputs() buildpack.GenerateInputs {
 func (g *Generator) copyDockerfiles(dockerfiles []buildpack.DockerfileInfo) error {
 	for _, dockerfile := range dockerfiles {
 		targetDir := filepath.Join(g.GeneratedDir, dockerfile.Kind, launch.EscapeID(dockerfile.ExtensionID))
-		var targetPath = filepath.Join(targetDir, "Dockerfile")
+		targetPath := filepath.Join(targetDir, "Dockerfile")
+
+		if g.PlatformAPI.AtLeast("0.13") {
+			targetDir = filepath.Join(g.GeneratedDir, launch.EscapeID(dockerfile.ExtensionID))
+			targetPath = filepath.Join(targetDir, filepath.Base(dockerfile.Path))
```
natalieparellano, line 176: Isn't this a no-op?

phase/generator.go
```diff
@@ -176,6 +197,19 @@ func (g *Generator) copyDockerfiles(dockerfiles []buildpack.DockerfileInfo) erro
 	return nil
 }
 
+func (g *Generator) copyContexts(contexts []buildpack.ContextInfo) error {
```
natalieparellano, line 200: Is this also not a no-op? I thought that was the point of providing `filepath.Join(g.GeneratedDir, launch.EscapeID(contextDir.ExtensionID))` as the `CNB_OUTPUT_DIR`

buildpack/buildcontext.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// ContextInfo captures info about the context used for build- and run-image extensions
+type ContextInfo struct {
+	ExtensionID string
+	Kind        string
+	Path        string
+}
+
+func findContexts(d ExtDescriptor, extOutputDir string, logger log.Logger) ([]ContextInfo, error) {
```
natalieparellano, line 28: It would be really nice if we could unit test this somehow

natalieparellano (commented):
@modulo11 thank you for this. In general this looks good. I still need to go through the unit tests a bit more carefully to understand what is tested. 

It would be awesome if we could add an acceptance test (by adding a new fixture or symlinking to the child fixtures [here](https://github.com/buildpacks/lifecycle/tree/main/acceptance/testdata/extender/container/layers/generated/)). I actually think (without updating the fixtures) the current acceptance tests are going to start failing when we add 0.13 to the list of supported Platform APIs [here](https://github.com/buildpacks/lifecycle/blob/be5e24be9990fab639896150a006a0cd4d861db1/api/apis.go#L14). You could go ahead and do that as part of this PR if you want. When we add an (unreleased) API version we typically add a row to the table in the README, as shown in [this PR](https://github.com/buildpacks/lifecycle/pull/550/files).

LMK if I can be of any assistance here, especially with the acceptance test. Running our suite can be a bit tricky.

pbusko:
@natalieparellano we've addressed all the comments and introduced additional acceptance test with the extensions contexts

buildpack/generate.go
```diff
@@ -134,9 +136,15 @@ func readOutputFilesExt(d ExtDescriptor, extOutputDir string, extPlanIn Plan, lo
 		return GenerateOutputs{}, err
 	} else if found {
 		gr.Dockerfiles = append(gr.Dockerfiles, dfInfo)
+		logger.Debugf("Found '%d' Dockerfiles for processing", len(gr.Dockerfiles))
```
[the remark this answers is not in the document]
natalieparellano: If we could add another statement for run.Dockerfiles around line 133, that would be helpful

phase/generator_test.go
```diff
@@ -176,465 +174,696 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
...
-					dirStore.EXPECT().LookupExt("A", "v1").Return(&extA, nil)
-					executor.EXPECT().Generate(extA, gomock.Any(), gomock.Any()).Return(buildpack.GenerateOutputs{}, errors.New("some error"))
-
-					if _, err := generator.Generate(); err == nil {
-						t.Fatal("Expected error.\n")
-					} else if !strings.Contains(err.Error(), "some error") {
-						t.Fatalf("Incorrect error: %s\n", err)
-					}
-				})
+		when("using the platform API 0.13", func() {
```
natalieparellano, line 645: I'm not sure it's worth having a whole separate block for the 0.13 tests - it looks like if we run the 0.12 block with `generator.PlatformAPI = api.MustParse("0.13")` the only two tests that fail are:
* `it("copies Dockerfiles and extend-config.toml files to the correct locations", func() {`
* `descCondition: "a run.Dockerfile declares a new base image (only) and no run.Dockerfiles follow",
						descResult:    "returns the referenced base image, and sets extend to false",`

It might make our future lives easier to just conditionally set a variable for "dockerfile parent dir" depending on the Platform API and use that in the tests

phase/generator.go
```diff
@@ -154,10 +162,23 @@ func (g *Generator) getGenerateInputs() buildpack.GenerateInputs {
 func (g *Generator) copyDockerfiles(dockerfiles []buildpack.DockerfileInfo) error {
 	for _, dockerfile := range dockerfiles {
 		targetDir := filepath.Join(g.GeneratedDir, dockerfile.Kind, launch.EscapeID(dockerfile.ExtensionID))
-		var targetPath = filepath.Join(targetDir, "Dockerfile")
-		if dockerfile.Kind == buildpack.DockerfileKindRun && dockerfile.Ignore {
+		targetPath := filepath.Join(targetDir, "Dockerfile")
```
natalieparellano, line 165: Just a nit, but could we move the "target" code below the Platform API 0.13 block? This stuff:

```
		targetDir := filepath.Join(g.GeneratedDir, dockerfile.Kind, launch.EscapeID(dockerfile.ExtensionID))
		targetPath := filepath.Join(targetDir, "Dockerfile")
		if ignoreDockerfile {
			targetPath += ".ignore"
		}
```

I think that will make it easier to understand what's going on when viewing with fresh eyes

phase/extender_test.go
```diff
@@ -196,305 +196,622 @@ func testExtender(t *testing.T, when spec.G, it spec.S) {
...
-					secondDockerfileRebasable: false,
-					expectedImageSHA:          notRebasableSHA,
-				},
-			} {
-				tc := tc
-				when := when
-				desc := func(b bool) string {
-					if b {
-						return "rebasable"
+		when("using platform API 0.13", func() {
```
natalieparellano, line 507: Similar comment to https://github.com/buildpacks/lifecycle/pull/1276#discussion_r1489634499, I wonder if we could just make vars for "A build.Dockerfile path", "A build BuildContext" (etc) and make this more of a table test.

natalieparellano:
@pbusko @modulo11 thank you for this - I left a few comments but they are mostly nits about test organization. LMK what is your bandwidth for addressing these as it would be nice to cut an 0.19.0 RC soon. I'd be okay leaving https://github.com/buildpacks/lifecycle/pull/1276#discussion_r1489634499 and https://github.com/buildpacks/lifecycle/pull/1276#discussion_r1489675972 for later refactoring if we want to merge this one soon.

phil9909:
@natalieparellano I addressed the two small comments.
On the tests: I have some ideas how to change them, but there is no obviously superior way of doing this. So I would prefer to do this in a follow-up PR as you suggested, so we can start a fresh discussion there. Maybe even have two competing PRs.

natalieparellano (approved):

loewenstein (commented):
LGTM

Issue #1180, opened 2023-08-10
The `exporter` when writing extension layers should zero timestamps


natalieparellano:
### Summary

When using extensions to extend the run image, it appears that timestamps within the layers are not zeroed out. This hurts reproducibility.

---

##### Steps

1. Following the tutorial from https://github.com/buildpacks/docs/pull/581, modify the output image tag to be `localhost:5000/hello-extensions` (including the registry) and provide `--publish` to `pack build`
2. Use `crane manifest localhost:5000/hello-extensions` to get the digest of the extension layer (it should be the 6th layer)
3. Use `crane blob localhost:5000/hello-extensions@sha256:<extension layer digest> > blob.tar.gz
`
4. Expand the tarball and `ls -lt` from the root directory

##### Current behavior

The timestamps are not zeroed out:

```
$  ls -lt
total 0
drwxr-xr-x  3 narellano  staff   96 Aug 10 11:33 etc
drwxr-xr-x  3 narellano  staff   96 Jul 26 08:16 lib
drwxr-xr-x  6 narellano  staff  192 Jul 26 08:16 usr
drwxr-xr-x  3 narellano  staff   96 Jun 14 11:03 var
```

##### Expected behavior

The timestamps should be zeroed out.

---

##### lifecycle version

0.17.0

##### anything else?

We don't use `Reproducible: true` in kaniko for reasons having to do with the fact that we don't pull the complete base image (since we are already running in the context of the image that is being extended). See where we override `image.RetrieveRemoteImage` and set `InitialFSUnpacked: true`. Available options seem to be:

1. Try to modify kaniko
2. Have the exporter zero out timestamps when streaming the layer to the target (this may be easier)

PR #1289, opened 2024-02-08
Exporter zeroes timestamps when adding extension layers to the app image

natalieparellano:
### Summary

Before, we simply added the layer as output by kaniko (though we uncompressed the layer for daemon images).

After, we uncompress the layer so that we can read the tarball, then zero the mod time for each header. We can add the layer to the image uncompressed, as the specific imgutil layout/remote implementations will take care of compressing the layer where needed. The imgutil local package wants the layer uncompressed. 

We should investigate if it's possible to force kaniko not to do any compression - looking [here](https://github.com/GoogleContainerTools/kaniko/blob/11e2e9f40847754141ee8f1c009248facc651467/pkg/executor/build.go#L533-L535) and [here](https://github.com/google/go-containerregistry/blob/8dadbe76ff8c20d0e509406f04b7eade43baa6c1/pkg/v1/tarball/layer.go#L109) this seems intentionally not permitted.

#### Release notes

Exporter zeroes timestamps when adding extension layers to the app image

---

### Related

Resolves https://github.com/buildpacks/lifecycle/issues/1180


phase/exporter.go
```diff
@@ -362,34 +357,6 @@ func isLocalImage(workingImage imgutil.Image) bool {
 	return false
 }
 
-func uncompressLayerAt(layerPath string, toArtifactsDir string) (string, error) {
```
natalieparellano, line 365: IDK why I added this function, we could just call `layer.Uncompressed()`

natalieparellano:
IDK, I think Codecov is being too picky here (I even added a bunch of tests!)

phase/exporter.go
```diff
@@ -281,54 +279,51 @@ func (e *Exporter) addExtensionLayers(opts ExportOptions) error {
...
-		if localImage {
-			var calculatedDiffID string
-			if calculatedDiffID, err = uncompressLayerAt(layerPath, artifactsDir); err != nil {
-				return err
-			}
-			if calculatedDiffID != layerHex.String() {
-				return fmt.Errorf("digest of uncompressed layer from %s does not match expected value; found %q, expected %q", layerPath, calculatedDiffID, layerHex.String())
-			}
-			layerPath = filepath.Join(artifactsDir, calculatedDiffID)
+		if !blobExists {
```
jabrown85, line 287: If the blob doesn't exist we are skipping it and not throwing an error - is that expected? If so, maybe a comment explaining why would be good for future us!

jabrown85 (approved):
I added a question but looks good

phase/exporter.go
natalieparellano, line 287:
```suggestion
		if !blobExists {
		// the referenced blob is a run image (base image) layer, not an extension-provided layer
```

natalieparellano: Added a comment in https://github.com/buildpacks/lifecycle/pull/1289/commits/51050acd241dc0150df004bf17f08218b0ecb2a3

PR #1297, opened 2024-02-16
refactor `phase/extender_test.go`

phil9909:
### Summary

Refactor `phase/extender_test.go` as suggested in https://github.com/buildpacks/lifecycle/pull/1276#discussion_r1489675972

### Context

In #1276 we duplicated most of the test code in `phase/extender_test.go`, basically doing the same tests twice, once for Platform API `0.12` and once for `0.13`. The API versions expect a different folder layout, which requires changes to the test setup. The reduce the size of `phase/extender_test.go`, I:
- moved the code to create the folder layout into a helper function (it will check for the Platform API version and behave differently based on that)
- added a loop around the test cases looping over API versions.

Now we have every tests only once in code, while still being executed twice.

The drawback: The helper function is quite complex (especially for test code). 

A middle ground could be to keep the loop, but to the Platform API version switch in every test case.

@natalieparellano WDYT?

### DIFF
--- a/phase/extender_test.go
+++ b/phase/extender_test.go
@@ -196,623 +196,327 @@ func testExtender(t *testing.T, when spec.G, it spec.S) {
 			mockCtrl.Finish()
 		})
 
-		when("using the platform API 0.12", func() {
-			it.Before(func() {
-				extender.PlatformAPI = api.MustParse("0.12")
-			})
+		prepareDockerfile := func(id, kind, contextDir string) extend.Dockerfile {
+			basePath := filepath.Join(generatedDir, kind, id)
+			dockerfilePath := filepath.Join(basePath, "Dockerfile")
 
-			when("build base image", func() {
-				it("applies the Dockerfile with the expected args and opts", func() {
-					expectedDockerfileA := extend.Dockerfile{
-						Path: filepath.Join(generatedDir, "build", "A", "Dockerfile"),
-						Args: []extend.Arg{{Name: "argA", Value: "valueA"}},
-					}
-					h.Mkdir(t, filepath.Join(generatedDir, "build", "A"))
-					h.Mkfile(t, "some dockerfile content", filepath.Join(generatedDir, "build", "A", "Dockerfile"))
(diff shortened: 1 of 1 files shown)


phase/extender_test.go
```diff
@@ -196,623 +196,327 @@ func testExtender(t *testing.T, when spec.G, it spec.S) {
 			mockCtrl.Finish()
 		})
 
-		when("using the platform API 0.12", func() {
-			it.Before(func() {
-				extender.PlatformAPI = api.MustParse("0.12")
-			})
+		prepareDockerfile := func(id, kind, contextDir string) extend.Dockerfile {
```
natalieparellano, line 199:
```suggestion
		// prepares fixtures and expected data for extensions
		// On older platforms, the extender looks for Dockerfiles in <generated>/<kind>/<extension-id>/Dockerfile.
		// On newer platforms, it looks for Dockerfiles in <generated>/<extension-id>/kind.Dockerfile, and context directories are allowed.
		prepareDockerfile := func(id, kind, contextDir string) extend.Dockerfile {
```

phase/extender_test.go
```diff
@@ -196,623 +196,327 @@ func testExtender(t *testing.T, when spec.G, it spec.S) {
...
+			switch contextDir {
+			case "specific":
+				h.Mkdir(t, filepath.Join(basePath, fmt.Sprintf("context.%s", kind)))
+			case "shared":
+				h.Mkdir(t, filepath.Join(basePath, "context"))
+			case "app":
+				// nothing to do
+			default:
+				t.Fail()
+			}
```
natalieparellano, line 236: This might read easier if we move it into the Platform API 0.13 block (around line 205)

natalieparellano (approved):
@phil9909 thanks a lot for this! Despite the complexity of `prepareDockerfile` I think this will make our lives a lot easier. I added a comment suggestion to hopefully make it friendlier for newcomers.

phil9909:
Added your suggestions

PR #1301, opened 2024-02-20
refactor `phase/generator_test.go`

phil9909:
### Summary

Refactor `phase/generator_test.go` as suggested in https://github.com/buildpacks/lifecycle/pull/1276#discussion_r1489634499.
Related PR #1297

### Context

In #1276 we duplicated some of the test code in `phase/generator_test.go`, basically doing the same tests twice, once for Platform API `0.12` and once for `0.13`. The API versions expect a different folder layout, which requires changes to the test setup and assertions. The reduce the size of `phase/generator_test.go`, we:
- added a loop around the test cases looping over API versions.
- for API version >= 0.13 skip the copy tests
- added some `platformAPI.AtLeast("0.13")` switches where file paths are different

Now we have every tests only once in code, while still being executed twice (even the ones that we previously only had for platform API 0.12).

CC @natalieparellano

### DIFF
--- a/phase/generator_test.go
+++ b/phase/generator_test.go
@@ -2,6 +2,7 @@ package phase_test
 
 import (
 	"bytes"
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


natalieparellano (approved):
@phil9909 thank you for this! This should make the tests a lot easier to maintain going forward :)

PR #1304, opened 2024-02-29
Document deprecatedRunImage as deprecated

ryanbrainard:
### Summary

When looking at the flags for `rebase`, it is confusing because both `-image` and `-run-image` are documented as `reference to run image`. It's not until inspection of the source code that you realize `-image` is deprecated. This adds a `[deprecated]` marker to `-image` so it can be seen in the help output.

Looking at uses of `DeprecatedFlagRunImage`, this impacts both `rebase` and `export`.

#### Release notes

Document `-image` flag as deprecated in help output for `rebase` and `export`.


jabrown85 (approved):
I like this!

PR #1305, opened 2024-02-29
Fix rebase run-image resolution

ryanbrainard:
### Summary

Currently, if `-run-image` is not set, `io.buildpacks.lifecycle.metdata[runImage.reference]` is used. This does not follow the [Run Image Resolution spec](https://github.com/buildpacks/spec/blob/main/platform.md#run-image-resolution), which specifies using `io.buildpacks.lifecycle.metdata[runImage.image]` and optionally `io.buildpacks.lifecycle.metdata[runImage.mirrors]`. Because of this, it ends up making `lifecycle rebase` without the `-run-image` flag a no-op because the run image is pinned to the same version instead of getting the latest.

This change simplifies and unifies the behavior for before and after Platform Version 0.12 so that they both read the same run-image data type (just from different locations), and then validate and resolve mirrors the same.

Note, this is also consistent with [the way `pack rebase` resolves run-images](https://github.com/buildpacks/pack/blob/8b58c6eb9d1e7b40644b8d64017355d58253dc82/pkg/client/rebase.go#L80-L91); however, it does not switch on the Platform Version (I didn't want to change that here).

#### Release notes

Fix `lifecycle rebase` automatic run image resolution from `io.buildpacks.lifecycle.metdata` to read keys `runImage.image` (and optionally `runImage.mirrors`) instead of `runImage.reference`.

### Context

I did not see any relevant tests for this code, but please let me know if there's a test I should change or add. I did do various permutations of manual testing and it is working as expected.

### DIFF
--- a/cmd/lifecycle/rebaser.go
+++ b/cmd/lifecycle/rebaser.go
@@ -191,21 +191,19 @@ func (r *rebaseCmd) setAppImage() error {
 	}
 
 	if r.RunImageRef == "" {
-		if r.PlatformAPI.AtLeast("0.12") {
-			r.RunImageRef = md.RunImage.Reference
-			if r.RunImageRef != "" {
-				return nil
-			}
-		}
-
-		// for backwards compatibility, we need to fallback to the stack metadata
-		// fail if there is no run image metadata available from either location
-		if md.Stack == nil || md.Stack.RunImage.Image == "" {
+		var runImage files.RunImageForExport
+		if r.PlatformAPI.AtLeast("0.12") && md.RunImage.RunImageForExport.Image != "" {
+			runImage = md.RunImage.RunImageForExport
+		} else if md.Stack != nil && md.Stack.RunImage.Image != "" {
+			// for backwards compatibility, we need to fallback to the stack metadata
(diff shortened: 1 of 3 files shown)


jabrown85 (approved):

ryanbrainard:
> Thank you for this. It would be great if we could test this somehow - an acceptance test would probably be too heavy handed, but perhaps some of this logic could move to https://github.com/buildpacks/lifecycle/blob/main/platform/resolve_inputs.go (adding a function similar to [here](https://github.com/buildpacks/lifecycle/blob/main/platform/resolve_inputs.go#L131-L148))

Yes, I would love to add a test. I just looked at `resolve_inputs.go`, but that doesn't look like it'll work because this needs to happen after input resolution because of some caveots with `-daemon` (according to the comments). Looks like `platform/run_image.go` would be a good place where I can at least extract out the `switch` statement and put some tests around it. I'll push something up soon.

ryanbrainard:
@natalieparellano I added a test in https://github.com/buildpacks/lifecycle/pull/1305/commits/5a5511127c2496cab81fe8395cbe53f67d98d231

Issue #1186, opened 2023-08-23
Implementation of target env vars is misleading


natalieparellano:
### Summary

To provide `CNB_TARGET_*` vars in the buildpack env, the builder [concatenates](https://github.com/buildpacks/lifecycle/blob/7fda8e271f47d58b12df17a919ac8e9691f510b8/builder.go#L69) these to the lifecycle's env. This works because the `CNB_TARGET_*` vars are added to the "[allow list](https://github.com/buildpacks/lifecycle/blob/7fda8e271f47d58b12df17a919ac8e9691f510b8/env/build.go#L11-L15)" of env vars that the lifecycle will pass through to buildpacks.

However, this is misleading. Someone reading the code might think that `CNB_TARGET_*` vars need to be set in the lifecycle's env, but this is not the case (the analyzer reads this data from the run image).

We should use something like [`WithOverrides`](https://github.com/buildpacks/lifecycle/blob/7fda8e271f47d58b12df17a919ac8e9691f510b8/env/env.go#L92) to append target vars to the buildpack env in builder.go.

Note this functionality is [covered](https://github.com/buildpacks/lifecycle/blob/7fda8e271f47d58b12df17a919ac8e9691f510b8/acceptance/builder_test.go#L521-L525) by acceptance so it should be safe to refactor at the unit level.

##### lifecycle version

0.17.0+

##### platform version(s)

0.12+

UtkarshUmre:
I'm new to this project, but I'm interested in working on this good first issue. May I give it a try?

natalieparellano:
@UtkarshUmre we would love to have your help with this! Please let us know if there is any additional guidance we can provide that would be helpful

sarthaksarthak9:
@UtkarshUmre r u still working on this issue?

UtkarshUmre:
yes @sarthaksarthak9 I'm still working on this issue

sarthaksarthak9:
> yes @sarthaksarthak9 I'm still working on this issue

ok, If u need some help .. I would love to do!!

UtkarshUmre:
@sarthaksarthak9 thank you for offering your help. I believe I'm going to need assistance with this one. Where can I contact you?

sarthaksarthak9:
> @sarthaksarthak9 thank you for offering your help. I believe I'm going to need assistance with this one. Where can I contact you?

you can ask your query here onwards or can dm me on slack

edmorley:
[...]

PR #1309, opened 2024-03-04
Always set CNB_TARGET_* variables during detect, build, and generate

### LINKED ISSUE #1308: `CNB_TARGET_` environment variables are not set when expected
### Summary

In [Buildpack API 0.10](https://github.com/buildpacks/spec/blob/buildpack/0.10/buildpack.md?plain=1#L750-L760), it says:

> The following environment variables MUST be set by the lifecycle during the detect and build phases to describe the target runtime image, if inputs are provided.

But `lifecycle` fails to provide these `CNB_TARGET_*` environment variables if the Platform API is less than `0.12`:

https://github.com/buildpacks/lifecycle/blob/be881f29d5b4a33795e1b61bb262ce9ae439cc10/phase/analyzer.go#L84-L93

This means that a buildpack using Buildpack API 0.10 running on a platform using Platform API 0.9 won't have `CNB_TARGET_OS` and `CNB_TARGET_ARCH` set, even though these should be set according to the Buildpack API.

---

##### Steps

1. Use an older version of `pack`, like `v0.27.0`, that uses a Platform API version < 0.12
2. Run `pack build myapp --builder heroku/builder@sha256:5856c2e1f6c55ec5e289453b872f57f4ebe1a32a6da31f053cd27eddccbf21c5`
3. Observe that the build fails (some of the included buildpacks try to read `CNB_TARGET_OS`, which is not present)
4. Use a newer version of `pack` that uses Platform API version >= 0.12
5. Run `pack build myapp --builder heroku/builder@sha256:5856c2e1f6c55ec5e289453b872f57f4ebe1a32a6da31f053cd27eddccbf21c5` again.
6. Observe that the build doesn't error on missing environment variables.

##### Current behavior

Buildpacks using Buildpack API 0.10 may not receive `CNB_TARGET_OS` / `CNB_TARGET_PLATFORM` env vars when an older Platform API is in use.

##### Expected behavior

Buildpacks using Buildpack API 0.10 always receive `CNB_TARGET_OS` / `CNB_TARGET_PLATFORM` (regardless of Platform API version).

---

##### lifecycle version

`0.18.5` in my testing, but I think older versions are affected as well.

##### platform version(s)

Anything < 0.12.

natalieparellano:
when the Buildpack API version is at least 0.10

### Summary

Previously, we only set these variables when the Platform API version was at least 0.12. But, newer Buildpack APIs expect these variables regardless of the Platform API version. If we are on an older platform, derive the target variables from the base image OS.

#### Release notes

The lifecycle always set CNB_TARGET_* variables during detect, build, and generate when the Buildpack API version is at least 0.10

---

### Related

Resolves #1308 and #1186

---

### DIFF
--- a/buildpack/build.go
+++ b/buildpack/build.go
@@ -35,6 +35,7 @@ type BuildInputs struct {
 	LayersDir      string
 	PlatformDir    string
 	Env            BuildEnv
+	TargetEnv      []string
 	Out, Err       io.Writer
 	Plan           Plan
 }
@@ -150,6 +151,9 @@ func runBuildCmd(d BpDescriptor, bpLayersDir, planPath string, inputs BuildInput
 			EnvBpPlanPath+"="+planPath,
 			EnvLayersDir+"="+bpLayersDir,
 		)
+	}
+	if api.MustParse(d.API()).AtLeast("0.10") {
+		cmd.Env = append(cmd.Env, inputs.TargetEnv...)
 	}
 
 	if err = cmd.Run(); err != nil {
--- a/buildpack/detect.go
+++ b/buildpack/detect.go
@@ -30,6 +30,7 @@ type DetectInputs struct {
 	BuildConfigDir string
 	PlatformDir    string
 	Env            BuildEnv
+	TargetEnv      []string
 }
 
 type DetectOutputs struct {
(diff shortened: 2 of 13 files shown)


jabrown85 (approved):

edmorley:
@natalieparellano Thank you for fixing this! We're now using lifecycle 0.19.0 in our newer builders and it's resolved the issue. 

However, I've just realised we have an older builder that's stuck on `lifecycle` 0.17.x (since it has to support buildpacks using Buildpack API 0.6) that's also affected. I see [RFC-120](https://github.com/buildpacks/rfcs/blob/main/text/0120-cvebackports.md) only covers CVE backporting to older lifecycle releases - do you have any thoughts as to whether this bug fix might be something that could/should be backported? 

(I think in general backporting should be for exceptional circumstances only, however, I wonder if for this specific case it might be worth the effort, just so buildpack authors don't hold back from upgrading to newer Buildpack API versions and targets instead of stacks)

natalieparellano:
@edmorley that sounds reasonable to me - I created https://github.com/buildpacks/lifecycle/pull/1317

edmorley:
Thank you - much appreciated! 😄

PR #1314, opened 2024-03-08
Fix: log level and color level should be configurable via the env

natalieparellano:
### Summary

Currently, log level and color level are only configurable via flags (`-log-level` and `-no-color`).

This broke awhile back in an effort to consolidate some logic in the `platform` package.

#### Release notes

Log level and color level are configurable via the environment (in addition to flags)

---

### Context

Screenshot 2024-03-08 at 12 09 19 PM

### DIFF
--- a/cmd/lifecycle/cli/command.go
+++ b/cmd/lifecycle/cli/command.go
@@ -33,9 +33,10 @@ func Run(c Command, withPhaseName string, asSubcommand bool) {
 	var printVersion bool
 	FlagVersion(&printVersion)
 
-	// DefineFlags along with any function FlagXXX defines the flags that are considered valid
-	// `flagSet.Parse` reads their values (if provided)
-	// The command `c` (e.g., detectCmd) is at this point already populated with platform inputs from the environment and/or default values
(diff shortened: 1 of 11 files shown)


cmd/lifecycle/cli/command.go
```diff
@@ -47,22 +48,23 @@ func Run(c Command, withPhaseName string, asSubcommand bool) {
...
 		if arg[0:1] == "-" {
 			cmd.DefaultLogger.Warnf("Warning: unconsumed flag-like positional arg: \n\t%s\n\t This will not be interpreted as a flag.\n\t Did you mean to put this before the first positional argument?", arg)
 		}
 	}
 
+	cmd.DisableColor(c.Inputs().NoColor)
+	if err := cmd.DefaultLogger.SetLevel(c.Inputs().LogLevel); err != nil {
+		cmd.Exit(err)
+	}
+	cmd.DefaultLogger.Debugf("Starting %s...", withPhaseName)
```
natalieparellano, line 66: Hmm I don't know why I moved this - we should disable color before we print the warning

jabrown85 (approved):

PR #1315, opened 2024-03-08
Allow env vars to skip vendor specific keychain

jabrown85:
### Summary

Vendor keychains can be slow or fail. This allows platform operators to skip them entirely.

#### Release notes

Allow platform operators the ability to skip vendor specific keychain implementations by setting `CNB_REGISTRY_AUTH_KEYCHAIN_SKIP_AMAZON` or `CNB_REGISTRY_AUTH_KEYCHAIN_SKIP_AZURE`.

---

### Related

Resolves https://github.com/buildpacks/lifecycle/issues/1007#issuecomment-1561349524

---

### Context

The Amazon/ECR one is _terribly_ inefficient. It tries to query the metadata service and eventually fails after a few tries. This is not needed if the platform is providing the credentials via Default/Env. Selectively disabling seemed liked a decent enough idea.

### DIFF
--- a/auth/keychain.go
+++ b/auth/keychain.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"regexp"
+	"strings"
 
 	ecr "github.com/awslabs/amazon-ecr-credential-helper/ecr-login"
 	"github.com/chrismellard/docker-credential-acr-env/pkg/credhelper"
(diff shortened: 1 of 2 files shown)


natalieparellano:
@jabrown85 I think this makes a lot of sense - and, this would allow us to merge https://github.com/buildpacks/lifecycle/pull/887 with less trepidation that it could break certain environments (I think the original underlying issue there was fixed, and we just forgot about the PR, but still...).

natalieparellano (approved):

auth/keychain.go
```diff
@@ -34,14 +38,25 @@ func DefaultKeychain(images ...string) (authn.Keychain, error) {
...
 	}
 
+	keychains := []authn.Keychain{envKeychain, authn.DefaultKeychain}
+
+	if vendorKeychainEnabled("amazon") {
+		keychains = append(keychains, amazonKeychain)
+	}
+	if vendorKeychainEnabled("azure") {
+		keychains = append(keychains, azureKeychain)
+	}
```
matejvasek, line 48: Were the calls to `NewResolvedKeychain()` omitted deliberately?

PR #1333, opened 2024-04-09
Initialize default `LayerMetadataRestorer` and `SBOMRestorer` if none provided

pbusko:
### Summary

Initialize default `layer.LayerMetadataRestorer` and `layer.SBOMRestorer` structs if none provided.

#### Release notes

---

### Related

Resolves #___

---

### Context

At the moment, it is impossible to reuse `phase.Restorer` structure due to the usage of two internal structs: `layer.LayerMetadataRestorer` and `layer.SBOMRestorer`. Defaulting them with the default constructors would allow re-use of the `phase.Restorer` without opening up the internal package.

This change would greatly improve future work on https://github.com/cloudfoundry/community/pull/796

### DIFF
--- a/phase/restorer.go
+++ b/phase/restorer.go
@@ -35,6 +35,18 @@ func (r *Restorer) Restore(cache Cache) error {
 		return err
 	}
 
+	if r.LayerMetadataRestorer == nil {
+		r.LayerMetadataRestorer = layer.NewDefaultMetadataRestorer(r.LayersDir, false, r.Logger)
+	}
+
+	if r.SBOMRestorer == nil {
+		r.SBOMRestorer = layer.NewSBOMRestorer(layer.SBOMRestorerOpts{
+			LayersDir: r.LayersDir,
+			Logger:    r.Logger,
+			Nop:       false,
+		}, r.PlatformAPI)
+	}
+
 	layerSHAStore := layer.NewSHAStore()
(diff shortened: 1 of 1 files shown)


phase/restorer.go
```diff
@@ -35,6 +35,18 @@ func (r *Restorer) Restore(cache Cache) error {
 		return err
 	}
 
+	if r.LayerMetadataRestorer == nil {
+		r.LayerMetadataRestorer = layer.NewDefaultMetadataRestorer(r.LayersDir, false, r.Logger)
```
natalieparellano, line 39: Should "skip layers" not be a configurable option?

natalieparellano (commented):
I would love to see us move the logic for constructing a new restorer into a "factory" method similar to `ConnectedFactory.NewAnalyzer`, as part of our ongoing quest to move stuff out of the main package. But, that doesn't need to block improving this code now.

natalieparellano:
Rebasing against `main` should fix the Codecov failure

phase/restorer.go
pbusko: The idea was to make this change minimally invasive, with no changes to the signature of the public receiver functions.

c0d1ngm0nk3y: If you want to "skip layers", you can still provide a `LayerMetadataRestorer`.

natalieparellano (approved):

PR #1334, opened 2024-04-09
Fix auth by wrapping keychain in a ResolvedKeychain

natalieparellano:
Prior to https://github.com/buildpacks/lifecycle/pull/1315, all keychains passed to NewMultiKeychain were resolved keychains, which prevented the credentials from becoming inaccessible after the lifecycle dropped privileges.


jjbustamante (approved):
I saw this error Yesterday with @matejvasek, this change looks good to me according to our analysis. `NewResolvedKeychain` was removed in the previous commit and need to be restore

jabrown85 (approved):

natalieparellano:
Looks like the Codecov token is now required, this should be fixed in https://github.com/buildpacks/lifecycle/pull/1282 then we can merge this one

jabrown85:
You should be able to rebase now @natalieparellano

PR #1335, opened 2024-04-10
Bump imgutil and docker

natalieparellano:
#### Release notes

---

### Related

Resolves #___

---


natalieparellano:
Relies on a dev version of imgutil

phase/exporter.go
```diff
@@ -613,21 +613,7 @@ func (e *Exporter) addOrReuseBuildpackLayer(image imgutil.Image, layer layers.La
 }
 
 func (e *Exporter) addOrReuseExtensionLayer(image imgutil.Image, layer layers.Layer) (string, error) {
-	rc, err := image.GetLayer(layer.Digest)
-	if err != nil {
-		// FIXME: imgutil should declare an error type for missing layer
```
natalieparellano, line 618: A failing test due to the error string being different cause me to take a closer look at this method. I think the current code is wrong in the sense that the working image at this point is just the base image and so it will NEVER contain the extension layer. What we're after is if the PREVIOUS image contained the extension layer.

This effectively just avoids one call to `tarball.LayerFromFile`, so I'm not sure how much of an optimization it really is.

natalieparellano:
Needs https://github.com/buildpacks/imgutil/pull/266

phase/exporter.go
natalieparellano: After adding `AddOrReuseLayerWithHistory` to imgutil I decided to just use `AddLayerWithDiffIDAndHistory` here, as otherwise we need to update the "caching image" to re-implement `AddOrReuseLayerWithHistory` and that would effectively just call `AddLayerWithDiffIDAndHistory` anyway. We can cleanup the imgutil.Image interface to remove `AddOrReuseLayerWithHistory` if we don't foresee ever needing it.

jabrown85 (approved):

PR #1347, opened 2024-04-26
Read `/etc/os-release` file when distro information is not present in labels

### LINKED ISSUE #1336: Target distro name/version not derived from `/etc/os-release` when Docker labels aren't set
### Summary

The Platform API 0.12 spec says:

> - If target distribution data is missing, it will be inferred from `/etc/os-release` for Linux images

(https://github.com/buildpacks/spec/blob/platform/0.12/platform.md#analyzedtoml-toml)

As such `lifecycle` has an implementation for this:
https://github.com/buildpacks/lifecycle/blob/feb0d757c9bc16b85f19cb07c80dc28f0bdd8504/platform/target_data.go#L66-L81

However, it seems this implementation never actually gets used in practice since it's behind a `if tm.OS == ""` conditional, and `OS` is always available, so the `/etc/os-release` fallback isn't used even if other properties about the target (such as the distro name and version) are not available:
https://github.com/buildpacks/lifecycle/blob/feb0d757c9bc16b85f19cb07c80dc28f0bdd8504/platform/target_data.go#L15-L19
(Also, why does that code comment say build image - surely it should be analysing the run image?)

This means that for any run images where the `io.buildpacks.base.distro.*` Docker labels haven't been set (which is any run image that hasn't yet made the transition from stacks to targets), the `CNB_TARGET_DISTRO_NAME` and `CNB_TARGET_DISTRO_VERSION` env vars are correctly set in the buildpack environment.

This can be seen by using one of our older run images from before we added the relevant Docker labels - by passing `--run-image heroku/heroku:22-cnb.v122` to `pack build`.

---

##### Steps

1. `pack buildpack new testcase-targets --api 0.10 --targets "linux/amd64"`
1. `sed -i 's/exit 0/printenv | grep CNB_TARGET | sort/' testcase-targets/bin/build`
1. `pack build --builder heroku/builder:22 --run-image heroku/heroku:22-cnb.v122 --buildpack testcase-targets/ --path testcase-targets/ testcase-targets --verbose`

##### Current behavior

The `CNB_TARGET_DISTRO_NAME` and `CNB_TARGET_DISTRO_VERSION` env vars are not set correctly in the buildpack env:

```
$ pack build --builder heroku/builder:22 --run-image heroku/heroku:22-cnb.v122 --buildpack testcase-targets/ --path testcase-targets/ testcase-targets --verbose
...
===> ANALYZING
...
Run image info in analyzed metadata is: 
{"Reference":"6d5f868e91eba6a6ec22a9a9b9b31a2fd486a0f0f57082fd0864e0bf0b3e4395","Image":"heroku/heroku:22-cnb.v122","Extend":false,"target":{"os":"linux","arch":"amd64"}}
...
===> BUILDING
...
CNB_TARGET_ARCH=amd64
CNB_TARGET_ARCH_VARIANT=
CNB_TARGET_DISTRO_NAME=
CNB_TARGET_DISTRO_VERSION=
CNB_TARGET_OS=linux
...
```

##### Expected behavior

For the `CNB_TARGET_DISTRO_*` env vars to be set based on the contents of `/etc/os-release`, ie:

```
CNB_TARGET_DISTRO_NAME=ubuntu
CNB_TARGET_DISTRO_VERSION=22.04
```

...given that `/etc/os-release` exists and has the necessary `ID` and `VERSION_ID` values:

```
$ docker run --rm heroku/heroku:22-cnb.v122 cat /etc/os-release | rg ID= | sort
ID=ubuntu
VERSION_ID="22.04"
```

---

##### lifecycle version
`0.19.3`

##### platform version(s)

```
$ pack report
Pack:
  Version:  0.33.2+git-f2cffc4.build-5562
  OS/Arch:  darwin/arm64
```

The build is using Platform API 0.12 and Buildpack API 0.10.

pbusko:
### Summary

Related issue: https://github.com/buildpacks/lifecycle/issues/1336

#### Release notes

---

### Related

Resolves #1336

---

### DIFF
--- a/phase/builder_test.go
+++ b/phase/builder_test.go
@@ -191,8 +191,6 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_ARCH=amd64")
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_ARCH_VARIANT=")
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_OS=linux")
-					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_NAME=")
-					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_VERSION=")
 					return buildpack.BuildOutputs{}, nil
 				},
 			)
@@ -205,8 +203,6 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_ARCH=amd64")
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_ARCH_VARIANT=")
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_OS=linux")
-					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_NAME=")
-					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_VERSION=")
 					return buildpack.BuildOutputs{}, nil
 				})
 
--- a/phase/generator_test.go
+++ b/phase/generator_test.go
@@ -284,8 +284,6 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
 							h.AssertContains(t, inputs.TargetEnv,
 								"CNB_TARGET_ARCH=amd64",
 								"CNB_TARGET_ARCH_VARIANT=",
-								"CNB_TARGET_DISTRO_NAME=ubuntu",
-								"CNB_TARGET_DISTRO_VERSION=22.04",
 								"CNB_TARGET_OS=linux",
 							)
 							return buildpack.GenerateOutputs{Dockerfiles: []buildpack.DockerfileInfo{{ExtensionID: d.Extension.ID,
--- a/phase/generator_test.go
+++ b/phase/generator_test.go
@@ -284,8 +284,8 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
 							h.AssertContains(t, inputs.TargetEnv,
 								"CNB_TARGET_ARCH=amd64",
 								"CNB_TARGET_ARCH_VARIANT=",
-								"CNB_TARGET_DISTRO_NAME=",
-								"CNB_TARGET_DISTRO_VERSION=",
+								"CNB_TARGET_DISTRO_NAME=ubuntu",
+								"CNB_TARGET_DISTRO_VERSION=22.04",
 								"CNB_TARGET_OS=linux",
 							)
 							return buildpack.GenerateOutputs{Dockerfiles: []buildpack.DockerfileInfo{{ExtensionID: d.Extension.ID,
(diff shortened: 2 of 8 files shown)


platform/target_data.go
```diff
@@ -11,11 +11,11 @@ import (
...
 // EnvVarsFor fulfills the prophecy set forth in https://github.com/buildpacks/rfcs/blob/b8abe33f2bdc58792acf0bd094dc4ce3c8a54dbb/text/0096-remove-stacks-mixins.md?plain=1#L97
 // by returning an array of "VARIABLE=value" strings suitable for inclusion in your environment or complete breakfast.
-func EnvVarsFor(tm files.TargetMetadata, logger log.Logger) []string {
+func EnvVarsFor(d fsutil.Detector, tm files.TargetMetadata, logger log.Logger) []string {
 	// we should always have os & arch,
 	// if they are not populated try to get target information from the build-time base image
-	if tm.OS == "" {
-		GetTargetOSFromFileSystem(&fsutil.Detect{}, &tm, logger)
+	if tm.OS == "" || tm.Distro == nil {
+		GetTargetOSFromFileSystem(d, &tm, logger)
```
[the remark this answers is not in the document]
natalieparellano: That sounds great to me

pbusko:
@natalieparellano unfortunately the unit tests on windows are failing due to unknown to me reasons and I'm unable to debug it on that platform, seems like the change itself included in the PR does not have any effect on windows?

phase/builder_test.go
```diff
@@ -191,8 +191,8 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_ARCH=amd64")
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_ARCH_VARIANT=")
 					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_OS=linux")
-					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_NAME=")
-					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_VERSION=")
+					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_NAME=ubuntu")
+					h.AssertContains(t, inputs.TargetEnv, "CNB_TARGET_DISTRO_VERSION=22.04")
```
natalieparellano, line 195: I think these assertions are causing the tests to fail on Windows (and in face would only pass when run on an ubuntu/22.04 system). I think we can just remove them.

platform/target_data.go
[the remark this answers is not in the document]
natalieparellano: @pbusko would you be up for adding this?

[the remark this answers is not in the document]
pbusko: sure

phase/generator_test.go
```diff
@@ -284,8 +284,8 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
 							h.AssertContains(t, inputs.TargetEnv,
 								"CNB_TARGET_ARCH=amd64",
 								"CNB_TARGET_ARCH_VARIANT=",
-								"CNB_TARGET_DISTRO_NAME=",
-								"CNB_TARGET_DISTRO_VERSION=",
+								"CNB_TARGET_DISTRO_NAME=ubuntu",
+								"CNB_TARGET_DISTRO_VERSION=22.04",
```
natalieparellano, line 288:
```suggestion
```

phase/generator_test.go
```diff
@@ -296,8 +296,8 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
...
```
natalieparellano, line 300:
```suggestion
```

phase/generator_test.go
natalieparellano: The generator tests are skipped on Windows, which is why these aren't failing, but we don't want to tie ourselves to a particular environment for getting these tests to pass.

natalieparellano (approved):

PR #1352, opened 2024-05-10
Also read distro information from /etc/os-release when checking target compat

natalieparellano:
### Summary

This will fix an issue reported by our friends from Paketo. Without this fallback, when combined with the stricter validation in https://github.com/buildpacks/lifecycle/pull/1348, we could fail builds that we actually want to proceed with.

https://github.com/buildpacks/lifecycle/pull/1347 reads the file when providing target env vars to buildpacks during detect, but we also need to consider this info when deciding whether or not to run detect for the buildpack.

#### Release notes

The `detector`, when determining if a base image satisfies target constraints declared by a buildpack, also uses information from `/etc/os-release` when distro information is not present in the run image labels

---

### Related

Resolves #___

---

### Context

I moved some functions around in files, so ignore the first commit when parsing the diff (it just moves stuff doesn't change any logic)

### DIFF
--- a/phase/analyzer.go
+++ b/phase/analyzer.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/buildpacks/lifecycle/api"
 	"github.com/buildpacks/lifecycle/image"
-	"github.com/buildpacks/lifecycle/internal/fsutil"
 	"github.com/buildpacks/lifecycle/internal/layer"
 	"github.com/buildpacks/lifecycle/log"
 	"github.com/buildpacks/lifecycle/platform"
@@ -86,9 +85,6 @@ func (a *Analyzer) Analyze() (files.Analyzed, error) {
 			atm, err = platform.GetTargetMetadata(a.RunImage)
 			if err != nil {
 				return files.Analyzed{}, errors.Wrap(err, "unpacking metadata from image")
-			}
-			if atm.OS == "" {
-				platform.GetTargetOSFromFileSystem(&fsutil.Detect{}, atm, a.Logger)
 			}
 		}
 	}
--- a/phase/analyzer.go
+++ b/phase/analyzer.go
@@ -86,6 +86,9 @@ func (a *Analyzer) Analyze() (files.Analyzed, error) {
 			if err != nil {
 				return files.Analyzed{}, errors.Wrap(err, "unpacking metadata from image")
 			}
+			if atm.OS == "" {
+				return files.Analyzed{}, errors.New("failed to find OS in run image config")
+			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


phase/analyzer.go
```diff
@@ -87,9 +86,6 @@ func (a *Analyzer) Analyze() (files.Analyzed, error) {
 			if err != nil {
 				return files.Analyzed{}, errors.Wrap(err, "unpacking metadata from image")
 			}
-			if atm.OS == "" {
-				platform.GetTargetOSFromFileSystem(&fsutil.Detect{}, atm, a.Logger)
-			}
```
natalieparellano, line 92: Unrelated bug. This only runs if os is blank (which shouldn't happen) but it will also populate distro information and I don't think we can assume this will be the same as the run image as `analyze` often runs in the context of the lifecycle image.

platform/target_data_test.go
```diff
@@ -21,34 +21,61 @@ func TestTargetData(t *testing.T) {
...
 					when("missing os and arch", func() {
 						it("matches", func() {
-							h.AssertEq(t, platform.TargetSatisfiedForBuild(baseTarget, buildpack.TargetMetadata{OS: "", Arch: ""}), true)
+							h.AssertEq(t, platform.TargetSatisfiedForBuild(&d, baseTarget, buildpack.TargetMetadata{OS: "", Arch: ""}, &log.Logger{Handler: memory.New()}), true)
 						})
 					})
 
 					when("has distro information", func() {
 						it("does not match", func() {
-							h.AssertEq(t, platform.TargetSatisfiedForBuild(baseTarget, buildpack.TargetMetadata{OS: baseTarget.OS, Arch: baseTarget.Arch, ArchVariant: "MMX"}), true)
```
natalieparellano, line 46: Removed extra test

natalieparellano:
@edmorley @pbusko FYI

platform/target_data.go
```diff
@@ -64,25 +40,14 @@ func GetTargetMetadata(fromImage imgutil.Image) (*files.TargetMetadata, error) {
...
-		}
-	}
-}
-
 // TargetSatisfiedForBuild treats empty fields as wildcards and returns true if all populated fields match.
-func TargetSatisfiedForBuild(base files.TargetMetadata, module buildpack.TargetMetadata) bool {
+func TargetSatisfiedForBuild(d fsutil.Detector, base files.TargetMetadata, module buildpack.TargetMetadata, logger log.Logger) bool {
+	// ensure we have all available data
+	if base.OS == "" || base.Distro == nil {
+		logger.Info("target distro name/version labels not found, reading /etc/os-release file")
```
edmorley, line 47: In what cases could `OS` be the empty string? If the author of the run image created it in such a way that the image didn't have properly specified metadata? The OCI spec suggests `OS` is REQUIRED:
https://github.com/opencontainers/image-spec/blob/main/config.md#properties

It's just at the moment, the log message (whose wording exists prior to this PR) isn't consistent with the conditional. ie: The message says "target distro name/version labels", but the conditional suggests it should say "OS metadata, or target distro name/version labels" etc. (But if `OS` should always be present, perhaps the `OS` check can be moved to an earlier assertion that errors instead?)

natalieparellano: Good point. I added an error in 609e287c

PR #1354, opened 2024-05-13
More fixes for target compat checking during detect

natalieparellano:
### Summary

- If a buildpack does not specify targets, but DOES specify stack with ID io.buildpacks.stacks.bionic, only translate that to a target if the buildpack is on a newer Buildpack API (where stacks are deprecated, and we expect targets) OR if the buildpack only ever declared one stack. When multiple stacks are declared, specifying only one target actually reduces the number of builds that can succeed. Older Buildpack APIs don't actually specify stack -> target translation so we should do the thing that will work in the maximum number of builds.
- If a buildpack fails to specify os/arch (but specifies distro) still check targets
- If the run image fails to specify os/arch (this should not happen actually as we will fail during analyze) still check targets
- Fix typo in buildpack descriptor struct ~so that we actually get stack information~ (we still see it because of the struct field name)
- If we get distro information from /etc/os-release, persist this information to later invocations so that the log message printed when errors are encountered will be accurate
- Don't override inner `i` in loop (this should not actually affect the outer loop but is confusing)

#### Release notes

The `detector`, when checking target compatibility, will still perform the check if buildpacks fail to declare os/arch information in targets

### DIFF
--- a/acceptance/testdata/detector/Dockerfile
+++ b/acceptance/testdata/detector/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:jammy
 
 ARG cnb_uid=1234
 ARG cnb_gid=1000
--- a/buildpack/bp_descriptor.go
+++ b/buildpack/bp_descriptor.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 
+	"github.com/buildpacks/lifecycle/api"
 	"github.com/buildpacks/lifecycle/internal/encoding"
 )
 
@@ -69,7 +70,11 @@ func ReadBpDescriptor(path string) (*BpDescriptor, error) {
 	if len(descriptor.Targets) == 0 {
 		for _, stack := range descriptor.Stacks {
 			if stack.ID == "io.buildpacks.stacks.bionic" {
-				descriptor.Targets = append(descriptor.Targets, TargetMetadata{OS: "linux", Arch: "amd64", Distros: []OSDistro{{Name: "ubuntu", Version: "18.04"}}})
+				if api.MustParse(descriptor.API()).AtLeast("0.10") || len(descriptor.Stacks) == 1 {
+					descriptor.Targets = append(descriptor.Targets, TargetMetadata{OS: "linux", Arch: "amd64", Distros: []OSDistro{{Name: "ubuntu", Version: "18.04"}}})
+				}
+			} else if stack.ID == "*" {
+				descriptor.Targets = append(descriptor.Targets, TargetMetadata{}) // matches any
 			}
 		}
 	}
--- a/buildpack/bp_descriptor.go
+++ b/buildpack/bp_descriptor.go
@@ -70,8 +70,6 @@ func ReadBpDescriptor(path string) (*BpDescriptor, error) {
 		for _, stack := range descriptor.Stacks {
 			if stack.ID == "io.buildpacks.stacks.bionic" {
 				descriptor.Targets = append(descriptor.Targets, TargetMetadata{OS: "linux", Arch: "amd64", Distros: []OSDistro{{Name: "ubuntu", Version: "18.04"}}})
-			} else if stack.ID == "*" {
-				descriptor.Targets = append(descriptor.Targets, TargetMetadata{}) // matches any
 			}
 		}
 	}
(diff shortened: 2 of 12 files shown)


buildpack/bp_descriptor.go
```diff
@@ -70,8 +70,6 @@ func ReadBpDescriptor(path string) (*BpDescriptor, error) {
 		for _, stack := range descriptor.Stacks {
 			if stack.ID == "io.buildpacks.stacks.bionic" {
```
robdimsdale, line 71: should we add `jammy` given it's the primary stack that paketo supports?

phase/detector.go
```diff
@@ -192,26 +180,29 @@ func (d *Detector) detectGroup(group buildpack.Group, done []buildpack.GroupElem
...
 				return nil, nil, err
 			}
 		}
+
+		// Check target compatibility.
 		if d.PlatformAPI.AtLeast("0.12") {
-			targetMatch := false
-			if isWildcard(d.AnalyzeMD.RunImageTarget()) || hasWildcard(descriptor.TargetsList()) {
+			var targetMatch bool
+			if len(descriptor.TargetsList()) == 0 {
```
robdimsdale, line 194: Is this for backwards-compatibility with buildpacks that don't have a `targets` field? Is this spec-compliant?

robdimsdale (approved):

phase/detector.go
natalieparellano: In reality, this is for tests. In practice, the lifecycle will infer at least an OS when targets are missing. Here is what the spec has to say about matching:

> If the targets list is empty, tools reading buildpack.toml will assume:

> os = "linux" and arch = <matches any> if ./bin/build is present
os = "windows" and arch = <matches any> if ./bin/build.bat or ./bin/build.exe are present
Metadata specified in [[targets]] is validated against the runtime and build-time base images.

> A buildpack target satisfies a base image target when os, arch, and variant match and at least one distribution in distros (if provided) matches

acceptance/testdata/detector/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:jammy
```
natalieparellano, line 1: This change causes https://github.com/buildpacks/lifecycle/blob/1a1de0850a2e8cd34bedccf9856886077dd702e4/acceptance/detector_test.go#L163 to start failing when 

```
[[stacks]]
id = "io.buildpacks.stacks.bionic"

[[stacks]]
id = "io.buildpacks.stacks.jammy"
```
 is added to acceptance/testdata/detector/container/cnb/buildpacks/simple_buildpack/simple_buildpack_version/buildpack.toml without corresponding changes in how we read the descriptor

buildpack/bp_descriptor.go
```diff
@@ -17,7 +18,7 @@ type BpDescriptor struct {
 	Order       Order            `toml:"order"`
 	WithRootDir string           `toml:"-"`
 	Targets     []TargetMetadata `toml:"targets"`
-	Stacks      []StackMetadata  `tome:"stacks"` // just for backwards compat so we can check if it's the bionic stack, which we translate to a target
+	Stacks      []StackMetadata  `toml:"stacks"` // just for backwards compat so we can check if it's the bionic stack, which we translate to a target
```
natalieparellano, line 21: Worth noting that this typo while unsightly didn't cause errors because the struct field name matches the expected TOML key name

phase/detector.go
```diff
@@ -144,26 +144,14 @@ func (d *Detector) detectOrder(order buildpack.Order, done, next []buildpack.Gro
 	return nil, nil, ErrFailedDetection
 }
 
-// isWildcard returns true IFF the Arch and OS are unspecified, meaning that the target arch/os are "any"
-func isWildcard(t files.TargetMetadata) bool {
```
natalieparellano, line 148: In practice, `isWildcard` would never return true because we check that we at least have an OS during analyze

phase/detector.go
```diff
@@ -144,26 +144,14 @@ func (d *Detector) detectOrder(order buildpack.Order, done, next []buildpack.Gro
...
-	return t.Arch == "" && t.OS == ""
-}
-
-func hasWildcard(ts []buildpack.TargetMetadata) bool {
```
natalieparellano, line 152: It's possible to have a target without OS/arch but with a distribution so this function never made much sense

buildpack/bp_descriptor.go
robdimsdale: As discussed - we will opt for a len == 1 && stack == bionic check, as bionic is encoded in the spec.

robdimsdale (approved):

jabrown85 (approved):

PR #1338, opened 2024-04-12
Refactor lifecycle/cmd/restorer to move logic out of main and into phase package

natalieparellano:
#### Release notes

---

### Related

Resolves #___

---

PR #1364, opened 2024-06-17
Ensure read access to the run image selected by extensions

pbusko:
### Summary

When an extension switches the run image to one listed in the `run.toml`, the read access check is not triggered, as it is done in the analyze phase. This change will iterate over the requested run image and it's mirrors, to select the first accessible one.

#### Release notes

---

### Related

Resolves #___

---

### DIFF
--- a/cmd/lifecycle/restorer.go
+++ b/cmd/lifecycle/restorer.go
@@ -50,6 +50,7 @@ func (r *restoreCmd) DefineFlags() {
 		cli.FlagBuildImage(&r.BuildImageRef)
 	}
 	cli.FlagAnalyzedPath(&r.AnalyzedPath)
+	cli.FlagRunPath(&r.RunPath)
 	cli.FlagCacheDir(&r.CacheDir)
 	cli.FlagCacheImage(&r.CacheImageRef)
 	cli.FlagGID(&r.GID)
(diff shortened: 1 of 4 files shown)


jabrown85 (approved):

loewenstein-sap:
@jabrown85 With your approval, what's missing to get it merged and released? Do we need @natalieparellano's approval in addition?

jabrown85:
@loewenstein-sap yeah - I think I'd like a second set of eyes on this. What comes to mind now is that this phase may need registry auth, which isn't typically a thing in the untrusted phase executions. `CNB_REGISTRY_AUTH` is set for exporter/analyzer/restorer but is not mounted for builder/detector. I'm not sure that value should be set but that also doesn't mean generator phase shouldn't check access for what should be assumed to be unauthenticated image access.

There are other registry related settings like `CNB_INSECURE_REGISTRIES`, `CNB_USE_LAYOUT`, `CNB_LAYOUT_DIR` to think about as well.

natalieparellano:
> What comes to mind now is that this phase may need registry auth, which isn't typically a thing in the untrusted phase executions.

This is true. Can we move this check to the restore phase?

loewenstein-sap:
@natalieparellano ^^

natalieparellano:
https://github.com/buildpacks/lifecycle/pull/1338 would have made this PR much easier to test :) unfortunately, that PR is blocked on me having some bandwidth to add units.

Apologies for being slow here, I've just returned from vacation. I'll take a look today.

cmd/lifecycle/restorer.go
```diff
@@ -50,6 +50,7 @@ func (r *restoreCmd) DefineFlags() {
 		cli.FlagBuildImage(&r.BuildImageRef)
 	}
 	cli.FlagAnalyzedPath(&r.AnalyzedPath)
+	cli.FlagRunPath(&r.RunPath)
```
natalieparellano, line 53: I am sorry to raise the specter of another Platform API version bump, but I think we need to gate this on Platform 0.14 also (an raise another spec PR).

natalieparellano (approved):

natalieparellano:
I may have been too hasty in merging this one and https://github.com/buildpacks/lifecycle/pull/1346, as we're due for a go version / dependencies patch and this should be in the next minor...

PR #1399, opened 2024-09-11
Adjust exporter App Layer log output

jabrown85:
### Summary
The 1/N app layer logs are emitted _after_ the layers have been added. This can sometimes be an expensive operation depending on the layer size and export target.

When trying to understand performance during builds, this wording implies the previous log line before the app layers was responsible for the time between the log lines. Adjusting the wording will make it a bit clearer that the time between the two log lines includes both steps.

Before:
```bash
03:50:00 Adding layer 'buildpacksio/lifecycle:launch.sbom'
03:51:00 Adding 5/5 app layer(s)
````

After:
```bash
03:50:00 Adding layer 'buildpacksio/lifecycle:launch.sbom'
03:51:00 Added 5/5 app layer(s)
````

#### Release notes

Adjust exporter App Layer log output

---

### Context

If we are more interested in going the other direction, where we keep "Adding 1/N app layers" and emitting multiple times or something I'm happy to iterate.

### DIFF
--- a/phase/exporter.go
+++ b/phase/exporter.go
@@ -446,6 +446,8 @@ func (e *Exporter) addLauncherLayers(opts ExportOptions, buildMD *files.BuildMet
 
 func (e *Exporter) addAppLayers(opts ExportOptions, slices []layers.Slice, meta *files.LayersMetadata) error {
 	// creating app layers (slices + app dir)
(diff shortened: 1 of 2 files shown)


edmorley (approved):
Good spot - thank you!

natalieparellano (approved):

PR #1401, opened 2024-09-13
Update README for 0.20.x

natalieparellano:
#### Release notes

---

### Related

Resolves #___

---


edmorley (approved):
Good spot! 

Out of curiosity, do you know when we might want the next round of platform/buildpack API deprecations to take place? (Similar to https://github.com/buildpacks/rfcs/blob/main/text/0110-deprecate-apis.md)

natalieparellano:
> do you know when we might want the next round of platform/buildpack API deprecations to take place

Oh gosh, I do not know. I suppose it would start with an RFC similar to https://github.com/buildpacks/rfcs/blob/main/text/0110-deprecate-apis.md

Sadly, I think some folks have still not upgraded from the pre-0.7 APIs 😭 Maybe we should do a survey?

On the platform side, it would be nice to remove at least 0.7 and 0.8. This gets us over the breaking change released in 0.9: "Legacy BOM information is removed from the `io.buildpacks.build.metadata` label". The next breaking change is in 0.12 (stack removal) but I'd say we're probably not quite there yet. WDYT?

On the buildpack side, we could aspire to remove 0.7 and 0.8. This gets us over shell removal, released in 0.9, but I imagine that might be a heavy lift for some folks and we'd want a long window with deprecation warnings, etc. https://github.com/buildpacks/profile will help, but we'll probably want to implement https://github.com/buildpacks/rfcs/blob/main/text/0101-system-buildpacks-in-builder.md to make it easier for folks to include this helper.

We could at least start the RFC to get the conversation going. Thoughts here are very welcome :)

natalieparellano:
@edmorley I like that - maybe something (perhaps an info message, less noisy than a warning) that's like "FYI - your buildpack version is superseded. Check out https://buildpacks.io/docs/for-buildpack-authors/how-to/migrate/"

For posterity, I remembered that https://github.com/buildpacks/profile is needed in order for us to remove shell support in the launcher. It isn't directly tied to buildpacks upgrading or not upgrading.

PR #1414, opened 2024-10-28
Bump docker to silence non-impactful CVE 

natalieparellano:
(https://github.com/moby/moby/security/advisories/GHSA-v23v-6jw2-98fq)

#### Release notes

---

### Related

Resolves #___

---


jabrown85 (approved):

natalieparellano:
Ugh this may be blocked on upstream dependencies getting updated:
* We need docker at least v27.1.0, which requires buildkit at least v0.15.0
* This is incompatible with the latest version of kaniko, which uses buildkit v0.14.1

Maybe we can contribute the dependency bump in kaniko. For now, I'll put this PR in draft, and add an entry to our .grype.yaml file to ignore the CVE during scans since it is non-impactful

Issue #1171, opened 2023-08-01
Modularize Makefile


dlion:
### Description
As mentioned here https://github.com/buildpacks/lifecycle/pull/1169#issuecomment-1658732291 our Makefile is getting bigger and bigger, we need to refactor and modularize it.
It will reduce duplicate code, improve the overall maintenance and readability.

### Proposed solution
Use different `.mk` modules and include/use them inside the Makefile as already done here: https://github.com/buildpacks/lifecycle/pull/1169

PR #1438, opened 2024-12-08
Makefile: Create dynamic build targets

gogolok:
### Summary
Use a Makefile function to create dynamic build targets so it's simple to add new GOOS/GOARCH combinations if required. The duplication of Makefile code is reduced.

### Related

#1171 

### Context

My 'overall goal' is still to get some FreeBSD support as attempted in #1271 and #1087 .

I thought it might make sense to help out with #1171 and come up with a simpler Makefile so that other GOOS/GOARCH can be added more easily. I'm not a Makefile expert.

I've left windows out since it might get removed soon.
I've left out the darwin build targets for the time being.

Other Makefile targets like the package targets can be implemented in a similar way once this attempt is accepted of being worth it.

I wonder (and this contradicts this attempt in this PR) whether the Makefile should just work with GOOS and GOARCH set outside the Makefile and the .github workflows to create releases should actually have several `make build` (or make package or ...) runs to create individual binaries/packages/whatsoever.

For this PR and the attempt at making the Makefile more dynamic I see the following advantages and disadvantages:

#### Advantages
- Code duplication is reduced.
- No separate *.mk files.

#### Disadvantages
- It might be harder to understand the new Makefile logic.
- It seems like there is no 'Makefile feature' to review the 'first pass' of the Makefile. It would be nice to see the new targets the new function actually generates. At least I'm not aware of such a Makefile feature.
- If you want to build binaries for currently unsupported GOOS/GOARCH, you would need to patch the Makefile (FreeBSD case for example at the moment).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,12 +38,9 @@ GOFILES := $(shell $(GOCMD) run tools$/lister$/main.go)
 
 all: test build package
 
-build: build-linux-amd64 build-linux-arm64 build-linux-ppc64le build-linux-s390x
+GOOS_ARCHS = linux/amd64 linux/arm64 linux/ppc64le linux/s390x
 
-build-linux-amd64: build-linux-amd64-lifecycle build-linux-amd64-symlinks build-linux-amd64-launcher
-build-linux-arm64: build-linux-arm64-lifecycle build-linux-arm64-symlinks build-linux-arm64-launcher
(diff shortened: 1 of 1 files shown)


jabrown85:
Nice - I'm not sure which approach I prefer either. I think I'm fine with either approach having only seen the diff for the dynamic one here.

This PR will have to wait until we've release a last windows lifecycle release in the next week or two.

jabrown85 (approved):

PR #1442, opened 2024-12-10
Bump go-dependencies containerd, golang.org/x/sync and golang.org/x/sys

gogolok:
### Summary
Picked version updates from #1440 that are working.

### Context

It looks like the go-dependencies buildkit, docker are blocked by GoogleContainerTools/kaniko, therefore I've picked the versions that can be upgraded right away.


jabrown85 (approved):

jabrown85 (approved):
re-approving after DCO fix

PR #1473, opened 2025-02-28
go 1.24

jabrown85:
### Summary

Update to go 1.24

#### Release notes

Update to go 1.24

### Context

I think this could delay the go toolchain weirdness we've seen elsewhere. When I run the build on 1.24 I don't get the toolchain directive put into go.mod. If I run on the latest 1.23 version, I do. That toolchain directive breaks our "do nothing to update go patch versions" flow.


natalieparellano:
Need to update the linter most likely

jabrown85:
@natalieparellano 👀 if you get a chance

PR #1525, opened 2025-08-11
Send stdout and stderr to the same stream (build and generate)

jabrown85:
### Summary

To ensure ordering of stdout and stderr in build and generate phases, send the stdout and stderr of the executing command to the same stream (stdout).

This will have an outside effect of anyone consuming `lifecycle` as what was previously in stderr is now in stdout.

#### Release notes

- `lifecycle/builder` and `lifecycle/generate` will no longer send buildpack stderr to stderr, but instead stdout to preserve ordering.

---

### Related

Resolves https://github.com/buildpacks/pack/issues/2330

---

### Context

I looked into and tried a few other options, but ultimately didn't find a path that allowed for us to keep the streams separate and also guarantee ordering. I believe ordering of buildpack produced logs to be more important to more users. We already use a buffer for `detector` but for different reasons (hidden by default).

### DIFF
--- a/buildpack/build.go
+++ b/buildpack/build.go
@@ -132,7 +132,7 @@ func runBuildCmd(d BpDescriptor, bpLayersDir, planPath string, inputs BuildInput
 	) // #nosec G204
 	cmd.Dir = inputs.AppDir
 	cmd.Stdout = inputs.Out
-	cmd.Stderr = inputs.Err
+	cmd.Stderr = inputs.Out
 
 	var err error
 	if d.Buildpack.ClearEnv {
--- a/buildpack/generate.go
+++ b/buildpack/generate.go
@@ -88,7 +88,7 @@ func runGenerateCmd(d ExtDescriptor, extOutputDir, planPath string, inputs Gener
 	) // #nosec G204
 	cmd.Dir = inputs.AppDir
 	cmd.Stdout = inputs.Out
-	cmd.Stderr = inputs.Err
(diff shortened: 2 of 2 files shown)


jkutner (approved):

edmorley:
Thank you for this!

I agree that (a) combining both output streams makes the most sense to solve this, (b) picking stdout (rather than stderr) for the combined output is best.

(Internally we've previously had some discussion about making the buildpack logging utility helpers all default to stderr to work around this, but I think the unix "stderr is for humans" best practice really only applies when the output from stdout is expected to be machine consumable/easily piped, and that doesn't really apply to plain logs. In the future if `pack build` gains a `--format json` or some other pipeable output, then we can always have that mode send progress bars/warnings/any other output to stderr and leave the main `pack build` mode untouched.)

schneems:
Seems good to standardize. Thanks! We had some buildpacks with output to stdout/stderr indicating different things (like they used stderr to indicate only warnings and wanted to assert no warnings). Merging them to one stream provides more consistency. The `pack build ... > log.txt` is a prime use case and I agree it overrules the traditional stdout/stderr split.

I have a vauge desire that Ed hinted at to introduce some more meaningful machine-usable outputs from pack. Like being able to retrieve the input image name or other metadata about the build itself. I don't have a strong need today, but imagine something like https://github.com/rubygems/rubygems/pull/8728 for pack to reduce/remove the need of people to grep unstructured data would be nice (once we have some strong use-cases).

PR #1654, opened 2026-05-27
Bump github.com/google/go-containerregistry to v0.21.6

jabrown85:
go-containerregistry v0.21.6 introduced an `isImageConfig` guard in `mutate.Image.compute()` that skips config re-serialization when the manifest's Config.MediaType is not a recognized image config type. The extender tests were using a bare `&v1.Manifest{Layers: ...}` with an empty Config.MediaType, causing `mutate.ConfigFile` to no longer embed the config digest into the manifest — producing a single identical image digest regardless of config content.

Fix the tests by setting `Config.MediaType: types.DockerConfigJSON` on the fake manifest so that config changes are properly reflected in the computed image digest, and update the expected SHA values accordingly.


AidanDelaney (approved):
LGTM. However, I can't see why the upstream introduced this change?

jjbustamante (approved):
LGTM

hone (approved):

PR #1661, opened 2026-06-01
ci: disable failing s390x tests

jabrown85:
ZVSI provisioning fails with `forbidden: user does not have access to selected profile`. Commenting out the workflow until IBM Cloud access is restored.

Ref: https://github.com/buildpacks/lifecycle/actions/runs/26774897508/job/78923792930


hone (approved):
I'm happy to unblock this for now, but we should decide if we're going to fix the s390x tests or remove them all together in the near future.

PR #1667, opened 2026-06-08
Pin go-containerregistry to v0.21.5

### LINKED ISSUE #1662: v0.21.11: incorrect diffIDs in image config due to go-containerregistry v0.21.6 isImageConfig guard causing CannotPullContainerError on containerd
### Summary
  `v0.21.11` produces images with incorrect `diffID` values in the image config,
  causing ECS/containerd to reject the image on pull with `wrong diff id ... calculated on extraction`.

  The issue was introduced by PR #1654 (bump `go-containerregistry` to `v0.21.6`).
  That library version added an `isImageConfig` guard in `mutate.Image.compute()` that skips
  config re-serialization when `Config.MediaType` is not a recognized image config type.
  When the lifecycle exports layers where this condition is triggered, the diffIDs in the
  image config end up equal to the compressed layer digest instead of the uncompressed hash
  which is invalid and rejected by containerd at runtime.

  ---

  ### Reproduction

  ##### Steps
  
  1. Build a Spring Boot application using `spring-boot:build-image` with builder
     `paketobuildpacks/builder-jammy-base` or `paketobuildpacks/builder-noble-java-tiny`
     on a version that ships lifecycle `v0.21.11`
  2. Push the resulting image to a registry (e.g. Amazon ECR)
  3. Deploy the image to ECS (or any containerd-based runtime)

  ##### Current behavior 
  ECS fails to start the container with:
  CannotPullContainerError: ref pull has been retried 1 time(s): wrong diff id
  "sha256:a3aa99b46093c48a7b86403deb152a45891bad3a0039f066d47d68d06aaa7206" 
  calculated on extraction
  "sha256:8bba68e7621928237aa6d6e2c680cb9572c942ee23c2adabd22016bb67cb938d",
  desc "sha256:a3aa99b46093c48a7b86403deb152a45891bad3a0039f066d47d68d06aaa7206"

  Note that `desc` (compressed digest) equals the expected `diffID` confirming that the
  uncompressed hash was never correctly written into the image config.

  Running `docker image inspect <image> --format '{{json .RootFS.Layers}}'` locally
  confirms the affected layer's diffID matches its compressed digest.

  ##### Expected behavior
  The image config contains correct diffIDs (uncompressed layer hashes), and the image
  starts successfully on containerd-based runtimes.
  
  ---

  ### Context

  `v0.21.11` (broken) -  `v0.21.10` not tested, `v0.21.9` confirmed working.
  
  ##### platform version(s)
  - Builder: `paketobuildpacks/builder-jammy-base:latest` (sha256:c0bb1adc...)
  - Docker Engine: `29.4.1`
  - containerd: `v2.2.3`
  - Runtime: Amazon ECS
  
  ##### anything else?
  The fix in PR #1654 only updated the extender tests to set `Config.MediaType: types.DockerConfigJSON`,
  but the production code path that exports layers for real container images appears to hit the same
  `isImageConfig` guard under certain conditions leaving diffIDs incorrect in the final image manifest.
  
  Workaround: pin the builder to a version that ships lifecycle `v0.21.9`,
  e.g. `paketobuildpacks/builder-noble-java-tiny:0.0.140`.

jabrown85:
### Summary

go-containerregistry v0.21.6 introduced an isImageConfig guard in mutate.Image.compute() that only updates the image config's RootFS.DiffIDs when the config media type is a recognized image config type (DockerConfigJSON/OCIConfigJSON). When the working image presents an empty/unrecognized config media type, appended layers are written to manifest.Layers but their diff IDs are omitted from the config, so the exported image fails to pull with 'wrong diff id ... calculated on extraction' on containerd/ECS.

This reverts the functional change from #1654 (the ggcr v0.21.5 -> v0.21.6 bump) and pins ggcr at v0.21.5, where the RootFS.DiffIDs update is unconditional. Also adds a Dependabot ignore for go-containerregistry so the bump is not reintroduced before the upstream regression is resolved.

#### Release notes

- Revert `ggcr` bump to fix reported regression #1662 

---

### Related

Resolves #1662

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,13 @@ updates:
       update-types:
         - "minor"
         - "patch"
+  ignore:
+    # Pinned to v0.21.5: v0.21.6 introduced an isImageConfig guard in
+    # mutate.Image.compute() that omits RootFS.DiffIDs from the image config
+    # when the config media type is unrecognized, producing images that fail
+    # to pull with "wrong diff id". See buildpacks/lifecycle#1662.
+    # Do not bump until the upstream regression is resolved.
+    - dependency-name: "github.com/google/go-containerregistry"
 - package-ecosystem: "github-actions"
   directory: "/"
   schedule:
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/docker/docker v28.5.2+incompatible
 	github.com/golang/mock v1.6.0
 	github.com/google/go-cmp v0.7.0
-	github.com/google/go-containerregistry v0.21.6
+	github.com/google/go-containerregistry v0.21.5
 	github.com/google/uuid v1.6.0
 	github.com/heroku/color v0.0.6
 	github.com/moby/buildkit v0.30.0
@@ -103,6 +103,7 @@ require (
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
 	github.com/containerd/platforms v1.0.0-rc.4 // indirect
+	github.com/containerd/stargz-snapshotter/estargz v0.18.2 // indirect
 	github.com/containerd/typeurl/v2 v2.2.3 // indirect
 	github.com/curioswitch/go-reassign v0.3.0 // indirect
 	github.com/daixiang0/gci v0.13.7 // indirect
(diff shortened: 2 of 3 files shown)


jabrown85:
I reported the issue [here](https://github.com/google/go-containerregistry/issues/2339) as well - but will continue hunting for why on our end (imgutil/lifecycle handling)

hone (approved):

PR #1675, opened 2026-06-17
Add retry logic for retrieving top layer SHA with registry mirrors

### LINKED ISSUE #1674: Lifecycle export fails intermittently with "get run image top layer SHA: image has no layers" when using a registry mirror
### Summary

`lifecycle` export phase fails intermittently with `get run image top layer SHA: image has no layers` when using a corporate Docker registry mirror. The failure is transient – retrying the same build succeeds.

Originally reported in paketo-buildpacks/spring-boot#605, where it was suggested that this is a lifecycle-level issue rather than specific to the Spring Boot buildpack.

---

##### Steps

1. Use a Spring Boot project with Gradle 9.4.1 and Java 25.
2. Configure the build to use `paketobuildpacks/builder-jammy-full:latest` as the builder.
3. Set environment variables `BP_DOCKER_REGISTRY_MIRROR` and `DOCKER_REGISTRY_MIRROR` to point to a corporate mirror (internal).
4. Run `./gradlew bootBuildImage` multiple times (e.g., in a CI pipeline). The failure appears sporadically (roughly 1 in 5–10 builds).

##### Current behavior

During the `EXPORTING` phase, the build fails with:

```
ERROR: failed to export: get run image top layer SHA: image has no layers
```

The run image (`paketobuildpacks/run-jammy-full:latest`) exists and is accessible (pulled successfully in previous steps), but the export step cannot obtain its layer information. The issue seems to be that the mirror sometimes returns an incomplete manifest or fails to serve the platform‑specific image (`linux/amd64`) correctly. The lifecycle does not retry or handle this transient registry error gracefully.

##### Expected behavior

The export step should either:
- Succeed consistently if the run image is available, or
- Retry fetching the run image layers (with exponential backoff) when a transient registry error occurs, instead of failing the build immediately.

---

##### lifecycle version

v0.21.8 (as shown in the build logs)

##### platform version(s)

- Gradle 9.4.1
- Java 25
- Builder image: `paketobuildpacks/builder-jammy-full:latest`
- Run image: `paketobuildpacks/run-jammy-full:latest`
- Registry mirror: corporate mirror (internal, accessed via `BP_DOCKER_REGISTRY_MIRROR` and `DOCKER_REGISTRY_MIRROR`)
- Platform: `linux/amd64`

##### anything else?

- The issue occurs only when using a registry mirror. When pulling directly from Docker Hub, the build always succeeds.
- The same error can be reproduced with other Paketo builders (e.g., `paketobuildpacks/builder-jammy-base`).
- Buildpacks involved (from the log, though likely irrelevant):
  - `paketo-buildpacks/ca-certificates` 3.12.2
  - `paketo-buildpacks/bellsoft-liberica` 11.7.1
  - `paketo-buildpacks/syft` 2.32.0
  - `paketo-buildpacks/executable-jar` 6.15.2
  - `paketo-buildpacks/dist-zip` 5.12.2
  - `paketo-buildpacks/spring-boot` 5.36.2
- Full build log can be provided upon request (it contains internal data that would need sanitizing).

Original discussion: https://github.com/paketo-buildpacks/spring-boot/issues/605

vpelikh:
### Summary

This PR adds retry logic to mitigate intermittent failures when retrieving the top layer SHA from registry mirrors. Registry mirrors may temporarily return incomplete manifests as layers sync, causing the `image has no layers` error. The retry mechanism gives the mirror time to stabilize.

#### Release notes

When using registry mirrors, the lifecycle will now retry top layer retrieval up to 5 times with exponential backoff before failing.

---

### Related

Resolves #1674

---

### Context

- Added `OpenRemoteImage` function in `phase/retry.go` with 5 retry attempts
- Exponential backoff delays: 100ms → 200ms → 500ms → 1s → 2s
- Exporter and Rebaser now use `OpenRemoteImage` instead of direct `TopLayer()` calls
- All existing tests pass

### DIFF
--- a/cmd/lifecycle/exporter.go
+++ b/cmd/lifecycle/exporter.go
@@ -217,18 +217,18 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore phase.Cache, analyz
 
 	g.Go(func() error {
 		report, err := exporter.Export(phase.ExportOptions{
-			AdditionalNames:     e.AdditionalTags,
-			AppDir:              e.AppDir,
-			DefaultProcessType:  e.DefaultProcessType,
-			ExecEnv:             e.ExecEnv,
-			ExtendedDir:         e.ExtendedDir,
-			LauncherConfig:      launcherConfig(e.LauncherPath, e.LauncherSBOMDir),
-			LayersDir:           e.LayersDir,
-			OrigMetadata:        analyzedMD.LayersMetadata,
-			Project:             projectMD,
-			RunImageRef:         runImageID,
-			RunImageForExport:   runImageForExport,
-			WorkingImage:        appImage,
+			AdditionalNames:    e.AdditionalTags,
+			AppDir:             e.AppDir,
+			DefaultProcessType: e.DefaultProcessType,
+			ExecEnv:            e.ExecEnv,
+			ExtendedDir:        e.ExtendedDir,
+			LauncherConfig:     launcherConfig(e.LauncherPath, e.LauncherSBOMDir),
+			LayersDir:          e.LayersDir,
+			OrigMetadata:       analyzedMD.LayersMetadata,
+			Project:            projectMD,
+			RunImageRef:        runImageID,
+			RunImageForExport:  runImageForExport,
+			WorkingImage:       appImage,
 		})
 		if err != nil {
 			return cmd.FailErrCode(err, e.CodeFor(platform.ExportError), "export")
@@ -345,7 +345,11 @@ func (e *exportCmd) initRemoteAppImage(analyzedMD files.Analyzed) (imgutil.Image
 	}
 
 	appImage, err := phase.OpenRemoteImage(cmd.DefaultLogger, func() (imgutil.Image, error) {
-		return remote.NewImage(e.OutputImageRef, e.keychain, appOpts...)
+		return remote.NewImage(
+			e.OutputImageRef,
+			e.keychain,
+			appOpts...,
+		)
 	})
 	if err != nil {
 		return nil, "", cmd.FailErr(err, "create new app image")
@@ -372,6 +376,7 @@ func (e *exportCmd) initRemoteAppImage(analyzedMD files.Analyzed) (imgutil.Image
 	if err != nil {
 		return nil, "", cmd.FailErr(err, "get run image ID")
 	}
+
 	return appImage, runImageID, nil
 }
 
--- a/cmd/lifecycle/exporter.go
+++ b/cmd/lifecycle/exporter.go
@@ -195,8 +195,9 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore phase.Cache, analyz
 	}
 
 	var (
-		appImage   imgutil.Image
-		runImageID string
+		appImage            imgutil.Image
+		runImageID          string
+		workingImageFactory func() (imgutil.Image, error)
 	)
 	switch {
 	case e.UseLayout:
@@ -204,7 +205,7 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore phase.Cache, analyz
 	case e.UseDaemon:
 		appImage, runImageID, err = e.initDaemonAppImage(analyzedMD, cmd.DefaultLogger)
 	default:
-		appImage, runImageID, err = e.initRemoteAppImage(analyzedMD)
+		appImage, workingImageFactory, runImageID, err = e.initRemoteAppImage(analyzedMD)
 	}
 	if err != nil {
 		return err
@@ -217,18 +218,19 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore phase.Cache, analyz
 
 	g.Go(func() error {
 		report, err := exporter.Export(phase.ExportOptions{
-			AdditionalNames:    e.AdditionalTags,
-			AppDir:             e.AppDir,
-			DefaultProcessType: e.DefaultProcessType,
-			ExecEnv:            e.ExecEnv,
-			ExtendedDir:        e.ExtendedDir,
-			LauncherConfig:     launcherConfig(e.LauncherPath, e.LauncherSBOMDir),
-			LayersDir:          e.LayersDir,
-			OrigMetadata:       analyzedMD.LayersMetadata,
-			Project:            projectMD,
-			RunImageRef:        runImageID,
-			RunImageForExport:  runImageForExport,
-			WorkingImage:       appImage,
+			AdditionalNames:     e.AdditionalTags,
+			AppDir:              e.AppDir,
+			DefaultProcessType:  e.DefaultProcessType,
+			ExecEnv:             e.ExecEnv,
+			ExtendedDir:         e.ExtendedDir,
+			LauncherConfig:      launcherConfig(e.LauncherPath, e.LauncherSBOMDir),
+			LayersDir:           e.LayersDir,
+			OrigMetadata:        analyzedMD.LayersMetadata,
+			Project:             projectMD,
+			RunImageRef:         runImageID,
+			RunImageForExport:   runImageForExport,
+			WorkingImage:        appImage,
+			WorkingImageFactory: workingImageFactory,
 		})
 		if err != nil {
 			return cmd.FailErrCode(err, e.CodeFor(platform.ExportError), "export")
@@ -313,7 +315,7 @@ func (e *exportCmd) initDaemonAppImage(analyzedMD files.Analyzed, logger log.Log
 	return appImage, runImageID.String(), nil
 }
 
-func (e *exportCmd) initRemoteAppImage(analyzedMD files.Analyzed) (imgutil.Image, string, error) {
+func (e *exportCmd) initRemoteAppImage(analyzedMD files.Analyzed) (imgutil.Image, func() (imgutil.Image, error), string, error) {
 	var appOpts = []imgutil.ImageOption{
 		remote.FromBaseImage(e.RunImageRef),
 	}
@@ -321,7 +323,7 @@ func (e *exportCmd) initRemoteAppImage(analyzedMD files.Analyzed) (imgutil.Image
 	if e.supportsRunImageExtension() {
 		extendedConfig, err := e.getExtendedConfig(analyzedMD.RunImage)
 		if err != nil {
-			return nil, "", cmd.FailErr(err, "get extended image config")
+			return nil, nil, "", cmd.FailErr(err, "get extended image config")
 		}
 		if extendedConfig != nil {
 			cmd.DefaultLogger.Debugf("Using config from extensions...")
@@ -344,13 +346,16 @@ func (e *exportCmd) initRemoteAppImage(analyzedMD files.Analyzed) (imgutil.Image
 		appOpts = append(appOpts, remote.WithCreatedAt(e.customSourceDateEpoch()))
 	}
 
-	appImage, err := remote.NewImage(
-		e.OutputImageRef,
-		e.keychain,
-		appOpts...,
-	)
+	workingImageFactory := func() (imgutil.Image, error) {
+		return remote.NewImage(
+			e.OutputImageRef,
+			e.keychain,
+			appOpts...,
+		)
+	}
+	appImage, err := workingImageFactory()
 	if err != nil {
-		return nil, "", cmd.FailErr(err, "create new app image")
+		return nil, nil, "", cmd.FailErr(err, "create new app image")
 	}
 
 	runImageID, err := func() (string, error) {
@@ -372,10 +377,10 @@ func (e *exportCmd) initRemoteAppImage(analyzedMD files.Analyzed) (imgutil.Image
 		return runImageID.String(), nil
 	}()
 	if err != nil {
-		return nil, "", cmd.FailErr(err, "get run image ID")
+		return nil, nil, "", cmd.FailErr(err, "get run image ID")
 	}
 
-	return appImage, runImageID, nil
+	return appImage, workingImageFactory, runImageID, nil
 }
 
 func (e *exportCmd) initLayoutAppImage(analyzedMD files.Analyzed) (imgutil.Image, string, error) {
--- a/cmd/lifecycle/exporter.go
+++ b/cmd/lifecycle/exporter.go
@@ -195,9 +195,8 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore phase.Cache, analyz
 	}
 
 	var (
-		appImage            imgutil.Image
-		runImageID          string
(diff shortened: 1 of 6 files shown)


phase/exporter.go
```diff
@@ -111,10 +112,11 @@ func (e *Exporter) Export(opts ExportOptions) (files.Report, error) {
 	}
 
 	meta := files.LayersMetadata{}
-	meta.RunImage.TopLayer, err = opts.WorkingImage.TopLayer()
+	topLayer, err := TopLayerWithRetry(opts.WorkingImage, e.Logger)
```
jabrown85: I don't believe this will do what you want. I am fairly confident that go-containerregistry will end up serving the same manifest over and over ([src](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/image.go#L81-L83)). The working image would need to be re-initialized upon failure.

phase/exporter.go
```diff
@@ -672,3 +674,39 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+}
+
+// TopLayerWithRetry calls TopLayer() on an imgutil.Image with retry logic.
+// This handles transient registry errors when using registry mirrors.
+// The backoff delays are chosen to be conservative - registry mirrors may need time to sync.
+func TopLayerWithRetry(img imgutil.Image, logger log.Logger) (string, error) {
+	var lastErr error
+	for attempt := range topLayerMaxRetries {
+		topLayer, err := img.TopLayer()
+		if err == nil {
```
jabrown85: This is too broad of an error case check. Hard errors like 401/403 should not burn time retrying.

phase/exporter.go
```diff
@@ -672,3 +674,39 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+			if attempt > 0 {
+				logger.Debugf("Successfully retrieved top layer SHA after %d retries", attempt)
+			}
+			return topLayer, nil
+		}
+		lastErr = err
+
+		// Log the error and wait before retrying
+		if attempt < topLayerMaxRetries-1 {
```
jabrown85: I believe there is a bug here and we won't sleep/retry the 4th time correctly

phase/exporter.go
```diff
@@ -672,3 +674,39 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+	1 * time.Second,
+	2 * time.Second,
+}
+
+// TopLayerWithRetry calls TopLayer() on an imgutil.Image with retry logic.
+// This handles transient registry errors when using registry mirrors.
+// The backoff delays are chosen to be conservative - registry mirrors may need time to sync.
+func TopLayerWithRetry(img imgutil.Image, logger log.Logger) (string, error) {
+	var lastErr error
+	for attempt := range topLayerMaxRetries {
```
jabrown85: I think ranging over topLayerDelays could be easier to follow

phase/exporter.go
```diff
@@ -672,3 +674,39 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+				logger.Debugf("Successfully retrieved top layer SHA after %d retries", attempt)
+			}
+			return topLayer, nil
+		}
+		lastErr = err
+
+		// Log the error and wait before retrying
+		if attempt < topLayerMaxRetries-1 {
+			duration := topLayerDelays[attempt]
+			logger.Debugf("Attempt %d/%d: Failed to get top layer SHA: %v. Retrying in %v...", attempt+1, topLayerMaxRetries, err, duration)
```
jabrown85: We could consider using warn here so it is visible to the end users?

phase/exporter.go
```diff
@@ -672,3 +674,39 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+	200 * time.Millisecond,
+	500 * time.Millisecond,
+	1 * time.Second,
+	2 * time.Second,
+}
+
+// TopLayerWithRetry calls TopLayer() on an imgutil.Image with retry logic.
+// This handles transient registry errors when using registry mirrors.
+// The backoff delays are chosen to be conservative - registry mirrors may need time to sync.
+func TopLayerWithRetry(img imgutil.Image, logger log.Logger) (string, error) {
```
jabrown85: This method needs testing.

vpelikh: @jabrown85, I completely agree - this method definitely needs tests.

Since I'm not very experienced with Go, I'd appreciate some guidance on how to approach this properly. Could you please point me to:

1. Existing test examples in the repo (e.g., for exporter or detector) that mock imgutil.Image?
2. The recommended way to mock TopLayer() to simulate different error scenarios (both transient and permanent)?
3. Any helpers or patterns for testing retry logic (backoff, timeouts, etc.)?

Thanks in advance!

jabrown85:
```
  // stubImage embeds imgutil.Image so it satisfies the 14-method interface with a
  // nil base; only TopLayer is implemented. Each call pops the next scripted
  // result, so we can model "fail, fail, then succeed".
  type stubImage struct {
        imgutil.Image // nil — any other method call would panic, which is fine: we only call TopLayer
        results        []topLayerResult
        calls          int
  }

  type topLayerResult struct {
        sha string
        err error
  }

  func (s *stubImage) TopLayer() (string, error) {
        r := s.results[s.calls]
        s.calls++
        return r.sha, r.err
  }

func swapRetryTiming(t *testing.T, n int) (sleeps *[]time.Duration) {
        t.Helper()
        var recorded []time.Duration

        origDelays := topLayerDelays
        origSleep := topLayerSleep

        // n attempts, tiny non-zero delays we never actually wait on.
        topLayerDelays = make([]time.Duration, n)
        for i := range topLayerDelays {
                topLayerDelays[i] = time.Duration(i+1) * time.Millisecond
        }
        topLayerSleep = func(d time.Duration) { recorded = append(recorded, d) } // record, don't sleep

        t.Cleanup(func() {
                topLayerDelays = origDelays
                topLayerSleep = origSleep
        })
        return &recorded
  }
```
Something like this will give you the ability to write tests like 

```
        t.Run("succeeds after transient failures", func(t *testing.T) {
                sleeps := swapRetryTiming(t, 4)
                img := &stubImage{results: []topLayerResult{
                        {err: errNoLayers},
                        {err: errNoLayers},
                        {sha: "sha-1"},
                }}

                got, err := TopLayerWithRetry(img, logger)

                h.AssertNil(t, err)
                h.AssertEq(t, got, "sha-1")
                h.AssertEq(t, img.calls, 3)
                h.AssertEq(t, len(*sleeps), 2) // slept before retry 1 and retry 2
        })
```

jabrown85: `var topLayerSleep = time.Sleep` is what I would define to use in place of `time.Sleep` so you can override it in the tests like I did above

phase/exporter.go
vpelikh: Should be fixed now. Check it please.

phase/exporter.go
vpelikh: Added tests in retry_test.go. Check it please.

phase/exporter.go
vpelikh: Done.

phase/exporter.go
vpelikh: Nice suggestion. Fixed.

phase/exporter.go
vpelikh: Should be fixed.

phase/exporter.go
vpelikh: Fixed.

vpelikh (commented):
@jabrown85 , thank you for the thorough review! I've gone through all your comments and pushed the necessary fixes. The PR should be ready for another look when you have a moment.

Really appreciate your time and attention to detail!

phase/exporter.go
```diff
@@ -63,6 +67,7 @@ type LauncherConfig struct {
 	Metadata files.LauncherMetadata
 }
 
+// ExportOptions is the set of options for exporting an image.
```
vpelikh, line 66: @jabrown85, linting fail without comment for ExportOptions and Rebaser. Should I keep it?

cmd/lifecycle/exporter.go
```diff
@@ -217,18 +218,19 @@ func (e *exportCmd) export(group buildpack.Group, cacheStore phase.Cache, analyz
...
+			ExecEnv:             e.ExecEnv,
+			ExtendedDir:         e.ExtendedDir,
+			LauncherConfig:      launcherConfig(e.LauncherPath, e.LauncherSBOMDir),
+			LayersDir:           e.LayersDir,
+			OrigMetadata:        analyzedMD.LayersMetadata,
+			Project:             projectMD,
+			RunImageRef:         runImageID,
+			RunImageForExport:   runImageForExport,
+			WorkingImage:        appImage,
+			WorkingImageFactory: workingImageFactory,
```
jabrown85: This is getting more involved than I had hoped 😅 

I don't love the `WorkingImage` and `WorkingImageFactory`. It is basically duplication and has a smell to it.

cmd/lifecycle/rebaser.go
```diff
@@ -106,30 +106,33 @@ func (r *rebaseCmd) Exec() error {
...
-		)
+		newBaseImageFactory = func() (imgutil.Image, error) {
+			opts := append(image.GetInsecureOptions(r.InsecureRegistries), remote.FromBaseImage(r.RunImageRef))
+			return remote.NewImage(
+				r.RunImageRef,
+				r.keychain,
+				opts...,
+			)
+		}
+		newBaseImage, err = newBaseImageFactory()
```
jabrown85: Instead of creating a factory - can we do the sniff test of the remote image at this point? If `TopLayer` can successfully be called higher up we can eliminate the need to pass a factory down. This goes for exporter as well. The `initRemoteAppImage` could block waiting on `TopLayer()` perhaps.

jabrown85:
```
  func openRemoteImage(logger log.Logger, newImage func() (imgutil.Image, error)) (imgutil.Image, error) {
      var lastErr error
      for attempt := 0; attempt <= len(topLayerDelays); attempt++ {
          img, err := newImage()
          if err == nil {
              if _, err = img.TopLayer(); err == nil {
                  return img, nil // manifest is complete
              }
          }
          lastErr = err
          if !isRetryable(err) {
              return nil, err
          }
          if attempt < len(topLayerDelays) {
              topLayerSleep(topLayerDelays[attempt])
          }
      }
      return nil, lastErr
  }

// later

appImage, err := openRemoteImage(cmd.DefaultLogger, func() (imgutil.Image, error) {
      return remote.NewImage(e.OutputImageRef, e.keychain, appOpts...)
  })
```

Arguably we could put this validation/retry into `imgutil`'s `remote` package instead

vpelikh: Implemented the openRemoteImage approach you suggested. Now code is much simpler. Thank you!

phase/rebaser.go
```diff
@@ -66,7 +67,7 @@ func (r *Rebaser) Rebase(workingImage imgutil.Image, newBaseImage imgutil.Image,
 	// update metadata label
 	origMetadata.RunImage.TopLayer, err = newBaseImage.TopLayer()
 	if err != nil {
-		return files.RebaseReport{}, fmt.Errorf("get rebase run image top layer SHA: %w", err)
+		return files.RebaseReport{}, errors.Wrap(err, "get rebase run image top layer SHA")
```
jabrown85: is there a reason to make this change now with the refactor?

phase/exporter.go
```diff
@@ -672,3 +677,52 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+// go-containerregistry caches manifests, so each retry attempt creates a fresh image.
+// Non-retryable errors (401, 403) are returned immediately without retry.
+func OpenRemoteImage(logger log.Logger, newImage func() (imgutil.Image, error)) (imgutil.Image, error) {
+	var lastErr error
+	for attempt := 0; attempt <= len(topLayerDelays); attempt++ {
+		img, err := newImage()
+		if err == nil {
+			if _, err = img.TopLayer(); err == nil {
+				if attempt > 0 {
+					logger.Warnf("Successfully opened remote image after %d retries", attempt)
```
jabrown85: Maybe info since this is good news after previous bad news log
```suggestion
					logger.Infof("Successfully opened remote image after %d retries", attempt)
```

phase/exporter.go
```diff
@@ -672,3 +677,52 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+
+// topLayerSleep is the function used for sleeping between retries.
+// It can be replaced for testing.
+var topLayerSleep = time.Sleep
+
+// isRetryable returns true if the error is likely transient and should be retried.
+// 401 Unauthorized and 403 Forbidden are not retryable as they indicate auth/config issues.
+func isRetryable(err error) bool {
+	if tErr, ok := stderrors.AsType[*transport.Error](err); ok {
+		return tErr.StatusCode != http.StatusUnauthorized && tErr.StatusCode != http.StatusForbidden
```
jabrown85: Let's add 400/405/429.

phase/exporter.go
```diff
@@ -672,3 +677,52 @@ func (e *Exporter) addSBOMLaunchLayer(opts ExportOptions, meta *files.LayersMeta
...
+	}
+	return true
+}
+
+// OpenRemoteImage opens a remote image with retry logic for registry mirror transient errors.
+// go-containerregistry caches manifests, so each retry attempt creates a fresh image.
+// Non-retryable errors (401, 403) are returned immediately without retry.
+func OpenRemoteImage(logger log.Logger, newImage func() (imgutil.Image, error)) (imgutil.Image, error) {
```
jabrown85: Consider moving this to `phase/retry.go` to match your test and make the diff for exporter similar to the other cmd phases

jabrown85 (commented):
This is looking much better! A few nits and we'll be ready

phase/rebaser.go
vpelikh: Reverted.

phase/exporter.go
vpelikh: Done.

phase/exporter.go
vpelikh: Done.

phase/exporter.go
vpelikh: Done.

jabrown85 (approved):

PR #1698, opened 2026-08-10
fix: resolve reachable CVEs and document non-exploitable grype findings

### LINKED ISSUE #1697: CVE(s) found in v0.21.14
Latest lifecycle release v0.21.14 triggered CVE(s) from Grype. For further details, see: https://github.com/buildpacks/lifecycle/actions/runs/31350476480 json: {
  "id": "GO-2026-5970",
  "severity": "High",
  "description": "A norm.Iter can enter an infinite loop when handling input containing invalid UTF-8 bytes."
}
{
  "id": "GO-2026-5064",
  "severity": "Critical",
  "description": "containerd CRI checkpoint restore CDI annotation smuggling in github.com/containerd/containerd"
}
{
  "id": "GO-2026-5338",
  "severity": "Critical",
  "description": "containerd: CRI checkpoint import allows local image tag poisoning in github.com/containerd/containerd"
}
{
  "id": "GO-2026-5158",
  "severity": "Medium",
  "description": "Opentelemetry-go's baggage parsing no longer caps raw header length in go.opentelemetry.io/otel"
}
{
  "id": "GO-2026-5622",
  "severity": "Medium",
  "description": "Arbitrary host CRI log file read via symlink following in CRI checkpoint restore in github.com/containerd/containerd"
}
{
  "id": "GO-2026-5841",
  "severity": "Unknown",
  "description": "Providing a specially crafted dictionary to s2.NewDict and using it to encode data can make the encoder read out of bounds."
}
{
  "id": "GO-2026-5932",
  "severity": "Unknown",
  "description": "The golang.org/x/crypto/openpgp package is unsafe by design, has numerous known security issues, is not maintained, and should not be used.\n\nIf you are required to interoperate with OpenPGP systems and need a maintained package, consider github.com/ProtonMail/go-crypto/openpgp which is a maintained fork that aims to be a drop-in replacement for this package."
}

pujitha24:
### Summary

Grype flagged 7 CVEs against the v0.21.14 release image. For each, I checked whether the vulnerable code is actually reachable from lifecycle's build (via `go list -deps` / `go mod why`) rather than reacting to the module version alone:

- `golang.org/x/text/unicode/norm` (GO-2026-5970, High) is reachable via `internal/extend/kaniko` -> `osscontainertools/kaniko/pkg/executor` -> `spf13/afero`, and is the one finding that's actually reproducible against a binary built from current `main` (see Validation). Fixed by bumping `golang.org/x/text` to v0.40.0 (fix landed in 0.39.0).
- `go.opentelemetry.io/otel/baggage` (GO-2026-5158, Medium) is reachable via `auth` -> `docker/docker/registry` -> `otelhttp` -> `otel/propagation` -> `otel/baggage`. Bumped `go.opentelemetry.io/otel{,/metric,/trace}` to v1.45.0 (fix landed in 1.44.0) as defense-in-depth, even though it wasn't reproducible against a binary built from current `main`.
- `github.com/klauspost/compress/s2` (GO-2026-5841) is not imported anywhere in the build (only `zstd` from that module is used). Bumped to the fixed v1.18.7 anyway since it's a trivial, risk-free patch bump.
- The three containerd CRI checkpoint-restore findings (GO-2026-5064, GO-2026-5338, GO-2026-5622) are not exploitable: lifecycle only imports `containerd/platforms`, never the CRI plugin. There is also no fix available for the `github.com/containerd/containerd` v1 module line lifecycle depends on (only `containerd/containerd/v2` was patched). Documented as suppressions in `.grype.yaml`, following this repo's existing convention.
- `golang.org/x/crypto/openpgp` (GO-2026-5932) is never imported by lifecycle and has no fixed version (upstream advisory says to stop using the package, not upgrade). Documented as a suppression in `.grype.yaml`.

`go mod tidy` after these bumps also pulled forward some purely transitive/tooling versions (`golang.org/x/sync`, `x/sys`, `x/mod`, `x/net`, `x/tools`, `x/telemetry`, `go-logr/logr`) via minimal version selection; these are incidental to satisfying the direct bumps above, not independently chosen.

#### Release notes

Bumped `golang.org/x/text`, `go.opentelemetry.io/otel`, and `github.com/klauspost/compress` to resolve reachable CVEs, and documented several non-exploitable/no-fix-available Grype findings in `.grype.yaml`.

---

### Related

Resolves #1697

---

### Context

Validation: built a `linux/amd64` lifecycle binary from `main` before and after this change and scanned both with a local `grype` CLI against a same-day vulnerability DB. Before: 1 finding (`GO-2026-5970`, High, `golang.org/x/text` v0.37.0). After: 0 findings. The other 6 CVEs from the issue did not reproduce against either binary built from current `main` (the issue was filed against the older v0.21.14 release, whose dependency versions differ from `main`); the `.grype.yaml` entries are added defensively since they were explicitly called out in the issue and could resurface in a release scan.

`go build ./...` and `go vet ./...` are clean. `go test $(go list ./... | grep -v acceptance)` passes except for the `buildpack` and `image` packages, which fail identically on an unmodified checkout in this sandbox (no local Docker daemon; a missing helper binary needed by one generate test) — confirmed pre-existing/environmental, not caused by this change, by re-running the same tests against `git stash`.


jabrown85 (approved):

jabrown85:
A rebase is needed here @pujitha24 - then I can re-review and merge

jabrown85 (changes requested):
needs rebase

jabrown85 (approved):

